PR #3, opened 2015-12-06
use mkdirp instead of fs.mkdir to allow for scoped packages

cab:
https://docs.npmjs.com/getting-started/scoped-packages

Scoped packages exist in `node_modules/@package-scope/package-name`, so using `mkdirp` will create the top level directory.


sebmck:
Squash commits and this is GTM.

PR #5, opened 2015-12-14
Add "run" command to run an npm script in each package

RickWong:


bin/lerna.js
```diff
@@ -1,11 +1,6 @@
 #!/usr/bin/env node
 
 var args = process.argv.slice(2);
-if (args.length > 1) {
```
sebmck: This looks good although I'd prefer if the `args` assertion was present when the command isn't `run`.

RickWong: Hey thanks for the comment. How would you like to see this part? Perhaps the check could be moved to the individual command scripts?

RickWong: √ Done what you literally said. I've also added an "ls" command that I used to test the too-many-arguments warning with :)

patrickheeney:
This looks cool. I wonder if it could be a bit more flexible like `lerna run --- make test-ci` to support more than just `npm run`, although that would be a good default.

hzoo:
Is this still WIP, needs review, or ready?

RickWong:
@hzoo It's functionally ready. Not sure whether @kittens has more feedback.

kittens:
LGTM. We should probably add some integration tests where we just run a bunch of commands and check that the console output is right too.

thejameskyle:
Moving to #44

Issue #6, opened 2015-12-18
Windows paths are broken


antigremlin:
While generating index.js files Lerna doesn't escape the backslashes. This produces files like this:

```js
module.exports = require("D:\src\babel\babel\packages\babel-messages");
```

PR #7, opened 2015-12-18
Escape backslashes in paths when generating index.js

antigremlin:
The solution is primitive, just a regex replace. It doesn't do full ES6
string escaping (is there even a standard way to do this?)

Closes: #6


antigremlin:
Oh, thanks! Now I know how to escape strings the right way :)

PR #10, opened 2016-01-07
Add private package support

patrickheeney:
I am looking to use lerna to manage our monorepo, but we don't have packages published on `npm`. This causes `npm install` to fail on `lerna bootstrap` because the package was not found on the registry. 

I added a simple check to see if `private: true` is set in `package.json`. I am not sure if this is the best way to do it because it bypasses the version check. However, in my case I don't have a version because my dependencies look like this:

```
# packages/some-other-package/package.json
  "dependencies": {
    "some-local-dependency": "file:../some-local-dependency"
  },
```

Is there a better way to do this?

### DIFF
--- a/lib/commands/bootstrap.js
+++ b/lib/commands/bootstrap.js
@@ -43,13 +43,14 @@ exports.execute = function (config) {
 
       tasks.push(function (done) {
         async.each(packages, function (sub, done) {
+          var priv = root.pkg.private || false;
           var ver = false;
           if (root.pkg.dependencies) ver = root.pkg.dependencies[sub.name];
           if (root.pkg.devDependencies && !ver) ver = root.pkg.devDependencies[sub.name];
           if (!ver) return done();
 
           // ensure that this is referring to a local package
-          if (ver[0] !== "^" || ver[1] !== config.currentVersion[0]) return done();
+          if (!priv && (ver[0] !== "^" || ver[1] !== config.currentVersion[0])) return done();
 
           var linkSrc = path.join(config.packagesLoc, sub.folder);
           var linkDest = path.join(nodeModulesLoc, sub.name);


thejameskyle:
I'm confused, you have some packages in your lerna repo that are private and others that are not?

thejameskyle:
Closing due to inactivity.

patrickheeney:
@thejameskyle Sorry, I missed this notification. Yes I have a private mono repo with private packages and some that get published to npm as public packages. The use case for this all the reasons you would use `file://` or private packages.

thejameskyle:
I don't understand the use case of having public and private packages. If things are public then their source code should be public too. I don't want to support mixing these

patrickheeney:
There are dozens of use cases for private packages which I won't get into here. Not all my private packages are published to npm, but I would think supporting private packages would fall in line with the purpose of the tool: `A tool for managing JavaScript projects with multiple packages.`.

patrickheeney:
Perhaps you are also thinking that the mono repo is open source, it's not. We have a private git repo using lerna to manage it. We have some packages that sync to open source, but the mono repo is not. Lerna does not currently support private packages of any sort, this PR fixes that.

thejameskyle:
I use lerna for a ton of private packages, it supports them fine. What you are asking for is local dependencies and mixing of local and remote, and I don't get the use case for that.

patrickheeney:
We have some local dependencies to the project, that are not published to npm but bundled with the project itself. Npm supports `file://` based dependencies so I am looking for lerna to support it as well. Without this simple change, the packages do not get bootstrapped for local development.

thejameskyle:
Could you create an example repo with the structure you have in mind?

patrickheeney:
I am not sure why the directory structure would matter with this, but I reckon something like:

```
packages/foo
packages/foo/internal_modules/bar
packages/baz

// packages/foo/package.json
"dependencies": {
  "bar": "file:./internal_modules/bar"
}
```

If you don't understand the use case, I don't want to keep pushing for local dependencies support. Perhaps I should look into other projects or rolling my own.

thejameskyle:
Okay good luck

kittens:
Feel free. We don't get paid to work on this :)

patrickheeney:
@kittens This is a PR that I took the time to write and submit? I wanted to contribute to the community and this repository, but I don't know how else to explain the importance of npm local packages, which this adds support for.

patrickheeney:
In case there was some confusion: My comments about rolling my own or searching for other projects was not intended to be snarky, but rather ask if my time would be better spent finding an alternative solution vs attempting to get this re-opened.

kittens:
@patrickheeney Ah right, my bad.

@thejameskyle I don't see why we shouldn't just support this flag, especially since the addition is so simple. We've even had some packages inside `babel/babel` that were internal and weren't meant for public consumption.

hzoo:
Sounds like a low complexity change to support; will be better when we have more tests?

patrickheeney:
@kittens @hzoo It sounds like you are both in support of this (private / local packages). Should this be closed? If the PR in its current format can't be accepted, should I open an issue to discuss potential implementations of this feature?

doug-wade:
fwiw, we've got a couple of private packages in react-server, c.f. https://github.com/redfin/react-server/blob/master/packages/react-server-integration-tests/package.json and https://github.com/redfin/react-server/blob/master/packages/react-server-examples/package.json

PR #15, opened 2016-01-17
Command to show what packages will be updated

hzoo:
Fixes https://github.com/kittens/lerna/issues/12

New command: `lerna updated` (open to suggestions for the name: `diff, changed, dryrun`) 
Add linting config from babel

There's some duplicated code (utility functions) that I didn't bother to share.

We should probably just print out the changed packages by default anyway when you run `lerna publish` right? We currently print it at the end.

---

For babel https://github.com/babel/babel/compare/v6.4.4...master

screen shot 2016-01-16 at 10 39 45 pm

### DIFF
--- a/lib/commands/publish.js
+++ b/lib/commands/publish.js
@@ -1,10 +1,13 @@
 var progressBar = require("../progress-bar");
+var updated     = require("./updated");
 var readline    = require("readline-sync");
 var semver      = require("semver");
 var chalk       = require("chalk");
 var child       = require("child_process");
 var async       = require("async");
 var fs          = require("fs");
+
+var checkUpdatedPackages = updated.checkUpdatedPackages;
 
 exports.description = "Publish updated packages to npm";
 
@@ -21,7 +24,9 @@ exports.execute = function (config) {
   //
 
   try {
-    checkUpdatedPackages();
+    changedPackages = checkUpdatedPackages(config);
+    console.log('Packages to be updated');
+    console.log(changedPackages);
     updateChangedPackages();
     updateTag();
     publish();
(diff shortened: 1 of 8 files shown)


lib/commands/publish.js
```diff
@@ -21,7 +24,9 @@ exports.execute = function (config) {
   //
 
   try {
-    checkUpdatedPackages();
+    changedPackages = checkUpdatedPackages(config);
+    console.log('Packages to be updated');
+    console.log(changedPackages);
```
hzoo: Printing this out as part of `lerna publish`

lib/commands/updated.js
```diff
@@ -0,0 +1,62 @@
+var progressBar = require("../progress-bar");
+var chalk       = require("chalk");
+var child       = require("child_process");
+var fs          = require("fs");
+
+exports.description = "Check which packages have changed since the last release";
+
+exports.execute = function (config) {
+  console.log(exports.checkUpdatedPackages(config));
```
hzoo: All this does is console.log the packages

lib/commands/publish.js
thejameskyle: How about a nicer format like: 

```js
changedPackages.map(pkg =>  `- ${pkg}`).join('\n');
```

hzoo: Sounds good!

lib/commands/publish.js
```diff
@@ -21,7 +24,11 @@ exports.execute = function (config) {
   //
 
   try {
-    checkUpdatedPackages();
+    changedPackages = checkUpdatedPackages(config);
+    console.log('Packages to be updated');
+    console.log(changedPackages.map(function(pkg) {
+      return "- " + pkg;
+    }).join('\n'));
```
thejameskyle: @kittens is going to complain about the single quotes.

NoNameProvided:
// Off
@hzoo what do you use to have that styled breadcrumbs in the first line of your console output?

lib/commands/publish.js
hzoo: Yeah I forgot to run lint after I pasted your snippet - I just added travis which will run lint for now

PR #19, opened 2016-01-22
Fix checkUpdatedPackages for repos without any tags

### LINKED ISSUE #14: `git rev-list --tags --max-count=1` fails
It wants me to put `master` after rev-list in order to function.

thejameskyle:
Fixes #14

### DIFF
--- a/lib/commands/updated.js
+++ b/lib/commands/updated.js
@@ -26,14 +26,25 @@ exports.checkUpdatedPackages = function (config) {
 
   var tick = progressBar(packageNames.length);
 
-  var lastTagCommit = execSync("git rev-list --tags --max-count=1");
-  var lastTag       = execSync("git describe " + lastTagCommit);
+  var hasTags = !!execSync("git tags");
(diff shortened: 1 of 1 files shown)


lib/commands/updated.js
```diff
@@ -26,8 +26,14 @@ exports.checkUpdatedPackages = function (config) {
 
   var tick = progressBar(packageNames.length);
 
-  var lastTagCommit = execSync("git rev-list --tags --max-count=1");
-  var lastTag       = execSync("git describe " + lastTagCommit);
+  var hasTags = !!execSync("git tags");
```
hzoo: Is this supposed to be `git tag`, since `git tags` isn't a command I don't think

thejameskyle:
```
➜  lerna git:(master) ✗ git tags
v1.0.1
v1.0.2
v1.0.3
v1.0.4
v1.0.5
v1.1.0
```

PR #22, opened 2016-01-23
Added support for handling @org/package

dendril:
Hey everyone!

I was having problems with packages which are not named as the folder containing the code (my personal scenario are scoped packages). So instead of use the folder's name for everything, I'm using the `package.json{name}` when running `npm dist-tag`.

Do you think we could do something to handle this scenario?

Any feedback will be welcome!

### DIFF
--- a/lib/commands/publish.js
+++ b/lib/commands/publish.js
@@ -26,8 +26,8 @@ exports.execute = function (config) {
   try {
     changedPackages = checkUpdatedPackages(config);
     console.log("Packages to be updated");
-    console.log(changedPackages.map(function(pkg) {
-      return "- " + pkg;
+    console.log(changedPackages.map(function(pkgObject) {
+      return "- " + pkgObject.pkgName;
     }).join("\n"));
     updateChangedPackages();
     updateTag();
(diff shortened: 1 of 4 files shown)


dendril:
Is there any chance to get some attention from the project maintainers? I want to prevent any kind of diverge from the official version :smile: 

Thanks again.

thejameskyle:
Sorry, I'm rewriting a lot right now, this is going to fall out of date, but I'll make sure its something that we do.

dendril:
Cool! let me know if I can give you a hand in some way.

thejameskyle:
Going to take on this PR in the refactored codebase. Is the only thing here that lerna needs to use the `package.json#name` for more things? Do you know which places those are?

dendril:
```
execSync("npm dist-tag rm " + pkgObject.pkgName + " prerelease");
execSync("npm dist-tag add " + pkgObject.pkgName + "@" + NEW_VERSION + " stable");
execSync("npm dist-tag add " + pkgObject.pkgName + "@" + NEW_VERSION + " latest");
```

thejameskyle:
Alright, I'll make that happen

dendril:
Yay! Thanks!

thejameskyle:
Moving to: #33

dchambers:
That's remarkably good timing! I just tried to use Lerna for the first time but found it doesn't currently work because I'm using scoped package names. Looking forward to a release containing the fix :smile:.

PR #23, opened 2016-01-23
WIP: add lerna changelog

### LINKED ISSUE #18: lerna changelog
Could print out a simple changelog of commits since the last tag/release?

Can be on by default after `lerna publish` but can be run seperately.

Example: ```git log `git describe --tags --abbrev=0`..HEAD --pretty=format:' * %s (%an)' | grep -v 'Merge pull request'```

hzoo:
Fixes #18

So far, I'm just getting a list of commits that are either the merge commit or a commit to master directly.

screen shot 2016-01-23 at 10 09 44 pm

Running: 
` GITHUB_AUTH=auth-token-here GITHUB_ORG=babel GITHUB_REPO=babel ./node_modules/.bin/lerna changelog`

Right now we error if you pass more than 2 args so we'll either have to support multiple args, do more `process.env` stuff, or add a config file?

### DIFF
--- a/lib/commands/changelog.js
+++ b/lib/commands/changelog.js
@@ -1,0 +1,197 @@
+var child       = require("child_process");
+var progressBar = require("../progress-bar");
+
+var org  = process.env.GITHUB_ORG;
+var repo = process.env.GITHUB_REPO;
+
+var tags = [
+  "tag: breaking change",
+  "tag: spec compliancy",
+  "tag: new feature",
+  "tag: bug fix",
+  "tag: documentation",
+  "tag: internal",
+  "tag: polish"
+];
+
+exports.description = "Creates a basic changelog categorized by designated github labels";
+
+function getLastTag() {
+  return execSync("git describe --abbrev=0 --tags");
+}
+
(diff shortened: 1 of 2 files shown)


hzoo:
Code sucks but at least it prints out the labels now.

hzoo:
Almost done but it's needs to be generalized (just want to get it working for babel)

lib/commands/changelog.js
```diff
@@ -0,0 +1,191 @@
+var child       = require("child_process");
+var progressBar = require("../progress-bar");
+var updated     = require("./updated");
+
+var getLastTag = updated.getLastTag;
+var org  = process.env.GITHUB_ORG;
+var repo = process.env.GITHUB_REPO;
+
+var tags = [
```
hzoo: Just need to not hardcode this in

hzoo:
Didn't get to updating the readme but I would do that as well

Maybe a screenshot or example of what it looks like (or just link to the next babel changelog release when we do it).

hzoo:
Can pr the one in the repo instead of my branch.

PR #24, opened 2016-01-25
be able to change the npm dist tag

hzoo:


kittens:
We should add some command line flags instead of using environment variables.

hzoo:
Yeah that would probably be better.

PR #34, opened 2016-02-01
Symlink instead of creating files

thejameskyle:
I'm not sure if there was any reason not to just symlink dependencies, but it causes issues like this: https://github.com/kittens/lerna/issues/11


hzoo:
It definitely was symlinks before - I forgot it it had to do with multiple copies of a package/babel-traverse or something - probably need to ask seb since I can't recall.

thejameskyle:
Talking to seb about it right now, seems to be working fine for the Babel repo and every test I can come up with. Idk if it was fixed by #31

thejameskyle:
yolo

hzoo:
i'm :innocent:

PR #36, opened 2016-02-02
Add Independent mode

### LINKED ISSUE #20: Independent mode
Talked to @kittens about this the other day.

I have a project where I want all the packages to be in the same repo, to be tested together and for versions to be bumped for inter-dependencies. However we don't want all the versions to be locked together.

I'm suggesting adding an `--independent` flag (name can be whatever) that will not force the same version on all packages and will instead prompt for versioning in each package that has been changed.

thejameskyle:
Resolves #20

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -15,8 +15,14 @@ var cli = meow([
   "  updated    Check which packages have changed since the last release",
   "  run        Run npm script in each package",
   "  ls         List all public packages",
-  ""
-]);
+  "",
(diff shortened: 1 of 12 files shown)


thejameskyle:
@hzoo Would you feel okay merging this now? I've been depending on this branch at work and it's been working fine.

thejameskyle:
I'm just going to merge this for now since the next release is going to be 2.0 and I want to make sure everything works with it.

PR #44, opened 2016-02-11
Run and ls

thejameskyle:
Moving from #5


hzoo:
Can update the readme for this one as well

thejameskyle:
Actually, I think we should completely rewrite the README before 2.0

hzoo:
Kk, we suddenly have a lot more commands

hzoo:
Oh maybe `ls` could provide more info (or maybe that is a different command)? It's the same as updated except all packages

It could print like: name, last version (useful for independent mode?), if changed, last update, last tag date, etc?

thejameskyle:
Sure, want to make a PR? Otherwise I can get to it tomorrow

PR #45, opened 2016-02-11
Add lerna diff command

### LINKED ISSUE #41: lerna diff
Whenever I run `lerna updated` and find something I didn't expect in there I immediately want to see the diff. It would be great if lerna could easily show me that.

```sh
$ lerna diff
$ lerna diff package-name
```

thejameskyle:
Resolves #41

### DIFF
--- a/lib/commands/diff/index.js
+++ b/lib/commands/diff/index.js
@@ -5,18 +5,18 @@ var path  = require("path");
 var fs    = require("fs");
 
 module.exports = function (config, cli) {
-  var package = cli.input[1];
+  var packageName = cli.input[1];
   var filePath = "packages";
 
-  if (package) {
-    var pkgLoc = path.join(config.packagesLoc, package, "package.json");
+  if (packageName) {
+    var pkgLoc = path.join(config.packagesLoc, packageName, "package.json");
 
     if (!fs.existsSync(pkgLoc)) {
-      console.log(chalk.red("Package '" + package + "' does not exist."));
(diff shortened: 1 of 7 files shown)


hzoo:
Do we need to add an entry to meow? Also update the readme with usage

lib/commands/diff/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+    if (!fs.existsSync(pkgLoc)) {
+      console.log(chalk.red("Package '" + packageName + "' does not exist."));
+      process.exit(1);
+    }
+
+    filePath += "/" + packageName;
+  }
+
+  var diff = child.spawn("git", ["diff", "--color", filePath]);
```
hzoo: https://github.com/kittens/lerna/blob/master/lib/utils/checkUpdatedPackages.js#L25

Do we need to check against the last tag? `lastTag = execSync("git describe --tags " + lastTagCommit);`

thejameskyle: > Do we want to check against the last tag?

Oh right duh.

> Or is that another cli parameter to pass in?

I think this should just be a simple command to see everything since the last release

hzoo: Ok, I saw in the OP you said last tag too. Yeah for sure

hzoo: I think this is good to go in after you add the last tag stuff (I pasted it from another command, and you can make it a util if you want), or I can finish it up

PR #46, opened 2016-02-11
Switch to publishing temp tag instead of "prerelease"

thejameskyle:
Thought I'd open this to discuss since we were talking about pre-releases elsewhere and theres no reason we need to override the "prerelease" tag.


hzoo:
Ok sure. If we are goina allow pre-releases then we should change this to temp (since the point of doing it was in case some of the packages failed to publish and we want to make sure they are all published at once).

thejameskyle:
Do you mean I should change it from `lerna-temp` to `temp`?

hzoo:
Nah the name doesn't really matter since it's going to be deleted, I was just saying I agree with doing it

thejameskyle:
Oh okay cool

hzoo:
We'll probably catch it when the README is redone, but just noting here we should update the readme since its `lerna-temp` now

PR #57, opened 2016-02-13
Add travis badge

hzoo:
For when we turn on travis (more useful when we have tests but why not)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@
   A tool for managing JavaScript projects with multiple packages.
 </p>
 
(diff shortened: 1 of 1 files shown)


thejameskyle:
heh... I think I'll just leave this open until there's actually CI and tests.

hzoo:
:smile:, I think @kittens needs to turn travis on for the repo too (right now it only checks linting)

hzoo:
Travis is enabled now and runs `npm run ci` which is just lint atm

PR #60, opened 2016-02-20
disable progress bar in non-TTY environments

ianwremmel:
If I try to use lerna inside a docker container (and possibly under certain Jenkins configurations), it fails with something `like this.stream.clear() is undefined`. I've found a few reference scattered around the Internet to fix it, but most imply node-progress has already fixed the problem (which is clearly not the case).


forivall:
:+1: was trying to run travis tests locally under vagrant

PR #65, opened 2016-02-22
Publish as a dev/canary release

hzoo:
Testing in https://github.com/hzoo/hzoo-npm-publish.

Right now I don't know how to make this automatically run in travis since you can't just do `npm publish` and have to use their deploy feature (I think)

I just ran it locally (or we can set this up on our instance so that it listens for a merge and then will run `./node_modules/.bin/lerna publish --canary`.

We can change the name/implementation; just testing.

screen shot 2016-02-22 at 10 03 07 am

screen shot 2016-02-22 at 10 06 00 am

- [x] Fix: `checkUpdatedPackages` does a git diff on the last tag (but I'm not creating a tag here, so either we do need to create a git tag or we figure out how to make sure we don't publish all packages unless that is necessary for this feature to work) - we could check the commits related to the merge PR and git diff that only.

screen shot 2016-02-22 at 11 33 36 pm

### DIFF
--- a/lib/commands/bootstrap/bootstrapPackages.js
+++ b/lib/commands/bootstrap/bootstrapPackages.js
@@ -1,9 +1,9 @@
 var linkDependencies = require("./linkDependencies");
 var getPackages      = require("../../utils/getPackages");
 
-module.exports = function bootstrapPackages(packagesLoc, currentVersion, independent, callback) {
+module.exports = function bootstrapPackages(packagesLoc, currentVersion, flags, callback) {
   var packages = getPackages(packagesLoc);
-  linkDependencies(packages, packagesLoc, currentVersion, independent, function(err) {
+  linkDependencies(packages, packagesLoc, currentVersion, flags, function(err) {
     if (err) return callback(err);
     callback(null, packages);
   });
--- a/lib/commands/publish/index.js
+++ b/lib/commands/publish/index.js
@@ -8,7 +8,7 @@ module.exports = function publish(config) {
     config.versionLoc,
     config.packagesLoc,
     config.flags,
-    function(err, changedPackages, versions) {
+    function(err, changedPackages) {
       if (err) {
         logger.log("error", "Errored while publishing.", false, err);
         exit(1);
@@ -16,7 +16,7 @@ module.exports = function publish(config) {
         var message = "Successfully published:";
 
         changedPackages.forEach(function(pkg) {
-          message += "\n - " + pkg.name + "@" + versions[pkg.name];
+          message += "\n - " + pkg.name + "@" + pkg.pkg.version;
         });
 
         logger.log("success", message);
--- a/lib/commands/publish/index.js
+++ b/lib/commands/publish/index.js
@@ -7,7 +7,7 @@ module.exports = function publish(config) {
     config.currentVersion,
     config.versionLoc,
     config.packagesLoc,
-    config.independent,
+    config.flags,
     function(err, changedPackages, versions) {
       if (err) {
         logger.log("error", "Errored while publishing.", false, err);
--- a/lib/commands/publish/npmPublishSemiAtomic.js
+++ b/lib/commands/publish/npmPublishSemiAtomic.js
@@ -10,7 +10,7 @@ module.exports = function npmPublishSemiAtomic(packages, packagesLoc, versions, 
     if (flags.canary) {
       console.log("Reseting git state");
       // reset since the package.json files are changed
-      execSync("git reset HEAD --hard");
+      execSync("git checkout -- packages/*/package.json");
     }
 
     // Now that we've successfully published a prerelease set it as the latest
--- a/lib/commands/publish/npmPublishSemiAtomic.js
+++ b/lib/commands/publish/npmPublishSemiAtomic.js
@@ -1,12 +1,19 @@
 var npmUpdateAsLatest = require("./npmUpdateAsLatest");
 var npmPublishAsTemp  = require("./npmPublishAsTemp");
+var execSync          = require("../../utils/execSync");
 
-module.exports = function npmPublishSemiAtomic(packages, packagesLoc, versions, callback) {
+module.exports = function npmPublishSemiAtomic(packages, packagesLoc, versions, flags, callback) {
   // Start by publishing a prerelease that we'll bump later to be more atomic
   npmPublishAsTemp(packages, packagesLoc, function (err) {
     if (err) return callback(err);
 
+    if (flags.canary) {
+      console.log("Reseting git state");
+      // reset since the package.json files are changed
+      execSync("git reset HEAD --hard");
+    }
+
     // Now that we've successfully published a prerelease set it as the latest
-    npmUpdateAsLatest(packages, versions, callback);
+    npmUpdateAsLatest(packages, versions, flags, callback);
   });
 };
--- a/lib/commands/publish/npmUpdateAsLatest.js
+++ b/lib/commands/publish/npmUpdateAsLatest.js
@@ -12,7 +12,7 @@ module.exports = function npmUpdateAsLatest(changedPackages, versions, flags, ca
     return function (done) {
       while (true) {
         try {
-          if (~execSync("npm dist-tag ls " + pkg.name).indexOf("lerna-temp")) {
+          if (npmUtils.checkDistTag(pkg.name, "lerna-temp")) {
             logger.log("info", "Removing temporary npm dist-tag");
             npmUtils.removeDistTag(pkg.name, "lerna-temp");
           }
--- a/lib/commands/publish/npmUpdateAsLatest.js
+++ b/lib/commands/publish/npmUpdateAsLatest.js
@@ -22,7 +22,7 @@ module.exports = function npmUpdateAsLatest(changedPackages, versions, flags, ca
             npmUtils.addDistTag(pkg.name, versions[pkg.name], process.env.NPM_DIST_TAG);
           } else if (flags.canary) {
             logger.log("info", "Adding canary npm dist-tag");
-            npmUtils.addDistTag(pkg.name, versions[pkg.name], "canary");
+            npmUtils.addDistTag(pkg.name, pkg.pkg.version, "canary");
           } else {
             logger.log("info", "Adding latest npm dist-tag");
             npmUtils.addDistTag(pkg.name, versions[pkg.name], "latest");
--- a/lib/commands/publish/npmUpdateAsLatest.js
+++ b/lib/commands/publish/npmUpdateAsLatest.js
@@ -3,7 +3,7 @@ var npmUtils    = require("../../utils/npmUtils");
 var logger      = require("../../utils/logger");
 var async       = require("async");
 
-module.exports = function npmUpdateAsLatest(changedPackages, versions, callback) {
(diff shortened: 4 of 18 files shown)


hzoo:
Ok got it to successfully publish only the commits related to a merge.

lib/utils/checkUpdatedPackages.js
```diff
@@ -20,11 +21,14 @@ module.exports = function checkUpdatedPackages(packagesLoc) {
 
   var hasTags = gitHasTags();
   var lastTagCommit;
-  var lastTag;
+  var commits;
 
-  if (hasTags) {
+  if (flags.canary) {
+    var lastSHA = getLastSHA();
+    commits = lastSHA.slice(0, 8) + "^.." + lastSHA.slice(0, 8);
```
thejameskyle: Can you pull this into it's own descriptive function?

hzoo: Yeah good idea

lib/commands/publish/npmPublishSemiAtomic.js
```diff
@@ -1,12 +1,19 @@
...
+var execSync          = require("../../utils/execSync");
 
-module.exports = function npmPublishSemiAtomic(packages, packagesLoc, versions, callback) {
+module.exports = function npmPublishSemiAtomic(packages, packagesLoc, versions, flags, callback) {
   // Start by publishing a prerelease that we'll bump later to be more atomic
   npmPublishAsTemp(packages, packagesLoc, function (err) {
     if (err) return callback(err);
 
+    if (flags.canary) {
+      console.log("Reseting git state");
```
hzoo: Just commenting here if it was missed (the package.json files will be modified (dependencies and the version)), but we aren't commiting or creating a tag so the git state isn't empty so we reset.

lib/commands/publish/npmUpdateAsLatest.js
```diff
@@ -12,10 +12,15 @@ module.exports = function npmUpdateAsLatest(changedPackages, versions, callback)
       while (true) {
         try {
           console.log("Removing temporary npm dist-tag");
-          execSync("npm dist-tag rm " + pkg.name + " lerna-temp");
+          if (~execSync("npm dist-tag ls " + pkg.name).indexOf("lerna-temp")) {
```
hzoo: Might not be necessary but just checks if its there so it doesn't error out

lib/commands/publish/publishChangedPackages.js
```diff
@@ -96,9 +97,13 @@ module.exports = function publishChangedPackages(
 
   try {
     // Find all changed packages to publish
-    changedPackages = checkUpdatedPackages(packagesLoc);
+    changedPackages = checkUpdatedPackages(packagesLoc, flags);
 
-    if (!independent) {
+    if (flags.canary) {
+      // stick with the current version
+      masterVersion = currentVersion;
```
hzoo: Don't update the version file, and use the current version

lib/commands/publish/publishChangedPackages.js
```diff
@@ -106,8 +111,11 @@ module.exports = function publishChangedPackages(
     }
 
     changedPackages.forEach(function(pkg) {
-      if (independent) {
+      if (flags.independent) {
         versions[pkg.name] = userGetVersion(pkg.version, pkg.name);
+      } else if (flags.canary) {
+        // set the master version plus sha of the current commit
+        versions[pkg.name] = masterVersion + "-canary." + getLastSHA().slice(0, 8);
```
hzoo: appends `-canary.sha`

lib/commands/publish/updateChangedPackages.js
```diff
@@ -1,7 +1,15 @@
 var getPackageLocation = require("../../utils/getPackageLocation");
-var gitAddFile            = require("../../utils/gitAddFile");
+var getPackages        = require("../../utils/getPackages");
+var gitAddFile         = require("../../utils/gitAddFile");
 var fs                 = require("fs");
 
+function allPackages(packagesLoc) {
```
hzoo: Could refactor `ls/index.js` to return or something instead of this

lib/commands/publish/updateChangedPackages.js
```diff
@@ -19,10 +27,12 @@ function updateDepsObject(deps, changedPackages, versions) {
   }
 }
 
-module.exports = function updateChangedPackages(changedPackages, packagesLoc, versions) {
+module.exports = function updateChangedPackages(changedPackages, packagesLoc, versions, flags) {
   var changedFiles = [];
 
-  changedPackages.forEach(function (pkg) {
+  var packages = flags.canary ? allPackages(packagesLoc) : changedPackages;
```
hzoo: Do we want to do this by default? (Is this related to a current issue?)

If I only modify `pkg1` and other pkgs depend on it (say pkg2), then if we publish something that isn't covered in the range then all those dependencies wouldn't be updated.

Ex: in the case of babel-generator, if we only publish that then I think the users who want to use the newer version of it but only use babel-core wouldn't get the update?

lib/commands/publish/npmPublishSemiAtomic.js
thejameskyle: Err... This really worries me, I don't want to do this since it's a very destructive operation. Is there a way you could pass the canary flag into `npmPublishAsTemp` and avoid having to do this?

hzoo:
Ok getting complicated.. can probably refactor a lot

screen shot 2016-02-24 at 12 17 48 am

Now it will publish dependencies that use the packages that have been updated (babel-core if babel-generator updates)

lib/commands/publish/npmPublishSemiAtomic.js
```diff
@@ -1,12 +1,19 @@
...
+      // reset since the package.json files are changed
+      execSync("git checkout -- */package.json");
```
thejameskyle: Just to be more specific can this be `packages/*/package.json`?

hzoo: :+1:

lib/utils/getCurrentSHA.js
```diff
@@ -0,0 +1,5 @@
+var execSync = require("./execSync");
+
+module.exports = function getCurrentSHA() {
```
thejameskyle: Want to put this in gitUtils?

lib/utils/checkUpdatedPackages.js
```diff
@@ -20,11 +26,14 @@ module.exports = function checkUpdatedPackages(packagesLoc) {
 
   var hasTags = gitUtils.hasTags();
   var lastTagCommit;
-  var lastTag;
+  var commits;
 
-  if (hasTags) {
+  if (flags.canary) {
+    var currentSHA = gitUtils.getCurrentSHA();
```
hzoo: Instead of the `getCurrentSHA`, we could change this to be a default and either prompt the user or even better just use a passed in sha from the flag (flag.canary). Would be cool to support just passing in a PR number too

lib/commands/bootstrap/bootstrapPackages.js
```diff
@@ -1,9 +1,9 @@
 var linkDependencies = require("./linkDependencies");
 var getPackages      = require("../../utils/getPackages");
 
-module.exports = function bootstrapPackages(packagesLoc, currentVersion, independent, callback) {
+module.exports = function bootstrapPackages(packagesLoc, currentVersion, flags, callback) {
```
hzoo: If we want we can move all these params into an object

lib/commands/publish/index.js
```diff
@@ -16,7 +16,7 @@ module.exports = function publish(config) {
         var message = "Successfully published:";
 
         changedPackages.forEach(function(pkg) {
-          message += "\n - " + pkg.name + "@" + versions[pkg.name];
+          message += "\n - " + pkg.name + "@" + pkg.pkg.version;
```
hzoo: We can use this here since it should be set correctly

lib/commands/publish/publishChangedPackages.js
```diff
@@ -59,7 +60,7 @@ function gitRemoveTaggedVersions(tags) {
 }
 
 function logChangedPackages(packages, versions) {
-  logger.log("info", "Packages to be updated");
+  logger.log("info", "Packages to be updated based on changes");
```
hzoo: could change the wording here

lib/commands/publish/publishChangedPackages.js
```diff
@@ -89,8 +94,11 @@ module.exports = function publishChangedPackages(
     }
 
     changedPackages.forEach(function(pkg) {
-      if (independent) {
+      if (flags.independent) {
         versions[pkg.name] = userGetVersion(pkg.version, pkg.name);
+      } else if (flags.canary) {
+        // set the master version plus sha of the current commit
+        versions[pkg.name] = masterVersion + "-canary." + gitUtils.getCurrentSHA().slice(0, 8);
```
hzoo: I guess we could move this out into a function

lib/commands/publish/npmUpdateAsLatest.js
```diff
@@ -12,13 +12,19 @@ module.exports = function npmUpdateAsLatest(changedPackages, versions, callback)
...
+            logger.log("info", "Removing temporary npm dist-tag");
+            npmUtils.removeDistTag(pkg.name, "lerna-temp");
+          }
+
           if (process.env.NPM_DIST_TAG) {
             logger.log("info", "Adding " + process.env.NPM_DIST_TAG + " npm dist-tag");
             npmUtils.addDistTag(pkg.name, versions[pkg.name], process.env.NPM_DIST_TAG);
+          } else if (flags.canary) {
+            logger.log("info", "Adding canary npm dist-tag");
+            npmUtils.addDistTag(pkg.name, pkg.pkg.version, "canary");
```
hzoo: If `pkg.pkg.version` has already been set, we can just use that everywhere instead of `versions[pkg.name]` right? In this case, the cross dependent versions are undefined (unless I should modify that instead)

thejameskyle: Are we actually updating the version in `pkg.pkg.version` or are we just updating the file?

hzoo: https://github.com/hzoo/lerna/blob/publish-dev/lib/commands/publish/updateChangedPackages.js#L46-L58

in `updateChangedPackages` we used to only change the updated packages.

I changed it to run on all packages, and then set the version in `pkg.pkg.version = version;` at the end we rewrite the package.json

hzoo:
I did find you that you can do `^1.1.1-canary`. Could change all the dependencies to that instead of the exact release. (or `canary` dist tag)

screen shot 2016-02-27 at 11 44 00 am

lib/commands/publish/npmPublishSemiAtomic.js
thejameskyle: Can you change this to use the logger?

dendril:
@hzoo @thejameskyle I'm pretty new versioning packages using `npm` and feels pretty conformable with the canary strategy over custom tags in npm.

The question is, does it make sense to keep `NPM_DIST_TAG` over `--canary` or `--tag name`?

Feel free to remove the comment if make noise.

Thanks!

thejameskyle:
@dendril want to create a separate issue for that?

dendril:
Done #77.

Thanks!

PR #66, opened 2016-02-22
Fix diff output colors

evocateur:
Without this patch, in OS X `bash`,  `default-pager` outputs a bunch of escaped color codes:

```
ESC[1mdiff --git a/packages/my-awesome-pkg/.npmignore b/packages/my-awesome-pkg/.npmignoreESC[m
ESC[1mnew file mode 100644ESC[m
ESC[1mindex 0000000..281df39ESC[m
ESC[1m--- /dev/nullESC[m
ESC[1m+++ b/packages/my-awesome-pkg/.npmignoreESC[m
ESC[36m@@ -0,0 +1,2 @@ESC[m
ESC[32m+ESC[mESC[32msrcESC[m
ESC[32m+ESC[mESC[32mtestESC[m
ESC[1mdiff --git a/packages/my-awesome-pkg/README.md b/packages/my-awesome-pkg/README.mdESC[m
ESC[1mnew file mode 100644ESC[m
ESC[1mindex 0000000..b326138ESC[m
ESC[1m--- /dev/nullESC[m
ESC[1m+++ b/packages/my-awesome-pkg/README.mdESC[m
ESC[36m@@ -0,0 +1 @@ESC[m
ESC[32m+ESC[mESC[32m# my-awesome-pkgESC[m
```

Providing `less -R` as the pager makes it display the correct colors.

I suppose one could consult `process.env.PAGER` to make sure we aren't overriding user preferences, but that's problematic because `export PAGER='less -R'` doesn't work with `default-pager`'s method of defaulting. We'd have to split the args manually, etc, and that's extremely error prone for very little gain.

### DIFF
--- a/lib/commands/diff/index.js
+++ b/lib/commands/diff/index.js
@@ -35,11 +35,7 @@ module.exports = function (config, cli) {
     lastCommit = gitGetFirstCommit();
   }
 
-  var diff = child.spawn("git", ["diff", lastCommit, "--color", filePath]);
-  var pagerOpts = {
-    pager: "less",
-    args: ["-R"] // parse raw control characters
-  };
-
-  diff.stdout.pipe(pager(pagerOpts));
+  child.spawn("git", ["diff", lastCommit, "--color", filePath], {
+    stdio: "inherit"
+  });
 };
--- a/lib/commands/diff/index.js
+++ b/lib/commands/diff/index.js
@@ -36,6 +36,10 @@ module.exports = function (config, cli) {
   }
 
   var diff = child.spawn("git", ["diff", lastCommit, "--color", filePath]);
+  var pagerOpts = {
+    pager: "less",
+    args: ["-R"] // parse raw control characters
+  };
 
-  diff.stdout.pipe(pager());
+  diff.stdout.pipe(pager(pagerOpts));
 };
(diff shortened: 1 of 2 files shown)


thejameskyle:
Hmm, kinda annoying that this isn't a default that can be overridden. Is there a way to test if a pager supports colors?

lib/commands/diff/index.js
```diff
@@ -36,6 +36,10 @@ module.exports = function (config, cli) {
   }
 
   var diff = child.spawn("git", ["diff", lastCommit, "--color", filePath]);
+  var pagerOpts = {
+    pager: "less",
+    args: ["-R"] // parse raw control characters
+  };
```
thejameskyle: Maybe this could somehow default to `process.env.PAGER`

evocateur:
A simpler workaround is to not use `default-pager` at all, and just inherit `stdio`.

```js
child.spawn("git", ["diff", lastCommit, "--color", filePath], {
  stdio: "inherit"
});
```

The advantage to this approach is that people who wish to configure their `git`'s pager behaviour will be respected, and the default should "just work" as it doesn't attempt to be fancy (`git` is pretty decent at colors on a TTY by default).

Now that I think about it some more, I think inheriting `stdio` is the way to go.

lib/commands/diff/index.js
evocateur: One problem with `process.env.PAGER` is that not everyone will have it, contrary to the expectation of `default-pager`. Inheriting `stdio` does not have this problem.

thejameskyle:
Hmm makes sense, want to update your PR

evocateur:
Let me know if you want me to squash the commits, should be good-to-go now.

thejameskyle:
Thanks

evocateur:
Glad I could help :)

PR #75, opened 2016-03-02
Expose public api

### LINKED ISSUE #74: Using lerna as a library
Hi, I'm trying to build a couple of tools on top of lerna, but I don't want to duplicate logic inside of lerna. Especially since it seems to be in flux right now and I don't want to keep updating things.

It would be cool if you could `require("lerna")` and call methods like `getPackagesDirectory()` or [`getPackages()`](https://github.com/kittens/lerna/blob/master/lib/utils/getPackages.js).

I can open a PR for this

thatjessicakelly:
Resolves #74

I made more changes than I probably needed to. It seemed like most utils were grouped into stuff like `gitUtils` and `fsUtils`, but the path stuff was mixed up everywhere, so I put it all in the same place.

I also added some tests for the public api.


thejameskyle:
Mind rebasing?

thatjessicakelly:
done

PR #84, opened 2016-03-10
Adds optional argument to specify master version when publishing

dijs:
Added so we could bypass user input during publishing.


ahfarmer:
This change would really help me - any chance it can be merged @thejameskyle ?

I'm trying to use lerna in an automated publish system. (Alongside [commitizen](https://commitizen.github.io/cz-cli/) and [conventional-recommended-bump](https://github.com/stevemao/conventional-recommended-bump) )

hzoo:
Since this skips user input it sounds like https://github.com/lerna/lerna/pull/135? I think we can close this in favor of that

thejameskyle:
`--confirm` does not skip the input prompts, just the confirmations.

hzoo:
Ok sounds good. Not sure we want to call it the `master version` although not sure I have better names - maybe even `--repo-version` or `--monorepo-version`?

dijs:
So what is the change I need to make here?

hzoo:
I'm just not sure about the name for the flag, and then also a test or to for the option would be nice

hzoo:
Great, I think the last thing is to update the docs with the new option, maybe after https://github.com/lerna/lerna#--yes?

dijs:
Done

hzoo:
Awesome, thanks a lot! 🎉

PR #85, opened 2016-03-11
Create lerna init command

### LINKED ISSUE #47: lerna init
Rather than running the [init script](https://github.com/kittens/lerna/blob/master/lib/init.js) every single time, I think it'd be good to split that out into a `lerna init` command that did the same thing.

When you run any other command in a directory that hasn't been initialized using `lerna init` then it would just error.

This prevents weirdness when accidentally running lerna in a subdirectory which currently will just setup a new lerna repo.

thejameskyle:
Resolves #47

This will simplify a lot of things that I want to do soon like `lerna.json`, then once `lerna.json` is in we won't have to use the `cwd` anymore and we can recursively look up the tree to find a `lerna.json` or `package.json` with a `lerna` field.

This also improves the experience of `--independent`, see the warning in `lib/init.js`.

```
`VERSION` does not exist, have you run `lerna init`? Or maybe you meant to run this with `--independent` or `-i`?
```

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -14,6 +14,7 @@ var cli = meow([
   "  publish    Publish updated packages to npm",
   "  updated    Check which packages have changed since the last release",
   "  diff       Diff all packages or a single package since the last release",
+  "  init       Initialize a lerna repo",
   "  run        Run npm script in each package",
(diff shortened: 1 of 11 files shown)


hzoo:
Ok :+1:

PR #86, opened 2016-03-11
Use git directory instead of cwd

thejameskyle:
This just allows you to run lerna from any directory inside a git repo.


hzoo:
:+1:

PR #90, opened 2016-03-23
Add -a to 'git commit'

ahfarmer:
This is useful if you want to integrate lerna with other services.

I'm running [conventional-changelog-cli](https://github.com/stevemao/conventional-changelog-cli) and then `lerna`. With this change, `lerna` will include the changelog in it's version commit.


thejameskyle:
We don't want to do this because we want files to be manually added because it causes problems otherwise. Try manually `git add` the files before running lerna.

ahfarmer:
Oh yeah, duh. Thanks!

PR #92, opened 2016-03-25
Fixes the progress bar not working

kentaromiura:
Fixes the progress bar not working and erroring when spawning lerna because stdout is not a TTY when called from process.spawn, node-progress was failing when trying to call this.stream.clearLine

### DIFF
--- a/lib/progress-bar.js
+++ b/lib/progress-bar.js
@@ -2,14 +2,30 @@ var ProgressBar = require("progress");
 var pad         = require("pad");
 
 module.exports = function (total) {
+
+  var COLUMNS = process.stdout.columns || 100
+  if (! process.stderr.cursorTo) {
+    process.stderr.columns = COLUMNS;
+    process.stderr.isTTY = true;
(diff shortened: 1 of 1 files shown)


thejameskyle:
Should this maybe be fixed within node-progress?

kentaromiura:
by looking at it node-progress seems to be an abandoned project

thejameskyle:
I just got added to the project by tj, if you want to make a PR we can continue over there

reggi:
What's the status here? Could this fix my issue?

https://github.com/lerna/lerna/issues/164

kentaromiura:
I didn't follow-up because I fixed the problem by spawning the process with `{stdio:'inherit'}`, so that the progress bar receive a TTY;
I think it may fix your issue as it mocks the TTY commands needed by the progressbar

hzoo:
What's up with this PR?

gamtiq:
I have tested the fix in Windows (cygwin). It works fine.

Issue #27, opened 2016-01-25
Should changed packages include cross-depencies?


thejameskyle:
Right now when publishing a new version we:

1. Find all the changed packages by checking their diffs. [ref](https://github.com/kittens/lerna/blob/master/lib/commands/updated.js#L49-L54)
1. Cross check these "changed packages" for cross-deps and update them in `package.json`. [ref](https://github.com/kittens/lerna/blob/master/lib/commands/publish.js#L83-L104)
1. However, we don't update cross-dependencies outside of the original changed packages. ([ref1](https://github.com/kittens/lerna/blob/master/lib/commands/publish.js#L94-L95), [ref2](https://github.com/kittens/lerna/blob/master/lib/commands/publish.js#L88-L89))

I feel like this will cause issues with major versions. Because we're relying on npm's `^x.x.x` version matching for minor and patch versions outside of the changed packages.

hzoo:
Yeah I think you're right on that.
Could we do that only if it's a major version change? Thats kinda what `FORCE_VERSION` does (but it would update all packages).

thejameskyle:
Only doing it on major versions sounds like the best solution. The question is if we recursively look up changed packages or bump them all.

I'm concerned about how this would interact with #20

thejameskyle:
Fixed in 2.0.0-beta.4

PR #95, opened 2016-04-02
[WIP] Find all dependents to update

thejameskyle:
Working on #27 

cc @hzoo @kittens @cpojer

### DIFF
--- a/lib/commands/updated/index.js
+++ b/lib/commands/updated/index.js
@@ -1,12 +1,23 @@
 var checkUpdatedPackages = require("../../utils/checkUpdatedPackages");
-var logger = require("../../utils/logger");
+var findAllDepdendents   = require("../../utils/findAllDependents");
+var packageUtils         = require("../../utils/packageUtils");
+var logger               = require("../../utils/logger");
+
+function format(packages) {
+  return packages.map(function(pkg) {
+    return "- " + pkg.name;
+  }).join("\n")
+}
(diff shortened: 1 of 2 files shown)


thejameskyle:
Outputs:

```
➜  __lerna git:(master) lerna updated
Lerna updated v1.1.1
Checking for updated packages...
Changed Packages:
- lerna-test-package-2
Dependent Packages:
- lerna-test-package-1
```

hzoo:
Thoughts: Do we care about listing out the specific dependents for each package or just listing them all out?

thejameskyle:
I'm not sure what you mean for either question

hzoo:
Sorry, so for the first question I mean the way we format the dependent packages could be like

```js
Changed Packages
- pkg1
  - pkg-dep1
  - pkg-dep2
- pkg2
  - pkg2-dep1
```

thejameskyle:
I'm not sure how to format that because you can have packages with more than one parent changed

hzoo:
Hmm yeah probably better to do the simple thing. Seeing a dep graph would be would nice for our monorepo though. I did see someone posted [one](https://graphcommons.com/graphs/a67121c8-dbd8-43f4-a8bf-dd2e725c4278) on twitter although it looks deleted

thejameskyle:
Fixed in 2.0.0-beta.4

PR #100, opened 2016-04-11
add FAQ with our first FAQ!

cdaringe:
# problem statement
- i didn't understand why my packages weren't being linked as claimed by the lerna docs

# investigation
- per investigation, linking does not happen in the traditional sense, and there is a caret range requirement for the linking.

# solution
- clarify in docs how this process occurs

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,3 +67,7 @@ Lerna projects operate on a single version line. The version is kept in the file
 at the root of your project. When you run `lerna publish`, if a module has been updated
 since the last time a release was made, it will be updated to the new version you're
 releasing. This means that you only publish a new version of a package when you need to.
+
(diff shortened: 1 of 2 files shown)


Chrisui:
Purely out of interest - why isn't npm linking used?

thejameskyle:
@Chrisui npm would break, but we might be able to make it happen with https://github.com/kittens/lerna/pull/101 in place now.

cdaringe:
still debating, folks?  just curious :)

hzoo:
Updated the readme, and added https://github.com/lerna/lerna#how-bootstrap-works - which I think covers this?

cdaringe:
yep

PR #110, opened 2016-04-20
Only push current branch on publish

### LINKED ISSUE #88: Git push only current branch
We're currently pushing every branch on lerna publish which breaks the publishing process if you have branches that are rejected.

```
Publishing tagged packages...
Setting latest npm tags...
Removing temporary npm dist-tag
Adding latest npm dist-tag
npm WARN dist-tag add latest is already set to version 1.0.0
To ssh://git@xxxx.git
   aafeff6..f74cc37  master -> master
 ! [rejected]        form -> form (non-fast-forward)
```

thejameskyle:
Fixes #88

### DIFF
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -40,7 +40,7 @@ export default class GitUtilities {
 
   @logger.logifySync
   static pushWithTags(tags) {
-    ChildProcessUtilities.execSync("git push");
+    ChildProcessUtilities.execSync("git push origin " + GitUtilities.getCurrentBranch());
     ChildProcessUtilities.execSync("git push origin " + tags.join(" "));
   }
 
@@ -67,5 +67,10 @@ export default class GitUtilities {
   @logger.logifySync
   static checkoutChanges(changes) {
     ChildProcessUtilities.execSync("git checkout -- " + changes);
+  }
+
+  @logger.logifySync
+  static getCurrentBranch() {
+    return ChildProcessUtilities.execSync("git symbolic-ref --short HEAD");
   }
 }
--- a/test/PublishCommand.js
+++ b/test/PublishCommand.js
@@ -166,7 +166,7 @@ describe("PublishCommand", () => {
           { args: ["npm dist-tag rm package-4 lerna-temp"] },
           { args: ["npm dist-tag add package-4@1.1.0 latest"] },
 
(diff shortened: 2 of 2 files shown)


test/PublishCommand.js
```diff
@@ -71,7 +71,8 @@ describe("PublishCommand", () => {
           { args: ["npm dist-tag rm package-4 lerna-temp"] },
           { args: ["npm dist-tag add package-4@1.0.1 latest"] },
 
-          { args: ["git push"] },
+          { args: ["git symbolic-ref --short HEAD"], returns: "master", returns: "master" },
```
hzoo: I see - there's 2 `returns: "master"` and fails lint

src/GitUtilities.js
```diff
@@ -40,7 +40,7 @@ export default class GitUtilities {
 
   @logger.logifySync
   static pushWithTags(tags) {
-    ChildProcessUtilities.execSync("git push");
+    ChildProcessUtilities.execSync("git push origin " + GitUtilities.getCurrentBranch());
```
hzoo: Not a bad assumption, but we assume the branch is always `origin` in these 2 lines - I guess that could go in config (https://github.com/kittens/lerna/issues/42)?

test/PublishCommand.js
thejameskyle: :expressionless:

test/PublishCommand.js
```diff
@@ -165,7 +166,8 @@ describe("PublishCommand", () => {
           { args: ["npm dist-tag rm package-4 lerna-temp"] },
           { args: ["npm dist-tag add package-4@1.1.0 latest"] },
 
-          { args: ["git push"] },
+          { args: ["git symbolic-ref --short HEAD"], returns: "master", returns: "master" },
```
hzoo: sorry 😢  same here

hzoo:
👍

src/GitUtilities.js
thejameskyle: Yeah, we could attempt to lookup what the tracked remote branch is:

```
$ repo git:(branch): git for-each-ref --format='%(upstream:short)' $(git symbolic-ref -q HEAD)
> origin/master
```

I don't know how reliable that would be though

hzoo: I think it's cool we can just note it in the help command, and the readme

PR #111, opened 2016-04-20
Add --skip-git flag

### LINKED ISSUE #106: Option to not push/tag a release.
On Jest we currently have a more involved release and push process. We are publishing more than just `packages/` and we are landing through our Facebook sync scripts. An option to avoid commit and pushing a diff would be great.

thejameskyle:
Resolves #106

This causes publish not to `git add/commit/tag/push` at all

cc @cpojer

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -75,7 +75,9 @@ export default class PublishCommand extends Command {
       }
 
       this.updateUpdatedPackages();
-      this.commitAndtagUpdates();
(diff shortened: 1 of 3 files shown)


src/commands/PublishCommand.js
```diff
@@ -262,7 +266,7 @@ export default class PublishCommand extends Command {
     });
   }
 
-  commitAndtagUpdates() {
+  commitAndTagUpdates() {
```
hzoo: 👍

PR #116, opened 2016-04-27
Replace FORCE_VERSION with --force-version flag

thejameskyle:
I don't really like this name. Thoughts @hzoo?

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -21,7 +21,8 @@ var cli = meow([
   "  --independent, -i    Version packages independently",
   "  --canary, -c         Publish packages after every successful merge using the sha as part of the tag",
   "  --skip-git           Skip commiting, tagging, and pushing git changes (only affects publish)",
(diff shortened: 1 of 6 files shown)


thejameskyle:
@cpojer suggested `--force-publish` which I like

hzoo:
Yeah that sounds good - kind of like force push

bin/lerna.js
```diff
@@ -20,7 +20,8 @@ var cli = meow([
   "Options:",
   "  --independent, -i  Version packages independently",
   "  --canary, -c       Publish packages after every successful merge using the sha as part of the tag",
-  "  --skip-git         Skip commiting, tagging, and pushing git changes (only affects publish)"
+  "  --skip-git         Skip commiting, tagging, and pushing git changes (only affects publish)",
+  "  --force-version    Force a new version for the specified packages (comma-seperated) or all packages using *"
```
hzoo: I guess we can update this too - Force a new publish.

Could also explain that it bypasses the updated/git diff check if that's helpful

PR #118, opened 2016-05-02
Add publishConfig and ignore option for publishing

thejameskyle:
Basically:
```
$ git diff --name-only
packages/package-a/src/source-file.js
packages/package-b/test/test-file.js
```

Then use `minimatch` to filter out files in the diff that `lerna.json#publishConfig.ignore` is ignoring


thejameskyle:
I need to add tests to this

hzoo:
👍

PR #120, opened 2016-05-06
Support jsnext:main in packages

nolanlawson:
Currently Lerna cannot be used with tools like [Rollup](https://github.com/rollup/rollup), insofar as it depends on the `jsnext:main` field to denote exported ES6 modules, and Lerna writes its own `package.json` for each dependency and omits the `jsnext:main` field.

For my own use case (PouchDB), I would like to build up modules that have both a `main` and a `jsnext:main`. That way, users can choose either CommonJS or ES modules, but internally we'll use `jsnext:main` so that we can get the benefits of Rollup's super-efficient bundler.

This fix adds support for `jsnext:main`. The idea is that, if an existing `package.json` exposes a `jsnext:main` field, then the Lerna-created `package.json` will also have one, with a `module.js` that looks like this:

```js
import _ from 'package';
export default _;
```
Packages that don't have a `jsnext:main` will have the same behavior as before.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -88,7 +88,7 @@ export default class BootstrapCommand extends Command {
 
     const packageJsonFileContents = JSON.stringify(newPkgJson, null, "  ");
     const indexJsFileContents = "module.exports = require(" + JSON.stringify(src) + ");";
-    const moduleJsFileContents = "import _ from " + JSON.stringify(src) + ";\nexport default _;";
+    const moduleJsFileContents = "export * from " + JSON.stringify(src) + ";";
 
     async.parallel([
       callback => {
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -73,21 +73,32 @@ export default class BootstrapCommand extends Command {
     const srcPackageJsonLocation = path.join(src, "package.json");
     const destPackageJsonLocation = path.join(dest, "package.json");
     const destIndexJsLocation = path.join(dest, "index.js");
+    const destModuleJsLocation = path.join(dest, "module.js");
+    const pkg = require(srcPackageJsonLocation);
 
-    const packageJsonFileContents = JSON.stringify({
+    const newPkgJson = {
       name: name,
-      version: require(srcPackageJsonLocation).version
-    }, null, "  ");
+      version: pkg.version,
+      main: "index.js"
+    };
 
+    if (pkg["jsnext:main"]) { // support jsnext:main convention for ES modules
+      newPkgJson["jsnext:main"] = "module.js";
+    }
+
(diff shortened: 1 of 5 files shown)


hzoo:
lgtm if we want to support jsnext:main (why not)

nolanlawson:
Hm, it just occurred to me that this only works for `default` exports, not non-`default` exports. 😕

thejameskyle:
Is `jsnext:main` a rollup-specific thing?

src/commands/BootstrapCommand.js
```diff
@@ -73,21 +73,32 @@ export default class BootstrapCommand extends Command {
...
-      }
+    if (pkg["jsnext:main"]) { // support jsnext:main convention for ES modules
+      newPkgJson["jsnext:main"] = "module.js";
+    }
 
-      FileSystemUtilities.writeFile(destIndexJsLocation, indexJsFileContents, callback);
-    });
+    const packageJsonFileContents = JSON.stringify(newPkgJson, null, "  ");
+    const indexJsFileContents = "module.exports = require(" + JSON.stringify(src) + ");";
+    const moduleJsFileContents = "import _ from " + JSON.stringify(src) + ";\nexport default _;";
```
thejameskyle: `export * from src` is the correct way to expose everything

nolanlawson: Thanks! I'll add a fix for that.

nolanlawson:
`jsnext:main` is currently only supported by Rollup and System.JS, yes. Axel Rauschmayer has a recent [a blog post where he mentions it](http://www.2ality.com/2016/03/multi-platform-npm-packages.html) among other solutions to the ES6 problem.

nolanlawson:
Done! ^

thejameskyle:
I wonder if this is a good opportunity to create better hooks into lerna for 3rd-party integration stuff like this. Rather than trying to build in support for a bunch of different things, we can build in support for extending Lerna.

nolanlawson:
At some point it would be nice to have support for ES6 modules, via `jsnext:main` or `.mjs` or whatever the Node community settles on (looks like `.mjs` for now). Maybe another option would be to allow a config in Lerna to either always expose CJS or always expose ES6.

thejameskyle:
Yeah, once the node community decides what they want lerna will definitely support it.

nolanlawson:
If you'd like to add a hook to support arbitrary injection of files, that would be great, although I can't imagine a very streamlined API for it. Maybe it's just best if I fork Lerna in the meantime, and then when Node figures out modules, Lerna can just go with the standard solution.

Closing for now. Please reopen if you wish.

hzoo:
I'm ok with doing this, so depends on james. - but if we want to do it as a plugin system yeah you'll need to wait

PR #123, opened 2016-05-09
style(typo): seperated -> separated

joscha:


thejameskyle:
Ugh I always spell this word wrong. Thanks

Issue #128, opened 2016-05-12
UpdatedPackagesCollector.isPackageDependentOf is super slow


thejameskyle:
https://github.com/kittens/lerna/blob/master/src/UpdatedPackagesCollector.js#L81-L91

I'm pretty sure this check is like `O(n!)` (in the worst case) because I didn't really bother optimizing it. For repos with tons of packages that have tons of cross-dependents this even causes a `maximum call stack exceeded` error.

A fast solution might be to memoize the checks. Although that doesn't necessarily fix the max call stack issue.

In order to solve this completely we need to switch to an iterative algorithm that doesn't check the same packages multiple times.

hzoo:
The recursive issue was fixed and there's a simple cache now

PR #129, opened 2016-05-13
Add cache for checking updated dependencies

hzoo:
https://github.com/lerna/lerna/issues/128

Probably not the best but it seems to work on Babel (where the max callstack error happened).

Examples:
babel-types -> all packages updated
babel -> babel only (no other dependencies)
stage-3 preset -> stage-0, stage-1, stage-2, stage-3

### DIFF
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -79,19 +79,40 @@ export default class UpdatedPackagesCollector {
   }
 
   isPackageDependentOf(packageName, dependency) {
-    var dependencies = this.packageGraph.get(packageName).dependencies;
+    if (!this.cache[packageName]) {
+      this.cache[packageName] = {};
+    }
+
+    if (this.cache[packageName][dependency] === "dependent") {
+      return true;
+    } else if (this.cache[packageName][dependency] === "visited") {
+      return false;
+    }
+
+    let dependencies = this.packageGraph.get(packageName).dependencies;
 
     if (dependencies.indexOf(dependency) > -1) {
+      this.cache[packageName][dependency] = "dependent";
       return true;
     }
 
-    return !!find(dependencies, dep => {
-      return this.isPackageDependentOf(dep, dependency);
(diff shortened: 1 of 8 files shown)


src/UpdatedPackagesCollector.js
```diff
@@ -79,19 +79,40 @@ export default class UpdatedPackagesCollector {
...
+    }
+
+    let dependencies = this.packageGraph.get(packageName).dependencies;
 
     if (dependencies.indexOf(dependency) > -1) {
+      this.cache[packageName][dependency] = "dependent";
       return true;
     }
 
-    return !!find(dependencies, dep => {
```
hzoo: I think find was previously only getting the first dependency?

src/UpdatedPackagesCollector.js
```diff
@@ -79,19 +79,40 @@ export default class UpdatedPackagesCollector {
...
-      return this.isPackageDependentOf(dep, dependency);
+    this.cache[packageName][dependency] = "visited";
```
hzoo: I think it was the case for if a package A depends on package B, and then package B depends on package A

src/UpdatedPackagesCollector.js
thejameskyle: Well it would stop as soon as it found a path where A depended on B. But I guess that doesn't necessarily speed it up because every path is going to be checked anyways.

There has to be a way of representing this data that makes it easy to check every path by simply folding dependencies over. But caching works well enough.

src/UpdatedPackagesCollector.js
thejameskyle: I don't think I previously handled cycles

thejameskyle: Should we add a test case for this?

hzoo: Ok added some tests? Just copied the other one and made 2 packages depend on each other (which causes the issue)

thejameskyle:
👍

Issue #127, opened 2016-05-12
Add --skip-publish option, like --skip-git


eirslett:
The idea would be that you could run `lerna publish --skip-publish` (a somewhat strange command if you read it like that) and it would do the git update, git tag, git push etc., but not `npm publish` the packages.

The motivation is that you could then setup Travis CI to do the releasing*. Travis already supports [vanilla npm publish](https://docs.travis-ci.com/user/deployment/npm/), but one could easily write a lerna publisher, that ran `lerna publish --skip-git` instead of `npm publish`.

So `.travis.yml` would look like this:

```yml
# build config etc. first

# deploy config
deploy:
  provider: lerna
  email: email@example.com
  api_key:
    secure: XXX
  on:
    tags: true
    repo: some/repo
```

It would also require an update to [https://github.com/travis-ci/dpl](Dpl), which is Travis' deployment tool, so that it could support lerna.

Otherwise, I have to tell you, Lerna is great! I use it for publishing [zipkin-js](https://github.com/openzipkin/zipkin-js) and the user experience is very smooth.

(*I hate releasing stuff from my Windows computer, because it does weird things to file permissions, and also I tend to mess up my local repo, so it's not a clean build like on Travis. And of course, I forget to run the tests too)

hzoo:
Since `lerna publish --skip-publish` does read weird, it might be better to split it into a separate command that just updates git

thejameskyle:
Running `lerna publish --skip-publish` and then `lerna publish --skip-git` doesn't make a lot of sense. You'd effectively be bumping versions twice, and the only solution otherwise would be for one to pick up where the last one left off which would be a weird behavior to introduce for the common-case.

It sounds like what you really want is a way to specify the next versions to be published by lerna inside CI without having to follow the prompts. sort of like:

```sh
# non-independent:
$ lerna publish --specify-version minor
# independent:
$ lerna publish --specify-version package-1:major package-2:minor package-3:1.2.3
```

How you choose to get these versions into CI is a bit outside of the scope of Lerna, but you could easily write a script to do it.

eirslett:
@thejameskyle I think you may have misunderstood me a bit, this is what I want to achieve:

1) Run `lerna [some-command]` (maybe the command would be called prepare or prepare-publish?) locally/manually/from a developer machine, answer the prompts to choose a new version, and it will update package.json files, commit the changes, tag the commit and push it, like this:

[...]

PR #130, opened 2016-05-14
Add 'lerna exec' command

eirslett:
As opposed to 'lerna run', which runs
scripts defined in package.json, exec will
let you execute an inbuilt npm command directly
in each package, e.g.

- `lerna exec rebuild`
- `lerna exec publish`
- `lerna exec upgrade`

This is response to #127 and implements one of the solutions @thejameskyle suggests. What do you think?

### DIFF
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -1,4 +1,4 @@
-import NpmUtilities from "../NpmUtilities";
+import ChildProcessUtilities from "../ChildProcessUtilities";
 import Command from "../Command";
 import async from "async";
 
@@ -8,7 +8,7 @@ export default class ExecCommand extends Command {
     this.args = this.input.slice(1);
 
     if (!this.command) {
-      callback(new Error("You must specify which npm command to run."));
+      callback(new Error("You must specify which command to run."));
       return;
     }
 
@@ -22,9 +22,9 @@ export default class ExecCommand extends Command {
   }
 
   runCommandInPackage(pkg, callback) {
-    NpmUtilities.execInDir(this.command, this.args, pkg.location, (err, stdout) => {
+    ChildProcessUtilities.exec(this.command, { cwd: pkg.location }, (err, stdout) => {
       if (err) {
-        this.logger.error(`Errored while running npm command '${this.command}' in '${pkg.name}'`, err);
+        this.logger.error(`Errored while running command '${this.command}' in '${pkg.name}'`, err);
       } else {
         this.logger.info(stdout);
       }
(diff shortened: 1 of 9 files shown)


thejameskyle:
I meant `exec` as a way of running any command, not just npm commands.

For example:

```sh
$ lerna exec pwd
/path/to/repo/packages/package-1
/path/to/repo/packages/package-2
/path/to/repo/packages/package-3
```

eirslett:
Ok, I updated the PR now! "lerna exec pwd" works as intended. It's still two commits, but they can easily be squashed on merge if that's better!

src/commands/ExecCommand.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  execute(callback) {
+    async.parallelLimit(this.packages.map(pkg => cb => {
+      this.runCommandInPackage(pkg, cb);
+    }), 4, callback);
+  }
+
+  runCommandInPackage(pkg, callback) {
+    ChildProcessUtilities.exec(this.command, { cwd: pkg.location }, (err, stdout) => {
```
thejameskyle: This should be handling more than just a single argument. It needs to handle things like:

```
lerna exec ls -la subdirectory 
```

It'd probably be better if you exposed the entire `process.argv` on `this.argv` inside commands and then trim the `lerna exec` from the beginning simply passing the rest through.

thejameskyle:
Thinking more about this could you make it similar to how you pass flags to `npm run` commands in that you need `--` before it?

```sh
$ lerna exec -- ls -ls subdirectory
```

This will make it easier to add flags in the future:

```sh
$ lerna exec --debug -- ls -la subdirectory

File: src/commands/ExecCommand.js [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+import Command from "../Command";
+import async from "async";
+
+export default class ExecCommand extends Command {
+  initialize(callback) {
+    this.command = this.input[0];
+    this.args = this.input.slice(1);
+
+    if (!this.command) {
+      callback(new Error("You must specify which npm command to run."));
```
thejameskyle: This error message is no longer valid

eirslett: fixed!

joscha:
This is awesome, is there anything else that needs to be done to bring this in a mergeable state? Like the `--` functionality and/or a test with a command that has a few arguments?

thejameskyle:
Yeah, I would like to see both of those things

joscha:
@thejameskyle you can find the feature in #151 including a couple more tests and a fix for the test in this PR (it doesn't actually work properly).

eirslett:
Thanks @joscha , I'll close this in favor of #151 .

PR #131, opened 2016-05-14
lerna init should add lerna as a devDependency

hzoo:
I think we want this as a devDep right? Also do we want to add `^` in front as well?

Also, this means we don't want users to use it as a global npm package?

cc @thejameskyle

### DIFF
--- a/src/commands/InitCommand.js
+++ b/src/commands/InitCommand.js
@@ -36,7 +36,7 @@ export default class InitCommand extends Command {
     // if (!packageJson.private) packageJson.private = true;
     if (!packageJson.dependencies) packageJson.dependencies = {};
 
-    objectAssignSorted(packageJson.dependencies, {
+    objectAssignSorted(packageJson.devDependencies, {
       lerna: this.lernaVersion
     });
 


thejameskyle:
How does `^` interact with `x.x.x-beta.x`?

And my plan in the future is to make the global CLI use the local version if it exists which would be the preferred way.

hzoo:
If you do `^x.x.x-beta` it should update all betas. http://semver.npmjs.com/

screen shot 2016-05-15 at 9 55 14 pm

thejameskyle:
I don't know about the carrot range because I want the version in `lerna.json` to always match the version in `devDependencies`. We can either lock it down to the specific version or open them both up using `semver.satisfies`. What do you think?

thejameskyle:
I'm going to merge this as is for now

hzoo:
Ok lets go with locking it down atm then and reevaluate `^` later. Similar to [flowconfig](http://flowtype.org/docs/advanced-configuration.html#version) although in that case if you don't specify a version it allows any version and otherwise errors if it doesn't match

PR #135, opened 2016-05-16
Implements #134 - add a --confirm flag

jpnelson:
This adds a `--confirm` flag for usage in CI that automatically answers the publish confirmation prompt.

```
lerna publish --canary --confirm

Lerna v2.0.0-beta.9
Independent Versioning Mode
Publishing canary build
Checking for updated packages...

Changes:
- component-a: 0.0.3 => 0.0.3-canary.2c247805
- component-b: 0.0.2 => 0.0.2-canary.2c247805
- component-c: 0.0.2 => 0.0.2-canary.2c247805

Assuming confirmation.

Publishing packages to npm...
Reseting git state
Successfully published:
 - component-a@0.0.3-canary.2c247805
 - component-b@0.0.2-canary.2c247805
 - component-c@0.0.2-canary.2c247805
```

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -22,7 +22,8 @@ var cli = meow([
   "  --canary, -c         Publish packages after every successful merge using the sha as part of the tag",
   "  --skip-git           Skip commiting, tagging, and pushing git changes (only affects publish)",
   "  --npm-tag [tagname]  Publish packages with the specified npm dist-tag",
-  "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)"
+  "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)",
+  "  --confirm            Skip all confirmation prompts"
 ], {
   alias: {
     independent: "i",
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -214,9 +214,14 @@ export default class PublishCommand extends Command {
     }).join("\n"));
     this.logger.newLine();
 
-    PromptUtilities.confirm("Are you sure you want to publish the above changes?", confirm => {
-      callback(null, confirm);
-    });
+    if (!this.flags.confirm) {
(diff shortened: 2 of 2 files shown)


thejameskyle:
So we can do this as a one off, or we can modify `PromptUtilities.confirm` to accept the cli flags, and if `--confirm` or `--yes` or whatever it's named is passed we can just continue on.

Depends on if we want to implement this globally. @hzoo what do you think?

joscha:
Either way would work for us I think - a generic one might make more sense, however then we need to find a way to propagate CLI flags to the `PromptUtilities`.

joscha:
@thejameskyle could you please tell us if this is in a mergeable state or if you would want us to move the flag detection into the `PromptUtilities` logic in order for this to be integrated?

thejameskyle:
It's in a mergeable state, I just want to think through the usage of prompts that aren't yes/no. Like the independent versioning. Maybe it's acceptable to just ignore that...

thejameskyle:
Actually could you rename `--confirm` to `--yes` to match stuff like npm's `--yes` flag or the `yes` bash command?

joscha:
@thejameskyle f49ff60 has the changes

bin/lerna.js
```diff
@@ -22,7 +22,8 @@ var cli = meow([
   "  --canary, -c         Publish packages after every successful merge using the sha as part of the tag",
   "  --skip-git           Skip commiting, tagging, and pushing git changes (only affects publish)",
   "  --npm-tag [tagname]  Publish packages with the specified npm dist-tag",
-  "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)"
+  "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)",
+  "  --yes               Skip all confirmation prompts"
```
hzoo: maybe a space to align Skip?

hzoo:
I think we can add a test for this?

joscha:
@hzoo I made the fix you asked for and added a test in #169

Issue #143, opened 2016-05-18
Allow running command for a specific package


joscha:
Currently our setup uses `lerna run ...` for anything related to all packages (like running the tests for example). However in development, we'd like to be able to run a command for a single specified package.
Reason for this is, that our main repository contains the dependencies to some build tools (for example Karma) that we'd like to re-use.
For example one of our packages looks like this:
```json
...
scripts: {
    "test": "karma start ../../karma.conf.js --single-run",
    "test/watch": "../../node_modules/.bin/karma start ../../karma.conf.js",
    "prepublish": "webpack"
}
...
```

whereas `lerna run test` and `lerna run prepublish` work fine for all packages, the `test/watch` script would only make sense for a single package. However using the binary short form is not possible, because the package itself doesn't actually depend on karma, only our base package does, which means we need to link to the binary ourselves (`../../node_modules/.bin/...`).
Ideally we would be able to run something like

```console
lerna run-single my-component test/watch
```

Is this something you would accept a pull request for?

thejameskyle:
Hmm, I'd almost rather extending the existing `lerna run` command to accept some way of filtering packages.

```sh
$ lerna run --matches-packages-idk-what-to-name-this-flag component-*
```

joscha:
@thejameskyle I thought about that, the downside is that currently anything after the first argument to `lerna` is treated as the target script name - adding another flag in between would mean that we potentially some script out there might break, but I guess as 2.0 is still in the flow that would be fine.

maybe `lerna run --restrict-to component-* my-script`? I'd be happy to provide a PR for this feature.

thejameskyle:
> the downside is that currently anything after the first argument to lerna is treated as the target script name

I'm pretty sure `meow` takes care of that for us. But idk

And yeah, I'd be happy to support this in some form. Lets see if you face any challenges when making a PR and work it out there.

joscha:
@thejameskyle I opened #145, please let me know if there are any changes you'd like me to make in order to integrate it.

PR #145, opened 2016-05-18
feat: --restrict-to={glob} for "run" command

joscha:
This adds a flag `--restrict-to` that takes a glob to restrict a command to be run to the packages matched by the glob.

implements #143

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -22,6 +22,7 @@ var cli = meow([
   "  --canary, -c         Publish packages after every successful merge using the sha as part of the tag",
   "  --skip-git           Skip commiting, tagging, and pushing git changes (only affects publish)",
   "  --npm-tag [tagname]  Publish packages with the specified npm dist-tag",
+  "  --restrict-to [package glob] Restricts a command to run only packages matching the given glob (Works only in combination with the 'run' command).",
   "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)"
 ], {
   alias: {
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -15,15 +15,15 @@ export default class RunCommand extends Command {
 
     let packagesToRunCommandIn = this.packages;
 
-    if (this.flags.restrictTo) {
+    if (typeof this.flags.restrictTo !== "undefined") {
       this.logger.info(`Restricting to packages that match '${this.flags.restrictTo}'`);
       packagesToRunCommandIn = packagesToRunCommandIn
         .filter(pkg => minimatch(pkg.name, this.flags.restrictTo));
-    }
 
-    if (!packagesToRunCommandIn.length) {
-      callback(new Error(`No packages found that match '${this.flags.restrictTo}'`));
-      return;
+      if (!packagesToRunCommandIn.length) {
+        callback(new Error(`No packages found that match '${this.flags.restrictTo}'`));
+        return;
+      }
     }
 
     this.packagesWithScript = packagesToRunCommandIn
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -16,9 +16,9 @@ export default class RunCommand extends Command {
     let packagesToRunCommandIn = this.packages;
 
     if (this.flags.restrictTo) {
-        this.logger.info(`Restricting to packages that match '${this.flags.restrictTo}'`);
-        packagesToRunCommandIn = packagesToRunCommandIn
-            .filter(pkg => minimatch(pkg.name, this.flags.restrictTo));
+      this.logger.info(`Restricting to packages that match '${this.flags.restrictTo}'`);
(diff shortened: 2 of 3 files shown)


thejameskyle:
@joscha Can you rebase this onto master?

joscha:
@thejameskyle just did, sorry about that.

src/commands/RunCommand.js
```diff
@@ -12,7 +13,20 @@ export default class RunCommand extends Command {
       return;
     }
 
-    this.packagesWithScript = this.packages
+    let packagesToRunCommandIn = this.packages;
+
+    if (this.flags.restrictTo && `${this.flags.restrictTo}`.length > 0) {
```
thejameskyle: I think this can just be: `if (this.flags.restrictTo) {`

test/RunCommand.js
```diff
@@ -33,4 +33,43 @@ describe("RunCommand", () => {
...
+        callback();
+    });
+
+    runCommand.runCommand(exitWithCode(0, () => {
+        assert.deepEqual(ranInPackages, [
+            "package-1",
+            "package-3"
+        ]);
+        done();
+    }));
```
thejameskyle: Can you fix the indentation in this file.

@hzoo does jscs do any kind of auto-correction of code style? Or is there another tool that can do that for us?

src/commands/RunCommand.js
joscha: I thought about that, but then `lerna run test --restrict-to "` would not run anything - if that is the expected behaviour I am happy to change it.

bin/lerna.js
```diff
@@ -22,6 +22,7 @@ var cli = meow([
   "  --canary, -c         Publish packages after every successful merge using the sha as part of the tag",
   "  --skip-git           Skip commiting, tagging, and pushing git changes (only affects publish)",
   "  --npm-tag [tagname]  Publish packages with the specified npm dist-tag",
+  "  --restrict-to [package glob] Restricts a command to run only packages matching the given glob (Works only in combination with the 'run' command).",
```
thejameskyle: I'm not sure about the name `--restrict-to`, any other ideas?

---

Also (separate from this PR), we really should fine a better way of documenting flags within the CLI help.

src/commands/RunCommand.js
joscha: O wait, never mind, I'll change it 👯

thejameskyle: I think it's better that it doesn't run anything than trying to run everything

test/RunCommand.js
joscha: we could add an `.editorconfig` file, or you can add the rules to `eslint`.

joscha: indentation fixed in 38b101b

src/commands/RunCommand.js
joscha: simplified in f137619

bin/lerna.js
joscha: * `--include-packages`
* `--match`
* `--only`

@jpnelson any other ideas?

test/RunCommand.js
thejameskyle: Yeah, I just don't like style linting, I just want tools that automatically and silently fix it for you.

joscha: with editorconfig if the editor respected it, it would happen automatically. Alternatively I guess we could have a pre-commit hook that runs eslint with the `--fix` flag to do so, however that is quite a bit of overhead I think.

joscha: I opened #146 to add an eslint rule for a start.

src/commands/RunCommand.js
joscha: just saw your comment @thejameskyle - if you want to not run anything in case an empty restriction is passed, then I think I need to add another check. Right now it will behave as if no restriction was given when passing an empty one (which I think might feel more natural).

thejameskyle: I would consider `--restrict-to` without anything specified to be invalid input with an error that says so

joscha:
@thejameskyle I made the changes for not running anything if `--restrict-to` is given but empty, they are in 36f22b5. Do you think we can brainstorm about the name again and wrap this up? I personally like `--run-in` as it conveys that its something todo with the `run` command, however if we want to re-use the flag for other commands, `--restrict-to` or `--only-in` might be better candidates.

test/RunCommand.js
hzoo: Yeah `jscs` has autofix for all it's whitespace rules, and eslint supports it for some rules now as well.

bin/lerna.js
jpnelson: no better ones, `--run-in` makes sense to me though

joscha: I changed it to `--scope` in #152

joscha:
#152 should be favored to this one.

thejameskyle:
Alright, lets move to #152 then

PR #146, opened 2016-05-18
chore: enforce proper indentation

joscha:
This adds an eslint rule to make sure we use proper indentation.

cc @thejameskyle

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*]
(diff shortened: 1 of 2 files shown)


thejameskyle:
You want to add an `.editorconfig` file too while you're at it?

joscha:
will do, one sec.

thejameskyle:
Thanks!

PR #148, opened 2016-05-18
Add option to ignore packages when bootstrapping (fixes #117)

### LINKED ISSUE #117: Config option for ignoring packages to bootstrap
https://github.com/babel/babel/pull/3438#discussion_r57091960

rygine:
* Added `bootstrap-ignore` flag to CLI
* Check `bootstrapConfig.ignore` key in `lerna.json`

The CLI flag will override the `bootstrapConfig` option in `lerna.json`. Both the flag and the option allow one or more packages to be specified. The CLI flag accepts comma-separated values while the `bootstrapConfig` option can also accept an array.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -5,6 +5,7 @@ import semver from "semver";
 import async from "async";
 import find from "lodash.find";
 import path from "path";
+import minimatch from "minimatch";
 
 export default class BootstrapCommand extends Command {
   initialize(callback) {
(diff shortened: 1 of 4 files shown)


rygine:
Since `minmatch` is already included, the glob matching came for free and can easily be removed. I think another unit test should be added to include usage of `lerna.json` and `bootstrapConfig`. I didn't see one for the publish command. Thoughts?

src/commands/BootstrapCommand.js
```diff
@@ -27,7 +40,7 @@ export default class BootstrapCommand extends Command {
     this.progressBar.init(this.packages.length);
     this.logger.info("Linking all dependencies");
 
-    async.parallelLimit(this.packages.map(pkg => done => {
+    async.parallelLimit(this.packages.filter(pkg => !pkg.ignored).map(pkg => done => {
```
thejameskyle: I would rather structure this differently where we don't pass down ignore packages into the package graph and instead we do the check here if the package has been ignored.

This should remove a lot of the complexity of this feature.

rygine: In other words, don't add an ignored package to the packages array in `PackageUtilities.getPackages`?

thejameskyle: > I haven't really thought through the implementation. However, removing it from the packageGraph means you just never have access to that information which could cause problems down the road when we try to implement something else so I guess I would prefer that it didn't.

My comment here was about completely removing packages that have been ignored from the packageGraph. Which you have not done here.

I'm more concerned about having the logic for checking for ignored packages inside `PackageUtilities.getPackages` and `Package.ignored`. Those functions shouldn't have to be modified in order to implement this.

rygine: Ahh, that makes sense. Will update.

rygine:
@thejameskyle updated. looks much cleaner. thanks for the suggestions. if this works for you, i'll add a few more tests to cover invalid formats of the flag/config value and using a config value from `lerna.json`.

rygine:
closing this PR from my personal account and re-opening from company account. sorry for the confusion.

Issue #150, opened 2016-05-19
npm run uses exec, rather than an async exec


jpnelson:
Would be nice if it used child_process.spawn, and we'd get streaming log output 🐉 

https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L33

PR #151, opened 2016-05-19
Feature/lerna exec

joscha:
This implements a proper `lerna exec` command that allows you to both pass flags to the lerna runtime and run commands in each package with multiple parameters, e.g.:
`lerna exec --debug -- ls -la node_modules`

closes #130 

cc @thejameskyle 
Also, after this is merged I'd like to enhance #145 to also change the target packages of `lerna exec`.

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -1,4 +1,5 @@
 import child from "child_process";
+import objectAssign from "object-assign";
 
 export default class ChildProcessUtilities {
   static exec(command, opts, callback) {
@@ -17,9 +18,9 @@ export default class ChildProcessUtilities {
     }).trim();
   }
 
-  static spawn(command, args, callback) {
-    child.spawn(command, args, {
+  static spawn(command, args, opts, callback) {
+    child.spawn(command, args, objectAssign({
       stdio: "inherit"
-    }).on("close", callback);
+    }, opts)).on("close", callback);
   }
 }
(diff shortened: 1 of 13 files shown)


src/ChildProcessUtilities.js
```diff
@@ -1,4 +1,5 @@
 import child from "child_process";
+import objectAssign from "object-assign";
```
eirslett: Isn't it possible to use Object.assign directly? (I'm not sure whether babel polyfills it or not, to be honest)

eirslett:
Where is the logic that handles everything before `--` as flags to lerna, and everything after as the actual command?

src/ChildProcessUtilities.js
joscha: I guess it depends on the runtime - `object-assign` was available already, so I just used it.

joscha:
@eirslett it seems to be handled by `meow` automatically. Everything before the `--` lands in `.flags`.

src/ChildProcessUtilities.js
hzoo: We aren't using Babel polyfill or Babel runtime atm so no but we could

thejameskyle: I need it to support older versions of node too

jpnelson:
This will also fix #150

joscha:
#152 should be favored to this one.

thejameskyle:
Moving to #152

PR #152, opened 2016-05-20
lerna exec, lerna run and scoping

### LINKED ISSUE #143: Allow running command for a specific package

### LINKED ISSUE #150: npm run uses exec, rather than an async exec

joscha:
This introduces an abstract command named `ScopedCommand` which adheres to a given CLI flag named `scope` that allows to scope the `lerna run` and `lerna exec` command (I pulled both #145 and #151) into this PR to a subset of packages by giving a glob that matches the package name.

After merging this, the following can be used:
* `lerna run --scope my-component-* prepublish`
* `lerna exec --scope my-component-* -- ls -la node_modules`

fixes #143
closes #145 
fixes #150 
closes #151

@thejameskyle PTAL
@eirslett @jpnelson FYI


src/ScopedCommand.js
```diff
@@ -0,0 +1,26 @@
+import Command from "./Command";
+import minimatch from "minimatch";
+
+export default class ScopedCommand extends Command {
```
thejameskyle: I don't know if inheritance is the best way to solve this. It should be just as easy to have a util that does this and it will get us away from using inheritance which complicates things

thejameskyle:
I do like the `--scope` name for this, nice suggestion.

src/ScopedCommand.js
joscha: I think it is the perfect problem to be solved by inheritance. Especially as then the concerns are also clearly separated for the tests as well, but I am happy to change it to whatever form you like it in order to get it integrated. Would you like me to change it to a helper?

joscha:
@thejameskyle is there anything I can do to help this getting integrated into lerna? Would you like to me to change the inheritance-based code into a helper?

thejameskyle:
Sorry, I've fallen behind on reviewing PRs because of other stuff going on right now.

I would personally prefer using a helper for created a scoped set of packages using mini-match. The only reason I'm using inheritance today is to create a shared interface for every command.

joscha:
@thejameskyle I changed the implementation to use a helper. Just on a side note, I think it was a bit cleaner with inheritance. Anyway, the changes are in 5dd9e9f - I hope that brings it in a mergeable state.

paulyoung:
I'd love to be able to use this functionality. Is there anything I can do to help here?

joscha:
@paulyoung I don't think so - last time I talked to @thejameskyle he was just preoccupied with other things - maybe that little github email nudge will bring him back 😉

thejameskyle:
> maybe that little github email nudge will bring him back

lol, I don't think you understand how many notifications I get 😜

I'm happy with this

hzoo:
Awesome! @joscha can you do a doc pr (or i'll do it later)

joscha:
@hzoo will do one - give me a few minutes!

joscha:
@hzoo its in #197 - I also found that the concurrency PR #163 was merged in the meantime, so I added a fix that re-uses concurrency in the code that was just merged.

acostaf:
Reopen please --scope is not working on "lerna": "^3.4.0"

npm run this "build:dep": "lerna run --scope common build"

I get lerna ERR! EFILTER No packages remain after filtering [ 'common' ]

my Lerna file   
"packages": [
    "common",
...
  ],

evocateur:
@acostaf please open a new issue with complete reproduction details, thanks

Issue #133, opened 2016-05-15
[Feature] Scripting Lerna


jamiebuilds:
Any good CLI tool can be composed with others, so I would like to focus on that with Lerna.

Stuff like:

Making commands like `lerna ls` or `lerna updated` have easy to parse output.

``` sh
$ lerna ls
package-1
package-2

# instead of:

$ lerna ls
Lerna v2.0.0-beta.6
Independent Versioning Mode
- package-1
- package-2
```

``` sh
$ lerna updated
package-1
package-2

# instead of

$ lerna updated
Lerna v2.0.0-beta.6
Independent Versioning Mode
Checking for updated packages...

- package-1
- package-2
```

> This will require moving debug information behind a `--debug` flag.

There's also `lerna exec` in #130 

``` sh
$ lerna exec pwd
/path/to/repo/packages/package-1
/path/to/repo/packages/package-2
```

I would also like to be able to expose different config-like information from lerna:

``` sh
$ lerna get repo.version
2.0.0
# or
independent
$ lerna get paths.root
/path/to/repo
$ lerna get paths.packages
/path/to/repo/packages
```

This is all my ideas for now, but I'm definitely interested in anything that makes Lerna more scriptable.

joscha:
Maybe we could adapt the git `--porcelain` flag for that. That way current scripts would continue working, just the output with --porcelain would allow composition and hide all debug output.

jpnelson:
I'd prefer @joscha 's suggestion if it's the difference between a major version and a minor version

thejameskyle:
What would break with current scripts? Are people using the output of Lerna already?

joscha:
can't speak for other people, but we do:
```shell
# Get all packages so we now which ones to iterate through
ALL_PACKAGES=$($LERNA_LOC exec pwd\
| sed 1,2d)
```

light24bulbs:
the --json flag covers this use case now, but unfortunately there isn't quite enough data in there.  It would be nice if `ls` also listed the package `path`

evocateur:
@light24bulbs A lot of scriptable options been [implemented](https://github.com/lerna/lerna/commit/2e204af3f10d30f0fca37891aed721ddf15cf667) [recently](https://github.com/lerna/lerna/commit/6ecdd83d14f3b9d0cac62cc68b4066ec554e79ce), but not released yet. Trying to finish #1091 first.

JamesHenry:
Hi Folks 👋 

[...]

PR #153, opened 2016-05-20
feat: add version command

joscha:
This allows users of lerna to print the used version:
`lerna --version` or `lerna -v`


thejameskyle:
Hmm, this conflicts with the `--version/-v` that gets created by meow to show the installed version of lerna that you are executing. But I could see this working with the `lerna get` command I described in https://github.com/lerna/lerna/issues/133

joscha:
would you like me to rewrite this into `lerna get version`?

PR #163, opened 2016-05-25
Add --concurrency flag

### LINKED ISSUE #162: Add a --concurrency flag
We need some way to specify number of threads for `lerna run` and the like (running into problems with git commands and locking that means we need to do stuff synchronously).

jpnelson:
Fixes #162

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -14,6 +14,7 @@ export default class Command {
     this.repository = new Repository();
     this.progressBar = progressBar;
     this.logger = logger;
+    this.concurrency = flags.concurrency === undefined ? 4 : flags.concurrency;
   }
 
   run() {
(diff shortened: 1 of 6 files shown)


src/Command.js
```diff
@@ -14,6 +14,7 @@ export default class Command {
     this.repository = new Repository();
     this.progressBar = progressBar;
     this.logger = logger;
+    this.concurrency = (!flags || flags.concurrency === undefined) ? 4 : flags.concurrency;
```
thejameskyle: > `(!flags ||`

I'm pretty sure flags is always an object

thejameskyle:
Nice, thanks

src/Command.js
jpnelson: It was broken in the tests actually – `flags` was null. I imagine that'd also be the case if you used commands programatically.

I also check for undefined specifically so that it's not `0 || 4`, so that if you specify `0` you get an error, rather than it just using `4`.

joscha: I think it is good that it would throw, however I think me might need to add a test to make sure it doesn't accidentally get refactored away the next time somebody reads over that code?!

joscha:
I updated the branch in #170 and also added a few tests around the empty/NaN/negative concurrency.

thejameskyle:
Merged in #170

PR #165, opened 2016-05-25
Fix --force-publish

thejameskyle:
😩


hzoo:
Was it just copied wrong? https://github.com/lerna/lerna/pull/116/commits/ad9dc98c4d544f69d4c3068220989413ff499ebf

Should we remove - https://github.com/lerna/lerna/blob/master/bin/lerna.js#L30

thejameskyle:
It was renamed improperly and I want to leave the alias since thats what it used to be called

PR #168, opened 2016-05-26
Add option to ignore packages when bootstrapping (fixes #117)

### LINKED ISSUE #117: Config option for ignoring packages to bootstrap

rygine:
* Added `ignore` flag to CLI
* Check `bootstrapConfig.ignore` key in `lerna.json`

The CLI flag will override the `bootstrapConfig` option in `lerna.json`. Both the flag and the option accept a glob pattern of package names to ignore.

this is the new PR from my company's github org

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -66,7 +66,13 @@ export default class PackageUtilities {
   */
   static filterPackages(packages, glob, negate = false) {
     if (typeof glob !== "undefined") {
-      packages = packages.filter(pkg => negate ^ minimatch(pkg.name, glob));
+      packages = packages.filter(pkg => {
+        if (negate) {
+          return !minimatch(pkg.name, glob)
+        } else {
+          return minimatch(pkg.name, glob);
+        }
+      });
 
       if (!packages.length) {
         throw new Error(`No packages found that match '${glob}'`);
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -60,12 +60,13 @@ export default class PackageUtilities {
   *
   * @param {!Array.<Package>} packages The packages to filter
   * @param {String} glob The glob to match the package name against
+  * @param {Boolean} negate Negate glob pattern matches
   * @return {Array.<Package>} The packages with a name matching the glob
   * @throws in case a given glob would produce an empty list of packages
   */
-  static filterPackages(packages, glob) {
+  static filterPackages(packages, glob, negate = false) {
     if (typeof glob !== "undefined") {
-      packages = packages.filter(pkg => minimatch(pkg.name, glob));
+      packages = packages.filter(pkg => negate ^ minimatch(pkg.name, glob));
 
       if (!packages.length) {
         throw new Error(`No packages found that match '${glob}'`);
--- a/README.md
+++ b/README.md
@@ -158,6 +158,8 @@ When run, this command will:
 Currently, what Lerna does to link internal dependencies is replace the
 `node_modules/package-x` with a link to the actual file in the repo.
 
+`lerna bootstrap` respects the `--ignore` flag (see below).
+
 #### How `bootstrap` works
 
 Let's use `babel` as an example.
@@ -425,7 +427,7 @@ $ lerna publish --concurrency 1
 
 #### --scope [glob]
 
-Allows to scope an command only affect a subset of packages.
+Scopes a command to a subset of packages.
 
 ```sh
 $ lerna exec --scope my-component -- ls -la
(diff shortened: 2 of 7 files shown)


joscha:
I think potentially we could use the `--scope` parameter introduced in #152 for this as well.

rygine:
@joscha hmm, `--scope` seems more suited to be a whitelist, where this would be a blacklist. maybe a more general `--ignore` flag would serve this purpose. i understand that negative matches could be applied to scope, but perhaps it would be confusing as it would not be readily apparent that `--scope` could be used that way.

joscha:
@rygine you are right, I guess it might be confusing ;)

thejameskyle:
Sorry about the confusion with flag names, I don't want it to turn into a mess, but haven't really thought about it holistically.

rygine:
the more i think about it, the more inclined i am make the flag name more generic. i think `ignore` would work well here. that way, it _could_ apply to other commands, such as `lerna run`. the `ignore` flag would be opposite of the `scope` flag, unless that's renamed to something else like `only`. it makes sense to have both of these options when running commands.

joscha:
I am happy to work on it to make it work for both black- and whitelisting.
It could be a generic `--include` and `--exclude` for example and the last
of the two options given trumps the one before.

Cheers,
Joscha

rygine:
@joscha: I've added another commit that renames the flag to `exclude` as per your suggestion. I think that makes sense. Each command will have to process the flag separately. Once your PR is done, I think I could use your `filterPackages` utility in place of what I currently have.

@thejameskyle: I remember you saying that you didn't want an `ignore` option to apply to all commands out of the box. I agree, and with this new generic naming (`include` and `exclude`), I think we can move forward and apply the flags to certain commands when it makes sense.

paulyoung:
This would be really useful. Please let me know if I can help in any way.

joscha:
@rygine @thejameskyle does that mean I should rename the flag in #152 to `--include`? I am happy to do that if that finally makes it mergeable and integrated.

hzoo:
this LGTM

rygine:
@hzoo @joscha @thejameskyle i've updated the PR to include usage of `PackageUtilities.filterPackages` from the latest commit. i had to slightly modify it to allow for negation. perhaps there's a better way of doing this.

i settled on `--ignore` for the flag, and placed it beneath `scope` in the CLI help, with a similar message.

the new `scope` flag could easily be applied to the `bootstrap` command, and truthfully, i'm going to need it for my current project. the same goes for the `ignore` flag and other commands.

i can do a PR for adding scope to the bootstrap command after this is merged.

src/PackageUtilities.js
```diff
@@ -60,12 +60,13 @@ export default class PackageUtilities {
...
   * @param {String} glob The glob to match the package name against
+  * @param {Boolean} negate Negate glob pattern matches
   * @return {Array.<Package>} The packages with a name matching the glob
   * @throws in case a given glob would produce an empty list of packages
   */
-  static filterPackages(packages, glob) {
+  static filterPackages(packages, glob, negate = false) {
     if (typeof glob !== "undefined") {
-      packages = packages.filter(pkg => minimatch(pkg.name, glob));
+      packages = packages.filter(pkg => negate ? !minimatch(pkg.name, glob) : minimatch(pkg.name, glob));
```
joscha: isn't that an xor? `negate ^ minimatch(pkg.name, glob)` should do it I think.

joscha:
@rygine nice! What if you specify both `--ignore` and `--scope`? I guess ignore would trump scope?

rygine:
@joscha i think specifying both should throw an error so that there's no confusion as to what gets precedent.

src/PackageUtilities.js
rygine: i knew there was a better way. this works! thanks. amending...

src/PackageUtilities.js
```diff
@@ -60,12 +60,13 @@ export default class PackageUtilities {
...
   * @param {String} glob The glob to match the package name against
+  * @param {Boolean} negate Negate glob pattern matches
   * @return {Array.<Package>} The packages with a name matching the glob
   * @throws in case a given glob would produce an empty list of packages
   */
-  static filterPackages(packages, glob) {
+  static filterPackages(packages, glob, negate = false) {
     if (typeof glob !== "undefined") {
-      packages = packages.filter(pkg => minimatch(pkg.name, glob));
+      packages = packages.filter(pkg => negate ^ minimatch(pkg.name, glob));
```
thejameskyle: I think theres approximately 7 people in the universe that know what the `^` operator does. I'd rather have something longer that expresses what happens better.

joscha: 😂 I think the name of the variable kinda gives the functionality away though...

joscha:
@rygine fair enough, I guess the use case for such a thing is very slim anyway, as you can always express one through the other or just use a more specific glob.

src/PackageUtilities.js
rygine: should i go back to the ternary operator i had before?

hzoo:
👍  Can you add a corresponding readme update?

rygine:
@hzoo docs added

src/PackageUtilities.js
hzoo: Either that or just a if else

```js
packages.filter(pkg => {
  if (negate) {

  } else {

  }
}
```

loganfsmyth: I know what they do and how to read them if I need to, but I also 100% agree that they are unreadable and unclear unless you're actually doing bitwise math.

simonkberg: if `minimatch` returns a boolean, couldn't `^` simply be replaced with `!==`? in case the operator is the issue.

hzoo:
Can you rebase again?

rygine:
done

rygine:
need to update the stub in the a unit test. should be good now.

PR #172, opened 2016-06-01
Use alpha version instead of canary

### LINKED ISSUE #166: Use alpha version instead of canary
As we learned `-canary` takes precedence over `-beta` when installing `^1.0.0-beta.x` so we should update canary mode to use `alpha` instead

hzoo:
Fixes #166 

Could potentially be an option in lernaConfig with a default of alpha? I think its fine to do just this atm though.

This is breaking change if people have been using it before though?

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,7 @@ $ lerna publish --canary
 
 The `canary` flag will publish packages after every successful merge using the sha as part of the tag.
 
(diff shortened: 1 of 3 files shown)


thejameskyle:
We've made a few breaking changes in the `-beta` releases, I'm fine doing this.

PR #173, opened 2016-06-02
Lerna import

gigabo:
The problem:  I have a package in a stand-alone repository that I want to pull into a lerna repo.

The solution: `lerna import`.

Imports a package from a foreign repository _with commit history_.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,13 +384,20 @@ $ lerna exec --scope my-component -- ls -la
 $ lerna exec --concurrency 1 -- ls -la
 ```
 
-### Import
+### import
 
 ```sh
 $ lerna import <path-to-external-repository>
 ```
 
-Import the package at `<path-to-external-repository>`, with commit history.
+Import the package at `<path-to-external-repository>`, with commit history,
+into `packages/<package-name>`.  Original commit authors, dates and messages
+are preserved.  Commits are applied to the current branch.
+
+This is useful for gathering pre-exising standalone packages into a Lerna
+repo.  Each commit is modified to make changes relative to the package
+directory.  So, for example, the commit that added `package.json` will
+instead add `packages/<package-name>/package.json`.
 
 ## Misc
 
--- a/README.md
+++ b/README.md
@@ -384,6 +384,14 @@ $ lerna exec --scope my-component -- ls -la
 $ lerna exec --concurrency 1 -- ls -la
 ```
 
+### Import
+
+```sh
+$ lerna import <path-to-foreign-repository>
+```
+
+Import the package at `<path-to-foreign-repository>`, with commit history.
+
 ## Misc
 
 Lerna will log to a `lerna-debug.log` file (same as `npm-debug.log`) when it encounters an error running a command.
--- a/README.md
+++ b/README.md
@@ -394,7 +394,7 @@ Import the package at `<path-to-external-repository>`, with commit history,
 into `packages/<package-name>`.  Original commit authors, dates and messages
 are preserved.  Commits are applied to the current branch.
 
-This is useful for gathering pre-exising standalone packages into a Lerna
+This is useful for gathering pre-existing standalone packages into a Lerna
 repo.  Each commit is modified to make changes relative to the package
 directory.  So, for example, the commit that added `package.json` will
 instead add `packages/<package-name>/package.json`.
--- a/README.md
+++ b/README.md
@@ -391,7 +391,7 @@ $ lerna import <path-to-external-repository>
 ```
 
 Import the package at `<path-to-external-repository>`, with commit history,
-into `packages/<package-name>`.  Original commit authors, dates and messages
+into `packages/<directory-name>`.  Original commit authors, dates and messages
 are preserved.  Commits are applied to the current branch.
 
 This is useful for gathering pre-existing standalone packages into a Lerna
--- a/README.md
+++ b/README.md
@@ -387,10 +387,10 @@ $ lerna exec --concurrency 1 -- ls -la
 ### Import
 
 ```sh
-$ lerna import <path-to-foreign-repository>
+$ lerna import <path-to-external-repository>
 ```
 
-Import the package at `<path-to-foreign-repository>`, with commit history.
+Import the package at `<path-to-external-repository>`, with commit history.
 
 ## Misc
 
--- a/src/commands/ImportCommand.js
+++ b/src/commands/ImportCommand.js
@@ -1,0 +1,76 @@
+import fs from "fs";
+import path from "path";
+import child from "child_process";
+import Command from "../Command";
+import progressBar from "../progressBar";
+import PromptUtilities from "../PromptUtilities";
+
+export default class ImportCommand extends Command {
+  initialize(callback) {
+    const inputPath = this.input[0];
+
+    if (!inputPath) {
+      callback(new Error("Missing argument: Path to foreign repository"));
+    }
+
+    const foreignRepoPath = path.resolve(inputPath);
+
+    try {
+      const stats = fs.statSync(foreignRepoPath);
+      if (!stats.isDirectory()) {
+        throw new Error(`Input path "${inputPath}" is not a directory`);
+      }
+      const packageJson = path.join(foreignRepoPath, "package.json");
+      const packageName = require(packageJson).name;
+      if (!packageName) {
+        throw new Error(`No package name specified in "${packageJson}"`);
+      }
+      this.targetDir = "packages/"+packageName;
+    } catch (e) {
+      callback(e);
+    }
+
+    this.foreignExecOpts = {
+      encoding: "utf8",
+      cwd: foreignRepoPath
+    };
+
+    this.commits = this.foreignExecSync(`git log --format="%h"`).split("\n").reverse();
+
+    if (!this.commits.length) {
+      callback(new Error(`No git commits to import at "${inputPath}"`));
+    }
+
+    this.logger.info(`About to import ${this.commits.length} commits into from ${inputPath} into ${this.targetDir}`);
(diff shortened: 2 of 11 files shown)


gigabo:
Pretty sweet progress bar experience with enough git history to import:

tty

thejameskyle:
I really really like this. Thanks for doing this.

Can you use the name "external" instead of "foreign"?

src/commands/ImportCommand.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  execute(callback) {
+    const replacement = "$1/"+this.targetDir;
+
+    progressBar.init(this.commits.length);
+
+    this.commits.forEach(sha => {
+      progressBar.tick(sha);
+      const patch = this.foreignExecSync(`git format-patch -1 ${sha} --stdout`)
```
thejameskyle: Can you add a comment explaining what is happening here?

src/commands/ImportCommand.js
```diff
@@ -0,0 +1,76 @@
...
+      const packageName = require(packageJson).name;
+      if (!packageName) {
+        throw new Error(`No package name specified in "${packageJson}"`);
+      }
+      this.targetDir = "packages/"+packageName;
+    } catch (e) {
+      callback(e);
+    }
+
+    this.foreignExecOpts = {
```
thejameskyle: What is this for?

gigabo:
> Can you use the name "external" instead of "foreign"?

Yes, that's a better name.  Done.

src/commands/ImportCommand.js
gigabo: `externalExecSync` is also called from the `execute` method, so the options need to persist until then.

Could, alternatively, attach only `externalRepoPath` to `this` and build up the exec options inside `externalExecSync`, if you'd prefer.

README.md
```diff
@@ -291,6 +291,14 @@ $ lerna run build
 
 Run an [npm script](https://docs.npmjs.com/misc/scripts) in each package that contains that script.
 
+### Import
+
+```sh
+$ lerna import <path-to-external-repository>
+```
+
+Import the package at `<path-to-external-repository>`, with commit history.
```
hzoo: If we want to be more verbose here, you can mention that the package x is imported into `packages/x` and maybe an example. You can even use lerna-changelog as the example?

gigabo: Added some more detail here.  I worry that using lerna-changelog as an example might be confusing, since it has "lerna" in the name, so I continued to use placeholders.

I also lower-cased the section heading to match the changes from @kaicataldo over the weekend (which I've picked up via rebase).

hzoo: Yeah you're right - maybe babel would of been a better example but all good

README.md
```diff
@@ -318,6 +318,21 @@ $ lerna run build
...
+
+```sh
+$ lerna import <path-to-external-repository>
+```
+
+Import the package at `<path-to-external-repository>`, with commit history,
+into `packages/<package-name>`.  Original commit authors, dates and messages
+are preserved.  Commits are applied to the current branch.
+
+This is useful for gathering pre-exising standalone packages into a Lerna
```
hzoo: `exising` -> `existing`

src/commands/ImportCommand.js
```diff
@@ -0,0 +1,83 @@
...
+      const stats = fs.statSync(externalRepoPath);
+      if (!stats.isDirectory()) {
+        throw new Error(`Input path "${inputPath}" is not a directory`);
+      }
+      const packageJson = path.join(externalRepoPath, "package.json");
+      const packageName = require(packageJson).name;
+      if (!packageName) {
+        throw new Error(`No package name specified in "${packageJson}"`);
+      }
+      this.targetDir = "packages/"+packageName;
```
hzoo: can do `packages/${packageName}` here

gigabo: Sure, though I've tended not to when concatenating a single string constant with a single variable, since a template string is actually longer in that case.  How about I add some space around the operator?  That's the same length, but involves less punctuation.  Seems to [fit](https://github.com/lerna/lerna/blob/master/bin/lerna.js#L43) [with](https://github.com/lerna/lerna/blob/master/src/Command.js#L23) [context](https://github.com/lerna/lerna/blob/master/src/GitUtilities.js#L12), too.

README.md
gigabo: Whoops!  Thanks.

src/commands/ImportCommand.js
hzoo: Yeah the style has been to usually have whitespace between operators (we should add `space-infix-ops`)

gigabo: > we should add `space-infix-ops`

#188

hzoo:
👍

thejameskyle:
Can you add a test that `--yes` works?

gigabo:
`--yes` was not supported.  Added support and a test.

gigabo:
Looks like the latest build may have had a spurious failure in one environment.  I'm not able to retry.

thejameskyle:
> `--yes` was not supported. 

Weird, I thought `PromptUtilities.confirm` supported that all the time

hzoo:
There's a merge conflict b/c of the last PR

gigabo:
Rebased.

thejameskyle:
Oh fucking hell, merged another one of your PRs and it needs another rebase. 🃏

hzoo:
blah I hope it's not because of the readme

gigabo:
Rebased again.

> blah I hope it's not because of the readme

Nope, lerna.js and index.js this time. 😛

hzoo:
Would be nice to use lerna here and do a canary release - can try a publish manually

src/commands/ImportCommand.js
```diff
@@ -0,0 +1,88 @@
...
+
+    const externalRepoPath = path.resolve(inputPath);
+
+    try {
+      const stats = fs.statSync(externalRepoPath);
+      if (!stats.isDirectory()) {
+        throw new Error(`Input path "${inputPath}" is not a directory`);
+      }
+      const packageJson = path.join(externalRepoPath, "package.json");
+      const packageName = require(packageJson).name;
```
hzoo: Do we want to make a specific error message here or no? (if the package path doesn't have a package.json)

```
Errored while running ImportCommand.initialize
Error: Cannot find module '/module/package.json'
```

hzoo:
Testing it out locally - do we want have an option to add a prefix or no? Like in babel everything is called `babel-`, and I assume when you import the packages are in like `babel/generator` and when you import you might want it to be called `babel-generator` (or for scoped modules)

Not a blocker and can be a further improvement

hzoo:
Another simple error we can add in is if you have already imported that repo in before (either by name or git log, package.json, etc)

giuseppeg:
@hzoo correct me if I am wrong but generally speaking existing packages already have a name (which may include a prefix). If one is importing `babel-generator` will lerna change the package name?

thejameskyle:
> maybe like lerna import ../generator --prefix generator or just the next be the new folder/package name like lerna import ../generator babel-generator `lerna import ../generator --name babel-generator

I would prefer just setting a new name

hzoo:
Yeah I think so as well (new name) similar to `git` with branches/checkout

gigabo:
> packageName should be the folderName of the imported package not the package name itself

That makes sense to me.  I hope they're generally the same, but if they're not then choosing the old directory name for the new directory seems right.

> Another simple error we can add in is if you have already imported that repo in before

Ooh, good one.

gigabo:
Rebased, added some tests, did [these two things](https://github.com/lerna/lerna/pull/173#issuecomment-225643676) and got it running in node 0.10.

src/commands/ImportCommand.js
```diff
@@ -0,0 +1,104 @@
...
+    } catch (e) {
+      if (e.code === "ENOENT") {
+        return callback(new Error(`No repository found at "${inputPath}"`));
+      }
+      return callback(e);
+    }
+
+    this.targetDir = "packages/" + externalRepoBase;
+
+    let targetDirExists;
```
giuseppeg: this block could be
```javascript
try {
  fs.statSync(this.targetDir);
} catch (e) { 
  return callback(new Error(`Target directory already exists "${this.targetDir}"`));
}
```
right?

README.md
```diff
@@ -384,6 +384,21 @@ $ lerna exec --scope my-component -- ls -la
...
 ```
 
+### import
+
+```sh
+$ lerna import <path-to-external-repository>
+```
+
+Import the package at `<path-to-external-repository>`, with commit history,
+into `packages/<package-name>`.  Original commit authors, dates and messages
```
giuseppeg: now it imports to `packages/<package-src-folder-name>`

src/commands/ImportCommand.js
gigabo: Yeah, I spent some time trying to come up with a simpler check here.  What you posted above is nicer looking, but it is actually the _inverse_ of what we want: It returns an error when the target directory does _not_ exist.

giuseppeg: Oh right my bad, read the code again now.
How about 
```javascript
try {
  if (fs.statSync(this.targetDir)) {
    return callback(new Error(`Target directory already exists "${this.targetDir}"`));
  }
} catch (e) { /* Pass */ }
```
This is a nit anyway :)

gigabo: I like it!  Always glad to be rid of a variable. 👹

hzoo:
There's a docs fix to add, but other than that I'm good with landing this and we can improve as it's tested/used?

gigabo:
Need to rebase again and back out the changes to support node 0.10.

gigabo:
Rebased and updated docs.

Issue #176, opened 2016-06-05
[Feature] Cross linking packages breaks identity checks in external dependencies


charlieschwabacher:
Imagine I have a lerna repo w/ two packages _a_ and _b_ that depend on each other, and each depend on an external package from npm, _c_.  If I bootstrap and then run code inside _a_, I will end up requiring two different copies of _c_ - the one in `packages/a/node_modules` AND the one in `packages/b/node_modules`.  This breaks any identity checks in _c_.

I'm running into this problem w/ graphql as the external dependency and a failing `instanceof` check.

One way to avoid this problem would be to add shared external dependencies to the root package.json, install them to `node_modules`, and link them from `packages/a/node_modules` and `packages/b/node_modules` in the same way that lerna links the internal dependencies.

Is this a change you would consider?  Or is there another solution for this that I have overlooked?

charlieschwabacher:
I was able to work around my issue by adding a dummy graphql package w/

package.json:
```
{
  "private": true,
  "name": "graphql",
  "version": "0.6.0"
}
```

and index.js:
```
module.exports = require('../../node_modules/graphql');
```

I would still love a cleaner solution, but this is at least a temporary solution.

hzoo:
Will link https://twitter.com/gr2m/status/738454783854563330 as a reference since it seems to be a general problem too

> Is there a CLI tool that lets me git clone multiple repositories, runs npm install and links them based on "dependencies" in package.json?

thejameskyle:
Yeah I've run into this exact issue with React related stuff. I wouldn't mind adding a feature that allowed you to whitelist external dependencies that need to share a single instance that could be linked from the root `node_modules`.

I think we'd have to use actual symlinks though... which we may need to explore again

charlieschwabacher:
@thejameskyle - why do you say we would have to use symlinks?

thejameskyle:
Because you would need to symlink it into each package.

rygine:
this would be fixed using the updated bootstrapping process i've created, as long as they share the same version of the external dependency.

more info here: https://github.com/lerna/lerna/issues/215

charlieschwabacher:
@thejameskyle what I mean is why symlinks here vs proxies for cross linking the local packages?

gigabo:
Oh, this didn't actually wind up getting addressed by #237.

The functionality that was dropped is tracked in #334.

ingro:
[...]

PR #179, opened 2016-06-06
[Feature] Update bootstrap to link external dependencies when present in the root node_modules

charlieschwabacher:
Resolves https://github.com/lerna/lerna/issues/176

Before installing external dependencies for packages, this makes bootstrap check to see if there is a package in the root node_modules directory that will satisfy the dependency.  If there is, it is linked in the same way that the local packages are cross linked.  If there is not, the package is installed through npm normally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+!test/**/node_modules
 *.log
 lib
 tmp
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -8,6 +8,7 @@ export default class Repository {
 
     this.lernaJsonLocation = path.join(this.rootPath, "lerna.json");
     this.packageJsonLocation = path.join(this.rootPath, "package.json");
+    this.nodeModulesLocation = path.join(this.rootPath, "node_modules");
     this.packagesLocation = path.join(this.rootPath, "packages");
 
     // Legacy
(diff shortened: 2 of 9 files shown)


charlieschwabacher:
A test has failed on Travis - it passes locally for me and doesn't look related to these changes.  Have you seen that failure before / do you have any guidance?

charlieschwabacher:
Looks like tests re-ran, thanks 👍

thejameskyle:
Without real symlinking this will end up breaking a lot of external packages that use other paths.

charlieschwabacher:
What is the reasoning behind the proxying vs symlinking the local packages?  It seems like `createLinkedDependency` could be updated to make the symlink instead of writing files.. I assume there was some issue w/ that that forced the proxying though?

thejameskyle:
Yeah, there was an issue with npm creating a bunch of recursive directories. Maybe #263 will fix that

rygine:
@charlieschwabacher does this work with PR #237? how does this affect `npm outdated`? what are you using to bundle your source?

i ran into a similar issue in webpack that was resolved with the right configuration.

noherczeg:
This has been somewhat sort of covered in #507 right @gigabo ?

evocateur:
I believe this has been addressed by the `--hoist` option.

PR #180, opened 2016-06-06
Add *.swp files to .gitignore

gigabo:
This makes `git status` much nicer for vim users.


hzoo:
You can probably do a global gitignore (don't know how well it works) but totally fine with merging too

PR #181, opened 2016-06-06
Emit partial stderr output from child on maxBuffer exceeded

gigabo:
If the error from `child.exec` is just that the child process emitted too much
on stderr, then that stderr output is likely to be useful.  Emit as much as
possible, with a warning that it's partial output.

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -4,6 +4,14 @@ export default class ChildProcessUtilities {
   static exec(command, opts, callback) {
     return child.exec(command, opts, (err, stdout, stderr) => {
       if (err != null) {
+
+        // If the error from `child.exec` is just that the child process
+        // emitted too much on stderr, then that stderr output is likely to
+        // be useful.
+        if (/^stderr maxBuffer exceeded/.test(err.message)) {
+          err = `Error: ${err.message}.  Partial output follows:\n\n${stderr}`;
+        }
+
         callback(err || stderr);
       } else {
         callback(null, stdout);


gigabo:
I was seeing output like this from `lerna run test`:

image


This patch let me _see_ the `stderr` output, which helped me track down what was actually going wrong.

gigabo:
Looks like a test failed on one of the environments.  `npm run ci` passed locally (and passes on the other two environments in travis).  Noticed that #179 also had an unexpected failure in one build environment, though the error output looks different.  Something intermittent going on?  I'm unable to restart the test to check.

gigabo:
Thanks for rerunning the tests.

hzoo:
👍

src/ChildProcessUtilities.js
```diff
@@ -4,6 +4,14 @@ export default class ChildProcessUtilities {
...
       if (err != null) {
+
+        // If the error from `child.exec` is just that the child process
+        // emitted too much on stderr, then that stderr output is likely to
+        // be useful.
+        if (/^stderr maxBuffer exceeded/.test(err.message)) {
+          err = `Error: ${err.message}.  Partial output follows:\n\n${stderr}`;
+        }
+
         callback(err || stderr);
```
thejameskyle: There's probably a better way to format this callback with the err and stderr output. But this is fine for now

hzoo: We can merge and leave a TODO/issue to cleanup

gigabo: Hey, I don't want to let this slip through the cracks, but I'm actually not sure what the TODO is here. 😅

PR #182, opened 2016-06-07
Run git init on lerna init

### LINKED ISSUE #175: Lerna assumes the repo is already been initialized with `git init`.
I think if you lerna init in a folder that isn't a git repo already it errors? It's an edge case but we can run `git init` if it isn't

if we do this, we can remove the the line I added in the init section of the readme

kaicataldo:
fixes #175 

Thoughts on this approach?

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -13,9 +13,9 @@ export default class Command {
     this.flags = flags;
 
     this.lernaVersion = require("../package.json").version;
-    this.repository = new Repository();
-    this.progressBar = progressBar;
     this.logger = logger;
+    this.repository = new Repository(this.logger);
+    this.progressBar = progressBar;
     this.concurrency = (!flags || flags.concurrency === undefined) ? DEFAULT_CONCURRENCY : Math.max(1, +flags.concurrency || DEFAULT_CONCURRENCY);
   }
 
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -3,6 +3,11 @@ import logger from "./logger";
 
 export default class GitUtilities {
   @logger.logifySync
+  static checkIfInitialized() {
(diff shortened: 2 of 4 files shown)


src/Repository.js
```diff
@@ -3,9 +3,16 @@ import FileSystemUtilities from "./FileSystemUtilities";
 import path from "path";
 
 export default class Repository {
-  constructor() {
-    this.rootPath = GitUtilities.getTopLevelDirectory();
```
kaicataldo: Error is thrown here

src/Command.js
```diff
@@ -13,9 +13,9 @@ export default class Command {
     this.flags = flags;
 
     this.lernaVersion = require("../package.json").version;
-    this.repository = new Repository();
-    this.progressBar = progressBar;
     this.logger = logger;
+    this.repository = new Repository(this.logger);
```
kaicataldo: Passing in the `logger` so we can log when initializing the Git repo

src/Repository.js
hzoo: Don't think we have any tests for Repository yet

hzoo:
Awesome!

src/GitUtilities.js
```diff
@@ -3,6 +3,11 @@ import logger from "./logger";
 
 export default class GitUtilities {
   @logger.logifySync
+  static checkIfInitialized() {
```
thejameskyle: The try-catch should happen in this util. It should be:

```js
static isInitialized() {
  try {
    ChildProcessUtilities.execSync("git rev-parse");
    return true;
  } catch (err) {
    return false;
  }
}
```

thejameskyle:
#190

src/GitUtilities.js
kaicataldo: That makes sense - thanks for the explanation!

PR #189, opened 2016-06-07
Update .gitignore

rygine:
This will ignore more files that are specific to an OS or IDE.


rygine:
I'm using webstorm, which has an /.idea folder. This should handle most other editors as well as some OS-specific stuff.

PR #190, opened 2016-06-08
Cleanup git init in lerna init

thejameskyle:


thejameskyle:
#182

hzoo:
nice, makes sense

Issue #121, opened 2016-05-06
[Feature] Does not work with optionalDependencies


nolanlawson:
If one module is an `optionalDependency` of the other (rather than a `dependency`), then Lerna does not link them together.

Our use-case for this in PouchDB is that sometimes we have native dependencies (e.g. `leveldown`) that users struggle to install on certain hardware, so we make it optional so it doesn't completely block them.

I'll try to get to this when I get a chance, but filing the issue to make a note of it.

thejameskyle:
This should just require updating:

- https://github.com/kittens/lerna/blob/3f84f9a19c76043441655120b66f65db46883997/src/commands/PublishCommand.js#L240-L243
- https://github.com/kittens/lerna/blob/9246f288b57122bbf1121d4b6910c2232baf35c5/src/Package.js#L30-L49

nolanlawson:
Nice! Probably just a user error on my part.

hzoo:
It's good now?

nolanlawson:
I'll check when I get a chance, but it sounds like it!

thejameskyle:
No it's definitely not supported. I was saying that in order to support it those places need to be updated.

hzoo:
If that's it, https://github.com/lerna/lerna/pull/191

PR #191, opened 2016-06-08
[Feature] Add optionalDependencies

hzoo:
Fixes https://github.com/lerna/lerna/issues/121


thejameskyle:
So optionalDependencies is supposed to make it so that the install does not fail if a optional dependency build fails. I'm not sure how to make that work here

nolanlawson:
@thejameskyle Since that feature is usually designed for users rather than developers (e.g. user is on Windows 7 64-bit, developer is on Mac OS X), I think it would be fine for the time being if it just failed if the optional dependency fails.

thejameskyle:
That seems like a poor experience when it goes wrong though

gigabo:
This seems like a good first step.  If people start running into failures due to optional deps then it seems not hard to handle them separately during bootstrap and just warn on failure or something.

@hzoo - This LGTM. 👍

bigtimebuddy:
Any updates on this PR? Would love to see it in?

doug-wade:
I'd really rather we hold off merging this until we ensure that optionalDependencies are supported _as optional_, i.e. that `bootstrap` doesn't fail if they aren't installed.

jstejada:
Any updates on this PR? Would love to see it in

hzoo:
Not sure what to do with this, closing

gtamas:
Would it be possible to reopen this issue?
Could you at least add support for this as an optional feature (disabled by default)? For example, how about adding a command line flag? That way this wouldn't cause any issues and people could enable it whenever they need it.

evocateur:
We need to stop mangling the package.json so much during install, this is just a bandaid.

a-magdy:
For those who are still looking for a solution

I pushed a lerna fork (w)ith (o)ptional (d)epdencies [lerna-wod](https://www.npmjs.com/package/lerna-wod) with the fix proposed in this PR

It is working for me locally but I had to use `npx lerna-wod` instead of `lerna` when I wanted to use it

Hope it helps

MaxMilton:
Would have been nice to see this in the main package :cry:

evocateur:
Implemented in b73e19df

PR #192, opened 2016-06-08
Handle packages with binaries

rygine:
Minor refactor of some methods for easier to read code using `async.series`

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -51,6 +51,17 @@ export default class FileSystemUtilities {
     rimraf(filePath, callback);
   }
 
+  @logger.logifyAsync
+  static symlink(src, dest, type = "file", callback) {
+    fs.lstat(dest, (err) => {
+      if (!err) {
+        fs.unlink(dest, () => fs.symlink(src, dest, type, callback));
+      } else {
+        fs.symlink(src, dest, type, callback);
+      }
+    });
+  }
+
   @logger.logifySync
   static unlinkSync(filePath) {
     fs.unlink(filePath);
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -6,7 +6,6 @@ import semver from "semver";
 import async from "async";
 import find from "lodash.find";
 import path from "path";
-import fs from "fs";
 
 export default class BootstrapCommand extends Command {
   initialize(callback) {
@@ -56,12 +55,29 @@ export default class BootstrapCommand extends Command {
       });
 
       async.parallelLimit(batch.map(pkg => done => {
-        async.series([
+        // async actions to bootstrap each package
+        const actions = [
           cb => FileSystemUtilities.mkdirp(pkg.nodeModulesLocation, cb),
           cb => this.installExternalPackages(pkg, cb),
           cb => this.linkDependenciesForPackage(pkg, cb),
-          cb => this.runPrepublishForPackage(pkg, cb),
-        ], err => {
+        ];
+        // array of dependencies with binaries
+        const depsWithBin = [];
+        // find matching dependencies with binaries
+        this.packages.forEach(dependency => {
+          // for each matched dependency with binaries
+          if (this.hasMatchingDependency(pkg, dependency) && dependency.bin) {
+            depsWithBin.push(dependency);
+          }
+        });
+        // if package has dependencies with binaries
+        if (depsWithBin.length) {
+          // symlink binaries
+          actions.push(cb => this.symlinkDependencyBinaries(pkg, depsWithBin, cb));
+        }
+        // add prepublish action
+        actions.push(cb => this.runPrepublishForPackage(pkg, cb));
+        async.series(actions, err => {
           this.progressBar.tick(pkg.name);
           donePackages[pkg.name] = true;
           todoPackages.splice(todoPackages.indexOf(pkg), 1);
@@ -96,15 +112,7 @@ export default class BootstrapCommand extends Command {
       const linkSrc = dependency.location;
       const linkDest = path.join(pkg.nodeModulesLocation, dependency.name);
 
-      this.createLinkedDependency(linkSrc, linkDest, dependency.name, (err) => {
-        if (!err && dependency._package.bin) {
-          const srcBinLocation = path.join(linkSrc, dependency._package.bin);
-          const destBinLocation = path.join(pkg.nodeModulesLocation, ".bin");
-          this.createBinaryLink(srcBinLocation, destBinLocation, dependency.name, done);
-        } else {
-          done(err);
-        }
-      });
+      this.createLinkedDependency(linkSrc, linkDest, dependency.name, done);
     }, callback);
   }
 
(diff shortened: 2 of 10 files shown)


rygine:
At my company, we've written a few CLI tools that are accessed via NPM scripts in each package. Bootstrapping did not factor this in and they would fail to run.

src/commands/BootstrapCommand.js
```diff
@@ -49,24 +50,25 @@ export default class BootstrapCommand extends Command {
       const linkSrc = dependency.location;
       const linkDest = path.join(pkg.nodeModulesLocation, dependency.name);
 
-      this.createLinkedDependency(linkSrc, linkDest, dependency.name, done);
+      this.createLinkedDependency(linkSrc, linkDest, dependency.name, (err) => {
+        const bin = dependency._package.bin;
```
thejameskyle: This should check `err` at the beginning and call `done(err)` if it exists and stop

rygine: fixed

rygine:
i'm noticing that the unit tests are non-deterministic, both locally and travis

paulyoung:
This is sorely needed! Anything I can do to help move this along?

rygine:
@paulyoung, this functionality is actually covered in PR #237 as well.

src/commands/BootstrapCommand.js
```diff
@@ -49,24 +50,24 @@ export default class BootstrapCommand extends Command {
       const linkSrc = dependency.location;
       const linkDest = path.join(pkg.nodeModulesLocation, dependency.name);
 
-      this.createLinkedDependency(linkSrc, linkDest, dependency.name, done);
+      this.createLinkedDependency(linkSrc, linkDest, dependency.name, (err) => {
+        if (!err && dependency._package.bin) {
+          const srcBinLocation = path.join(linkSrc, dependency._package.bin);
```
gigabo: Would be nice to add a `bin` getter in src/Package.js for consistency. Then this could be:

```javascript
const srcBinLocation = path.join(linkSrc, dependency.bin);
```

src/commands/BootstrapCommand.js
```diff
@@ -49,24 +50,24 @@ export default class BootstrapCommand extends Command {
       const linkSrc = dependency.location;
       const linkDest = path.join(pkg.nodeModulesLocation, dependency.name);
 
-      this.createLinkedDependency(linkSrc, linkDest, dependency.name, done);
+      this.createLinkedDependency(linkSrc, linkDest, dependency.name, (err) => {
+        if (!err && dependency._package.bin) {
```
gigabo: `bin` can be [an object](https://docs.npmjs.com/files/package.json#bin).  Should probably support that here.

src/commands/BootstrapCommand.js
gigabo: Actually, `dependency.bin` [isn't guaranteed to be a string](https://github.com/lerna/lerna/pull/192/files#r68466171).

Could normalize inside the getter to always return an object.  Else the logic here will need some rework.

src/commands/BootstrapCommand.js
rygine: i've neglected this PR in favor of PR #237, which accounts for this and also add a `.bin` getter to `Package` as you mentioned above.

gigabo: Great!  So it should be easy to back port. 😁 

The functionality provided by _this_ PR is independently useful and, frankly, less controversial than #237.  I don't want it to get delayed by association.  I'd like to see this merge, independently, as soon as possible.  That would have the added benefit of reducing the scope of #237, which is pretty big.

rygine: Agreed. Will update.

rygine:
@gigabo This should do it. I also uncovered a potential issue with the `ignore` flag during bootstrap. Fixed that as well.

* Added `symlink` method to FileSystemUtilities
* Added `bin` property to Package
* Updated bootstrap to factor in dependencies with multiple binaries
* Added test for package with multiple binaries
* Refactored way ignored packages are determined (to be more like run command)
* Updated unit test for ignored package

When a package is ignored while bootstrapping, it should be completely taken out of the equation. This update alters the previous behavior of not bootstrapping an ignored package, but still treating it as though it was part of the process. If a package is ignored, it should be treated as an external dependency. The updated unit test reflects this.

gigabo:
> If a package is ignored, it should be treated as an external dependency.

Hmm... that's actually not what I'd naively expect.  😕 

I'd expect "ignore package foo" to skip doing anything in `packages/foo`, but to do the _normal_ things in `package/bar`, including linking `foo` as an in-repo dep.  Am I just misunderstanding the `--ignore` flag?

rygine:
@gigabo that's a good question. what exactly does ignore mean?

my initial thought was that it doesn't get any of its dependencies "linked" in the bootstrap, but i'm not sure that makes sense.

for example, check out this scenario:

* packageA depends on packageB
* packageC depends on packageA

i run bootstrap and ignore packageA, which means that packageA does not have a "linked" or installed version of packageB. meanwhile, i bootstrap packageC and "link" to packageA like normal. now, when i import packageA into packageC, packageB will not be imported as no form of it exists in `packageA/node_modules/`.

under the new scenario, packageA would be installed as an external dependency of packageC and all would be well. i don't see much use for the `ignore` flag during bootstrap except for some edge-cases when you simply want to disregard a module.

it is a bit confusing to think about. once i realized how things would be bootstrapped when you ignore a package, i'm not sure the original way of doing it would actually work.

gigabo:
@rygine - The behavior or `--ignore` is worth examining, but seems unrelated to handling packages with binaries.  Can that part be pulled out into a separate issue or PR?

rygine:
@gigabo makes sense. amended the commit.

src/commands/BootstrapCommand.js
```diff
@@ -28,16 +42,11 @@ export default class BootstrapCommand extends Command {
     this.progressBar.init(this.packages.length);
     this.logger.info("Linking all dependencies");
 
-    const ignore = this.flags.ignore || this.repository.bootstrapConfig.ignore;
-
-    // Get a filtered list of packages that will be bootstrapped.
-    const todoPackages = PackageUtilities.filterPackages(this.packages, ignore, true);
-
-    // Get a trimmed down graph that includes only those packages.
-    const filteredGraph = PackageUtilities.getPackageGraph(todoPackages);
```
rygine: this is slightly unrelated to the PR, but i moved the package filtering to the `initialize` method to mimic the behavior of `lerna run` for consistency.

src/commands/BootstrapCommand.js
```diff
@@ -28,16 +42,11 @@ export default class BootstrapCommand extends Command {
...
+    // Get a trimmed down graph that includes only bootstrapping packages.
+    const filteredGraph = PackageUtilities.getPackageGraph(this.bootstrapPackages);
 
     // As packages are completed their names will go into this object.
-    const donePackages = {};
+    this.donePackages = {};
```
rygine: want to keep track of bootstrapped packages within the class to display the number at the end, see L#35.

src/commands/BootstrapCommand.js
gigabo: > unrelated to the PR

Generally speaking, smaller more targeted PRs are easier to review and therefore tend to get merged more quickly.

This PR adds much needed functionality, but it has gone through a number of iterations that each included a different set of unrelated changes.  These other changes may _also_ be valuable, but it would be easier to consider them on their own merit if they were showcased in dedicated PRs.

rygine: point taken. amended.

src/commands/BootstrapCommand.js
```diff
@@ -55,12 +55,28 @@ export default class BootstrapCommand extends Command {
...
+        this.packages.forEach(dependency => {
+          // for each matched dependency with binaries
+          if (this.hasMatchingDependency(pkg, dependency) && dependency.bin) {
+            depsWithBin.push(dependency);
+          }
+        });
+        // if package has dependencies with binaries
+        if (depsWithBin.length) {
+          // symlink binaries
+          actions.push(cb => this.symlinkDependencyBinaries(pkg, depsWithBin, cb));
```
gigabo: I think the binaries should be linked _before_ `runPrepublishForPackage`.

A package may have a `prepublish` script that uses a binary from a dependency.

src/commands/BootstrapCommand.js
```diff
@@ -128,13 +136,49 @@ export default class BootstrapCommand extends Command {
...
+   * Symlink package dependency binaries to the package's node_modules/.bin folder
+   * @param {Package} pkg
+   * @param {Array.<Package>} dependencies
+   * @param {Function} callback
+   */
+  symlinkDependencyBinaries(pkg, dependencies, callback) {
+    const actions = [];
+    dependencies.forEach(dependency => {
+      const { location, name, bin } = dependency;
+      const destPath = path.join(pkg.nodeModulesLocation);
```
gigabo: What does this `path.join` do?

src/FileSystemUtilities.js
```diff
@@ -51,6 +51,17 @@ export default class FileSystemUtilities {
     rimraf(filePath, callback);
   }
 
+  @logger.logifyAsync
+  static symlink(src, dest, callback, type = "file") {
```
gigabo: It doesn't look like the caller uses the `type` parameter?

All other asynchronous methods in `FileSystemUtilities` take the callback as the _final_ argument.

src/commands/BootstrapCommand.js
gigabo: Thanks, that makes it easier to read. 👍

src/commands/BootstrapCommand.js
rygine: good catch. this was leftover from some previous code.

src/FileSystemUtilities.js
rygine: ahh yes. will fix.

src/commands/BootstrapCommand.js
rygine: make sense.

rygine:
@gigabo I addressed your comments with an amended commit. For future reference, should I continue to amend commits, or just make another one for things like this?

gigabo:
> should I continue to amend commits, or just make another one for things like this?

From my perspective it's nice to have a new commit so I can see clearly what changed.  PRs are typically squashed during merge, so it won't clutter the commit history on master.

gigabo:
@hzoo, @thejameskyle - Anything else you'd like to see here?  I want to use this!

gigabo:
Taking https://github.com/lerna/lerna/pull/300#issuecomment-238647550 as a 👍  here.

Thanks for sticking with this @rygine.  This is a big improvement!

PR #194, opened 2016-06-08
Wait for child processes to exit before exiting from parent

gigabo:
I ran into this with `lerna run test`.  I have one test that fails quickly and another _expensive_ test that keeps on chugging after lerna exits.

With this patch I get this:
image

Then I can just `CTRL-C` instead of having to track down orphaned children and kill them.

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -46,7 +46,7 @@ export default class ChildProcessUtilities {
     let stderr = "";
 
     const childProcess = ChildProcessUtilities.registerChild(
-      child.spawn(command, args, objectAssign({
+      spawn(command, args, objectAssign({
         stdio: "inherit"
       }, opts))
         .on("error", () => {})
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -2,24 +2,33 @@ import child from "child_process";
 import spawn from "cross-spawn";
 import objectAssign from "object-assign";
 import syncExec from "sync-exec";
+import {EventEmitter} from "events";
+
+// This will hold ChildProcess objects until they exit.
+const children = [];
+
+// This is used to alert listeners when all children have exited.
+const emitter = new EventEmitter;
 
 export default class ChildProcessUtilities {
   static exec(command, opts, callback) {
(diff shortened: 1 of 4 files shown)


hzoo:
merge conflict here as well

gigabo:
Rebased.

gigabo:
Rebased again and made to work with node 0.10.

Tracking down and killing orphaned processes by hand is a pain.  Any objections here?

src/Command.js
```diff
@@ -147,12 +148,26 @@ export default class Command {
...
+    if (childProcessCount > 0) {
+      logger.info(
+        `Waiting for ${childProcessCount} child ` +
+        `process${childProcessCount === 1 ? "" : "es"} to exit. ` +
+        "CTRL-C to exit immediately."
+      );
+      ChildProcessUtilities.onAllExited(finish);
+    } else {
+      finish();
     }
```
thejameskyle: Instead of writing checks like this, it'd be easier to have an API:

```js
ChildProcessUtilities.poorlyNamed__maybeWaitForEverythingToExit(finish);
```

where `poorlyNamed__maybeWaitForEverythingToExit` did all of the above for you.

thejameskyle: Although I guess that doesn't matter if this is the only place this check is going to happen

gigabo:
Rebased again.

doug-wade:
👍

src/ChildProcessUtilities.js
```diff
@@ -50,4 +61,23 @@ export default class ChildProcessUtilities {
       childProcess.stderr.setEncoding("utf8").on("data", chunk => stderr += chunk);
     }
   }
+
+  static registerChild(child) {
+    children.push(child);
+    child.on("exit", () => {
+      children.splice(children.indexOf(child), 1);
```
thejameskyle: Could children just be a counter?

```js
children--;
if (children === 0) {
  // ...
}
```

gigabo: Yes.  I initially thought I was going to have to propagate the signal through to children on `CTRL-C`, but it turns out that happens automatically.

Pushed a patch that simplifies this (and rebased).

thejameskyle:
Looks good, once this passes travis it's good to merge

PR #196, opened 2016-06-08
Add `lerna clean` command

### LINKED ISSUE #195: Command to clear out node_modules directories
Something like `lerna clean` that just does `rm -rf packages/*/node_modules`

I find myself doing that every now and then because of npm sucking somehow.

gigabo:
Removes the `node_modules` directory from all packages.

Closes #195.

Or would [`lerna nuke`](https://github.com/lerna/lerna/issues/195#issuecomment-224743936) be better? 👹

### DIFF
--- a/src/commands/CleanCommand.js
+++ b/src/commands/CleanCommand.js
@@ -1,5 +1,6 @@
+import async from "async";
 import Command from "../Command";
-import ChildProcessUtilities from "../ChildProcessUtilities";
+import FileSystemUtilities from "../FileSystemUtilities";
 import PromptUtilities from "../PromptUtilities";
 import progressBar from "../progressBar";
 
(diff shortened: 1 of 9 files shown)


gigabo:
Output looks like:
image

thejameskyle:
Do you think its necessary to add a confirmation dialog before doing it?

gigabo:
Maybe not. but it _is_ `rm -rf`...

gigabo:
Added support for `--yes`, which reduces output:

```sh
$ lerna clean --yes
Lerna v2.0.0-beta.16
All clean!
```

Happy to remove the prompt altogether if you prefer, too.

thejameskyle:
I'm polling people on it https://twitter.com/thejameskyle/status/740692624839843840

joscha:
Why not a script in package.json that uses `lerna exec -- rm -rf ./node_modules`

thejameskyle:
> Why not a script in package.json that uses `lerna exec -- rm -rf ./node_modules`

Well everyone ends up doing exactly that, so why not just have it included for them?

joscha:
Sure, but then maybe it would be better to make this platform-independent with `FileSystemUtilities.rimraf` instead of calling out to the shell at all.

src/commands/CleanCommand.js
```diff
@@ -0,0 +1,36 @@
...
+      });
+    }
+  }
+
+  execute(callback) {
+    progressBar.init(this.packages.length);
+    this.packages.forEach(pkg => {
+      progressBar.tick(pkg.name);
+
+      ChildProcessUtilities.execSync("rm -rf " + pkg.nodeModulesLocation);
```
joscha: `FileSystemUtilities.rimraf` can be used here.

gigabo: :metal: Thanks, didn't see that.

joscha:
Nice work (8470e3d)!

hzoo:
👍

PR #197, opened 2016-06-09
Docs update exec run

joscha:
Adds information about `lerna exec` the `--scope` flag and fixes `lerna exec` to respect `--concurrency`.

@hzoo PTAL

refs #152


README.md
```diff
@@ -318,6 +318,40 @@ $ lerna run build
...
+`lerna run` respects the `--scope` flag (see below).
+
+```sh
+$ lerna run --scope my-component test
+```
+
+### exec
+
+```sh
+$ lerna exec -- [command] // runs the command in all packages
```
thejameskyle: comments should begin with `#` for the syntax highlighting to be correct

joscha: You are right, I just copied from the run command - both places fixed in de1828c

thejameskyle:
Thanks

Issue #64, opened 2016-02-22
[Bug] README.md is not published to npm


Deadarius:
When I try `lerna publish` - npm registry doesn't display readme, however when I use npm publish - it does.

thejameskyle:
The README where? Individual packages should have their own README, the primary one at the root of the repo will not be published anywhere.

gigabo:
We're actually seeing the same thing.  Assumed we were doing something wrong.  Hadn't considered that it may be related to deploying with `lerna`.

Here's our [package](https://github.com/redfin/react-server/tree/master/packages/react-server) and here's the corresponding [npm page](https://www.npmjs.com/package/react-server) with missing README.

thejameskyle:
Weird... we're literally just [`cd`-ing into the directory and running `npm publish`](https://github.com/kittens/lerna/blob/91a22ecb5bca5fd605209960621173840ea30677/lib/commands/publish/npmPublishAsTemp.js#L42)

gigabo:
Maybe something to do with the package not living at the repo root?

Seems more likely to be an `npm` bug than a `lerna` bug, honestly.

Deadarius:
Yes, I do have readmes... When I was trying to find a problem I noticed that you just run existing npm... have no idea what can be a problem here. Only aware of this issue in outdated version of npm.

Deadarius:
@gigabo But it works fine when I run ```npm publish``` directly...

ianwremmel:
I seem to be running into the same issue. I definitely have a new README, but the changes aren't showing up on NPM.

ianwremmel:
(by NPM, I mean https://www.npmjs.com/)

ianwremmel:
I have a hypothesis that the way lerna messes with dist-tags isn't triggering npm's normal website refresh job. I've opened a support ticket with npm, inc to investigate and will report back here if I hear anything.

Deadarius:
@ianwremmel thank you very much, I still suffer from this issue.

thejameskyle:
This bug is hard to debug as I've had no problems with README's (https://www.npmjs.com/package/cf-component-modal)

ianwremmel:
I heard back from npm. It is indeed a result of the dist-tags technique used by lerna. npm has an open issue: https://github.com/npm/newww/issues/389#issuecomment-188428605

ianwremmel:
[...]

PR #198, opened 2016-06-09
Add --skip-npm flag

gigabo:
Have Lerna update all versions and dependencies where required in `package.json` files without actually publishing the changes to NPM.

This is useful for publishing manually, which is necessary to do periodically to get README updates to show up on npmjs.com (until https://github.com/npm/newww/issues/389 is resolved).

This is a workaround for #64.

This was proposed in #127.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,21 @@ When run with this flag, `publish` will publish to npm without running any of th
 
 > Only publish to npm; skip committing, tagging, and pushing git changes (this only affects publish).
 
+#### --skip-npm
+
+```sh
+$ lerna publish --skip-npm
+```
+
+When run with this flag, `publish` will update all `package.json` package
+versions and dependency versions, but it will not actually publish the
+packages to npm.
+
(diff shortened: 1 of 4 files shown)


thejameskyle:
Nice, this looks awesome. Thanks!

README.md
```diff
@@ -244,6 +244,21 @@ When run with this flag, `publish` will publish to npm without running any of th
...
+
+```sh
+$ lerna publish --skip-npm
+```
+
+When run with this flag, `publish` will update all `package.json` package
+versions and dependency versions, but it will not actually publish the
+packages to npm.
+
+This flag can be combined with `--skip-git` to _just_ update versions and
```
hzoo: Could be useful to give the example of just wanting to push a readme update

dmfenton:
@thejameskyle @gigabo Is there a way to default this setting using lerna.json? We're using lerna for closed-source code that we don't want to publish to NPM at all.

Issue #125, opened 2016-05-09
Bootstrap only missing dependencies


thejameskyle:
In Lerna 2 we switch from running `npm install` in every package directory to installing all the packages individually in order to avoid installing local dependencies that don't exist on npm.

```
$ npm install dep-1@^1.0.0 dep-2@^2.0.0 ...
```

While this improves the initial bootstrap, this also slows down subsequence bootstraps because when running npm install with the dependencies specified will force npm to re-install things it already has in the local `node_modules`.

In order to speed this up we can replicate what npm normally does and check if a dependency is already installed before re-installing it making sure that it satisfies the version in `package.json`.

This means modifying [`BootstrapCommand#installExternalPackages`](https://github.com/kittens/lerna/blob/master/src/commands/BootstrapCommand.js#L93-L113) to perform this check and filter out installed packages.

hzoo:
Fixed in https://github.com/lerna/lerna/issues/199

PR #199, opened 2016-06-09
Only install missing external dependencies during bootstrap

gigabo:
This makes bootstrap in a repo that has been bootstrapped previously much faster.

This addresses #125.

### DIFF
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -7,60 +7,134 @@ import ChildProcessUtilities from "../src/ChildProcessUtilities";
 import BootstrapCommand from "../src/commands/BootstrapCommand";
 import exitWithCode from "./_exitWithCode";
 import initFixture from "./_initFixture";
+import stub from "./_stub";
 
 import assertStubbedCalls from "./_assertStubbedCalls";
 
 describe("BootstrapCommand", () => {
-  let testDir;
 
-  beforeEach(done => {
-    testDir = initFixture("BootstrapCommand/basic", done);
+  describe("dependencies between packages in the repo", () => {
+    let testDir;
+
+    beforeEach(done => {
+      testDir = initFixture("BootstrapCommand/basic", done);
+    });
+
+    it("should bootstrap files", done => {
+      const bootstrapCommand = new BootstrapCommand([], {});
+
+      bootstrapCommand.runValidations();
+      bootstrapCommand.runPreparations();
+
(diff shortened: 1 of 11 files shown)


test/BootstrapCommand.js
```diff
@@ -4,63 +4,179 @@ import path from "path";
...
+import FileSystemUtilities from "../src/FileSystemUtilities";
 import BootstrapCommand from "../src/commands/BootstrapCommand";
 import exitWithCode from "./_exitWithCode";
 import initFixture from "./_initFixture";
+import stub from "./_stub";
 
 import assertStubbedCalls from "./_assertStubbedCalls";
 
 describe("BootstrapCommand", () => {
-  let testDir;
```
gigabo: The diff is cluttered here due to indentation changes (added a nested `describe` layer).

I've left the original test unchanged, and added three additional tests.

hzoo: `?w=1` makes it a clean diff

gigabo: Beautiful.  Thank you.

thejameskyle:
Looks like the tests aren't stubbing out npm completely which is causing it to fail in CI

gigabo:
Yeah, that's strange.

It's stubbing `ChildProcessUtilities.exec`, which is what `NpmUtilities.installInDir` uses.

Not able to repro locally. 😕

thejameskyle:
I can take a look later. It might help if travis did `cat lerna-debug.log` if tests failed, so if you want to modify the script to do that it may help you debug CI.

gigabo:
That whole test was actually invalid.  It was intended to test that dependency _linking_ doesn't happen on warm bootstrap, but this patch actually didn't wind up addressing that (no real need -- it's cheap).

It was only incidentally passing locally because I had a [premature call to `done()`](https://github.com/gigabo/lerna/blob/57935e1a87b358d66ec5017880cdee69d52775d1/test/BootstrapCommand.js#L105) in there.  It was a race that was just never lost on my laptop.

The remaining two new tests exercise the functionality that is actually provided by this patch.

gigabo:
This, plus #204, takes our bootstrap on a warm repo from ~8 _minutes_ to a half _second_. 🚀

thejameskyle:
Should there be a flag to force it to reinstall things?

thejameskyle:
Actually no, I'll wait for someone to request that.

Issue #11, opened 2016-01-07
Nested directory support


patrickheeney:
I currently access packages like this `import { foo } from 'foo/bar';`. It seems the current bootstrap only creates `node_modules/foo/index.js` with `module.exports = require('/full/path/packages/foo');`. Thus when accessing `foo/bar`, it `Error: Cannot find module`. 

What would be the best path forward for supporting this? The only thing I can think of is to find all `*.js` paths and then create that path with matching exports.

thejameskyle:
@kittens any reason we don't just symlink everything?

thejameskyle:
This should be fixed by #34, but be warned that we're not sure if this will cause other issues and we might revert it with another major bump.

thejameskyle:
Symlinking causes other issues with npm, we'll have to figure something else out.

patrickheeney:
This seems to be outside the scope of the project, closing this.

thejameskyle:
It's not, we need to figure it out in another way

evocateur:
I've hit this too, it's especially nasty when you're trying to test stuff that is trying to import a subpath generated by Babel (`const foo = require('my-sibling-package/lib/foo');`). An approach like [npm-sync](https://github.com/neoziro/npm-sync) can work for simple use cases, but clearly doesn't scale for a `lerna`-based monorepo.

Symlinking is it's own ball of wax, since node's `require()` always calls `realpath` on the requested module path and thus it resolves the symlinks and now everything breaks because symlinks are. the. worst.

Maybe expand the [`createLinkedDepFiles`](https://github.com/kittens/lerna/blob/master/lib/commands/bootstrap/linkDependenciesForPackage.js#L5) method to loop over non-npmignored files and require those as well?

thejameskyle:
I want to narrow down the issue in npm and see if it can be patched.

patrickheeney:
@evocateur I thought of the same as well and prototyped locally. One issue I ran into that you need a watcher when adding new files as they would need that shim to link the dependency files.

@thejameskyle I am not sure that npm is the issue here or what the goal would be? The module code is frozen in npm so introducing new behavior to load files locally isn't something they are going to accept (don't have a link to cite this but its in a discussion about linking local modules).

thejameskyle:
@patrickheeney I tried [symlinking previously](https://github.com/kittens/lerna/pull/34) however `npm install` has weird behavior around installing recursively through symlinks which I don't think is intentional.

[...]

PR #204, opened 2016-06-09
[Feature] Support relative dependencies within the repo

gigabo:
We use [relative dependencies in private packages](https://github.com/redfin/react-server/blob/master/packages/react-server-test-pages/package.json#L21-L23) as a crude workaround for #11.

This generates warnings during bootstrap, and prevents us from receiving the full benefit of #199.

This patch causes bootstrap to treat relative dependencies as satisfactory.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -136,6 +136,8 @@ export default class BootstrapCommand extends Command {
   }
 
   isCompatableVersion(actual, expected) {
-    return semver.satisfies(actual, expected);
+    // If it's a relative dependency, then it is satisfied from within the
+    // repository, regardless of what the actual version of the module is.
+    return expected.startsWith("../") || semver.satisfies(actual, expected);
   }
 }
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -138,6 +138,6 @@ export default class BootstrapCommand extends Command {
   isCompatableVersion(actual, expected) {
     // If it's a relative dependency, then it is satisfied from within the
     // repository, regardless of what the actual version of the module is.
-    return expected.startsWith("../") || semver.satisfies(actual, expected);
+    return expected.match(/^..\//) || semver.satisfies(actual, expected);
   }
 }
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -63,4 +63,20 @@ describe("BootstrapCommand", () => {
       }
     }));
   });
+
+  describe("version compatibility", () => {
+
+    const checkPair = (actual, expected) => {
+      return new BootstrapCommand([], {}).isCompatableVersion(actual, expected);
+    };
+
+    it("should consider relative dependencies compatible", () => {
+      assert.ok(checkPair("1.0.1", "../peer-package"));
+    });
+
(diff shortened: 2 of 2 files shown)


thejameskyle:
How do the [relative dependencies](https://github.com/redfin/react-server/blob/master/packages/react-server-test-pages/package.json#L21-L23) work when publishing that module?

gigabo:
We only do it on private modules, so they don't get published.

gigabo:
Should probably actually be an _error_ to do it on a public repo...

patrickheeney:
I tried to add support for this in #10 but it was closed (also is outdated now). This may also work for my use case but it would need to support the other ways to declare relative dependencies:

```
"dependencies": {
    "some-local-dependency": "file:../some-local-dependency"
},
```

It may be worthwhile to use the same logic npm uses internally to determine if this is a local dependency or not.

gigabo:
> the other ways to declare relative dependencies

Oh yeah, good point.  Thanks for the link to `npm-package-arg`.

I've actually realized since opening this that local dependencies are a little trickier than I originally thought.  Specifically, running `npm install ../some-local-dependency` isn't actually correct if `some-local-dependency` _itself_ has non-local dependencies on packages that live in the repo.

For example with deps that look like:

```
package-a => "../package-b"
package-b => "package-c@^1.0.0"
```

If `package-c` lives in the repo it should be linked for `package-b`.

Not sure what to do about that case.

patrickheeney:
I think the first step is something like you have done, which is to make sure the local file based packages can be bootstrapped at all. I haven't looked into it since #10, but back then private / local packages were excluded from `lerna`. 

Then likely the bootstrap process needs to be modified to detect that `package-c` needs to be linked from `package-b`. That is likely related to my other issue #11 since I currently solved it by symlinking all package dependencies as a last step after `npm install` (see Global Cache / Symlink approach in that issue).

gigabo:
> back then private / local packages were excluded from lerna.

They're `npm install`ed now.  It actually works fine on the surface.  It's just with transitive dependencies that it starts to fall apart.

>  the bootstrap process needs to be modified to detect that package-c needs to be linked from package-b. 

Right, but _when_ is it linked?  It can't happen _before_ `npm install`, because the target in `node_modules` gets fresh copies of dependencies.  It can't happen _after_ `npm install`, because the dependency may be _used_ during install (in a `prepublish` script).  So, it would need to be linked _during_ install.  I don't know that `npm` provides the hooks we would need to do this without reimplementing `npm install` at a high level.

patrickheeney:
> Right, but when is it linked?

At the root level this is how I am doing it at the moment:

- Get a list of all project packages
- Symlink them in node_modules project node_modules so they can be resolved
- Diff local packages against package.json and pass non-local ones to npm install
- Do the same for each package

If each package is installed separately then all the other non-local dependencies, then in theory by the time the dependency packages are installed you already have the right local one.

I am also using `pnpm` because of global cache, faster, and easier to pass it CLI arguments. I don't think this easily supports multiple versions of the same local dependency (but could maybe?). I haven't had an issue with scripts but I imagine there is some scenario where this might not work?

evocateur:
@gigabo Can we close this PR now?

gigabo:
Yes!

PR #209, opened 2016-06-11
add-module-exports, set main file in package.json

hzoo:


.babelrc
```diff
@@ -1,4 +1,7 @@
 {
   "presets": ["es2015"],
-  "plugins": ["transform-decorators-legacy"]
+  "plugins": [
+    "transform-decorators-legacy",
+    "add-module-exports"
```
gigabo: Ooh, nice.  Didn't know about that.

hzoo: yeah basically `module.exports = exports['default'];` that was removed in babel 6

PR #210, opened 2016-06-11
Fix version mismatch in `lerna bootstrap`

### LINKED ISSUE #184: version mismatch: actual and expected are the wrong way round
`lerna bootstrap` lists version mismatches in the wrong order:

```
Version mismatch inside "package-B". Depends on "package-A@1.1.1" instead of "package-A@0.3.3".
```

should be:

```
Version mismatch inside "package-B". Depends on "package-A@0.3.3" instead of "package-A@1.1.1".
```

Problem lies in: [BootstrapCommand.hasMatchingDependency](https://github.com/lerna/lerna/blob/2a37e3793837dcb49fc3d016aeb323bf12327ece/src/commands/BootstrapCommand.js#L115)

sotojuan:
Fixes #184 (I think).

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -130,8 +130,8 @@ export default class BootstrapCommand extends Command {
     if (showWarning) {
       this.logger.warning(
         `Version mismatch inside "${pkg.name}". ` +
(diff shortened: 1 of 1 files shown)


hzoo:
Seems like it 🎅, thanks for the PR @sotojuan

PR #217, opened 2016-06-13
Expicitly ignore output from `npm install`

### LINKED ISSUE #213: maxBuffer exceeded because of using exec instead of spawn
My CI builds are failing at `lerna bootstrap` because the npm logging level is set to info instead of warn, which causes a lot of logs and a maxBuffer exceeded error as explained here http://stackoverflow.com/questions/23429499/stdout-buffer-issue-using-node-child-process.

This is because the default maxBuffer is used when using the exec command  (https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_child_process_exec_command_options_callback). This could be fixed by setting a bigger maxBuffer size (but then the next question is, how big this size should be...), or by using `child_process.spawn`.

gigabo:
This output was getting discarded, anyway.  This avoids buffering it
(and potentially failing due to exceeding `maxBuffer`).

This fixes #213.

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -4,13 +4,18 @@ import logger from "./logger";
 export default class NpmUtilities {
   @logger.logifyAsync
   static installInDir(directory, dependencies, callback) {
-    let command = "npm install";
+    let args = ["install"];
 
     if (dependencies) {
-      command += " " + dependencies.join(" ");
+      args = args.concat(dependencies);
     }
 
-    ChildProcessUtilities.exec(command, { cwd: directory }, callback);
+    const opts = {
+      cwd: directory,
+      stdio: "ignore"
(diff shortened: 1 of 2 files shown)


laurentdesmet:
Is there a reason to set the `stdio` to `ignore` instead of `inherit`? Now, there is no visual feedback during the `bootstrap` command, that can take a while. The drawback is of course that the 'npm install' commands are executed in parallel, which causes the output to be interleaved...

gigabo:
> Is there a reason to set the stdio to ignore instead of inherit?

Just preservation of existing behavior.  Currently the only feedback from `bootstrap` is the progress bar that ticks after each package completes.  The output from the `npm install` processes is captured and discarded.  This patch just skips the _capture_ part. ;)

PR #218, opened 2016-06-14
Add `linkedFiles.prefix` option to add @flow annotations to the linked files.

cpojer:
We'd like to add flow annotations to Jest (tomorrow at our internal hackathon :) ).

Flow only looks at files with an `@flow` annotation. If a file doesn't have it, it skips it entirely and doesn't infer types for that file. When using lerna, even if the specific package is flow typed, because the proxy file in between in `node_modules` doesn't have `@flow`, the package won't be typed across boundaries.

This diff adds a `linkedFiles.prefix` option to add such annotations (or anything else, really) to such linked files.

I also added it to the README and updated a test for it. I didn't want to duplicate too much test code and add more fixtures, so I added it to an existing bootstrap test which should be good enough. Let me know if you'd like a separate test, however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,9 @@ Running `lerna` without arguments will show all commands/options.
       "ignored-file",
       "*.md"
     ]
+  },
+  "linkedFiles": {
+    "prefix": "/**\n * @flow\n */"
   }
 }
 ```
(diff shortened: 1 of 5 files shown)


hzoo:
Sounds fine to me

cpojer:
thank you for the speedy merge <3

PR #235, opened 2016-06-20
Fix bootstrap install to use quotes around versions

### LINKED ISSUE #234: Bootstrap doesn't support different ways to specify package version
If I depend on a package like this:

```js
"package": "^1.0.0 || ^2.0.0"
```

The command that gets run is:

```sh
$ npm install package@^1.0.0 || ^2.0.0 ...
```

Which is only installing `package@^1.0.0` and everything after the `||` doesn't get installed

thejameskyle:
Fixes #234

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -109,7 +109,8 @@ export default class BootstrapCommand extends Command {
         return !this.hasDependencyInstalled(pkg, dependency);
       })
       .map(dependency => {
-        return dependency + "@\"" + allDependencies[dependency] + "\"";
+        // Needs "quotes" for versions like "^1.0.0 || ^2.0.0"
+        return `${dependency}@"${allDependencies[dependency]}"`;
       });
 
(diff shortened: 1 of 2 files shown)


src/commands/BootstrapCommand.js
```diff
@@ -109,7 +109,7 @@ export default class BootstrapCommand extends Command {
         return !this.hasDependencyInstalled(pkg, dependency);
       })
       .map(dependency => {
-        return dependency + "@" + allDependencies[dependency];
+        return dependency + "@\"" + allDependencies[dependency] + "\"";
```
gigabo: This is getting into the range where a template string might be more readable:

```javascript
return `${dependency}@"${allDependencies[dependency]}"`;
```

rygine: @gigabo i concur

hzoo: Could leave a comment of an example package version of why it needs the quotes

PR #236, opened 2016-06-20
Add windows support

laurentdesmet:
See #226

Issue #160, opened 2016-05-23
bootstrap installs package peerDependencies when npm install doesn't


javamonn:
Hi all,

I'm running lerna-2.0.0-beta.13 and npm 3.3.12.

It appears that running `lerna bootstrap` will install modules listed as package peerDependencies within the package. Is there a way to configure the behavior of `bootstrap` to match that of `npm@3` and not install peerDependencies by default?

hzoo:
Yeah that probably would have to be an option, I guess opt-in to install peer to match npm 3?

javamonn:
I think there's a separate issue here after playing around a bit more. Agreed that there should be opt-in to match `npm@3` not installing peer dependencies by default, but due to how lerna links local packages - requiring the absolute path of the package rather than fully "installing" it within the `node_modules` of the package listing it as a dependency, the peer dependencies of the depended upon package can never be satisfied as it exists within a different directory tree - `bootstrap` installing peer dependencies by default makes sense. 

This approach works for the majority of dependencies that can be duplicated but I'm currently having an issue with a package `A` that lists react as a peer, and a package `B` that lists react and package `A` as deps. 

After running `bootstrap`, package `B` breaks as react has been included twice. 

After running `boostrap` and pruning the installed  `react` from package `A`'s `node_modules`, package `B` will break on requiring package `A` as it can't resolve the react dependency. This issue doesn't exist when doing a full `npm install B` within `A`.

Is there a different build flow or organization I could use to get around this?

javamonn:
Ended up with a (hacky) working solution:

1. add `react`, `react-dom` to top-level package dependencies
2. `npm install` at top level
3. `lerna bootstrap` to link everything together
4. prune all installs of `react` and `react-dom` from submodule's node_modules with `rm -rf ./packages/*/node_modules/react && rm -rf ./packages/*/node_modules/react-dom`

This seems really fragile and I would love to hear any alternatives :)

hzoo:
That looks like what we're doing in [babel](https://github.com/babel/babel/blob/13c961d29d76ccd38b1fc61333a874072e9a8d6a/Makefile#L59-L62) actually. We don't want to link `babel-runtime` and want to use the 6.x version on npm so we depend on it at the top level and delete the other ones.

lukebatchelor:
@javamonn Thanks! You just saved us so much time trying to debug the same issue.

@hzoo is there any reason that you create the npm install command by appending all the dependencies instead of just running `npm install`?
[...]

Issue #126, opened 2016-05-10
[Feature] Configurable package locations


gigabo:
Hi,

I'd like organization beyond a flat `packages/` directory.

A gulp-style glob array seems natural:

``` json
{
        "lerna": "2.0.0-beta.7",
        "version": "0.2.3",
        "packages": [
                "packages/{core,plugins}/*/package.json",
                "build-tools/package.json",
                "integration-tests/**/package.json"
        ]
}
```

Objections?  Better ideas?

Happy to take it on.

thejameskyle:
I wouldn't be opposed to allowing the `packages` directory to be customizable, especially to account for nested directories. However I'm against allowing multiple directories because it would likely add a lot of complexity and I don't really see a huge benefit

gigabo:
Thanks for considering.  No real interest in just moving the directory here.  Feel free to close.

hzoo:
Ok going to close for now

mxstbr:
Would love to see this reopened, will gladly submit a PR implementing support for this if you see the use for it.

I have that exact use case – over at [Carte Blanche](https://github.com/carteb/carte-blanche) our file layout is a bit different, we have:

```
carte-blanche/
├── package.json
├── plugins/
│   ├── react/
│   │   └── package.json
│   └── source/
│       └── package.json
├── utils/
│   └── package.json
└── webpack-plugin/
    └── package.json
```

Where each plugin (`plugins/xyz/`) is its own package, the core (`webpack-plugin/`) is its own package and we have a few shared utils (`utils/`) that are their own package.

I think in our case it makes sense to be that way, so having an option to change that would be great:

```JS
{
    "packages": [
        "plugins/**/package.json",
        "{webpack-plugin,utils}/package.json"
    ]
}
```

If we want to keep it to a single glob, one could also express those as

```JS
{
    "packages": "{plugins,webpack-plugin,utils}/**/package.json"
}
```

though not sure how the first one is more complex. What do you think?

thejameskyle:
The problem I have with it is that there's no reason it doesn't work within the `packages` organization. You could just as easily have 

```
carte-blanche/
  packages/
    react/
    source/
    utils/
    webpack-plugin/
```

mxstbr:
Yeah I guess it works, but it's quite a bit of work to restructure the build processes – simply saying `"{plugins,webpack-plugin,utils}/**/package.json"` is a lot easier from that perspective.

doug-wade:
[...]

PR #237, opened 2016-06-21
Update bootstrap command (fixes #11, #176)

### LINKED ISSUE #11: Nested directory support

rygine:
* Symlink packages to each package's `node_modules` folder
* Symlink all binaries to packages that depend on them
* Minor refactor of bootstrap prepublish step
* Update `FileSystemUtilities.symlink` to allow type
* Move hasMatchingDependency and hasDependencyInstalled to Package
* Removed `linkedFiles` property of Repository
* Refactor bootstrap unit tests
* Add unit tests for `.hasMatchingDependency` and `.runScript` methods of Package


rygine:
this is a complete refactor of the bootstrap command. i haven't tested on Windows, but it seems like it would work with admin access. not sure the `junction` type is appropriate when linking directories. would love to get some feedback from Windows developers.

my company is using this right now, and it's working great for us (we're on Mac).

i'm sure there's optimizations that can be done. please take a look at the code and let me know what you think.

src/commands/BootstrapCommand.js
```diff
@@ -14,147 +14,345 @@ export default class BootstrapCommand extends Command {
...
+   * Install external dependencies for packages
+   * @param {Object} dependencies
+   * @param {Function} callback
+   */
+  installExternalDependencies(dependencies, callback) {
+    const actions = [];
+    Object.keys(dependencies)
+      .forEach(dest => {
+        const destLocation = dest === "__ROOT__" ?
+          this.repository.rootPath :
```
gigabo: This might be safer to put into `this.repository.packagesLocation`.  It's possible that the repo root has a dependency on a package from within the repo, which would conflict with the symlink.  If the `__ROOT__` deps went into `packages/` then they would have a clean directory with no potential for conflict.

gigabo:
This is awesome.  It's a big change.  It helps with the deep-require problem (#11), but we've got at least one module in `react-server` that doesn't play well with it somehow (still investigating).  I wonder how common that is...

gigabo:
This renders [`lerna exec npm outdated`](https://github.com/lerna/lerna/pull/245#issuecomment-228198964) useless. Npm reports all as `MISSING`. 😬

thejameskyle:
This worries me, it's completely changing how packages interact with `node_modules` in a breaking way.

rygine:
i could take out the "optimization" part and install dependencies in each package's folder like before, then symlink the packages to `packages/node_modules`. perhaps that would make more sense.

rygine:
@gigabo interesting that NPM doesn't traverse directories like Node to look for dependencies. i guess it would make the command a bit more complex than it needs to be. but i think it would be helpful in determining installed versions that same way as Node resolves dependencies.

rygine:
i'm thinking that maybe we could symlink the external dependencies to each package's `node_modules` folder, which i think would fix the `npm outdated` issue.

src/commands/BootstrapCommand.js
rygine: unfortunately, without a `package.json` in the `/packages` folder, NPM won't install anything there. using the `--prefix` and `--global` flags doesn't help much either as it creates `/lib` and `/etc` folders inside `/packages`.

sveisvei:
Im super-keen on getting support for linked binaries, as well as support for more than one entry/main file.

What is the blocker on this issue atm?

rygine:
@sveisvei this PR needs a refactor, which i'm hoping to sort out this week. PR #192 will handle binary linking and is ready to go.

doug-wade:
@giuseppeg I think that would work, but there is one rub: binaries on `PATH` take precedence over locally-installed dependencies, meaning that if you promote a dep to the root, then you can't override it for a local package.  What we do internally at Redfin is specify the path to the monorepo root in package.json, which isn't pretty, but it works, c.f.

```
{
  "scripts": {
    "test": "../../ava/cli.js tests.js"
  }
}
```

which means that if we want to upgrade some packages to a new version of `ava`, we can do so by using their locally-installed version.

giuseppeg:
@doug-wade are you sure about that?
```bash
PATH=$PATH:../../node_modules/.bin npm run my-script
```
works for me i.e. it uses the local bin.
Npm adds `./node_modules/.bin`, I append `../../node_modules/.bin` and if I remember correctly `$PATH` is read left to right

gigabo:
Hi @rygine - Looks like this has diverged from master a little.  Please rebase if you get a chance.

rygine:
@gigabo i'll work on this over the weekend. i have a few adjustments i'd like to make as well so that `npm outdated` works.

rygine:
@gigabo having issues with paths in windows. what am i missing? i thought `path.join` would handle things properly, but it seems like it's using `\\` as the path separator instead of `/`.

gigabo:
Weird.  Looks like the test expectation is `testDir`, which comes from `_initFixture`.  It's using `path.resolve()` to generate.  Maybe that produces different output from `path.join` on windows?

rygine:
`path.resolve` should produce the same thing as `path.join`. i don't have access to a windows machine to test on. not sure how to proceed. 😕

hzoo:
Maybe you can look at https://github.com/lerna/lerna/pull/236 for reference, not sure either

rygine:
i could potentially use `normalize-path`, but that seems hackish to me.

@laurentdesmet thoughts?

rygine:
@gigabo @hzoo finally found the issue!

[L#13](https://github.com/lerna/lerna/pull/237/commits/8bc9e0afc164cf167a049e048f22da478209881f#diff-b8678a10e51138ee78e70339fa5132f4R13) of `src/Repository.js` is getting the `rootPath` from running a git command. The result of running this command was a path that contained `/` (instead of `\`) as the path separator. Running the result through `path.resolve` seems to fix the issue (as seen in the above linked diff).

gigabo:
Nice find @rygine!

I'm pulling this down now to try it out and look through the diff carefully. 🔬

gigabo:
Hmm... ran into [a conflict in the root `node_modules/`](https://github.com/redfin/react-server/blob/master/package.json#L17) right off the bat. 😖 

image

Previously discussed this [here](https://github.com/lerna/lerna/pull/237#discussion_r68322890).  Sounds like the issue was the lack of a `package.json`.  Could we work around that by installing packages explicitly by name?

gigabo:
Actually, as I think about it more, using the `packages/` directory wouldn't make sense with [configurable package locations](https://github.com/lerna/lerna/issues/126)...

Maybe the right answer is just to say if you depend on a package from within the repo at the repo root you'll get the repo _version_ of that dependency going forward?  That would be consistent.

For now I'll just remove those deps and see where I get, but I think we should have some handling for that case one way or another before this lands.  Surely React Server isn't the only repo it would be an issue for.

rygine:
I think your case (and probably others) could be handled gracefully by checking the root `package.json` and comparing versions with the "local" one in the `/packages` folder. There's a couple cases to consider, but I think they could all be covered.

1. The monorepo depends on a version incompatible with the one in the `/packages` folder, which would lead to no symlink in the root `/node_modules` folder.
2. One or more packages depend on a version incompatible with the one in the root `package.json` (if it exists), which would symlink the local `/packages` version to modules that depend on it.

gigabo:
What is this I don't even...
image

Can't run `npm i ...` at the repo root after bootstrap?!

rygine:
I've installed a few modules manually at the root with no issues. Weird.

gigabo:
Just realized another minor pitfall with this approach: It's really easy to miss adding a dep to a package when it starts using it.  If `package-a` depends on `foo`, it will already be available to `package-b`.  I might just throw a `import foo from "foo"` into some file in `package-b` and forget to add the dep.

gigabo:
Oh, also: If I go into `packages/package-b` and run `npm i --save foo` it will put the repo into a different state than would be produced by bootstrap.  Seems like a recipe for weirdness. 😬

rygine:
indeed. this approach requires a lot of diligence with your dependency management, which may be counter-productive. it's been working great at my company. but, we're not using lerna for its intended purpose of managing a public-facing, open-source project with many NPM modules that are being published at regular intervals. we're not publishing any more, and most of the dependencies are shared between the modules. we're mostly interested in the bootstrapping process to link everything together.

to address your missing dependency scenario: if i forget to add a dependency to a package's `package.json` file, our CI would most likely pick it up with a build failure rather quickly (assuming we have test coverage).

it's just as easy to go back to the original approach for package dependencies while still symlinking packages to the root `node_modules` folder. maybe that's the best approach for now?

really looking forward to lerna becoming a lerna repo so we can all make commands that fit our individual organization, and share them with others to adopt or adapt to their liking.

nruhe:
Will this change allow the boostrap command to link dependencies that are in a nested folder structure inside of packages? E.g. /package/package-collection-1/... and /packages/package-collection-2/...?

rygine:
@nruhe i'm not sure i fully understand your question. can you elaborate a bit?

rygine:
@gigabo I changed things up a bit to be more like the current bootstrap while still symlinking packages. No longer symlinking to the repo root. Back to `packages/node_modules` to avoid issues with installing modules that are dependents of symlinked modules.

side note: after running this the first time, I found that more than a few dependencies weren't defined in some packages.  so much for previous theory! 😆 

random thought: i wonder how much power is consumed installing dependencies in each package's folder as opposed to having a centralized "cache". the same for running `lerna clean`. so much disk I/O.

final thought: AppVeyor is sooooooo sllllooooooooowwwwwwww 😒

gigabo:
Awesome, thanks @rygine.  I'll try this out.

Yeah, AppVeyor... 🐌

gigabo:
@rygine - I tried this out with React Server, and it works great!

My main concern now is that using `packages/node_modules/` is incompatible with [configurable package locations](https://github.com/lerna/lerna/issues/126).  Really there are only two viable options when packages could be anywhere in the repo:

- At the repo root
- In the packages themselves

Putting the links at the repo root turned out to cause some weird issues, but what about putting the links in the packages themselves?  Would that work?

In other words, could we just replace the proxy modules that Lerna _currently_ puts in each package's `node_modules/` with directory symlinks?

rygine:
I tried that initially and ran into some webpack issues. Also, Webstorm completely died. It doesn't seem to be able to handle circular symlinks. Fixed by excluding those folders.

I'll explore adding symlinks directly again.

A possible solution to configurable package locations with `packages/node_modules` is to exclude certain folders. Seems to be a pretty standard pattern.

rygine:
@gigabo I've added a commit to symlink directly in each package's `node_modules` folder.

This will symlink every directory in `/packages`. This could be optimized by walking the dependency tree to determine exactly which packages need a symlink. Simply symlinking packages listed in `package.json` is not enough if `package-1` depends on `package-2` which depends on `package-3`. This would mean `package-3` also needs to be symlinked to `package-1` even though it's not listed in `package.json`.

On a positive note, it seems to work just fine. I'm able to bootstrap all my packages and run a dev server for each of them without issue.

Unfortunately, after bootstrapping Webstorm can no longer be used. It simply cannot handle the circular symlinks even though all of my `node_modules` folders have been excluded from indexing. Not sure what I can do to solve this. This will affect many developers. :(

gigabo:
> Simply symlinking packages listed in `package.json` is not enough

Huh.  That's surprising.  Each package has access to its deps relative to its actual location.  Is node somehow caching the path through the symlink as the canonical location for a package and walking up that alternate tree to the root looking for deps?

> it seems to work just fine.

Awesome!  Will try it out soon. 🎉 

> Webstorm

wat

That's... terrible.

gigabo:
> Simply symlinking packages listed in package.json is not enough

Seems to be enough in React Server... what problems did you run into?

I tested with this patch, and it worked great:

```diff
diff --git a/src/commands/BootstrapCommand.js b/src/commands/BootstrapCommand.js
index 6a2a2b4..0ddb3f9 100644
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -193,7 +193,7 @@ export default class BootstrapCommand extends Command {
     const actions = [];
     this.filteredPackages.forEach((filteredPackage) => {
       const packageLocation = path.join(this.repository.packagesLocation, filteredPackage.name);
-      this.packages.map((pkg) => pkg.name)
+      Object.keys(filteredPackage.allDependencies)
         // filter out self
         .filter((pkg) => pkg !== filteredPackage.name)
         // filter out already install dependencies
```

That should help with webstorm, too, right?  There would only be cycles if there were actual cycles in the dependency graph.  That [_happens_](https://github.com/lerna/lerna/issues/299), but it seems not common (or always [intentional](https://github.com/lerna/lerna/issues/299#issuecomment-239658958)).

rygine:
It turns out my dependencies were not fully up to date. A side effect of having everything installed at the root. 😖 

I actually had it like your patch before. Doh!

Everything is working for me!

I need to update the unit tests and include the patch. Will update soon.

README.md
```diff
@@ -485,7 +485,7 @@ $ lerna run --scope toolbar-* test
 
 #### --ignore [glob]
 
-Excludes a subset of packages when running a command.
+Excludes a subset of packages when running the `bootstrap` command.
```
gigabo: Is `--ignore` bootstrap-only?  Would be nice if it worked for other commands...

rygine: It's currently `bootstrap` only based on this conversation: https://github.com/lerna/lerna/issues/117#issuecomment-218952045

I think everyone is open to making it available for more commands, we just haven't discussed which commands would use it.

gigabo: Got it.  Thanks for the background.

boxxxie:
i have an issue with this patch, when i have folders in node_modules that existed from before, they are not removed and replaced with symlinks. i have to manually delete these folders for the bootstrap to make symlinks.

gigabo:
Ooh, that's a good issue @boxxxie thanks for catching that!

@rygine This definitely needs to provide a smooth migration for existing repos.

rygine:
@boxxxie @gigabo `bootstrap` thinks you already have the module/package installed and therefore skips the symlinking. this is for handling the use-case of a package depending on a different version of the one currently in `/packages`. does that make sense?

possible solutions:

1. check the dependency version wanted in `package.json` and if it satisfies the one in `/packages`, remove the folder and then symlink
2. add something to the README to inform users that they should run `lerna clean` before switching to this new `bootstrap`

either way, we should add a note in the README letting users know exactly what is expected.

gigabo:
> check the dependency version wanted in `package.json` and if it satisfies the one in `/packages`, remove the folder and then symlink

This seems right.  If the repo package satisfies the dep then bootstrap should make a symlink.

An alternative would be to check if the repo package satisfies the dep and _warn_ if we want to make a symlink but something else is installed there.  Then a `lerna bootstrap --force` or something could overwrite.  Not sure if that level of caution is really necessary, though.

boxxxie:
thank you, I didn't think about  'lerna clean'

rygine:
@gigabo I added more intelligent checks for already installed dependencies. Check it out and let me know what you think. Need to add more unit tests.

src/FileSystemUtilities.js
```diff
@@ -65,6 +65,14 @@ export default class FileSystemUtilities {
 
   @logger.logifySync
   static unlinkSync(filePath) {
-    fs.unlink(filePath);
+    fs.unlinkSync(filePath);
+  }
+
+  @logger.logifySync
+  static isSymlink(path) {
+    const lstat = fs.lstatSync(path);
```
gigabo: I think this will throw if there's nothing at `path`.  Should this try/catch or guard with `pathExists.sync()`?

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +122,141 @@ export default class BootstrapCommand extends Command {
...
+                    "Replacing with symlink..."
+                  );
+                  // remove installed dependency
+                  actions.push((cb) => FileSystemUtilities.rimraf(installedDepLocation, cb));
+                } else {
+                  // dependency is not compatible with existing package, do nothing
+                  this.logger.warning(
+                    `${dependency} is already installed for ${filteredPackage.name}. ` +
+                    "No compatible package found. Skipping..."
+                  );
```
gigabo: Wait... haven't we already installed repo deps that don't satisfy semver as external dependencies by the time we get here?  We should be able to just filter those out above.

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +122,141 @@ export default class BootstrapCommand extends Command {
...
+                  skip = true;
+                }
+              }
```
gigabo: Should also skip if the symlink exists and is correct, right?

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +122,141 @@ export default class BootstrapCommand extends Command {
...
-    const externalPackages = Object.keys(allDependencies)
-      .filter((dependency) => {
-        const match = find(this.packages, (pkg) => {
-          return pkg.name === dependency;
+  /**
+   * Symlink all packages to the packages/node_modules directory
+   * Symlink package binaries to dependent packages' node_modules/.bin directory
+   * @param {Function} callback
+   */
+  symlinkPackages(callback) {
```
gigabo: Would be nice to have a progress bar in here, even though it's quick.

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +122,141 @@ export default class BootstrapCommand extends Command {
...
+                  // dependency is not compatible with existing package, do nothing
+                  this.logger.warning(
+                    `${dependency} is already installed for ${filteredPackage.name}. ` +
+                    "No compatible package found. Skipping..."
+                  );
+                  skip = true;
+                }
+              }
+            }
+            if (!skip) {
```
gigabo: Seems like this shouldn't need a flag...  Isn't the logic just:

```javascript
if (directoryExists) {
  warnAndemoveDirectory
}
if (wrongSymlinkExists) {
  warnAndRemoveSymlink
}
if (!symlinkExists) {
  makeSymlink
}
```

src/commands/BootstrapCommand.js
rygine: `skip = true` also skips creating the binary links, which we don't to do. it will just re-create the same symlink if it's already there.

src/commands/BootstrapCommand.js
rygine: i initially had one, but i never saw it. so i took it out.

src/commands/BootstrapCommand.js
rygine: we won't need the flag any more as that case will be covered by proper filtering, as you mentioned [above](https://github.com/lerna/lerna/pull/237#discussion_r76464399).

src/commands/BootstrapCommand.js
gigabo: Have you tried on babel? 😁

src/commands/BootstrapCommand.js
rygine: correct! i'll filter them out, which will simplify the logic.

src/commands/BootstrapCommand.js
rygine: the final test! 😆

src/FileSystemUtilities.js
rygine: `init` is the only command to use `unlinkSync`, and it checks for the file first. the confusing part is that it's calling `FileSystemUtilities.unlinkSync(versionLocation, "0.0.0");`.

not sure why that second argument is in there.

not sure this falls under this PR. i'll remove the change. it's out of scope and has no effect with these changes. i'll file a separate issue.

rygine:
@gigabo if you don't blink during the symlinking step, you _might_ see the progress bar 😜  (even on the babel repo)

gigabo:
Hmm... guess it's too fast.  Add a `setTimeout()` in there somewhere? 😈 

That's awesome.  Nice to have _something_ that doesn't need a progress bar.

README.md
```diff
@@ -185,10 +183,12 @@ Let's use `babel` as an example.
...
-- `babel-core/node_modules/babel-generator` is replaced with two files:
-  - A `package.json` with keys `name` and `version`
-  - An `index.js` file with the contents `module.exports = require("relative-path-to-babel-generator-in-the-lerna-repo")`
-- This links the `babel-generator` package in `node_modules` with the actual `babel-generator` files.
+- `packages/babel-core/node_modules/babel-generator` symlinks to `packages/babel-generator`
+- This allows nested directory imports
+
+**Note:** Circular dependencies result in circular symlinks which *may* impact your editor/IDE.
+
+[Webstorm](https://www.jetbrains.com/webstorm/) locks up when circular symlinks are present. To prevent this, add `node_modules` to the list of ignored files and folders in `Preferences | Editor | File Types | Ignored files and folders`.
```
bradleyayers: In IDEA it's `Ignore files and folders` (no **d** in Ignore), perhaps the same in WebStorm?

bradleyayers:
What's left to do on this before it's good to merge? FWIW I've tested it out and it works for me, and the strategy seems like a good balance of functionality and simplicity.

hzoo:
Thoughts on moving this to `lerna bootstrap-link` or moving the old bootstrap to `lerna boostrap-proxy`?

rygine:
@hzoo I think that could be an option. Is having 2 official bootstrap commands confusing to new users? Will this increase issues? Is there something the proxy method provides over (or in addition to) the linking strategy? 

My original thinking was that a strategy could be applied individually to packages based on need, but that just adds another layer of complexity and configuration. See https://github.com/awakenetworks/lerna/blob/bootstrap-strategy/bootstrap_strategies.md.

Once lerna becomes a lerna repo, custom commands will allow new strategies to emerge and the proxy method can live on its own (or the linking method for that matter). I'm really looking forward to this.

There seems to be a definitive need for the linking strategy right now, and I'm not sure how far off the transition to a lerna-style repo will be.

https://github.com/northbrookjs/northbrook uses symlinks too.

bradleyayers:
The only remaining issue I've come across related to this is that `peerDependencies` are installed once-per-package, rather than once globally. Turns out this is existing behaviour (#160) and this PR preserves that behaviour — so it seems fine to move forward with this.

I did have some teething problems switching to this, but that was due to some bespoke scripts I had which manually created symlinks in `node_modules` and confused the logic in this PR. I wouldn't expect most people to do such things, and if they have they shouldn't be surprised by the problems.

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +121,139 @@ export default class BootstrapCommand extends Command {
...
+          // get path to dependency and its package.json
+          const dependencyLocation = path.join(this.repository.packagesLocation, dependency);
+          const dependencyPackageJsonLocation = path.join(dependencyLocation, "package.json");
+          // ignore dependencies without a package.json file
+          if (!FileSystemUtilities.existsSync(dependencyPackageJsonLocation)) {
+            this.logger.error(`Unable to find package.json for ${dependency} dependency. Skipping...`);
+          } else {
+            const installedDepLocation = path.join(filteredPackage.nodeModulesLocation, dependency);
+            // check if dependency is already installed
+            if (FileSystemUtilities.existsSync(installedDepLocation)) {
```
gigabo: Hi @rygine - Just tried this out on my primary React Server repo (been testing on a fresh clone until now).

Ran into some unexpected errors:

image

I think there a few things going on here:

- The `isDepSymlink !== dependencyLocation` check is true when `isDepSymlink` is false.
- I _think_ this whole block should be looking at `dependencyLinkLocation` (defined below) instead of `installedDepLocation`.  In other words it looks like we're intending to check what exists in the location where we want to put a symlink, but are instead checking what exists at the actual package location...

rygine: @gigabo `dependencyLinkLocation` === `installedDepLocation`. my money is on this being a logic error.

i've cleaned it up a bit:

* use the same path variable for the installed dep. location and symlink
* always use the built-in `nodeModulesLocation` on the package (missed this one)
* check for `false` value instead of falsy value (should've known to do this)
* add more clarity when `package.json` of a dep. is missing

`react-server` is bootstrapping successfully on my end now.

gigabo: Yep, that did the trick!

gigabo:
So... I think this is finally ready to go.  Looks good to me! 👍

seansfkelley:
I just tried this PR per the requests on #320 and it doesn't quite work! It looks like the symlinks are using the package name as their target, rather than the filesystem location. I hand-jammed some logging in to see what was up, here annotated for your consumption:

```
# The `Package` object we're linking dependencies for. Note that the location and the name don't line up.
{
  "_package": {
    "name": "@local/legacy-series-search-commons",
    "version": "0.0.0",
    "main": "index.js",
    "scripts": {
      "test": "exit 0"
    },
    "dependencies": {
      "@local/legacy-client-logging": "0.0.0",
      "@local/legacy-frozen-library-singletons": "0.0.0"
    }
  },
  "_location": "/Volumes/git/main/packages/legacy-main-series-search-commons"
}
# The package name of dependency we're trying to link in.
@local/legacy-client-logging
# Where we're looking for this dependency.
/Volumes/git/main/packages/@local/legacy-client-logging
# Where we're looking for this dependency's package.json.
/Volumes/git/main/packages/@local/legacy-client-logging/package.json
# Complain!
Unable to find package.json for @local/legacy-client-logging dependency of @local/legacy-series-search-commons,  Skipping...
```

It looks like originally, `createLinkedDependency` got a path that used `package.location`:  https://github.com/lerna/lerna/pull/237/files#diff-0d9bb533ff943e8e57f7829afef76ab5L115

But in this PR, I don't see mention of that anywhere in the analogous code: https://github.com/lerna/lerna/pull/237/files#diff-0d9bb533ff943e8e57f7829afef76ab5R206

I use scopes heavily, and sometimes the directory names don't match the package names (even the part after the scope) for Reasons.

gigabo:
Oh, good issue @seansfkelley!  Thanks for trying this out and tracking down the problem!

@rygine This will need to handle scoped packages.  Should be pretty simple.

rygine:
@seansfkelley good catch!
@gigabo working on it!

rygine:
@gigabo the regex is primitive, but i'm not sure it matters. ¯\(º_o)/¯ added a scoped package to the basic bootstrap fixture. updated tests.

src/commands/BootstrapCommand.js
```diff
@@ -238,7 +239,7 @@ export default class BootstrapCommand extends Command {
             packageActions.push((cb) => FileSystemUtilities.symlink(dependencyLocation, pkgDependencyLocation, "dir", cb));
             const dependencyPackageJson = require(dependencyPackageJsonLocation);
             if (dependencyPackageJson.bin) {
-              const destFolder = path.join(this.repository.packagesLocation, filteredPackage.name, "node_modules");
+              const destFolder = filteredPackage.nodeModulesLocation;
```
rygine: should've been using this from the start

rygine:
@seansfkelley I just pushed an update that should work for you and the "normal" use case

src/commands/BootstrapCommand.js
```diff
@@ -201,6 +201,7 @@ export default class BootstrapCommand extends Command {
           const match = find(this.packages, { name: dependency});
           return match && filteredPackage.hasMatchingDependency(match);
         })
+        .map((dependency) => dependency.replace(/^(@.*\/)(.*)$/, "$2"))
```
gigabo: Hmm... we should be able to get the location from the package object, which we can grab from the depedency graph.  That's a little more future proof against [configurable package locations](https://github.com/lerna/lerna/issues/126), too.

If you drop this map you should be able to get the package using the dependency name:

```javascript
const dependencyPackage = this.packageGraph.get(dependency);
const dependencyLocation = dependencyPackage.location;
```

rygine: i didn't even think to use the packageGraph! i'll update what i have now to use that.

gigabo: Heh, looks like you did that while I was typing this. :D

rygine:
@gigabo now using package graph

seansfkelley:
Just tried this out. It almost works, but not quite. Here's a sample after I ran it:

```
> cd legacy-search

> cat package.json
{
  "name": "@local/legacy-search",
  "version": "0.0.0",
  "main": "lib/index.js",
  "style": "style.css",
  "scripts": {
    "test": "exit 0"
  },
  "dependencies": {
    "@local/legacy-frozen-library-singletons": "0.0.0",
    "@local/legacy-client-logging": "0.0.0",
    "@local/legacy-series-search-commons": "0.0.0"
  }
}

> ls node_modules/@local
ls: node_modules/@local: No such file or directory

> ls node_modules | grep legacy 
legacy-main-client-logging
legacy-main-series-search-commons
legacy-frozen-library-singletons
```

Two issues with this:

- symlinks aren't put into a scoped `node_modules/@local` as would be appropriate
- symlinks are created with the name of the target directory, rather than the name of the target package

rygine:
@seansfkelley ok, i'm not very familiar with how scoped modules are installed.

is this right?

```
├ /legacy-search
   └ /node_modules
      └ /@local
         ├ /legacy-client-logging
         ├ /legacy-frozen-library-singletons
         └ /legacy-series-search-commons
```

seansfkelley:
Exactly; the scope is rendered as the directory that the `/` in the package name implies that it is.

rygine:
@seansfkelley ok, give it a try now

src/commands/BootstrapCommand.js
```diff
@@ -204,8 +204,8 @@ export default class BootstrapCommand extends Command {
         .forEach((dependency) => {
           // get Package of dependency
           const dependencyPackage = this.packageGraph.get(dependency).package;
-          // get path to dependency and its package.json
-          const dependencyLocation = dependencyPackage.location;
+          // get path to dependency and its scope
+          const { location: dependencyLocation, scope: dependencyScope } = dependencyPackage;
```
gigabo: This seems like it's getting more complicated than it needs to be...

`dependencyPackage.location` is where the module actually _is_.

`path.join(dependentPackage.nodeModulesLocation, dependencyPackage.name)` is where it needs to get _linked to_.

Seems like we have everything we need without running a regexp over anything?

rygine: with scopes involved, the destination folder is actually `node_modules/<@scope>/<name>`, not just `node_modules/<name>`. i wanted a way to extract the scope and use that for building the path.

gigabo: But the scope is part of the name, as returned by [`package.name`](https://github.com/lerna/lerna/blob/master/src/Package.js#L10-L12), right?

It's just returning the `name` value from the package.json, which has the scope in it.

rygine: > But the scope is part of the name, as returned by package.name, right?

😖  correct again! i've been looking at this code too long.

will update

src/commands/BootstrapCommand.js
```diff
@@ -205,7 +205,7 @@ export default class BootstrapCommand extends Command {
           // get Package of dependency
           const dependencyPackage = this.packageGraph.get(dependency).package;
           // get path to dependency and its scope
-          const { location: dependencyLocation, scope: dependencyScope } = dependencyPackage;
+          const { location: dependencyLocation } = dependencyPackage;
```
gigabo: This is cool, btw.  Hadn't seen this destructuring syntax before.

seansfkelley:
Just gave it a go with f11278a68f74d182f49634439b3681e50a29816e and it almost works (again!). The symlinks are in the right place, which generally works, except they take on the name of their destination folder, rather than the package (so in the example above, I end up with `node_modules/@local/legacy-main-client-logging` rather than `node_modules/@local/legacy-client-logging`).

src/commands/BootstrapCommand.js
rygine: found it while trying to figure how to destructure without conflicting with an already existing variable with the same name. cool stuff!

rygine:
@seansfkelley so your package names don't align with your directory names, right?

gigabo:
@seansfkelley - Yeah, f11278a68f74d182f49634439b3681e50a29816e won't do it.  18b61c32fbec18f6068c3ea72a00242c29b2f458 should be right.

seansfkelley:
Yup, looks like it works! 👍

src/commands/BootstrapCommand.js
```diff
@@ -100,150 +121,147 @@ export default class BootstrapCommand extends Command {
...
+            // check if dependency is already installed
+            if (FileSystemUtilities.existsSync(pkgDependencyLocation)) {
+              const isDepSymlink = FileSystemUtilities.isSymlink(pkgDependencyLocation);
+              // installed dependency is a symlink pointing to a different location
+              if (isDepSymlink !== false && isDepSymlink !== dependencyLocation) {
+                this.logger.warning(
+                  `Symlink already exists for ${dependency} dependency of ${filteredPackage.name}, ` +
+                  "but links to different location. Replacing with updated symlink..."
+                );
+              // installed dependency is not a symlink
```
bradleyayers: I think we're missing a line here:

```javascript
packageActions.push((cb) => FileSystemUtilities.rimraf(pkgDependencyLocation, cb));
```

rygine: hmmm. this should mean that there's already a symlink, but it's not pointing to the right location. this invalid symlink will be removed here: https://github.com/lerna/lerna/pull/237/files#diff-0d9bb533ff943e8e57f7829afef76ab5R250

the `createBinaryLink` method removes any existing symlink before creating a new one.

can you go into more detail about why it wasn't working?

bradleyayers: On the few lines above we do

```
// create package symlink
packageActions.push((cb) => FileSystemUtilities.symlink(
  dependencyLocation, pkgDependencyLocation, "dir", cb
));
```

I didn't step through to see if this is exactly where I get the `EEXISTS` error OS error, but I assume it is. Does that seem plausible to you?

bradleyayers:
I hit a new webpack bug using this pull request — https://github.com/webpack/webpack/issues/2952.

We were using `require.context()` to search for React Storybook files in all of our packages, and fell into the symlink cycle problem.

src/commands/BootstrapCommand.js
bradleyayers: @rygine would you like me to open a pull request to this to add this?

rygine: @bradleyayers that's the interesting part, `FileSystemUtilities.symlink` will remove an existing symlink before creating a new symlink. and if `pkgDependencyLocation` exists and is not a symlink, it's removed by `rimraf`.

is your repo opensource? i'd like to try it out with your code base.

bradleyayers: Sure:

```
git clone git@bitbucket.org:atlassian/atlaskit.git
cd atlaskit
git checkout 856a08d
npm install
```

gigabo:
@bradleyayers - Is https://github.com/webpack/webpack/issues/2952 a blocker for you?

bradleyayers:
@gigabo no I wrote [a webpack loader](https://bitbucket.org/atlassian/atlaskit/src/d8d7e0d4166e5242750699d0e106363140441f23/build/storybook/requireStories.js?at=bayers%2Ftypescript&fileviewer=file-view-default) to do the same job, instead of using `request.context()`

src/commands/BootstrapCommand.js
bradleyayers: Here's the error I get:

```
Lerna bootstrap...
Lerna v2.0.0-beta.26
Independent Versioning Mode
Bootstrapping 46 packages
Installing external dependencies
Symlinking packages and binaries
Symlink already exists for akutil-shared-styles dependency of ak-editor-bitbucket, but links to different location. Replacing with updated symlink...
Errored while running BootstrapCommand.execute
Error: EEXIST: file already exists, symlink '/Users/brad/projects/atlaskit/packages/akutil-shared-styles' -> '/Users/brad/projects/atlaskit/packages/ak-editor-bitbucket/node_modules/akutil-shared-styles'
    at Error (native)
```

rygine: @bradleyayers I'm unable to reproduce this on my end. I edited `postinstall.sh` @ L#5 to point to my local `lerna`, which is at the latest commit of my `enhanced-bootstrap` branch. Everything works fine.

rygine: what's weird is that `akutil-shared-styles` is installed normally during the `Installing external dependencies...` step.

bradleyayers: What platform are you on?

rygine: mac, el cap

bradleyayers: I've `git reset --hard` to the latest commit, and am running against that any it works fine now. Perhaps I have a dodgy merge…

Sorry for the confusion and goose chasing!

bradleyayers:
The problem that I thought existed was a ghost. This PR looks all good and is ready for merge + release from my perspective.

gigabo:
Okay, looks like the issues raised by @seansfkelley and @bradleyayers have been addressed.

So... 👍 (again).

@hzoo, @thejameskyle This is a big one.  Solves a lot of problems, but probably introduces new issues we haven't thought of yet.  On balance seems worth moving forward.  Ready?

hzoo:
@rygine Thanks for your patience and work on this one! Hopefully we don't have to use the proxy method anymore

laurentdesmet:
First of all, nice work @rygine (and others that helped with this one), having support for deep-require within Lerna is awesome :-). At work we make heavily use of Lerna and lately we're experiencing performance problems (`lerna bootstrap` and webpack builds running slow) because we have a lot of packages with recurring/common packages (we're building a react/redux based application and this means every package depends on react, redux, redux-observable, etc.). 

I saw that this PR mentioned at some point an optimization where common dependencies (or all dependencies) of each package were moved to the root of the repo. What is the exact reason of not including this optimization? Because it would definitely be helpful for big Lerna projects (actually now, we're applying this strategy by removing commonly used dependencies from all the `package.json`s dependencies before `lerna bootstrap` and adding them back afterwards). I'm aware of the problem with commands in the `.bin` folder, and #245, but are there other issues beside these that I'm missing?

seansfkelley:
@laurentdesmet I have a similar problem with Lerna + Webpack. I figured a neat way to solve it might be to have a Webpack resolver plugin that looks at the filesystem ancestors of a particular module when that module is required, checking if any ancestor has a package that fits the desired version range (preferring those closer to the root), and using that one instead. It wouldn't help install times, but it would reduce duplication/bloat in the final distributable. To my knowledge, such a plugin doesn't exist, but I've been mulling over how to make one.

gigabo:
@laurentdesmet - Thanks for calling that out!

Hoisting common deps to the repo root would definitely be a big improvement for Lerna.

Filed #334 to track.

rsolomon:
Thank you SO much for this commit. It resolves a huge slew of issues I was having with Lerna, including:

* Relative CSS imports between modules
* Exporting and importing flow types
* Reducing boilerplate exports for component library modules

A really great improvement to the usability of Lerna, and upgrading was almost completely seamless. Again, great work.

AlastairTaft:
Just ran into this, I loved lerna specifically because it didn't use symlinks. Symlinks make developing with webpack really hard, it'll start giving you annoying errors like `Module not found: Error: Cannot resolve module 'babel'` because the path of the package is now a level up from your project, this is caused by node's default behaviour of `require` which I think is wrong, see the issue [here](https://github.com/nodejs/node/issues/3402). 

I guess I'm using lerna as an edge case as I like to use it as my main way of developing peer dependent packages to avoid the side effects of the above node 'require' issue. 

So my main gripe is node's require which I don't think they will change, so I wanted a way to develop without using `npm link`. Node have a cli option to `--preserve-simlinks` which helps alleviate the problem, I opened [an issue](https://github.com/nodejs/node/issues/8509) to turn this on permanently. Would be nice having a dev setup that doesn't require loads of custom configuration to get just right though.

bradleyayers:
Could you provide a bit more information about the structure of your project? It's surprising to me that you are having issues with babel.

AlastairTaft:
So I have my website located in `root/packages/luna`. The webpack config inside this folder has a babel loader that looks like this

```es6
loaders: [
      {
        test: /\.jsx?$/, 
        loader: 'babel',
        exclude: /node_modules/,
      }
      // ...
```

Now I have another package as a dependency to luna in the folder `root/packages/cms-controls`. Webpack is now scanning this `cms-controls` folder because that exclude doesn't pick it up anymore because it's symlinked. I could avoid the problem by changing the exclude to `exclude: /(node_modules|cms-controls)/` but that's not a nice solution.

bradleyayers:
It's curious to me that you *wouldn't* want cms-controls to be loaded via babel. In my project all packages use roughly the same conventions and are expected to be loaded in a similar way. 

I'm guessing cms-controls has a different webpack loader configuration, and luna's is not adequate to be able to load modules from cms-controls?

AlastairTaft:
All my third party packages are pre compiled with babel. So they have a src folder but then when I do npm publish everything gets babel'd into a lib folder (when developing `npm start` setups up a watch to do this). So anything consuming my package doesn't require babel it's just ES5.

AlastairTaft:
I think even if I wanted it to be consumed by babel it wouldn't work because the error says it can't find babel now, because it's using the physical path not the symlinked path. So I don't think the problem is with how I've setup my package.

bradleyayers:
The way [our repo](bitbucket.org/atlassian/atlaskit) is setup, is for all packages in `/packages` to be loaded by each other using the same webpack config for development, and then published to a `dist/` directory separately for consuming via ES5. We make use of different package.json mains for this. You can use `webpack` to point `src/index.js` (ES6), and `main` to point to `lib/index.js`.

rygine:
@AlastairTaft you may need to add a resolver to your webpack config.

this is what i have in my webpack config:

```javascript
resolve: {
  root: path.join(process.cwd(), 'node_modules'),
},
resolveLoader: {
  root: path.join(process.cwd(), 'node_modules'),
},
```

not sure if this is your issue, but this helped me.

AlastairTaft:
@bradleyayers Assuming the dist contents for each package are generated with webpack, how would you handle peer dependencies, say you used React, would you ignore it from the bundle? That also got me wondering say if componentA and componentB both depended on packageA, you'd end up with packageA in both component bundles? So you'd inflate your final bundle size? 

I just use babel for third party components so it doesn't bundle the dependencies. I guess you could do that same with webpack by telling it to ignore bundling everything in the node_modules folder.

@rygine Thanks for that, that did seem to work.

bradleyayers:
@AlastairTaft we have a few webpack outputs. In one of them we don't bundle any dependencies , instead we just declare `dependencies` in `package.json`, [mark all of those as external in our webpack config](https://bitbucket.org/atlassian/atlaskit/src/7317ac6142e2c86d16e3b2c1b45387bf3fe6c9ab/build/webpack/externalsMatcher.js?at=master&fileviewer=file-view-default), and let the consumer download those and bundle them however they wish.

We do produce other bundles that *do* include dependencies, and for those common dependencies are indeed duplicated.

seansfkelley:
@rygine when you were experiencing trouble in Webstorm, did you file a ticket on their tracker about it? I'm running into issues with Webstorm as well (https://youtrack.jetbrains.com/issue/WEB-23488#comment=27-1638599) and got myself into a bad state that's not clear if it's caused by circular symlinks or too many {{node_modules}}.

rygine:
@seansfkelley i didn't file a ticket as there were already a few related items. i've added `node_modules` to my ignored files and folders and everything seems to be fine.

seansfkelley:
I did that too, but I use Typescript and that causes it's own host of problems. Time to switch to VS Code!

Issue #240, opened 2016-06-22
Only publish explicitly updated packages


hzoo:
This is lerna 1.x behavior

In babel if we update `babel-types` for a simple type update, it will only update `babel-types` which is ok because we use `^` everywhere for versioning.

In lerna 2.x however, updating `babel-types` will cause all 117 packages to update because babel-types becomes a dependency of all packages if you keep checking nested package.json dependencies.

Could just be a config value/flag?

I don't have a good name for it but just have to change https://github.com/lerna/lerna/blob/master/src/UpdatedPackagesCollector.js#L130 to something like `return this.updatedPackages[pkg.name] || (this.flags.updatedOnly && this.dependents[pkg.name]) || this.flags.canary;`

thejameskyle:
https://github.com/lerna/lerna/pull/241#issuecomment-232465217

PR #241, opened 2016-06-22
Add `onlyExplicitUpdates` flag so that only packages that have change…

hzoo:
…d have version bumps rather than all packages that depend on the updated packages

Just picked a name

Ref https://github.com/lerna/lerna/issues/240

Doing this for backwards compat with how Babel has been released but maybe we don't want to do this?

cc @thejameskyle @loganfsmyth

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,3 +476,16 @@ The `ignore` flag, when used with the `bootstrap` command, can also be set in `l
 
 > Hint: The glob is matched against the package name defined in `package.json`,
 > not the directory name the package lives in.
+
+#### --only-explicit-updates
+
+Only will bump versions for packages that have been updated explicitly rather than cross-dependencies.
+
+> This may not make sense for a major version bump since other packages that depend on the updated packages wouldn't be updated.
+
+```sh
+$ lerna updated --only-explicit-updates
+$ lerna publish --only-explicit-updates
+```
+
+Ex: in Babel, `babel-types` is depended upon by all packages in the monorepo (over 100). However, Babel uses `^` for most of it's dependencies so it isn't necessary to bump the versions of all packages if only `babel-types` is updated. This option allows only the packages that have been explicitly updated to make a new version.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -461,3 +461,14 @@ The `ignore` flag, when used with the `bootstrap` command, can also be set in `l
 
 > Hint: The glob is matched against the package name defined in `package.json`,
 > not the directory name the package lives in.
+
+#### --onlyExplicitUpdates
+
+Only will bump versions for packages that have been updated explicitly rather than cross-dependencies.
```
gigabo: Might be worth including a warning against using this with a major version bump.

README.md
```diff
@@ -461,3 +461,14 @@ The `ignore` flag, when used with the `bootstrap` command, can also be set in `l
 
 > Hint: The glob is matched against the package name defined in `package.json`,
 > not the directory name the package lives in.
+
+#### --onlyExplicitUpdates
```
gigabo: With `meow` should this be `--only-explicit-updates`?

hzoo: Yeah that was looking weird

README.md
```diff
@@ -461,3 +461,16 @@ The `ignore` flag, when used with the `bootstrap` command, can also be set in `l
 
 > Hint: The glob is matched against the package name defined in `package.json`,
 > not the directory name the package lives in.
+
+#### --only-explicit-updates
```
gigabo: Nice, I like the "only" at the front.

gigabo:
Makes sense.  I'll use it.  What I'd _really_ like is for this to be the default behavior for semver-compatible version bumps, but I guess that's tricky... especially with `--independent`. 😬

hzoo:
Yeah we can talk about doing it by default later

thejameskyle:
This is more complex. If you have a package that gets a major bump and its dependents don't get bumped as well then they are no longer linked packages because they don't satisfy semver ranges.

I would revert this until everything is covered

hzoo:
Yeah that is an issue with non-independent mode in general since bumping 1 package means all packages will need to be updated to major version in the next update?

thejameskyle:
I think that is an issue regardless, I think it's best we just force all packages to be published to avoid weird states

PR #244, opened 2016-06-23
Test on all major versions of Node

hzoo:
wait till after dropping node 0.10, 0.12

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
----
 git:
   depth: 1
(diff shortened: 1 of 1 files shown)


thejameskyle:
> wait till after dropping node 0.10, 0.12

why

hzoo:
Ah it's probably since I removed 0.12 from the tests. No reason

hzoo:
updated

PR #250, opened 2016-06-24
Run `prepublish` scripts during bootstrap

gigabo:
Bootstrap should leave packages with usable dependencies.

If a dependency's `prepublish` script hasn't been run then it may not be `require()`-able.

> [The prepublish script will be run before publishing, so that users can consume the functionality without requiring them to compile it themselves. In dev mode (ie, locally running npm install), it'll run this script as well, so that you can test it easily.](https://docs.npmjs.com/files/package.json#devdependencies)

Furthermore, if packages use each other _during_ `prepublish`, then order of execution is important.

This, for example, is not a viable workaround:

```bash
$ lerna bootstrap
$ lerna run prepublish
```

Because a package may require a not-yet-built module due to incorrect order of execution.

This patch runs each package's `prepublish` during bootstrap, in dependency order.

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -77,6 +77,9 @@ export default class PackageUtilities {
       if (!packages.length) {
         throw new Error(`No packages found that match '${glob}'`);
       }
+    } else {
+      // Always return a copy.
+      packages = packages.slice();
     }
     return packages;
   }
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -30,19 +30,62 @@ export default class BootstrapCommand extends Command {
 
     const ignore = this.flags.ignore || this.repository.bootstrapConfig.ignore;
 
-    async.parallelLimit(PackageUtilities.filterPackages(this.packages, ignore, true).map(pkg => done => {
-      async.series([
-        cb => FileSystemUtilities.mkdirp(pkg.nodeModulesLocation, cb),
-        cb => this.installExternalPackages(pkg, cb),
-        cb => this.linkDependenciesForPackage(pkg, cb)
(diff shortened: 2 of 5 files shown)


joscha:
nice, this would actually solve some problems we have around this!

src/PackageUtilities.js
```diff
@@ -77,6 +77,9 @@ export default class PackageUtilities {
       if (!packages.length) {
         throw new Error(`No packages found that match '${glob}'`);
       }
+    } else {
+      // Always return a copy.
+      packages = packages.slice();
```
thejameskyle: 👍

src/commands/BootstrapCommand.js
```diff
@@ -30,19 +30,62 @@ export default class BootstrapCommand extends Command {
...
+    // dependent.  So the bootstrap proceeds in batches of packages where each
+    // batch includes all packages that have no remaining un-bootstrapped
+    // dependencies within the repo.
+    const bootstrapBatch = () => {
+
+      // Get all packages that have no remaining dependencies within the repo
+      // that haven't yet been bootstrapped.
+      const batch = todoPackages.filter(pkg => {
+        const node = filteredGraph.get(pkg.name);
+        return !node.dependencies.filter(dep => !donePackages[dep]).length;
```
thejameskyle: How does this handle dependencies of dependencies that haven't been bootstrapped?

gigabo: Packages remain queued until their immediate dependencies have been bootstrapped.

Given:

```
A -> B
B -> D
C -> D
```

Batch order will be:

```
[D]
[B, C]
[A]
```

This batch scheduler assumes that the dependency graph is acyclical.

Issue #229, opened 2016-06-17
[Bug] Not passing a value for `--repo-version` sets the package.json version to `true`


jeffdavidgreen:
In `PublishCommand.js`, when setting the flag for `--repo-version`, if you don't set a value for the version, it changes all the updated packages version numbers to true.

The code I'm referring to is here:

``` javascript
if (this.flags.repoVersion) {
    return callback(null, {
        version: this.flags.repoVersion
    });
}
```

The command I ran is `lerna publish --repo-version`. I was thinking about this and I assumed that if I did not specify a version, then maybe the code could use the version found in each `package.json` file for the individual packages in the repo. For our workflow, we cannot have the prompts for the deployment scripts, so we were thinking about having developers increment the versions manually using `lerna publish --skip-npm` and then running on the build and deployment server `lerna bootstrap && lerna publish --skip-git --yes --repo-version`, but then we ran across this problem.

One solution I thought of is to include logic where if you don't add a version number as a value for `--repo-version`, then it will use the package.json version for each package.

The other solution is to include a new flag like `--package-version` that will do what I'm hoping to accomplish.

davidfateh:
I am also seeing this bug.

hzoo:
`getVersionsForUpdates` gets the next version to update to. If you're package is currently v1.0.0 and update it and do `lerna publish --repo-version` it doesn't know what version you want to update to. If you picked the same version you would get an error that it is trying to publish an existing version.

If you look at the other if/else branches in that code, it does 4 different checks otherwise (independent vs non-independent and canary vs non-canary). Another option that I want to add in would be using a github label/commit prefix to specify patch/minor/major so you wouldn't need a prompt.

davidfateh:
A commit prefix might be helpful for when a user wants to update a version but doesn't want to use the prompt... Let's say the dev makes a PR, updates a package version and simply wants to use Lerna to then publish that package after the PR (with version number) has been approved.

hzoo:
> so we were thinking about having developers increment the versions manually using lerna publish --skip-npm

I see, so you already updated the package.json version and want a way to use that, sounds ok to me

davidfateh:
> I see, so you already updated the package.json version and want a way to use that, sounds ok to me


That seems to allow the dev to determine the individual package version while still utilizing lerna for its mass-publishing capabilities. This could help automate flows with build tools instead of having the build tools interface with the prompt (or some version of that).

davidfateh:
[...]

PR #252, opened 2016-06-25
Added --package-version

davidfateh:
This is in reference to #229 on the discussion to include a `--package-version` command.

**What this does:**
Running `lerna publish --package-version` will assume that each package that has changed has the correct version already in its package.json file.

**Why have it?**
Here's my use case. My team has a repo that has multiple packages in it and is managed by Lerna. However, when a teammate submits a PR for a package (within the main repo) they will discuss with the team, in the PR, what kind of version this package should be (major, minor, patch). This way the team can decide on the correct version for the package and leave less room for versioning error.

**The main part**
When the PR is accepted and merged to master, our build tools will run Lerna and use this `--package-version` command to publish our package updates without having to interface with the Lerna versioning prompt that one would get in normal mode.

I hope this makes sense. All feedback is appreciated, thank you and I hope this is useful for others!

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -142,6 +142,18 @@ export default class PublishCommand extends Command {
       });
     }
 
+    // Declare that each package has an updated and correct version
+    if (this.flags.packageVersion) {
+      const versions = {};
+      const version = this.globalVersion;
+
+      this.updates.forEach(update => {
+        versions[update.package.name] = update.package.version;
+      });
+
+      return callback(null, { version, versions });
+    }
+
     // Non-Independent Canary Mode
     if (!this.repository.isIndependent() && this.flags.canary) {
       const version = this.globalVersion + this.getCanaryVersionSuffix();
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -142,14 +142,14 @@ export default class PublishCommand extends Command {
       });
     }
 
-    // Declare that each package has an updated and correct version
+    // Declare that each package already has an updated and correct version
     if (this.flags.packageVersion) {
-      const versions = {};
       const version = this.globalVersion;
 
-      this.updates.forEach(update => {
-        versions[update.package.name] = update.package.version;
-      });
+      const versions = this.updates.reduce((vers, update) => {
+        vers[update.package.name] = update.package.version;
+        return vers;
+      }, {});
 
       return callback(null, { version, versions });
     }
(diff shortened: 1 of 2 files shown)


thejameskyle:
`--package-version` is ambiguous to me, suggestions for a better name?

davidfateh:
The idea is, when passed this flag, Lerna will assume the packages in project have the correct version already. I'm going to list some names that come to mind based on that idea.

1. --use-package-versions
2. --get-package-versions
4. --imply-versions
5. --imply-package-versions

Let me know if any of these make sense; I'm open to suggestion here. If any more names come to me I'll comment again.

thejameskyle:
I wonder if this can be combined with `--repo-version` to make something more pleasant. Even allowing you to specify a version for the prompts.

```sh
$ lerna publish --version [repo,package,major,minor,patch,1.2.3]
```

davidfateh:
I am specifically looking to bypass the prompts and use the versions found within each package.json.

thejameskyle:
I'm aware of that, I've read what you've written and what I suggested covers that use case.

davidfateh:
Sorry, I think I misinterpreted what you meant. 

To me it seems like your suggestion would require a command to be run that would look something like `lerna publish --version package` and that would to what do what I'm proposing `--package-version` to do?

thejameskyle:
Yes, that's what it would do, I'd rather have one flag that combines multiple options of the same type of action than have an individual flag for every single different use case.

davidfateh:
I see, sorry for the confusion. 

That seems like a bigger dig to me. I wouldn't mind taking a crack at it but is this something that requires a bigger discussion?

jeffdavidgreen:
I like the idea of having one flag to rule them all. I think this helps with the situation of a user trying to do `lerna publish --repo-version --package-version` and not knowing which one takes precedence.

doug-wade:
I dig the `lerna publish --version [repo,package,major,minor,patch,1.2.3]` api -- I think its significantly clearer and easier than two separate args.

thejameskyle:
@davidfateh No need to land everything. If you just want to implement the following, that would be fine.

```sh
$ lerna publish --version [repo,package]
# so that
$ lerna publish --repo-version
# becomes
$ lerna publish --version repo
# and
$ lerna publish --package-version
# becomes
$ lerna publish --version package
```

davidfateh:
@thejameskyle Sounds good! I will start on this soon as I'd like to see this functionality come to Lerna. Thanks for taking the time to review this and provide feedback.

hzoo:
Looking at the other publish flags, should we also change `lerna publish --canary` to `lerna publish --version canary` or something?

davidfateh:
Yeah I think that would get included. I forget (I'll look later) but I think canary would be needed for repo version no?

davidfateh:
It looks like canary could be added as an argument to the `--version` command. 

And to clarify, it isn't needed for repo-version like my comment asks above.

davidfateh:
I've created another PR with the updates that were suggested here.

https://github.com/lerna/lerna/pull/262

thejameskyle:
> Looking at the other publish flags, should we also change lerna publish --canary to lerna publish --version canary or something?

@hzoo I don't know how that would work with https://github.com/lerna/lerna/issues/277#issuecomment-232456862

Issue #174, opened 2016-06-03
[Internal] Lerna should be a lerna repo


gigabo:
Straw man proposal: Every lerna command should be a dedicated package in a lerna repo.
- `lerna-bootstrap`
- `lerna-publish`
- ...

They should live along side:
- `lerna-core`: Exposes a programmatic API (`Package`, `Command`, `Repository`, etc).
- `lerna-cli`: Exposes commands via `lerna <command>`.

Benefits:
- Lerna can serve as its own demonstration.
- Can [share test helpers](https://github.com/lerna/lerna-changelog/pull/18#issuecomment-223697607) and other infrastructure between packages.
- Can generate a changelog using [`lerna-changelog`](https://github.com/lerna/lerna-changelog).
- Self hosting is cool.  So is eating your own dog food.
- Commands can be pluggable.
  - A dev can publish `lerna-frobnicate` and let people `lerna frobnicate`.
  - Awesome commands can be easily pulled into the project with [`lerna import`](https://github.com/lerna/lerna/pull/173).

There it is.  Tear it apart. :feelsgood:

thejameskyle:
Hmm that would be interesting to play around with

Alxandr:
This has been sitting here for more than half a year with one comment and lots of positive reaction-emojis. Should this maybe be added to the new v2.0 milestone? Also, do you want someone to work on this?

evocateur:
I'd be interested in driving this, I have some hacking time at work coming up soon.

dylang:
Would be helpful to have an api to get the array of package paths.

light24bulbs:
What is the state of the programmatic API?  Was any progress made on invoking lerna from nodejs directly?

evocateur:
Aside from a lack of explicit documentation, all of the individual commands are callable from node directly. All of the internals previously exposed as named exports are also available as distinct packages.

That being said, there is no (useful) main entry of the `lerna` package itself, and that could certainly be improved.

light24bulbs:
@evocateur that's interesting!  Are you saying I can do `const lerna = require('lerna')`?  Doesn't seem to be working for me. 
What are the commands that are exposed?  For posterity, can you give an example of invoking a Lerna command from node?

Scriptability could go a long way towards making Lerna more extensible.  If you look at a package like Webpack, it eventually won out because it was SO extendable.

evocateur:
@light24bulbs 

> Are you saying I can do `const lerna = require('lerna')`?

[...]

Issue #138, opened 2016-05-16
[Feature] Using Lerna Locally


thejameskyle:
I want to encourage people to run lerna locally, but it's kinda annoying using a CLI based tool with tons of commands locally (you end up aliasing all of them to npm scripts or something else and no gross)

So instead I want to make it so that the global `/usr/bin/lerna` will run the local `./node_modules/.bin/lerna` if it exists.

It'd be nice if this could work independently of versioning, so any version of the global CLI can run any version of the local CLI.

I'm considering just making this another package `lerna-global` that is super dumb: (pseudo code)

``` js
require('child_process').exec(process.argv.slice(1).unshift('./node_modules/.bin/lerna'));
```

thejameskyle:
https://github.com/timoxley/npm-run

scriptjs:
I am wholeheartedly behind running locally but in the way all other modules currently work instead of any tricks with 'usr/bin'. Personally, I dislike -g with anything node and avoid preferring anything global.

I don't use `/usr/bin` at all when I must install something global. bins from global installs are instead executed from  `/.local/bin`. Global installs do not need sudo in this case. Like many that develop commercially, we rely on sinopia/verdaccio. The npm client is often upgraded more often than node.

thejameskyle:
I want to encourage as many people to install locally as possible. I wouldn't mind having a separate `lerna-cli` package just to push people in the right direction. Other packages like gulp have had success doing this.

scriptjs:
@thejameskyle babel-cli also

thejameskyle:
babel-cli doesn't resolve to the local version if you have one installed, it depends on babel-core itself

tjbenton:
All you have to do is add `./node_modules/.bin` or `$(npm bin)` to your `PATH` in your bash files. I know this works quite well for mac, because I've been using it for a couple of years and it's allowed me to not install any global dependencies for my projects.

Here's how to add it to your path in your dot files

```sh
PATH=./node_modules/.bin:$PATH # adds locally installed node modules to the path
export PATH
```

You can also make an [function](http://stackoverflow.com/questions/9679932/how-to-use-package-installed-locally-in-node-modules) for it if you don't like messing with the path

```sh
function npm-exec {
   $(npm bin)/$@  
}
```


If you're using a `Makefile` You can put this at the top of it and that also adds it to your path.

```make
PATH := $(npm bin):$(PATH)
SHELL := /bin/bash
```

So the following would work 

```make
bootstrap:
[...]

PR #255, opened 2016-06-27
[Feature] Add watch command

rygine:
The `watch` command will watch for changes in your package folders and run a script when it detects a change. It expect at least 2 arguments: `[glob]` and `[command]`.

Example: `lerna watch 'src/**/*.js' build`

In the example above, `npm run build <path/to/changed/file>` will run in the package folder of the file that changed. By default, the relative path to the changed file will be passed as the first argument of the command. You may specify additional arguments to include when running the watch command.

Example: `lerna watch 'src/**/*.js' test arg1 arg2 arg3`

This will run `npm run test <path/to/changed/file> arg1 arg2 arg3`.


rygine:
A few things to note:

* the colorized output when a file changes is up for debate
* unit tests aren't included as i don't think they are needed
* when an error occurs, the watcher is stopped and the error is passed to the callback
* the first argument is the relative path of the changed file, this could be an absolute path

joscha:
can't this be done via `lerna exec -- pwd | sed 1,2d` and passing the returned folders to chokidar?

thejameskyle:
ehh– you don't have a watch task running to see whenever you need to run `npm install` I don't see a need for this

rygine:
@thejameskyle the watch command here is intended to run the build command within packages as i'm making changes. this is currently not possible with babel. sure, i can watch files in 1 package and have them build, but not all packages at the same time. am i missing something?

this way, i'm free to work on multiple dependent packages at the same time and have them rebuild and hot reload into my webpack dev server. i find this highly useful.

i never intended the watcher to run `npm install` and don't see a use case for that.

doug-wade:
Since `npm install` calls `prepublish` internally, you could use `npm install` to run the build in watch mode (and check dependencies as a side-effect).

rygine:
@doug-wade i don't understand your use case. if all of my modules are bootstrapped, and i want to start watching for changes at an arbitrary time, how do i accomplish that?

doug-wade:
@rygine Consider [this commit](https://github.com/redfin/react-server/pull/302/commits/be2be2ab93dfda2c699c80645c02c85b23167d9b) that didn't merge to React Server.  Initially, as part of `prepublish`, I babelified the sources of my babel plugin, and then exported the transpiled javascript as `main`.  @thejameskyle says that we don't need a watch task to see when we need to run `npm install`, but in this case, a `watch` mode that would detect changes to `src` and run `npm install` because [`npm install` runs `prepublish`](https://docs.npmjs.com/misc/scripts#description).  With your implementation, it would make more sense to use `watch` to run `npm run prepublish`, but even running `npm install` would be a huge improvement.

I've been implementing watch tasks as scripts, and then using `lerna exec -- npm run watch`, which works pretty okay, but is a lot of boilerplate.

rygine:
@doug-wade running `npm install` after a source file change makes absolutely no sense to me. i have no idea why it was even mentioned as a use case here. this is simply for transpiling changed source files so that it's picked up by a dev server during development.

if i'm working on `packageA` which depends on `packageB` and `packageC`, i'd want to see the realtime changes of `packageB' or `packageC`. with this watcher, my webpack setup would pick up the changes and hot reload them for me.

note: the `main` entry for all my packages points to a transpiled version in the `/dist` folder.

paulyoung:
A `lerna watch` command seems incredibly useful.

I tried `lerna exec -- npm run watch` and it seemed to get stuck on the first package since the `watch` command doesn't exit.

In my case I would probably set `"prepublish": "npm run clean; npm run build"` and then run `lerna watch 'src/**/*.js' prepublish`.

I'm currently manually building individual packages when I make changes so that other dependent packages can get the latest. This is sometimes several levels deep.

If there's a way to achieve the same thing without these changes, that would be great and I'd love to know how.

Perhaps I'm missing something, but I'm afraid I don't at all follow why `npm install` was mentioned.

thejameskyle:
> the watch command here is intended to run the build command within packages as i'm making changes

I missed this, when I first read this I thought it was a watch version of `lerna bootstrap`. I blame having to wake up at 5:30am this morning

joscha:
I like the idea of watching, but it seems to me as if this does not necessarily need to be a first-class citizen of lerna, as it can be composed, e.g.
`lerna exec -- pwd | sed 1,2d | sed -e 's/$/\/**\/*.js/' | xargs npm run watch -- --files`

rygine:
@joscha is lerna not all about convenience and ease of use? i believe the `clean` command is a product of this thinking. :) while i'm sure some bash magic can achieve a similar end, it can look rather confusing to those not familiar with the command line. not to mention, the above command may not work on all platforms, creating a need to conjure up another way to achieve the same task.

how are you currently handling file changes during development?

i've only rarely used `sed` and in the most basic way. a quick googling tells me it's not something you master in a few minutes. i was unable to get your command to do anything. i'm sure it's something on my end.

joscha:
I get the convenience part ;) I think for me mass file watching is a relict of the past in most cases. Modern build tools would allow you to pass some entry points and watch the whole dependency graph, no matter what exact location those files are in - globs and directory thinking make refactors harder, so I prefer an `import { x } from 'module'` (that module gets then added to the watched graph) over an explicit `npm run watch module/**/*.js`.

rygine:
@thejameskyle @hzoo @gigabo @joscha is this still something to be considered for lerna right now? i know there's a push to gets things merged so that #174 can be implemented. in which case,  i can close this now and it can come back later as `lerna-command-watch`.

nruhe:
I would definitely like to know the status on this. We have major need of the ability to run our prepublish script after changes so that the requiring repos can get the latest content in dist.

rygine:
@nruhe that's exactly what my team is doing with this command. i just rebased. i'd love to get more feedback from others.

gigabo:
Yeah, this actually seems really sweet to me.

Our workflow on React Server involves a lot of:

```bash
$ cd ../<other-package> && npm run prepublish && cd - && npm start
```

This would really help streamline that.

> i still need to add documentation.

And tests? 😁

rygine:
yes, tests too!

i'm thinking the only thing i could test is the value of `this.packagesToWatch` for packages that don't have the script. everything else seems to be covered.

src/commands/WatchCommand.js
```diff
@@ -0,0 +1,72 @@
...
+      const [pkg] = path.relative(this.repository.packagesLocation, src).split(path.sep);
+      const packageLocation = path.join(this.repository.packagesLocation, pkg);
+      const srcFile = path.relative(packageLocation, src);
+      console.log(`${chalk.cyan(`${pkg}/${srcFile}`)} changed, running ${chalk.yellow(this.script)} script`);
+      NpmUtilities.runScriptInDir(this.script, [srcFile].concat(this.args), packageLocation, (err, stdout) => {
+        if (err) {
+          this.logger.error(err);
+        } else {
+          this.logger.info(stdout);
+        }
```
gigabo: Probably good to get `stdout` on error here, too.

Can just follow the pattern [you established](https://github.com/lerna/lerna/pull/265/files) in `RunCommand.js`.

src/commands/WatchCommand.js
```diff
@@ -0,0 +1,72 @@
...
+  execute() {
+    this.logger.info(`Watching for changes in package files that match '${this.glob}'`);
+    const globPath = (pkg) => path.join(this.repository.packagesLocation, pkg.name, this.glob);
+    // watch all package files matching passed glob
+    chokidar.watch(this.packagesToWatch.map(globPath)).on("change", (src) => {
+      const [pkg] = path.relative(this.repository.packagesLocation, src).split(path.sep);
+      const packageLocation = path.join(this.repository.packagesLocation, pkg);
+      const srcFile = path.relative(packageLocation, src);
+      console.log(`${chalk.cyan(`${pkg}/${srcFile}`)} changed, running ${chalk.yellow(this.script)} script`);
+      NpmUtilities.runScriptInDir(this.script, [srcFile].concat(this.args), packageLocation, (err, stdout) => {
```
gigabo: Hmm... does it always make sense to automatically pass `srcFile` as an argument?

What if I'm just doing `lerna watch src/* prepublish`?

rygine: Good point. My thought was that `srcFile` would just be ignored if it's not needed. In my environment, I transpile single files as they are changed as Webpack freaks out when you remove and replace the `/dist` folder while it's running.

Would it make sense to have a flag for this?

Here's the watch command I run:

```
lerna watch 'src/**/*.@(js|css)' build -- '-- --incremental'
```

I know it looks weird. We have a tool that wraps `babel` and by default, it cleans `/dist`, then transpiles everything in `/src` to `/dist`. The `incremental` flag take a source file argument and just transpiles that.

src/commands/WatchCommand.js
rygine: Good catch!

src/commands/WatchCommand.js
gigabo: Yeah, the file argument breaks if it's not expected:

image

One option would be to make the file available as an environment variable, similar to [this](https://github.com/lerna/lerna/pull/276/files).

So if I wanted to use it I could say:

```bash
lerna watch src/* eslint -- --fix $LERNA_CHANGED_FILE
```

That would also allow custom placement of the source file argument _within_ the arguments array, so it's not just forced to be at the beginning.

rygine:
Should the `watch` command also take flag values from `lerna.json`?

```js
{
  "lerna": "2.0.0-beta.26",
  "version": "1.0.0",
  "watchConfig": {
    "ignore": "my-ignored-component"
  }
}
```

gigabo:
Maybe?  Maybe add it if someone requests?

rygine:
@thejameskyle does this feature fit into the scope of what Lerna is trying to do?

thejameskyle:
Yeah, the addition of watch tools makes it a bit heavy of a dependency but that could be easily fixed with in the long term plan I had for Lerna having a global CLI that calls out to an installed package (https://github.com/lerna/lerna/issues/138). The global cli can have all the watch tools installed globally and be used in tandem. Will just require some refactoring later.

nruhe:
What's the ETA on merging this?

nruhe:
Sorry to keep pushing on this one, but we're in the process of adopting a monorepo and we -really- need the ability to have the system run build/prepublish commands on changes (e.g. npm prepublish). @thejameskyle can you comment on whether this will be added any time soon or if we should find an alternative?

gigabo:
@rygine Want to dust this off? 😁

rygine:
sure thing. will take a look in the next few days.

nruhe:
Thank you very much. :)

rygine:
@nruhe @gigabo i've updated the PR. i _think_ it's good to go. all tests seem to pass.

@gigabo i don't have a test for when the `ignore` and `scope` flags are set. i didn't see this being tested in other commands that support both. is this something we should be testing?

gigabo:
> i don't have a test for when the `ignore` and `scope` flags are set.

Always nice to have more tests, but I don't think we need to hold this PR up for that.

reggi:
Just tried to run this and got an error :(

```
npm i https://github.com/awakenetworks/lerna.git#a251ead423c77fc8e13185f70a052eaa8d6adf93 -g
```

```
Thomass-MBP-2:wonder thomas$ lerna
module.js:472
    throw err;
    ^

Error: Cannot find module '../lib/index'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/thomas/.nvm/versions/node/v7.2.1/lib/node_modules/lerna/bin/lerna.js:3:13)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
```

rygine:
@gigabo the AppVeyor failure seems to be completely unrelated. :/

reggi:
This might be useful! https://github.com/mateodelnorte/loop-things

doug-wade:
@rygine the appveyor failure is likely unrelated.  We've [dropped the version of node that failed](https://github.com/lerna/lerna/commit/79b29a30a5b60f3701d59c8ccd70e21febe38d43) -- wanna rebase and try again?

bin/lerna.js
```diff
@@ -10,25 +10,26 @@ var cli = meow([
...
+  "  publish                Publish updated packages to npm",
+  "  updated                Check which packages have changed since the last release",
+  "  import                 Import a package with git history from an external repository",
+  "  clean                  Remove the node_modules directory from all packages",
+  "  diff                   Diff all packages or a single package since the last release",
+  "  init                   Initialize a lerna repo",
+  "  run                    Run npm script in each package",
+  "  exec                   Run a command in each package",
+  "  ls                     List all public packages",
+  "  watch [glob] [script]  Watch for changes to package files and run a script when a change is detected",
```
evocateur: Oy, we really have to get away from this fiddly manual formatting. (urge to refactor into [yargs](https://github.com/yargs/yargs/) rising...)

None of the other commands list their positional arguments here, can we stay consistent with that until we move to a better argparser?

Bnaya:
Maybe i've missed something in all of the discussion,
But:
Why not add support running npm tasks in the packages in parallel? 
say, each package has it own `watch` task, and in the task you have `babel -w`

lerna exec claims to run tasks concurrently, but as i see it block on the first package watch command

gigabo:
@Bnaya Yep, that's an alternative way to get at good watch behavior in a Lerna repo.

There's some discussion about that here: https://github.com/lerna/lerna/issues/446.

PR #257, opened 2016-06-28
Bootstrap: copy `main` property from package.json

LoicMahieu:


gigabo:
Wait... doesn't this break the proxy module?  Bootstrap is still _writing_ to an `index.js` that requires the relative path to the actual module.  This will tell node to look somewhere _else_ for the proxy module's main file.

Why does the proxy need to mimic the actual module's structure at all?

gigabo:
So it turns out this _doesn't_ break the proxy module, because node falls back to `index.js` as a default if the module pointed to by `"main"` in package.json isn't found.

I'm still left wondering what the point of this is?

hzoo:
Yeah there wasn't any description/context

thejameskyle:
oh man– I was fucking everything up this morning.

PR #261, opened 2016-06-29
Just emit usage when run with no command

gigabo:
Skip the "Invalid lerna command: undefined" error message.

image
<p align="center">⬇️ </p>
image


hzoo:
Oh 👍

PR #262, opened 2016-06-30
[Feature] Adding the --update-version command

davidfateh:
This PR is in reference to [another PR that I opened recently](https://github.com/lerna/lerna/pull/252).

**What this adds**
I've added an `--update-version (-u)` command that can take two arguments right now `[repo, package]`.

`--update-version package` is the main addition here. In the reference PR I describe the use cases for it.

I've updated the documentation to reflect my changes.

**What this removes**
`--repo-version` has been removed as a main flag and is now `--update-version "x.x.x"`

Please let me know if this makes sense. Any and all feedback is appreciated!


bin/lerna.js
```diff
@@ -30,13 +30,14 @@ var cli = meow([
   "  --ignore [glob]      Ignores packages with names matching the given glob (Works only in combination with the 'bootstrap' command).",
   "  --force-publish      Force publish for the specified packages (comma-separated) or all packages using * (skips the git diff check for changed packages)",
   "  --yes                Skip all confirmation prompts",
-  "  --repo-version       Specify repo version to publish",
+  "  --update-version [repo, package, canary]  Specify the type of version update",
```
thejameskyle:
```
--update-version [type]       Specify the type of version update: repo, package, canary
```

src/commands/PublishCommand.js
```diff
@@ -135,13 +135,59 @@ export default class PublishCommand extends Command {
     });
   }
 
-  getVersionsForUpdates(callback) {
-    if (this.flags.repoVersion) {
+  handleVersionCommand(callback) {
+    const flagParts = this.flags.updateVersion.split(" ");
+    const flag = flagParts[0];
+    const flagArg = flagParts[1];
```
davidfateh: This could get condensed to:
```javascript
const [flag, flagArg] = flagParts;
```

bin/lerna.js
davidfateh: Ok, I will change this.

thejameskyle:
I'm worried about having canary as an option here because of https://github.com/lerna/lerna/issues/277#issuecomment-232456862

davidfateh:
I can remove canary from this option. To me canary makes more sense as its own flag and this PR might make the usage of canary more confusing since you can use it as its own flag here or as an argument passed to `--update-version`

thejameskyle:
Yeah, I would prefer to just remove it.

davidfateh:
Ok I've removed the canary option and fixed up the options description stuff.

README.md
```diff
@@ -290,11 +290,14 @@ $ lerna publish --canary --yes
 When run with this flag, `publish` will skip all confirmation prompts.
 Useful in [Continuous integration (CI)](https://en.wikipedia.org/wiki/Continuous_integration) to automatically answer the publish confirmation prompt.
 
-#### --repo-version
+#### --update-version [repo, package], -u
 
 ```sh
-$ lerna publish --repo-version 1.0.1
-# applies version and skips `Select a new version for...` prompt
+$ lerna publish --update-version "repo x.x.x"
```
thejameskyle: Why `--update-version "repo x.x.x"` instead of just `--update-version x.x.x`

davidfateh: My thought process was to explicitly declare the repo command. It can be changed though. We can simply say if the flag for `--update-version` isn't `package` then test it's a valid semver and move on with `repo` logic

thejameskyle: > We can simply say if the flag for --update-version isn't package then test it's a valid semver and move on with repo logic

That sounds like a better api

davidfateh:
`--update-version` now takes either `package` or a semver.

`--update-version 1.0.1` is the same as the now removed `--repo-version 1.0.1`

davidfateh:
Thinking about this some more, we could change `--update-version` back to `--repo-version` so the change isn't breaking but I think the name would then not be completely accurate.

gandazgul:
Can this be revived? We would like to user lerna for publishing as well but this is a requirement for us.

davidfateh:
I'm willing to take another look at this and fix the merge conflicts.

guscastro:
@davidfateh I'd be really interested in this change. Is it possible to revive it? I'm more than willing to pick it up myself if your time doesn't allow for it. My thoughts are to keep `--repo-version` as an alias for `--update-version` for backwards compatibility, but to then update the README recommending `--update-version`.

davidfateh:
@guscastro Let me look at it tonight. The PR is really out of date, but if it's easily salvageable I will update it and include your suggestion about the naming scheme. If not, I'll let you know, don't want to hold up progress for someone else!

davidfateh:
I didn't get a change to look last night. If you think you can tackle it go ahead, otherwise I'm not sure when I'll get to it.

PR #264, opened 2016-07-05
Use `logifySync` decorator for `FileSystemUtilities.readdirSync`

gigabo:
This just happened to catch my eye.


rygine:
noticed that too. thought it may be a typo.

thejameskyle:
Thanks

PR #287, opened 2016-07-29
[Bug] do no try to publish private packages in canary releases

### LINKED ISSUE #286: Lerna attempts to publish private packages
I'm pretty sure 1.x would ignore packages with `private:true` in package.json, but 2.x spits out the following:

```
npm ERR! This package has been marked as private
npm ERR! Remove the 'private' field from the package.json to publish it.
```

ianwremmel:
Fixes #286

### DIFF
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -127,7 +127,19 @@ export default class UpdatedPackagesCollector {
 
   collectUpdates() {
     return this.packages.filter(pkg => {
-      return this.updatedPackages[pkg.name] || (this.flags.onlyExplicitUpdates ? false : this.dependents[pkg.name]) || this.flags.canary;
+      if (this.updatedPackages[pkg.name]) {
+        return true;
+      }
+
+      if (!this.flags.onlyExplicitUpdates && this.dependents[pkg.name]) {
+        return true;
+      }
+
+      if (this.flags.canary && !pkg.isPrivate()) {
+        return true;
+      }
+
+      return false;
     }).map(pkg => {
       return new Update(pkg);
     });


src/UpdatedPackagesCollector.js
```diff
@@ -127,7 +127,19 @@ export default class UpdatedPackagesCollector {
...
+
+      if (!this.flags.onlyExplicitUpdates && this.dependents[pkg.name]) {
+        return true;
+      }
+
+      if (this.flags.canary && !pkg.isPrivate()) {
+        return true;
+      }
+
+      return false;
```
gigabo: Thanks for cleaning this up.  Easier to read this way.

src/UpdatedPackagesCollector.js
```diff
@@ -127,7 +127,19 @@ export default class UpdatedPackagesCollector {
...
+      if (this.updatedPackages[pkg.name]) {
+        return true;
+      }
+
+      if (!this.flags.onlyExplicitUpdates && this.dependents[pkg.name]) {
+        return true;
+      }
+
+      if (this.flags.canary && !pkg.isPrivate()) {
+        return true;
```
hzoo: Should we just return false if it's private or are we allowing private in the other cases?

edvinerikson: I think we should exit early if it's private. The down part with putting the check here is that private packages don't get local dependencies updates when `lerna publish` is called.
e.g

```
packageA
{
  "name": "A",
  "private": true,
  "dependencies": {
    "B": "^1.0.0",
  }
}

packageB
{
  "name": "B"
}
```

if we run `lerna publish` with this patch, the `packageB` dependency in `packageA`  will still be `1.0.0` even if we update `packageB` to `2.0.0`.

ianwremmel: Yea, I actually just ran into that problem. I've been using my fork for the last few days and one of my dev modules was needlessly marked as private, thus leading to canary versions going out of sync and breaking bootstrap.

hzoo: 👍  Ok do you want to modify the pr to do that? Don't think we have any tests for this method yet

ianwremmel: I'd be happy to, but I'm not entirely sure what you had in mind. I'd still like to use private modules with canary releases, but if I understand @edvinerikson's comment correctly, he's suggesting disabling canary if the repo contains private modules.

My use case for private modules is example apps. I explicitly don't want to publish my example code via npm, because installing examples would be kinda silly, but i want the example app's dependencies to update correctly.

hzoo: Oh I see I misunderstood. I thought you were agreeing with disabling canary with an early return

edvinerikson: No I still want canary to work even if some modules are private. My idea is to not implement the check in this method at all. Instead we should implement it when we: prompt new versions (no need to bump a new semver for private modules IMO) and when we run git/npm commands. This way I think we can get the dependency updates to work but skip git tags, version bumping (dependency bumping will still work) and npm publish.

e.g here: https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L385
https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L336
https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L227
https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L173
...and probably some other places.. 

Should probably change `this.updates` to something else for these, e.g `this.publicUpdates`

gigabo:
@ianwremmel - Is this covered by #321?

ianwremmel:
Seems plausible...I was running into enough issues with canary releases (and my build process in general), that I stopped using them for now. I won't be able to verify for a while, but I'm happy to close this PR and the related issue for the time being.

doug-wade:
> I'm happy to close this PR and the related issue for the time being

closing

PR #301, opened 2016-08-08
Add a note about common devDependencies to README

gigabo:
The ability to factor common `devDependencies` out into the repo root is a
useful feature of Lerna.  Add a section to the README explaining the benefits
and caveats.

This was suggested in https://github.com/boennemann/alle/issues/1.


hzoo:
Looks good to me! whoo merging on the ✈️

Issue #299, opened 2016-08-06
lerna bootstrap crashes (infinite loop)


bebraw:
For some weird reason, `lerna bootstrap` crashes in my project (dep issue?). Here's how to repro:

1. Clone https://github.com/reactabular/reactabular
2. `npm install` at project root. This will take a while.
3. `lerna bootstrap` (assumes beta-24).

After 3. you should see something like this:

```bash
reactabular [master] $ lerna bootstrap
Lerna v2.0.0-beta.24
Linking all dependencies
reactabular-sticky                                 ╢███████████████████████████████████░░░░░░░░░░░╟/usr/local/lib/node_modules/lerna/lib/commands/BootstrapCommand.js:110
          return !node.dependencies.filter(function (dep) {
                                    ^

RangeError: Maximum call stack size exceeded
    at Array.filter (native)
    at /usr/local/lib/node_modules/lerna/lib/commands/BootstrapCommand.js:110:37
    at Array.filter (native)
    at bootstrapBatch (/usr/local/lib/node_modules/lerna/lib/commands/BootstrapCommand.js:108:34)
    at /usr/local/lib/node_modules/lerna/lib/commands/BootstrapCommand.js:134:13
    at /usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:726:13
    at /usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:52:16
    at replenish (/usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:314:29)
    at /usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:330:15
    at _parallel (/usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:717:9)
```

It could easily be my error, but even then lerna shouldn't crash.

gigabo:
Thanks for reporting @bebraw.  Definitely a deficiency with Lerna (that [I introduced](https://github.com/lerna/lerna/pull/250#discussion_r68813654) with a faulty assumption).  Thanks for reporting.  Will get it fixed up for you.

gigabo:
Hi @bebraw and @jsonmaur.  This should be fixed in `lerna@2.0.0-beta.25`.  Thanks for reporting!

bebraw:
Thanks for the fix. I realized the circular dependency was actually an issue in my project. Maybe it would be nice to have some mechanism (`lerna check-deps`) to warn since often you want to explicitly avoid those.

kevnk:
[...]

PR #302, opened 2016-08-08
Handle dependency graph cycles during bootstrap

gigabo:
The dependency-order bootstrap scheduler introduced in #250 assumes that the
dependency graph is acyclic.  This turns out not to be a valid assumption, as
demonstrated in #299.

This patch falls back to the package with the most dependents when bootstrap
is blocked by a cycle in the dependency graph.  It issues a warning that graph
cycles may cause instability during bootstrap if dependencies are used during
`prepublish`.

image

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -36,8 +36,13 @@ export default class BootstrapCommand extends Command {
     // Get a trimmed down graph that includes only those packages.
     const filteredGraph = PackageUtilities.getPackageGraph(todoPackages);
 
-    // As packages are completed their names will go into this object.
-    const donePackages = {};
+    // This maps package names to the number of packages that depend on them.
+    // As packages are completed their names will be removed from this object.
+    const pendingDeps = {};
+    todoPackages.forEach(pkg => filteredGraph.get(pkg.name).dependencies.forEach(dep => {
+      if (!pendingDeps[dep]) pendingDeps[dep] = 0;
+      pendingDeps[dep]++;
+    }));
 
     // Bootstrap runs the "prepublish" script in each package.  This script
     // may _use_ another package from the repo.  Therefore if a package in the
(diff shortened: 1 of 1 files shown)


doug-wade:
I'm not sure how much this matters, and maybe this is less closely related than it seems to me, but this strikes me as very different than how [node handles cycles in module loading](https://nodejs.org/api/modules.html#modules_cycles), where the first module is partially loaded, the second module requires the first module partially loaded, the second module finishes loading, and then the first module finishes loading.  Trying to emulate the way node modules are loaded when building modules probably doesn't make sense (at least, I'm inclined to say it doesn't), but the two ideas seem closely related enough in my mind, and the resolution implementation is different enough, that maybe it warrants a hint in the readme like the one that [exec has about nondeterminism](https://github.com/lerna/lerna#exec)?

gigabo:
Thanks for the link @doug-wade.  I hadn't seen that.

The situation here is a little different.  It's a _build_ time circular dependency rather than _require_ time.  In order for the `prepublish` script to run, its dependencies must already be installed.  In the context of `lerna bootstrap` that means we must have created the proxy module and run the `prepublish` script in the _dependency_ already.  If the dependency also depends on the dependent, then we don't know which of the two build first.  Either, both or neither may use the other during `prepublish`, so it may not _matter_ which is built first.  But since we're just guessing, and the order may _change_ between runs, we issue an instability warning.

As for the guess itself: I chose to pick the package with the most dependents.  This unlocks the most packages for the next batch, which increases potential concurrency and improves performance.

gigabo:
@hzoo, @thejameskyle - How does this patch look to you?

This has been [reported](https://github.com/lerna/lerna/issues/299) [twice](https://github.com/lerna/lerna/issues/306) now.  Should get _something_ out soon to unblock people.

hzoo:
👍

PR #318, opened 2016-08-22
Change bootstrap command to not install peerDependencies

loklaan:
**Reason**

`lerna bootstrap` deviates from the [peerDependencies behaviour][npmpeerdeps] found in `npm install` (npm@3)

---

This patch removes peerDependencies from `bootstrap`'s list of installable dependencies per package.

Tests added 💥 

Related:
- #160 First reporting the peer dep behaviour issue in `lerna bootstrap`
- #334 Discussion for lifting shared external packages to root directory (would be great for peer deps)

[npmpeerdeps]: https://docs.npmjs.com/files/package.json#peerdependencies

### DIFF
--- a/src/Package.js
+++ b/src/Package.js
@@ -49,7 +49,6 @@ export default class Package {
   get allDependencies() {
     return objectAssign(
       {},
-      this.peerDependencies,
       this.devDependencies,
       this.dependencies
     );
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -215,4 +215,43 @@ describe("BootstrapCommand", () => {
       }));
     });
   });
+
+  describe("peer dependencies in packages in the repo", () => {
+    let testDir;
+
+    beforeEach((done) => {
+      testDir = initFixture("BootstrapCommand/peer", done);
+    });
+
+    it("should not bootstrap ignored peer dependencies", (done) => {
+      const bootstrapCommand = new BootstrapCommand([], {
+        "ignore-peer-deps": true
+      });
+
+      bootstrapCommand.runValidations();
+      bootstrapCommand.runPreparations();
+
+      let installed = false;
+      stub(ChildProcessUtilities, "spawn", (command, args, options, callback) => {
+        // This should never be reached
+        assert.deepEqual(args, ["install", "external@^1.0.0"]);
+        assert.deepEqual(options, { cwd: path.join(testDir, "packages/package-1"), stdio: STDIO_OPT });
+        installed = true;
+        callback();
+      });
+
+      bootstrapCommand.runCommand(exitWithCode(0, (err) => {
+        if (err) return done(err);
+
+        try {
+          assert.ok(!pathExists.sync(path.join(testDir, "lerna-debug.log")), "lerna-debug.log should not exist");
+          assert.ok(!pathExists.sync(path.join(testDir, "packages/package-1/node_modules/package-2")), "The linkable peer dependency should not be installed");
+          assert.ok(!installed, "The external peer dependency should not be installed");
+          done();
+        } catch (err) {
+          done(err);
+        }
+      }));
+    });
+  });
 });
--- a/test/Package.js
+++ b/test/Package.js
@@ -75,8 +75,7 @@ describe("Package", () => {
     it("should return the combined dependencies", () => {
       assert.deepEqual(pkg.allDependencies, {
         "my-dependency": "^1.0.0",
-        "my-dev-dependency": "^1.0.0",
-        "my-peer-dependency": "^1.0.0"
+        "my-dev-dependency": "^1.0.0"
       });
     });
   });
--- a/test/fixtures/BootstrapCommand/peer/lerna.json
+++ b/test/fixtures/BootstrapCommand/peer/lerna.json
@@ -1,0 +1,4 @@
+{
+  "lerna": "__TEST_VERSION__",
+  "version": "1.0.0"
+}
--- a/test/fixtures/BootstrapCommand/peer/package.json
+++ b/test/fixtures/BootstrapCommand/peer/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "independent"
+}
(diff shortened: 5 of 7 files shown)


loklaan:
I don't know how to get the appveyor builds for older node version working. 🐬

gigabo:
Hi @loklaan - This looks great!  Thanks for adding tests.

The appveyor failures were spurious.  Took a couple of re-builds, but they're all green now.

This LGTM. 👍

loklaan:
Woot

Should I ping anyone else to get this merged in, @gigabo? 😄

loklaan:
I've updated this PR to work with 2.0.0-beta.28.

At the time I write this, an appveyor build for node 0.12 is breaking for no reason 😿

hzoo:
What was the reason for making this an option? I think we would want this as default if it's the same behavior as npm?

loklaan:
Yeah this is the [default behaviour of npm@3](https://docs.npmjs.com/files/package.json#peerdependencies).

In a project I am working on, we were running into issues when testing React components that were made from several packages; each sub-component had their own version of React running on the page. This was because the  `react` peerDependency in each package was being installed by `lerna bootstrap` and then used, making several React instances appear in a single page.

I thought that perhaps someone had chosen this non-conforming behaviour for a reason 💁, so I wrote the npm behaviour as opt in.

I'm all for making it the npm behavior the default in lerna, but that's not my decision to make!

AlastairTaft:
+1 for having this work by default. At the moment after I run bootstrap and then have to go through each package/node_modules and delete the peer dependencies.

thejameskyle:
Yeah I think this should be the default and there should be a flag for also installing peerDeps... maybe not even a flag if npm wouldn't even do that

loklaan:
@thejameskyle Changing the default behaviour sounds great. 👍 

Going forward, I don't think there should be a flag for the current behaviour of installing of the peerDeps in the individual packages.

If anyone currently needs that behaviour, it would be better to enable them in another way; hoisting peerDependencies to `packages/node_modules` or even the root of the lerna repo. I am happy to make another PR for that hoisting feature. It would be immediately useful to me. This could partially resolve #334.

For now... I think I am going to reuse this PR to change the default behaviour and fix some tests. If any maintainer would prefer I create a new PR and close this one, let me know!

thejameskyle:
What is the official recommended way from npm on how to deal with peerDeps as a package author? To install them as devDeps as well?

loklaan:
Afaik, looking at the docs, there is no recommendation from npm people.

The three ways that I know and have had success with:

1. `npm install` the peerDep manually
    * manual, but tooling like current lerna makes it effortless
    * breaks down when you have several in-development interdependent packages that all have a peerDep that is expected to be a singleton. ie React
2. mirror the peerDeps in the devDeps of `package.json`
    * `npm install` becomes the tool do remove manual-ness
    * same problem of parallel in-dev packages w/ singleton peerDeps
3. `npm install` higher up the node resolution path
    * unless you have tooling, it very manual
    * can fix the problem of parallel in-dev packages w/ singleton peerDeps

In the project I am working on, that uses lerna, we've tried all three and have settled with the 3rd item. Our approach was to add all of our peerDependencies to the root level `package.json`'s `devDependencies`.

rtsao:
Is this ready to be merged? This would be awesome to have.

cchamberlain:
@thejameskyle - I've not found a good use for peer dependencies. Most libs will have react as a dev and peer dependency but I always end up with multiple copies of react when using npm link and this approach. Personally, I write react libs as factories requiring React in to not burden myself and consumers with the multiple copies of React bugs.

AlastairTaft:
@cchamberlain Can you expand on how you do that, particular when you say libs as factories? I'd love to know as I always have problems developing with React as a peer dependency. And I get a ton of problems around using `npm link` as `require` uses the physical path not the symlinked path, I know there's a cli option to avoid this but it doesn't work with webpack and browserify.

cchamberlain:
@AlastairTaft - Sure, my default export is typically a function with signature:

**fooComponent.js (In npm foo-component lib) - 0 React npm dependencies**
```js
export default function fooComponent ( dependencies ) {
    const { React, ...otherHeavyDeps } = dependencies
    
    class FooComponent extends React.Component {
        /** implementation **/
    }

    return FooComponent
}
```

Then within the consumers app, you can use my FooComponent without it having any peer or dev dependencies on react:

**app.js**
```
import React from 'react'
import { render } from 'react-dom'
import fooComponent from 'foo-component'

// give factory the React / other dependencies, get back the component.
const FooComponent = fooComponent({ React })

render(<FooComponent />, document.getElementById('root'))
```

This pattern also gives a good place to provide configuration options for the library one level before props.  This example uses classes, but I always use [react-stamp](https://github.com/stampit-org/react-stamp) to create all my components (react-stamp employs the same factory philosophy).

For testing my react components, I'll create a separate project within the same lerna repo and that one takes on the heavy React dependencies.

Here's a [real world example](https://github.com/noderaider/gridiron/blob/master/packages/gridiron-react/src/lib/index.js) if you'd like to see it in larger scale.

AlastairTaft:
@cchamberlain That's a really nice idea. Thanks for taking the time to explain.

src/Package.js
```diff
@@ -49,7 +49,6 @@ export default class Package {
   get allDependencies() {
     return objectAssign(
       {},
-      this.peerDependencies,
```
thejameskyle: I'm tempted to say that we should leave `allDependencies` as is and have a `bootstrapDependencies` or filter out peerDeps. Maybe that's optimizing for the future too much though...

Actually... let's just do this, I don't think it'll cause any problems and if we need it in the future it's an easy refactor.

antonmedv:
Ok, and how to install peerDependencies now? For example I have webpack loader with peerDependencies inside lerna.

evocateur:
@antonmedv Just like you do in npm >= 3; that is, they are also specified as `devDependencies` for the package that specifies the `peerDependencies` (so tests work, etc). For package consumers, you install them separately.

antonmedv:
@evocateur `devDependencies` ok, got it.

PR #312, opened 2016-08-15
added style sheet handling for "styles.css"

boxxxie:
#311

PR #320, opened 2016-08-24
Respect package json "style" key

### LINKED ISSUE #311: Generate Index For Exposed CSS
It has become common practice to expose a package's main CSS file via the `style: ''` descriptor in _package.json_. I know that several of our packages expose a CSS file in addition to the main JS file. Lerna should generate an additional _index.css_ file that uses the `@import` syntax to link to the file pointed to in _package.json_. Bonus points if it could be flexible enough to support sass and less files as well (there are some bad habits in our repo...).

seansfkelley:
Fixes #311. I think this is more in line with the intent of #311 than #312 (@nruhe @boxxxie).

Caveat: this uses a CSS file with a relative path for `@import`, which assumes you are using a build system that will do something useful with this path instead of just dropping a `../../../foo.css` path onto the browser client that probably won't work. Absolute paths are a little tricky, as preprocessors seem to assume you're know what you're doing and will leave them as-is. In other words, relative paths seem to "just work" more transparently. (I assume you could tweak said build system or preprocessor to tolerate them if necessary.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,8 @@ Installs all of their dependencies and links any cross-dependencies.
 When run, this command will:
 
 1. Link together all Lerna `packages` that are dependencies of each other.
-  2. This doesn't currently use [npm link](https://docs.npmjs.com/cli/link) and instead uses a proxy to the actual package in the monorepo.
+  - This doesn't currently use [npm link](https://docs.npmjs.com/cli/link) and instead uses a proxy to the actual package in the monorepo.
+  - Lerna respects a filename specified as the `"style"` key in `package.json` and will link the named CSS file.
 2. `npm install` all external dependencies of each package.
 
 Currently, what Lerna does to link internal dependencies is replace the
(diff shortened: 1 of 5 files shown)


seansfkelley:
I'm using this locally with Webpack and it's working great. I don't have a loader that will read `style` and do the right thing, but if I (sigh) name my output file `style.css` to match the filename that this PR generates, it reads right through the `@import` and works perfectly.

seansfkelley:
@gigabo (tagging since you were on #312), looks like the AppVeyor build is failing with permissions errors. :(

gigabo:
Ugh, yeah I think that's spurious.  I'm actually not able to kick off rebuilds on appveyor.

@hzoo?

boxxxie:
this problem is completely solved by the enhanced-bootstrap pr. I use that instead and don't see a need for the pr anymore.

rygine:
@seansfkelley does PR #237 achieve the same result as this?

seansfkelley:
I'm going to tentatively close this because I'm pretty sure #237 should solve the problem once it fixes the issue I mentioned in a comment over there.

Issue #303, opened 2016-08-09
Bug : Lerna is publishing private packages when they have a dependency


yvele:
# Lerna v2.0.0-beta.24

First of all : Thank you for lerna :)

So.. Let's say I have 3 packages :

```json
{
  "name": "@scope/package-a",
  "private": false
}
```

```json
{
  "name": "@scope/package-b",
  "private": true
}
```

```json
{
  "name": "@scope/package-c",
  "private": true,
  "peerDependencies": {
    "@scope/package-b": "^0.1.0"
  }
}
```

Only **package-a** is public. When I run `lerna updated` I only get `- @scope/package-a` and that's what is expected.

But now when I switch **package-b** to public and I run `lerna updated` again, I get :

```none
Lerna v2.0.0-beta.24
Checking for updated packages...

- @scope/package-a
- @scope/package-b
- @scope/package-c
```

You can see that **package-c** is being published even it is explicitly private...

PS: When I remove peerDependencies from **package-c**, it's privacy is respected by lerna.

ianwremmel:
Same root cause as #286 and #287?

gigabo:
Just ran into this. :(

_Maybe_ related to #286?  I was not using `--canary`.

gigabo:
Just realized that previously all of our private packages were using _relative_ dependencies, like:

```javascript
    "package-b": "../package-b",
```

This was to work around #11.  But we recently eliminated deep requires, so we have normal dependencies that Lerna can link now.

I guess this has been an issue with private packages for a _while_. 😬

gigabo:
So, I _think_ the right behavior here would be to update the dependency versions in private packages, but not try to actually publish them.

What about `lerna updated`?  Private packages would be be _modified_ by `lerna publish`, so they'd get updated.  Maybe output with a `(private)` tag?

```
$ lerna updated
Lerna v2.0.0-beta.24
Checking for updated packages...

- generator-react-server
- react-server
- react-server-cli
- react-server-core-middleware
- react-server-data-bundle-cache
- react-server-integration-tests (private)
- react-server-test-pages (private)
```

gigabo:
Hi @yvele, #321 landed in `lerna@2.0.0-beta.28`.  That should get publish with private modules working correctly.  Please reopen (or start a new ticket) if you still run into trouble after upgrading.

PR #321, opened 2016-08-25
Don't try to publish private modules

gigabo:
Private modules may need to be updated during publish to keep their dependency
versions up to date, but they can't actually be published to the npm registry.

This addresses #303 (and #286 also, I think?)


rtsao:
This PR solves the issue for me

gigabo:
Going to merge this.  It's a real pain to publish in a repo with private modules without it.

PR #322, opened 2016-08-26
docs(lerna-publish): add hint to indicate which packages will be published

Scarysize:
If you use lerna without the intention to publish your packages on npm, it might be unclear, which packages are affected by `lerna publish`.

We use lerna internally and use `lerna publish` with `--skip-npm` to manage git releases. We accidentally had some of our packages marked as private, from when they were standalone. `lerna pubslish` and `lerna ls` gave us quite the headache, because it listed/published only "some" of our packages, though linking the packages worked just fine (`bootstrap`).


gigabo:
Thanks @Scarysize.  This will go nicely with #321.

PR #325, opened 2016-08-26
Make unlinkSync synchronous (fixes #324)

### LINKED ISSUE #324: FileSystemUtilities.unlinkSync is not synchronous
https://github.com/lerna/lerna/blob/master/src/FileSystemUtilities.js#L68

also, the `init` command calls `unlinkSync` with a second argument.

https://github.com/lerna/lerna/blob/master/src/commands/InitCommand.js#L87

rygine:
* Add path check
* Remove second argument in call from InitCommand

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -65,8 +65,6 @@ export default class FileSystemUtilities {
 
   @logger.logifySync
   static unlinkSync(filePath) {
-    if (pathExists(filePath)) {
-      fs.unlinkSync(filePath);
-    }
+    fs.unlinkSync(filePath);
   }
 }
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -65,6 +65,8 @@ export default class FileSystemUtilities {
 
   @logger.logifySync
   static unlinkSync(filePath) {
-    fs.unlink(filePath);
+    if (pathExists(filePath)) {
+      fs.unlinkSync(filePath);
+    }
   }
 }
--- a/src/commands/InitCommand.js
+++ b/src/commands/InitCommand.js
@@ -84,7 +84,7 @@ export default class InitCommand extends Command {
     const versionLocation = this.repository.versionLocation;
     if (FileSystemUtilities.existsSync(versionLocation)) {
       this.logger.info("Removing old VERSION file.");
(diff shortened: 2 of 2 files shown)


src/FileSystemUtilities.js
```diff
@@ -65,6 +65,8 @@ export default class FileSystemUtilities {
 
   @logger.logifySync
   static unlinkSync(filePath) {
-    fs.unlink(filePath);
+    if (pathExists(filePath)) {
```
rygine: do we even want this check?

src/commands/InitCommand.js
```diff
@@ -84,7 +84,7 @@ export default class InitCommand extends Command {
     const versionLocation = this.repository.versionLocation;
     if (FileSystemUtilities.existsSync(versionLocation)) {
       this.logger.info("Removing old VERSION file.");
-      FileSystemUtilities.unlinkSync(versionLocation, "0.0.0");
+      FileSystemUtilities.unlinkSync(versionLocation);
```
rygine: i don't see any version of `unlinkSync` in the history having a second argument. not sure why it's here.

src/FileSystemUtilities.js
gigabo: Good question.  Presumably if we're calling `unlinkSync` we should be reasonably sure that there's something to unlink, and it's an exception otherwise.  I think that's the logic behind the behavior of the `*Sync` methods in node's standard library.

src/commands/InitCommand.js
gigabo: Code [evolves](https://github.com/lerna/lerna/commit/1bc550438428afdb939ddf0a30907f891a3222e1). ;)

```diff
-        this.logger.info("Creating VERSION file.");
-        FileSystemUtilities.writeFileSync(versionLocation, "0.0.0");
...
+      this.logger.info("Removing old VERSION file.");
+      FileSystemUtilities.unlinkSync(versionLocation, "0.0.0");
```

This looks safe to remove.  Nice catch.

src/FileSystemUtilities.js
gigabo: Looking at this again, I'd say it's better to remove the guard for now.

That would be more consistent with the other `*Sync` methods in FileSystemUtilities, too.

gigabo:
Thanks @rygine!

Issue #219, opened 2016-06-14
Support for typescript packages


zkochan:
I would love to use lerna in a typescript monorepo.

It shouldn't be hard to add typescript support. Currently lerna adds a "symlink-package" to `node_modules` to reference the package in the monorepo. It has only an `index.js`. If the package has a typings file, lerna could also add a "symlink" to that file. It would be an `index.d.ts` file like this:

```ts
export * from '/home/user/src/monorepo/packages/foo/index.d.ts'
```

I can work on a pull request, if you have no objection about this enhancement

rygine:
this would be fixed using the updated bootstrapping process i've created.

more info here: https://github.com/lerna/lerna/issues/215

zkochan:
OK, great! I think I'll keep it open till that one is done and a typescript strategy is implemented.

thejameskyle:
If we are unable to get real symlinks working, I wouldn't mind a system for creating pseudo-linked files like this one, but I want it to be more generic.

calebmer:
What’s the status on this?

bradleyayers:
I'm also hitting the problem with trying to use TypeScript because everything from `package.json` is stripped except `name` and `main`.

I'm curious why real symlinks weren't/aren't a workable solution? Is there an issue specific to that that I can follow?

zkochan:
Hey @calebmer and @bradleyayers!
I think we have to subscripb to this issue: [Nested directory support](https://github.com/lerna/lerna/issues/11)

bradleyayers:
I propose we re-open this, as adding TypeScript basic support will be far easier than the nested directory support (which seems to have stalled for the moment). @zkochan's proposal is mostly good, but it doesn't cater for default exports, so instead the file should be:

```
import * as module from '/home/user/src/monorepo/packages/foo/index.d.ts';
export = module;
```

bradleyayers:
I've added a PR for this, keen to hear thoughts.

rygine:
@bradleyayers PR #237 has been very active lately and is ready for final review/merge. give it a try and see if it works for you!

bradleyayers:
Have tried and can confirm it works for me! Thanks for pointing me in the right direction.

danvk:
@bradleyayers @zkochan do you have examples you can point to of using lerna in a TS monorepo? Would be great to have a template for setting this up.

zkochan:
[...]

PR #326, opened 2016-08-27
Add support for TypeScript packages (#219).

bradleyayers:


bradleyayers:
Obsolete given #237 adds TypeScript support plus everything else!

PR #328, opened 2016-09-02
Make log levels more like npm

gigabo:
Currently Lerna has log _types_ and a logging verbose flag, which has
different default values for different types (e.g. `false` for "debug",
`true` for "error").  It also has log types that are close to, but don't
_quite_ match their npm counterparts.

This PR makes Lerna's logging configuration more similar to npm's.

It:

- Adds a `--loglevel` CLI option
- Drops the `verbose` argument to logger methods
- Renames some logger methods to be more like their `npm` equivalents
    - `logger.debug` => `logger.verbose`
    - `logger.warning` => `logger.warn` (also more like `console.warn`)
- Allows an optional error object to be included with any log level

The update to the README is basically cribbed from `npm help 7 config`.

Differences with `npm`:
- Lerna has a `success` log level in place of npm's `http`
- Lerna's default is `info`, while npm's default is `warn`

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,3 +524,9 @@ $ lerna publish --only-explicit-updates
 ```
 
 Ex: in Babel, `babel-types` is depended upon by all packages in the monorepo (over 100). However, Babel uses `^` for most of it's dependencies so it isn't necessary to bump the versions of all packages if only `babel-types` is updated. This option allows only the packages that have been explicitly updated to make a new version.
+
+#### --loglevel [silent|error|warn|success|info|verbose|silly]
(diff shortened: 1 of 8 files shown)


gigabo:
Default output is unchanged:

image

But now log level can be adjusted:

image

jimthedev:
This would be fantastic. As someone who is new to using lerna it was quite confusing to have stdout swallowed.

gigabo:
This [moved to Asini](https://github.com/asini/asini/pull/5), too.

PR #329, opened 2016-09-02
Indicate private modules in output from `updated`

gigabo:
This is helpful for figuring out what will actually be published.

Goes along with #321.

Looks like:

image

### DIFF
--- a/src/commands/UpdatedCommand.js
+++ b/src/commands/UpdatedCommand.js
@@ -16,7 +16,7 @@ export default class UpdatedCommand extends Command {
 
   execute(callback) {
     const formattedUpdates = this.updates
-      .map((update) => `- ${update.package.name}`)
+      .map((update) => `- ${update.package.name}${update.package.isPrivate() ? " (private)" : ""}`)
       .join("\n");
 
     this.logger.newLine();


hzoo:
👍

src/commands/UpdatedCommand.js
```diff
@@ -16,7 +16,7 @@ export default class UpdatedCommand extends Command {
 
   execute(callback) {
     const formattedUpdates = this.updates
-      .map((update) => `- ${update.package.name}`)
+      .map((update) => `- ${update.package.name}${update.package.isPrivate()?" (private)":""}`)
```
hzoo: maybe spaces between the operators though

gigabo: Oh, yeah.  Weird that the linter didn't catch that...

gigabo: Looks like that linter rule got dropped accidentally.  [Re-added](https://github.com/lerna/lerna/commit/309029397fc6f4a7bac7020af2ba4313c4bf1e4b) and fixed this up.

Nice catch @hzoo!

Issue #231, opened 2016-06-19
Customize `packages` folder


danderu:
Is it possible to place my modules in a different directory than `packages`? Let's se, if I don't like the name `packages` and I would place all of them under `modules`?. This line: https://github.com/lerna/lerna/blob/master/src/PackageUtilities.js#L15 could be reading some key from lerna.json, for instance. Makes sense?

Also, is it possible to link packages recursively under the `packages` folder For instance, with this structure?

```
packages
|_ui-components
|__input
|__button
|_apps
|__some-app
|__yet-another-app
|_my-app-core
```

thejameskyle:
> Is it possible to place my modules in a different directory than packages?

No but I'm happy to allow for that.

> Also, is it possible to link packages recursively under the packages folder For instance, with this structure?

No it is not possible, and I don't think it ever will be

zkochan:
If it will be configurable, I would suggest to allow specifying patterns.

jameswomack:
I would also love for it to be configurable. It's a minor issue, but one that I'm happy to open a PR for. Having a directory with such a similar name to package.json makes semantic sense—but makes ZSH autocomplete less effective. I use autocomplete to quickly open files in vim and package.json is one of the files I most frequently open with vim. I'm sure there are other better reasons to have it configurable too ;)

If I open a PR that allows a field folderName (or directoryName) in lerna.json to point to a directory that can be reached via `Path.resolve(Path.basename(lernaJSON, folderName)`, is that something you might merge?

zkochan:
I have a project in which the subpackages are all in the root directory. They're just prefixed with some word. That's why I'd need the possibility to specify a pattern, not just one dir name.

hzoo:
From http://staltz.com/setting-up-a-javascript-monorepo.html: packages are also in the root directory for https://github.com/cyclejs/cyclejs

gigabo:
@danderu I think #332 should provide what you're looking for here.

Please take a look if you get a chance.

thejameskyle:
See #365

PR #332, opened 2016-09-07
Add support for configurable package locations

gigabo:
This adds a `packages` configuration option to `lerna.json`:

```json
{
  "lerna": "2.0.0-beta.27",
  "version": "0.2.3",
  "packages": [
    {"glob": "packages/{core,plugins}/*/package.json"},
    {"glob": "build-tools/package.json"},
    {"glob": "integration-tests/**/package.json"}
  ]
}
```

The default is `[{"glob": "packages/*/package.json"}]`, which is equivalent to
the historical layout of a Lerna repo.

This addresses #126 and #231.

----
Still TODO:

- [ ] documentation
- [x] new tests

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -2,6 +2,7 @@ import FileSystemUtilities from "./FileSystemUtilities";
 import PackageGraph from "./PackageGraph";
 import Package from "./Package";
 import path from "path";
+import {sync as globSync} from "glob";
 import minimatch from "minimatch";
 
 export default class PackageUtilities {
@@ -27,25 +28,25 @@ export default class PackageUtilities {
     return require(PackageUtilities.getPackageConfigPath(packagesPath, name));
   }
 
-  static getPackages(packagesPath) {
+  static getPackages(repository) {
     const packages = [];
 
-    FileSystemUtilities.readdirSync(packagesPath).forEach((packageDirectory) => {
-      if (packageDirectory[0] === ".") {
-        return;
-      }
+    repository.packageConfigs.forEach(({glob}) => {
 
-      const packagePath = PackageUtilities.getPackagePath(packagesPath, packageDirectory);
-      const packageConfigPath = PackageUtilities.getPackageConfigPath(packagesPath, packageDirectory);
+      globSync(glob)
+        .map((fn) => path.resolve(fn))
+        .forEach((packageConfigPath) => {
+          const packagePath = path.dirname(packageConfigPath);
 
-      if (!FileSystemUtilities.existsSync(packageConfigPath)) {
-        return;
-      }
+          if (!FileSystemUtilities.existsSync(packageConfigPath)) {
+            return;
+          }
 
-      const packageJson = require(packageConfigPath);
-      const pkg = new Package(packageJson, packagePath);
+          const packageJson = require(packageConfigPath);
+          const pkg = new Package(packageJson, packagePath);
 
-      packages.push(pkg);
+          packages.push(pkg);
+        });
     });
 
     return packages;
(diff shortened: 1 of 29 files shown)


src/Repository.js
```diff
@@ -13,7 +13,7 @@ export default class Repository {
       GitUtilities.init();
     }
 
-    this.rootPath = GitUtilities.getTopLevelDirectory();
+    this.rootPath = path.resolve(GitUtilities.getTopLevelDirectory());
```
rygine: i was about to comment to add `path.resolve()` here 😉

gigabo: :feelsgood:

src/PackageUtilities.js
```diff
@@ -27,25 +28,25 @@ export default class PackageUtilities {
...
-        return;
-      }
+    repository.packageConfigs.forEach(({glob}) => {
 
-      const packagePath = PackageUtilities.getPackagePath(packagesPath, packageDirectory);
-      const packageConfigPath = PackageUtilities.getPackageConfigPath(packagesPath, packageDirectory);
+      globSync(glob)
+        .map((fn) => path.resolve(fn))
+        .forEach((packageConfigPath) => {
+          const packagePath = path.dirname(packageConfigPath);
```
gigabo: The section below just got indented a bit.  Easier to view with a [no whitespace diff](https://github.com/lerna/lerna/pull/332/files?w=1).

steelbrain:
I implemented it in a different way in https://github.com/motion/lerna/pull/1
thoughts?

gigabo:
@steelbrain Nice!  Looks like the main difference from a user perspective is this uses globs instead of directories?  It's also wrapping the items in objects (`{"glob":"..."}`).  My thought there is eventually we'll wind up with config for package groups.

steelbrain:
@gigabo The idea of globs is certainly tempting but in a realistic view, this could stray the package from it's main purpose. Having two separate package directories allows sorting packages based on their roles but still enforces them to be purely mono repo package directories, in our use case it was `apps` and `modules`.

If we allow users to mix lerna packages with other things in deep directories, it's only gonna get messy. but that's just my opinion :)

gigabo:
Oh, it's still all packages!

The globs resolve to `package.json` files for packages that are managed by Lerna.

steelbrain:
Ohhhh, in that case sounds good 👍

gigabo:
That reminds me... I'm missing documentation here!  Also new tests... 😅

seansfkelley:
Does this teach Lerna how to do anything with these package groups as a unit, or does it just allow you to split the "global Lerna namespace" across multiple directories? Example: if I have e.g. "apps" and "modules" as mentioned above, am I able to run things on only one glob, or do I still have to name packages in a way that `--scope` does what I want/name the npm scripts in a way that I can differentiate with `lerna run`?

zkochan:
This is ready, right? Just documentation has to be added.

nruhe:
If the only thing holding this change back is documentation, I'd be happy to volunteer to write it.

gigabo:
Oh, whoops meant to close this!  Moved to Asini.  Had some [issues](https://github.com/asini/asini/pull/22) beyond documentation.

smrq:
Huh? Is there any real reason this isn't being added to Lerna?

steelbrain:
I must be misunderstanding something, how are Lerna and Asini different?

gigabo:
Asini is a fork of Lerna.  I need features that are unavailable in Lerna and ran into a situation where I'm unfortunately no longer comfortable contributing here.

thejameskyle:
Moving to https://github.com/lerna/lerna/pull/365

PR #348, opened 2016-09-26
Use unprivileged alternatives to symlinks on Windows

### LINKED ISSUE #339: Add option to use NTFS junctions on Windows
On Windows, creation of symlinks requires administrator rights, whereas creation of NTFS junctions does not. Junctions are similar to hardlinks on Linux.

Currently, if you use Lerna without admin rights on Windows, you can encounter permissions errors similar to:

> Errored while running BootstrapCommand.execute
> Error: EPERM: operation not permitted, symlink 'C:\src\jest\packages\jest-runtime\bin\jest-runtime.js' -> 'C:\src\jest\packages\jest-repl\node_modules\.bin\jest-runtime'
>     at Error (native)
> -> failed running: C:\src\jest\node_modules\.bin\lerna.cmd bootstrap

motiz88:
Fixes #339 and allows `lerna bootstrap` to run as a regular user on Windows.

For the record, the calls to `normalize()` are there to remove any trailing slashes from the paths (I forget which had them - the `path.join` results or the `readlinkSync` results).

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,8 +154,7 @@ When run, this command will:
 
 1. `npm install` all external dependencies of each package.
 2. Symlink together all Lerna `packages` that are dependencies of each other.
-    * This requires Administrator privileges on Windows
-2. `npm prepublish` all bootstrapped packages.
+3. `npm prepublish` all bootstrapped packages.
 
 `lerna bootstrap` respects the `--ignore` flag (see below).
 
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "async": "^1.5.0",
     "chalk": "^1.1.1",
+    "cmd-shim": "^2.0.2",
     "cross-spawn": "^4.0.0",
     "inquirer": "^0.12.0",
     "lodash.find": "^4.3.0",
@@ -38,6 +39,7 @@
     "pad": "^1.0.0",
     "path-exists": "^2.1.0",
     "progress": "^1.1.8",
+    "read-cmd-shim": "^1.0.1",
     "rimraf": "^2.4.4",
     "semver": "^5.1.0",
     "signal-exit": "^2.1.2",
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -3,6 +3,9 @@ import logger from "./logger";
 import mkdirp from "mkdirp";
 import rimraf from "rimraf";
 import fs from "fs";
+import cmdShim from "cmd-shim";
+import readCmdShim from "read-cmd-shim";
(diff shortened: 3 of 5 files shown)


rygine:
i'm not sure this will work for the `symlink`'d binaries. it is my understanding that a `junction` symlink is for directories only.

thejameskyle:
@Daniel15 Do you know how to handle this ^

motiz88:
@rygine Quite right actually. Hmm. I've just noticed Windows will _create_ file junctions but then complain when they're actually read as files (which I guess the tests don't do?)

I suppose it's possible to emulate symlinked binaries on Windows by putting `foo.cmd` stubs in `node_modules` that forward to the real binaries. This approach would involve more Windows-specific code, but it would remove the requirement for Administrator privileges at bootstrap time. Happy to try and develop this further if there's any interest.

Daniel15:
Yeah, junctions are only for directories. You can instead create hard links
for files. I don't think hard links require admin permissions.

motiz88:
`npm` has [cmd-shim](https://github.com/ForbesLindesay/cmd-shim) for this purpose, by the way - might make sense to reuse it.

motiz88:
@Daniel15 I'm afraid hard links might not work in this case though (but I'm just thinking out loud here):

Suppose `foo` is hardlinked to `bar`, then `bar` is deleted and later recreated; `foo` and `bar` are now two distinct files, and any updates to one won't be reflected in the other.

I'm also not sure whether hard links can be detected with something like `lstat().isSymbolicLink()` which Lerna would need.

Daniel15:
Ahh, good catch!

> npm has cmd-shim for this purpose, by the way - might make sense to reuse it.

I think this approach would work. It seems reasonable to me. Rather than symlinking the "binaries", instead just generate new ones in the target location. They're simply small batch files and shell scripts that call Node with the correct arguments.

thejameskyle:
Awesome!

motiz88:
@thejameskyle merged this before I could even finish my comment explaining what I'd done 😁 

So anyway, just for the record - I co-opted `symlink()` and `isSymlink()` in `FileSystemUtilities` to use cmd-shim when appropriate without disturbing the rest of the code too much.

* `symlink` now recognizes a new value for the `type` argument - `"exec"`, which forwards to cmd-shim on Windows and acts like `file` elsewhere.

* `isSymlink` (which is really a misnomer) forwards to `read-cmd-shim` if we're on Windows and looking at a file that's not a symlink.

* The test for the binary linking feature now uses `FileSystemUtilities.isSymlink` rather than `fs.readlinkSync`.

PR #349, opened 2016-09-26
Handle spaces in paths correctly

motiz88:
This fixes the many test failures I encountered with lerna checked out at a path containing spaces.


motiz88:
Sorry, another change snuck in. Will reorganize the branch and reopen.

motiz88:
Hmm. this fails because `command-join` requires Node >= 4. What is the official line on Node version support here in Lerna?

hzoo:
Right now it's node 0.10 but we could remove support since support is ending. I mostly kept it since babel supports node 0.10 still

thejameskyle:
@motiz88 are you going to change the dependency to https://github.com/motiz88/command-join/commit/41dfad5710bd29770ece5fb18ff076619128c8db?

motiz88:
@thejameskyle That was the idea but it's not done yet. Is that the course of action you prefer for now?

thejameskyle:
If you get blocked on merging that into the main project you can do that, I would try making a PR first to see how they respond

motiz88:
https://github.com/seangenabe/command-join/pull/1

motiz88:
With seangenabe/command-join#1 merged, this now works on 0.10 and 0.12 too.

PR #350, opened 2016-09-26
Normalize newlines in lerna.json in publish tests

motiz88:
This along with #348 makes all tests pass on Windows.


motiz88:
Note that AppVeyor tests seem to pass without this change, but I've had problems when checking out Lerna on Windows with the default Git settings (line endings set to CRLF). This PR makes the tests more robust in the face of such settings.

PR #353, opened 2016-09-28
[Bug] Fix spawn with globs

### LINKED ISSUE #352: [Bug] Wildcard not working in Lerna exec
So I would like to symlink all CLI tools that are commonly used and stumbled upon the fact that the `exec` command doesn't expand wildcards.

`lerna exec -- ln -rs ../../node_modules/.bin/* ./node_modules/.bin/`
Expected result: symlinks of all files in the first passed directory to be in the second passed directory.
Result: a symlink named '_' linking to '../../../../node_modules/.bin/_' in the second passed directory.

Current workaround is using a script file like:

``` js
#!/usr/bin/env node
'use strict';

const exec = require('child_process').exec;

exec('ln -rs ../../node_modules/.bin/* ./node_modules/.bin/', function (error, stdout, stderr) {
  if (error) {
    console.error(`exec error: ${error}`);
    return;
  }
  console.log(stdout);
  console.log(stderr);
});
```

And running it with `lerna exec -- node ../../symlink.js`

thejameskyle:
Fixes #352


evocateur:
This has been fixed, as indicated in the linked issue

Issue #355, opened 2016-09-29
[Feature] link devDependencies binaries from root at bootstrap time


motiz88:
This has come up in a few places but I don't think there's an active issue for it. ~~I'd like to work on a PR if the idea is welcome.~~ Edit: I've started #357 to get a feel for this feature.

Basically I want `bootstrap` to make binaries provided by the root's devDependencies accessible in individual packages' npm scripts. Lerna has all the required machinery, and the same packages are already `require`able in (dev) code, so in my view this makes sense in terms of least surprise.

PR #357, opened 2016-09-29
[Feature] Link root devDependencies binaries in all packages at bootstrap time

motiz88:
Consider this a strawman for how #355 would work. I'd love to receive feedback on anything from the idea to the approach to the execution.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -34,6 +34,8 @@ export default class BootstrapCommand extends Command {
     async.series([
       // install external dependencies
       (cb) => this.installExternalDependencies(cb),
+      // symlink binaries from devDependencies
+      (cb) => this.symlinkDevBinaries(cb),
       // symlink packages and their binaries
       (cb) => this.symlinkPackages(cb),
       // prepublish bootstrapped packages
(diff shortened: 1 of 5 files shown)


DxCx:
Big +1 for progressing with this patch,
i was just about to do it myself and then i've found it..
already tested it and it works as expected.
Thanks @motiz88!

DxCx:
@hzoo @thejameskyle any chance it will be merged for beta.31?

thejameskyle:
I don't understand the need to do this. If you have dev scripts that you don't want to put in the top level package, why would you want the `devDependencies` at the top level? (other than file size which I don't think is a concern of lerna) If you want to something globally, I would do it globally.

DxCx:
I have 2 examples that im using:
1. The project im contributing into is written in typescript, i want to install compiler globally not to specify it for each package.
2. I have as devDepencey an external dependency which i want to ensure its exactly the same version for all packages.

This patch gives the power to define global tools/packages i dont see why its bad..

motiz88:
@thejameskyle I guess one argument would be least surprise like I mentioned in the issue - specifically symmetry with `require()`.

In terms of concrete use cases, consider a monorepo that is less homogenous than Babel - it may have _groups_ of packages built in a few different ways (in my case: some generate code from Protobuf definitions, others from pegjs grammars, and others are JS/Babel). I find it natural to want to do `lerna run build` and have it work, and at the same time I find it natural to put the build commands in the individual `package.json`s.

DxCx:
this is not progressing :( is there anything i can do to help progress with that?

thejameskyle:
I really don't want to add features like this unless there's a really good reason to do so. If we do stuff like this too often, Lerna starts to become magic.

DxCx:
So what do you suggest is the correct way of acheiving this behavior?
1. One of the libraries im using have to be the exact same for all of the packages.
2. I dint want to state stuff (such as typescript, mocha, chai) on every package since all of them need it as part of infrastructure

thejameskyle:
If you want to reference a single thing in each one why not just reference something higher up in the file system? `../../`

thejameskyle:
Also if something is being applied to every package, just do it at the top level. `mocha packages/*/test/**/*.js` or whatever

thejameskyle:
And finally, it's not that big of a deal to just install the same dependency into every single package. You would have done that anyways by having a bunch of separate repos.

DxCx:
it is much harder to keep dependencies updated once it is duplicated like you are suggesting.
working with top level is something we tried and there are issues with it.
if this patch is not going to be merged we are going to look for other solution..

thanks.

olegskl:
> If you want to reference a single thing in each one why not just reference something higher up in the file system? ../../

We cannot do this. When developing locally on windows we'll have to write "..\\..\\" which will fail for people using linux.

doug-wade:
@motiz88 @olegskl last night we merged a change that [hoists common dependencies to the monorepo root](https://github.com/lerna/lerna/pull/507).  I suspect that this fulfills your usecase.  It accomplishes much the same thing (having a core set of dependencies that are installed only once while being available to every package) while giving some other guarantees that I think are an improvement.  One of the things that is really important for Lerna projects is that they operate independently as if they were well formed npm packages.  This is important to make onboarding for new developers easier, since if they are already familiar with yarn or npm, they can develop on a single module in a lerna repo using the tools they are used to.  This also makes it possible to import and export modules from a lerna repo, since each module is well-formed on its own.  By symlinking devDependencies listed at the root to leaf modules, you're creating a module that can't be `npm install`ed and used (since the devDependencies aren't available), and a module that can't be exported (since its dependencies aren't listed), whereas by listing the full dependency closure in each module and then hoisting the dependencies, you have well-formed node_modules, singletons are respected, and you only install each dependency once.

If there's something I'm missing here, let me know, but I believe that we can decline this pr as redundant to #507.

DxCx:
Thanks @doug-wade !
i'll wait for the next lerna release and give it a try..

doug-wade:
@DxCx I have good news for you -- its out!  https://twitter.com/left_pad/status/826857968930992128

DxCx:
yeah i know @doug-wade was watching the repo for it :) thanks!
will test it soon and share the experiance...

DxCx:
great news!
i had out-of-the-box experience replacing the patched branch with beta.35 package.
works like a charm.
thanks alot to you @doug-wade and to the rest of lerna team.
IMO this PR can be closed unless @motiz88 isn't sufficient as i am..

PR #370, opened 2016-10-10
[Feature] lerna init assumes lerna in devDependencies

joscha:
The current implementation assumes lerna to be a devDependency. Our project has lerna as a dependency (because the shell project is only build tools, which makes lerna a dependency in that sense). However when we run `lerna init` (after a lerna upgrade for example), lerna gets added to the `devDependencies` and not updated in the dependencies. This PR fixes this.

### DIFF
--- a/src/commands/InitCommand.js
+++ b/src/commands/InitCommand.js
@@ -33,19 +33,27 @@ export default class InitCommand extends Command {
   ensurePackageJSON() {
     let {packageJsonLocation, packageJson} = this.repository;
 
-    if (!packageJson) packageJson = {};
-    // if (!packageJson.private) packageJson.private = true;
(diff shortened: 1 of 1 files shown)


src/commands/InitCommand.js
```diff
@@ -33,19 +33,27 @@ export default class InitCommand extends Command {
...
 
-    if (!packageJson) packageJson = {};
-    // if (!packageJson.private) packageJson.private = true;
-    if (!packageJson.devDependencies) packageJson.devDependencies = {};
-
-    objectAssignSorted(packageJson.devDependencies, {
-      lerna: this.lernaVersion
-    });
-
     if (!packageJson) {
```
joscha: this logic down here was flawed, as we update in case of `!packageJson` a few lines further up.

thejameskyle:
Closes #401

PR #372, opened 2016-10-11
Fix logifyAsync, logifySync decorators.

seansfkelley:
The decorator spec requires that a function decorator either:

- returns a new /descriptor/ for the property, or
- changes the fields on the given descriptor

The logify* methods instead returned a new function implementation, which
is not correct, and caused the decorations to be silently ignored.

decorator spec: https://github.com/wycats/javascript-decorators#detailed-design
function decorator example: https://github.com/jayphelps/core-decorators.js/blob/master/src/deprecate.js

This commit changes logifySync and logifyAsync in the following ways:

- they now have no return value, and instead modify the descriptor
- they now must be called with no arguments so they can properly capture
  the `this` value (without this, the function cannot find the logger instance)
- they now log at 'verbose', because they are extremely verbose
- remove 'error' flag from non-error logging
- they truncate their serialized arguments to (arbitrarily) 100 characters because
  otherwise they they OOM the node process

Additionally, one async method was erroneously decorated with logifySync.

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -14,47 +14,47 @@ function ensureEndsWithNewLine(string) {
 }
 
 export default class FileSystemUtilities {
-  @logger.logifySync
+  @logger.logifySync()
   static mkdirSync(filePath) {
     fs.mkdirSync(filePath);
   }
 
-  @logger.logifyAsync
+  @logger.logifyAsync()
   static mkdirp(filePath, callback) {
     mkdirp(filePath, callback);
   }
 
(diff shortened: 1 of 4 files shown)


seansfkelley:
AppVeyor's 0.10 build failed with permissions issues, per usual...

seansfkelley:
Hey @hzoo/@thejameskyle, this is a pretty bad bug to leave lying around. Can this (and hopefully #373 too) get merged sooner rather than later?

seansfkelley:
Thank you!

PR #373, opened 2016-10-11
[Feature] Log stdout when commands fail. Fixes #343.

### LINKED ISSUE #343: [Feature] pass through errors/warnings from npm tasks to terminal/lerna-debug.log
I have some local modules that declare peer dependencies. #318 keeps Lerna from installing peer dependencies, but since Lerna swallows all npm output, you don't get the informative message telling you whether or not your packages are violating or missing their peer dependencies. More generally, it seems like it'd be useful if Lerna had an option to pass through npm output, which would be useful not just for this kind of stuff but also (e.g.) cranking up the logging in CI builds to ease debugging build failures.

Re: `lerna-debug.log`, this _might_ be addressed by https://github.com/npm/npm/issues/10491, though I'm not sure how `npm run`, `npm-debug.log` and `lerna-debug.log` all relate to each other.

seansfkelley:
ChildProcessUtilities now returns stdout as the second argument of
a callback. logifyAsync will check for the presence of a second
argument and log it out in the case of an error, too.


thejameskyle:
Should we be combining `stdout` and `stderr` into a single string to keep things in order?

seansfkelley:
I hadn't thought of that, but yeah, I think that's the right thing to do. Will take another look Thursday.

seansfkelley:
@thejameskyle finally got around to making those changes.

PR #382, opened 2016-10-18
Add `--scope` to `ls` to determine which packages are being depended upon

Steven-Evans:
https://github.com/lerna/lerna/issues/381

Uses breadth-first search on the packageGraph to determine dependent packages. 
Removed dashes from output for easier cli parsing.
Added quiet flag to remove the lerna version message.

Issue #381, opened 2016-10-18
Adding `--scope` to `ls` or some way to determine which packages are depended upon for a single package


Steven-Evans:
Supposing a monorepo has multiple common dependent packages for services which are also in the same repository, it would be useful to know the complete list of packages that a particular service depends on. For example:

Service-1 depends on Auth
Auth depends on Common
Service-2 depends on Auth and Profile
Profile depends on Common

Then `lerna ls --scope Service-1` outputs something like:
Service-1
Auth
Common

Steven-Evans:
https://github.com/lerna/lerna/pull/382

lukebatchelor:
Just linking this here for completeness re: the quiet flag #133

ccrowhurstram:
Feature sounds very interesting... can you make sure it works with https://github.com/lerna/lerna/pull/365 when it lands (not sure if you need to anything, just a "heads up")

Steven-Evans:
I gave it a quick glance and it seems like it shouldn't be a problem. packages and packageGraph are still defined during runPreparations. 

As per my comment on the PR, I'm going to create a new PR to remove the --quiet flag since @thejameskyle suggested the intent to remove the lerna version line of output altogether, which is somewhat unrelated to this.

thejameskyle:
Moving to #384

PR #365, opened 2016-10-07
[Feature] Add support for configurable package locations

thejameskyle:
Just needs documentation. @nruhe you said you were interested? I would now, but I'm working on documentation for another project about to be open sourced

PR #384, opened 2016-10-19
[Feature] Add `--scope` to LsCommand to determine which packages are being depended upon

Steven-Evans:
Issue: https://github.com/lerna/lerna/issues/381
Previous PR: https://github.com/lerna/lerna/pull/382

Uses breadth-first search on the packageGraph to determine dependent packages. 
Removed dashes from output for easier cli parsing.
Removed quiet flag from previous pull request. It would have silenced the lerna version line of output but that should be it's own feature.

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -84,9 +84,7 @@ export default class PackageUtilities {
     return packages;
   }
 
-  static filterNonDependentPackages(packages, scope) {
-    const filteredGraph = PackageUtilities.getPackageGraph(packages);
-
+  static filterNonDependentPackages(packages, packageGraph, scope) {
     let dependentPackages = [];
     let fringe = [];
     fringe.push(scope);
@@ -95,7 +93,7 @@ export default class PackageUtilities {
     while (fringe.length !== 0) {
       let pkg = fringe.shift();
 
-      dependencies = filteredGraph.get(pkg).dependencies;
+      dependencies = packageGraph.get(pkg).dependencies;
 
       dependencies.forEach((dependency) => {
         if (!~dependentPackages.indexOf(dependency) && !~fringe.indexOf(dependency)) {
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -88,23 +88,22 @@ export default class PackageUtilities {
     const filteredGraph = PackageUtilities.getPackageGraph(packages);
 
     let dependentPackages = [];
-    let fringe = new Set();
-    fringe.add(scope);
+    let fringe = [];
+    fringe.push(scope);
 
     let dependencies;
-    while (fringe.size !== 0) {
-      let pkg = fringe.values().next().value;
+    while (fringe.length !== 0) {
+      let pkg = fringe.shift();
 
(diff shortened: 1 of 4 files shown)


src/PackageUtilities.js
```diff
@@ -83,4 +83,31 @@ export default class PackageUtilities {
...
+    fringe.push(scope);
+
+    let dependencies;
+    while (fringe.length !== 0) {
+      let pkg = fringe.shift();
+
+      dependencies = filteredGraph.get(pkg).dependencies;
+
+      dependencies.forEach((dependency) => {
+        if (!~dependentPackages.indexOf(dependency) && !~fringe.indexOf(dependency)) {
```
thejameskyle: Instead of using `!~` in these three places could you do this instead:

```js
dependentPackages.indexOf(dependency) < 0
```

thejameskyle:
Instead of using scope it makes more sense of me to use:

```diff
lerna ls [package]
```

src/PackageUtilities.js
Steven-Evans: Yep. It's definitely less cryptic than `!~`. 

Any plans to use ES7? It includes Array.prototype.includes which would be the ideal solution to checking for existence.

lukebatchelor:
Ah, I commented on #381 and didnt even think about it when I was doing #386, so these would be conflicting. I added --scope and --ignore support for bootstrap, exec, run, clean and ls to make them more consistent.

As the original ls command listed packages I would find it strange that `ls [package]` performed a different function. Do you think this could be a separate command? `lerna deps [package]` maybe?

Happy to change my PR, I only need the `bootstrap --scope` support but the refactoring made it nice to deduplicate repeated logic and inconsistent handling.

lukebatchelor:
I'm about to start work on #388 and the `filterNonDependentPackages` function looks like it will be super useful for that, so I'm probably going to merge this into that. 

I'm a little hesitant to do more work based on non-merged PRs but I think #384, #386, and #365 all look like they make sense, so hopefully are accepted

src/PackageUtilities.js
```diff
@@ -83,4 +83,29 @@ export default class PackageUtilities {
...
+          fringe.push(dependency);
+        }
+      });
+
+      dependentPackages.push(pkg);
+    }
+
+    return packages.filter((pkg) =>
+      dependentPackages.indexOf(pkg.name) >= 0
+    );
```
lukebatchelor: You probably want to `slice` here so we always return a copy (as in the other filter functions)

src/commands/LsCommand.js
```diff
@@ -1,15 +1,30 @@
 import Command from "../Command";
+import PackageUtilities from "../PackageUtilities";
 
 export default class LsCommand extends Command {
   initialize(callback) {
-    // Nothing to do...
+    if (this.input.length) {
```
thejameskyle: I'm not sure why you need to do all this work. You should already have the package graph which will allow you to find the tree of dependencies of a single package.

hzoo:
I guess this was done in https://github.com/lerna/lerna/pull/386 already? Thanks for the work thou!

PR #386, opened 2016-10-21
[Feature] Adds --scope and --ignore support for bootstrap, exec, run, clean and ls

### LINKED ISSUE #383: Support --scope flag in `lerna bootstap` command
It currently supports the --ignore flag and we'll be using that to work around this at the moment, but it would be nice for consistency and clean-ness if we could get this into lerna as well.

Will be working on the quick work around script tomorrow, if this is something that we would like in lerna I can start on the PR after that. If not, we'll just stick with the work around.

lukebatchelor:
Some refactoring to make --scope and --ignore more consistent across commands.

Fixes #383


lukebatchelor:
So, this was based on https://github.com/asini/asini/pull/8 which had some other changes not in Lerna which I've just realised came from #365. Lots of this was verbatim from the asini merge but there are a couple of differences from what was taken from #365. More than happy to help merge these two together in which ever order, just give me a yell. 👍

src/Command.js
```diff
@@ -100,9 +99,20 @@ export default class Command {
   }
 
   runPreparations() {
+    const scope = this.flags.scope || (this.configFlags && this.configFlags.scope);
+    const ignore = this.flags.ignore || (this.configFlags && this.configFlags.ignore);
```
lukebatchelor: The .configFlags is only set in Bootstrap right now.

README.md
```diff
@@ -486,15 +490,13 @@ $ lerna run --scope toolbar-* test
...
-Excludes a subset of packages when running the `bootstrap` command.
+Excludes a subset of packages when running a command.
 
 ```sh
 $ lerna bootstrap --ignore component-*
 ```
 
-The `ignore` flag, when used with the `bootstrap` command, can also be set in `lerna.json` under the `bootstrapConfig` key. The command-line flag will take precendence over this option. This flag is supported in `bootstrap` and `exec` commands.
-
-**Note**: If both `scope` and `ignore` are provided to `exec` command, `scope` takes precedence.
```
lukebatchelor: An intended side effect of this change is that `scope` wont take precedence anymore. Is there any reason to take precedence when you could simply obey both and throw an error if they would result in no packages being selected? I don't think it would be a common use case to be using both at the same time, but this would make more sense to me.

lukebatchelor:
@hzoo, @thejameskyle, Is there anything else I should do for this to get merged? I've got the next 3 days to be able to work on this pretty freely, and was hoping to do any iteration on this one, starting work merging it with #365 (my PR borrows quite a lot of the refactoring from that), and looking how we might do something with #388, probably/possibly using #384

README.md
joscha: @lukebatchelor When we developed the feature I suggested `ignore` to trump `scope`: https://github.com/lerna/lerna/pull/168#issuecomment-224874490 - not sure why and when that might have changed.

lukebatchelor: @joscha It changed in this PR, hopefully. Looking through that convo, I don't see a reason you cant accept both flags (as in this PR) and throw an error if the result would return no packages. 

Useful for cases like

```sh
npm install --ignore-scripts && lerna bootstrap --scope "ak*" --ignore "ak-editor*"
```

seansfkelley:
@hzoo / @thejameskyle can we merge this? Working around/waiting for prepublishing the entire world is making me sad. :(

src/PackageUtilities.js
```diff
@@ -56,7 +56,28 @@ export default class PackageUtilities {
...
+  * Filters a given set of packages and returns all packages that match the scope glob
+  * and do not match the ignore glob
+  *
+  * @param {!Array.<Package>} packages The packages to filter
+  * @param {Object} filters The scope and ignore filters.
+  * @param {String} filters.scope glob The glob to match the package name against
+  * @param {String} filters.ignore glob The glob to filter the package name against
+  * @return {Array.<Package>} The packages with a name matching the glob
+  */
+  static getFilteredPackages(packages, {scope, ignore}) {
```
seansfkelley: I would perhaps rename this to make it a bit more obviously different than `filterPackages`.

README.md
```diff
@@ -430,6 +433,7 @@ Running `lerna` without arguments will show all commands/options.
 - `lerna`: the current version of Lerna being used.
 - `version`: the current version of the repository.
 - `publishConfig.ignore`: an array of globs that won't be included in `lerna updated/publish`. Use this to prevent publishing a new version unnecessarily for changes, such as fixing a `README.md` typo.
+- `bootstrapConfig.ignore`: an glob that won't be bootstrapped when running the `lerna bootstrap` command.
```
seansfkelley: Should we mention `scope` here too?

src/Repository.js
```diff
@@ -43,7 +44,34 @@ export default class Repository {
...
+  get filteredPackages() {
+    if (!this._filteredPackages) {
+      this.buildPackageGraph();
+    }
+    return this._filteredPackages;
+  }
+
+  get packageGraph() {
+    if (!this._packageGraph) {
+      this.buildPackageGraph();
```
seansfkelley: Shouldn't these be called with arguments? I see `buildPackageGraph` is called in `runPreparations`, but that seems a bit like spooky action at a distance.

Perhaps `Repository` should only cache `packages`/`packageGraph` and filtering should be performed explicitly by the task that wants to filter (i.e. `Command`, as it does now).

lukebatchelor: Hmmm.. I liked the idea of pulling the filtering out of the commands just to make them consistent. Maybe these getters should each take optional { scope, ignore } flags?

lukebatchelor: Honestly it was done like this to match asini's PR (which originated from here), so I'd be happy for it to be either way

src/PackageUtilities.js
lukebatchelor: getScopedAndIgnoredPackages maybe? :/

I'll have a think of a better name

seansfkelley: Or maybe something along the lines of `applyFilteringRules` or something. Yeah, naming is hard. :(

README.md
seansfkelley: I think between (1) functionality consistency and (2) the fact that there is a use-case for "permanent" ignore rules, supporting permanent scope rules is a good idea. Principle of least surprise, and all that.

src/Repository.js
seansfkelley: So the big argument against this structure is that these are getters, but depending on the order that you invoke `repository.packageGraph` and `repository.buildPackageGraph(...)`, you can end up with permanently different values for `packageGraph` (or one of the other fields controlled by this). Spooky! 

There's also a hand-wavier argument about who should be _responsible_ for filtering; I figured that the "Repository" class should probably be a predictable, consistent, immutable representation of the state of the repo, and filtering should be applied elsewhere, namely, in the commands themselves. If you just pulled out the filtering behavior of `buildPackageGraph` to the root `Command` you should still get it across commands, yeah?

Lastly, you can't pass arguments to getters!

README.md
thejameskyle: Ignore should definitely take precedence over scope.

thejameskyle:
Sorry this was left around, rebase and I'll merge

README.md
lukebatchelor: Just to make sure we're saying the same thing, scope wont override ignore in this case. They work together. Scope is a whitelist, ignore is a blacklist. If scope exists, then no other packages other than the ones in scope are included. If ignore exists, no packages matched by ignore will get included. 

So, if you were matched by both, you would still get ignored.

src/Repository.js
seansfkelley: @lukebatchelor please fix the above before merging!

lukebatchelor: For sure!

lukebatchelor: Cool, ended up doing exactly what you said and moved the responsibility in PackageUtilities and it only gets called in one place still (Command.js).

Much better!

lukebatchelor:
Rebased and refactored. Just hoping Appveyor decides to play nicely today... ...

README.md
lukebatchelor: Added 👍

lukebatchelor:
@seansfkelley Wold you be able to do a quick review of this if you have time?

README.md
seansfkelley: Yeah, that's right.

lukebatchelor:
Ah, has to be an owner. Could you take a look when you're ready @thejameskyle ?

lukebatchelor:
@thejameskyle Just bumping this again. When you're free, cheers!

lukebatchelor:
@thejameskyle I've also updated the follow up to this #390 and that should be ready to go too.

Cheers 🍻

lukebatchelor:
😎 Should be ready @gigabo 

I merged master instead of rebasing, is that okay? Last time I rebased I had to fix each conflict of each commit manually for some reason. Didnt want to have to do that again for this and #390. 

(If anyone knows what I might be doing wrong, that would be great too!)

gigabo:
Thanks @lukebatchelor merging should be fine.  Going to squash it onto master either way. 👹 

FWIW the "fix each conflict of each commit manually" thing is actually a _feature_ of rebasing.  It allows the resulting commits to apply cleanly on top of the new base for a linear history.  It can be a pain when there are a lot of them, but sometimes I find it's actually easier to see how they should be resolved when they pop up in the context of the individual commits that introduced them.

So, in Asini we followed this change set up with https://github.com/asini/asini/pull/15 and then https://github.com/asini/asini/pull/18.  Would you be interested in/willing to port those over here? 😁

Issue #388, opened 2016-10-21
[Feature] Add a `--with-deps` flag to bootstrap --scope


lukebatchelor:
After playing with the results on #386 locally for a bit I've realised what would be most useful is if a scoped bootstrap command bootstrapped all the dependencies of packages first. 
- If used without `--scope` it would ignore this.
- Would need some logic to sort out the order of installation if more than one package is matched by `--scope`

The use case for us is lots of packages now taking a long to time prepublish and link. `--scope` helps a lot but not in the case where you've done a clean and just want to start working on one package. You would try to scope to the package you want but your dependencies wont be built.

I can see #384 being super useful for this. If we have a direction on #384 and #386 then I can start working on this next week if you like.

(`--with-deps` is a terrible name, any suggestion would be appreciated. `--recursive` maybe? `--bootstrap-deps`?)

lukebatchelor:
I was leaning towards that too, but I think keeping it like this would give more flexibility. If you want to prepublish deps, you can, but if you just want to clean and rebuild/link one package as fast as possible, you can too. 

Also makes sense from the semantics of --scope, "perform acertain command in a specific package". The only edge case is if your bootstrapping requires a package that you are building at the same time

lukebatchelor:
Maybe it should warn if you are bootstrapping a package that relies on a package you are bootstrapping?

seansfkelley:
Yeah, that also seems reasonable.

lukebatchelor:
Opinion: Should `--with-deps`/`--recursive`/whatever throw an error if `--scope` matches more than one package? The logic for the order to bootstrap them will get pretty complicated and it seems like more of an edge case than the case where you are simply bootstrapping one package. The other way of looking at it is that it would make `--scope`'s sematic meaning be a little less consistent  (`--scope` scopes a command to a set of packages matching a glob **unless** the `--with-deps` flag is set, the `--with-deps` flag is ignored if you don't pass the `--scope` flag, etc, etc). Just seems like it's making more exceptions to the rule.

Anyone have any opinions? I'm leaning towards Throwing if it matches more than one, just because I feel like this covers the majority of use-cases.

seansfkelley:
I would definitely not throw. The semantics of including the transient dependencies are well-defined, and throwing should be reserved for exceptional circumstances where you really cannot proceed (indeed, I think that Lerna throwing when nothing matches scope is bad UX).

[...]

PR #390, opened 2016-10-25
[Feature] Adds `--include-filtered-deps` flag for bootstrap command

lukebatchelor:
> Relies on #386, so the diff will include those changes also.

This adds a flag to the bootstrap command that ensures that **all** dependencies of any bootstrapped packages are also bootstrapped.

The use case for this is needing to "set up" a single package in a clean repository and not wanting to have to prepublish every single package. 

This has been a major pain point for us in a repository with only 61 packages.

Some quick numbers from using in our repo:

```
time npm install
```

Total: 9:45.04 (bootstraps 61 packages)

```
time (npm install --ignore-scripts && npm link lerna && lerna bootstrap --scope ak-field-base --include-deps)
```

Total: 2:22.19 (bootstraps 6 packages, 5 direct dependencies and 1 transitive dependency)


lukebatchelor:
Closes #388

lukebatchelor:
Question: Should this warn if using the `--include-deps` flag when not using `--scope` or `--ignore`? The resulting build will be identical, its just misusing a flag. Happy either way. 👍

lukebatchelor:
The second `Adds docs for --include-deps flag in readme` commit was to force appveyor to rebuild as it is strangely failing to install npm itself, still seems to be failing.

Anyone have any idea why that might be? It says it's a permission error when trying to rename a file and to re-run it as an admin.

lukebatchelor:
Playing with this a bit more locally, I'm wondering if this should also include dev dependencies? At the moment I just manually bootstrap those using the `--scope` flag because there isnt too many.

Anyone have any opinions on this? Happy to add it.

lukebatchelor:
I will definitely add the devDependencies at some point, its bitten me a few times using it locally now :P

ilkkahanninen:
I think it's a good idea to add devDependencies to --include-deps. That way it works as users expect it to work.

thejameskyle:
Hmm, `--include-deps` is ambiguous in certain commands I think. Can you rename it to `--include-filtered-deps` or something similarly descriptive?

lukebatchelor:
@thejameskyle It's not for filtered deps though, it's for transitive deps. 

`--include-transitive-deps` maybe? Bit verbose, but I'm only going to wrap it in scripts anyway. I should hopefully have time to rebase, rename and add the dev dependencies this week/weekend.

thejameskyle:
lol maybe `--include-filter-deps-deps`

lukebatchelor:
`--bootstrap-deps` maybe? Shorter and more descriptive/less ambiguous

thejameskyle:
So the original idea of `--include-filtered-deps` was to mean include the deps that were filtered out `--scope` and `--ignore`.

lukebatchelor:
Happy to go with `--include-filtered-deps`. Tried to make the change today but I have a feeling im stuffing up this rebasing... Going to do it after #386 is merged so I can be sure this is right.

Super simple fix, just need to change the `addDependencies` function to use Object.assign({}, pkg.dependencies, pkg.devDependencies)

lukebatchelor:
@gigabo Could you take  a look at this one as well when you have time?

This one is rebased on #386 and should be ready as well. It's pretty required to make #386 useful.

gigabo:
@lukebatchelor Please rebase this when you get a chance now that #386 has landed in master.

lukebatchelor:
Super weird, I ~did~ rebase on #386 and double checked the git log to make sure they had the same commits. Updating now

lukebatchelor:
Oops, sorry. Got pulled away yesterday, should be good to go @gigabo 
👍

seansfkelley:
Another thought on this topic (hopefully not derailing the PR too much in the process): I think this would be really valuable for all tasks to have, much like your filtering changes. Use case: `lerna run build --scope foo --toposort --include-filtered-deps`. Combined with #412, which I hope I get to today, this would allow me to rebuild `foo` and all of its dependencies in one shot in the right order (I have a Typescript project, so ordering matters, but a full rebuild of dependencies should be useful for non-Typescript projects too). This is very powerful, and likely has other usages that I haven't though of as well.

gigabo:
Hi @lukebatchelor yep, just getting back from the holidays.  Will take a look! 👀

spudly:
Quick question: would it make sense to change this to --include-dependencies'? I envision that at some point we might want a --include-dependents flag and "deps" could be either. 

Use case for --include-dependents would be to quickly install and run tests for any package that uses a package you're modifying.

lukebatchelor:
@spudly it was changed to `--include-filtered-deps` recently. Do you think this could still have be confusing?

seansfkelley:
I'm with @spudly on this: better to be entirely explicit, since it doesn't save many keystrokes that often anyway (so: `--include-filtered-dependencies`). Full disclosure: I was also thinking about an `--include-(filtered?)-dependents` flag (reason being that from my previous comment, it would be awesome to be able to automatically select dependents of my package for rebuilding to make sure things aren't broken).

spudly:
@lukebatchelor & @seansfkelley,

Yes, I do still think we need to explicitly call it `--include-filtered-dependencies`. I know it's long, which is annoying, but this way users know exactly what is going to happen without searching through the documentation, and it leaves room for similar flags down the road.

My vote is for `--include-filtered-dependencies` so that @seansfkelley or someone else can later add `--include-filtered-dependents`.

gigabo:
This looks good to me.

As for `deps` vs `dependents` I don't see that as a big issue.  `deps` is a common short form of `depencies`.  Don't really see it used for `dependents` much.  Not opposed to the long form if others feel strongly.

Thanks for your patience @lukebatchelor, and for the nice clean change set. 👍

lukebatchelor:
Sweet, well I'm happy with either as well. Personally I prefer deps because I spell dependency wrong approximately 100% of the time XD.

I'll leave the call up to you @gigabo / @seansfkelley ?

seansfkelley:
Everyone seems to be ambivalent, so I would err on the side of being explicit and just make it wordier. Reasonable?

lukebatchelor:
Sorry for the delay! Just managed to get some time to do this. Will rename this morning.

lukebatchelor:
@gigabo When you're ready 👌

gigabo:
Thanks @lukebatchelor!

PR #393, opened 2016-10-26
Expect packages config to be array of globs to directories containing package.json

bclinkinbeard:
Updated to address the discussion at https://github.com/lerna/lerna/pull/365#pullrequestreview-3931001


bclinkinbeard:
Is there anything I can do to get this looked at/approved?

PR #412, opened 2016-11-21
Make bootstrap, exec and run commands execute packages in dependency order by default

### LINKED ISSUE #342: [Feature] lerna run: topologically sort packages like bootstrap
[`bootstrap` topologically sorts packages](https://github.com/lerna/lerna/blob/143840eac0117e15c3a9180d8cc48ad830d34f47/src/commands/BootstrapCommand.js#L76) before installing them them to ensure dependencies exist before being required, which is awesome. This behavior would be valuable to have with the `run` command as well. I work on a Typescript-based project, and if I change the interface of a package and want to rebuild everything, I have to make sure I run things in the right order manually or shoehorn it into `prepublish` and depend on `bootstrap` to do it the right way. I currently do the latter, but it's somewhat clunky since I incur the entire cost of `bootstrap` just to recompile a couple things (since `bootstrap` isn't scope-able: #341) and it's not really documented behavior anyway. Consistency between commands also seems valuable (perhaps appropriate for `exec` as well?).

Code-wise, I'm thinking that `BootstrapCommand` could delegate to `RunCommand` (or they delegate to a shared method), and that this method could have some kind of `--toposort` flag. Not sure if said flag should be default on/off, nor if bootstrap should have a different default. I suggest the flag only because there are potential performance implications: I use `--concurrency 16` in CI; if everything was uncontrollably bottlenecked on a few dependencies when it didn't need to be I would be sad. If that wasn't a concern, unconditionally toposorting would probably be the way to go.

Alternately, a more complex toposorting could be used such that a concurrent task fetches the next package that has no unfinished dependents. This would make the flag less useful (though bottlenecking is still possible, it's less likely). This method would be slightly more complicated than just toposort-and-run (which is what bootstrap does now, though in a batched way) since you have to make sure dependencies have actually _finished_ before starting a dependent (since we're not single-threaded).

seansfkelley:


seansfkelley:
Permissions errors on the AppVeyor build per usual. 😞

seansfkelley:
This probably clashes awkwardly with #386; I would prefer that one get merged first and then I'll update this as appropriate.

sbryfcz:
Just wanted to say that I'd love #412 to get merged. Very typical in the micro-repos that I create to have a number of small packages and one that bundles them all into a single place. Would love to be able to run my build scripts and then perform a publish. But as you know, `lerna run` doesn't preserve the toposort order so things get built out of proper sequence. Thanks for all the hard work! Love this library.

gigabo:
@seansfkelley #386 has merged.  Please rebase this when you get a chance.

seansfkelley:
@gigabo updated!

src/commands/BootstrapCommand.js
```diff
@@ -28,8 +28,10 @@ export default class BootstrapCommand extends Command {
    * @param {Function} callback
    */
   bootstrapPackages(callback) {
-    this.filteredGraph = PackageUtilities.getPackageGraph(this.filteredPackages);
     this.logger.info(`Bootstrapping ${this.filteredPackages.length} packages`);
+    this.batchedPackages = this.flags.toposort
```
seansfkelley: Open question: bootstrap used to toposort by default. For consistency, I have tentatively made toposort opt-in-only across the board. Is this behavioral change (1) a good idea? and (2) sufficiently breaking for a beta version to justify special-casing bootstrap?

I am tempted to say (1) yes and (2) no, assuming there is a way we can call out to existing beta users that the default behavior has changed (maybe add a log line in bootstrap if the option was not specified?).

gigabo: Oh, boy.  Yeah, I appreciate that making it opt-in for bootstrap is consistent.

I guess a repo that _does_ need it can just add it to `lerna.json`.

seansfkelley: So are there any changes I should make here, or is it good to go?

seansfkelley: Ping @gigabo: thoughts?

seansfkelley: The most foolproof correct way to implement this flag would then be to default it on always, and provide a `--no-sort` option to disable it, no? (Agree that "topo" is probably noisy; I think "no sort" is phrased the most CLI-like though.)

gigabo: @seansfkelley That sounds great.

Thanks for the suggestion @doug-wade.

xaka:
Is there anything that's blocking this PR? It's becoming important for our team as we're adding more and more packages to monorepo, and doing fresh bootstrap isn't option anymore. The problem is when you have packages A and B (depends on A), and you introduce new package C being used by first two, you have to manually toposort everything in your head and do scoped bootstrap to save time.

gigabo:
@xaka I think @seansfkelley was planning to [make some tweaks](https://github.com/lerna/lerna/pull/412#discussion_r95000529) before we merge.

seansfkelley:
@gigabo it should be good to go now.

gigabo:
Awesome!  Thanks @seansfkelley!

gigabo:
@seansfkelley I updated the issue title to try to reflect the current state of what this provides.  Please improve it if you've got better wording.  It will go into the changelog entry for this PR.

seansfkelley:
Okay, changed (a bit wordier but more precise).

sbryfcz:
Awesome work! Looking forward to the merge.

xaka:
Would be nice to have another pre-release version with this change in 🎆 It's big and important one.

seansfkelley:
@doug-wade done!

PR #413, opened 2016-11-22
Update inquirer to latest version.

szarouski:
I thought that will fix issue with cmd on windows for me, but discovered that I just need to use newer version of node. Since I made an improvement I'll let you decide what to do with this commit.


szarouski:
So build failed for travis due to `stalled build` for v0.10 and for appveyor due to npm-cache issue for v0.12. Do you guys typically retrigger build in this case?

szarouski:
This time AppVeyor build passed, but Travis is stalled again just for v0.10 :(

thejameskyle:
Oh my god, when will someone create a CI service that doesn't get choked up all the time?

szarouski:
I tried to update this PR a few times and results are the same - 0.10 build freeze (only in travis). Perhaps it is time to retire 0.10? Greenkeeper thinks it is a good idea: https://github.com/IndigoUnited/node-cross-spawn/pull/47

gigabo:
I think those node 0.10 failures may actually be real.  It looks like [`inquirer@^1.0.0`](https://github.com/SBoudrias/Inquirer.js/releases/tag/v1.0.0) depends on native Promises.  This might be blocked on [dropping node v0.10 support](https://github.com/lerna/lerna/issues/418).

hzoo:
I'm ok with dropping node 0.10 if its necessary since yarn doesn't support it either, jest, eslint, etc. We just wouldn't be updating in Babel until we drop (hopefully soon)

szarouski:
I updated PR to drop older versions and updated inquirer to v2. Please feel free to decide what to do with this. Thanks.

evocateur:
We're on inquirer v3 now. Thanks for the contribution!

Issue #89, opened 2016-03-23
[Feature] Option instead of lastTag/lastRelease for updated packages


hzoo:
Right now we figure out what packages to publish based on a git diff of the lastTag.

However in the case of a backport (you might publish v5.x.x after a v.6.x.x) and now the lastTag is actually v5.x.x, so maybe you should be able to pass in a tag, otherwise default to the latest (we should log out the last tag if we aren't already or prompt it).

thejameskyle:
If you are in a separate branch for the older version I assume we can just change:

```diff
- git rev-list --tags --max-count=1           # get commit of last tag across branches
+ git rev-list -n 1 $(git describe --tags)    # get commit of last tag in current branch
```

[[ref](https://github.com/kittens/lerna/blob/master/lib/utils/gitUtils.js#L26)]

hzoo:
Would you ever want to do anything else such that you might want to specify the version manually? Probably not right?

What do you think about logging that info in the lerna updated/publish command as well? Basically something like `Checking updated files since v6.8.1`

And based on your suggestion, we just check the current branch - if it's `master` do the regular command, and if it's not do the 2nd one?

Gongreg:
Dumb, We got into this issue, that if we do changes on different branches for different versions, we need to use newest tag in that branch, not in whole project. Could we try to do a pull request for this feature? Maybe add passing some kind of parameter for turning this option on?

PR #415, opened 2016-11-23
Issue 89: Added flag option --branch, for comparing with current last version.

Gongreg:
Added flag option --branch, which checks for changes between last tag in current branch (instead of current latest tag in whole repository), it allows to publish older versions of project. (Issue #89).

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -81,15 +81,18 @@ export default class FileSystemUtilities {
   @logger.logifySync()
   static isSymlink(path) {
     const lstat = fs.lstatSync(path);
-    const isSymlink = lstat && lstat.isSymbolicLink()
+    let isSymlink = lstat && lstat.isSymbolicLink()
       ? fs.readlinkSync(path)
       : false;
-    if (process.platform === "win32" && lstat && lstat.isFile() && !isSymlink) {
-      try {
-        return resolve(dirname(path), readCmdShim.sync(path));
-      } catch (e) {
-        return false;
+    if (process.platform === "win32" && lstat) {
+      if (lstat.isFile() && !isSymlink) {
+        try {
+          return resolve(dirname(path), readCmdShim.sync(path));
+        } catch (e) {
+          return false;
+        }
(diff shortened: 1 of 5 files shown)


thejameskyle:
Does the code from this comment not solve your use case? https://github.com/lerna/lerna/issues/89#issuecomment-200172189

Gongreg:
I used your comment as a reference, 
```
git rev-list -n 1 $(git describe --tags) 
```

```
git rev-list -n 1 $(git describe --abbrev=0 --tags)
```

Only difference i had to add abbrev option.

gigabo:
Oh, this is cool!

Does it even need to be an option?  Seems like it would be useful as _default_ behavior, maybe with a warning message if you're not currently on `master`?

Gongreg:
I've added this as a option just for beginning (tried it out and it worked well for us), since tests for publish command looked like quite complicated. I could work on making it as a default behaviour/add message if you are not on master. :)

Gongreg:
So should I follow up with doing something more about this? I agree that it could be default behaviour, didn't have a reason not to use it (it works same as before if you are on master, if you check older tag/branch/commit you expect it to work with last version, so I am quite confident that we can make this default behaviour). Also will someone check the tests after the changes? It was quite difficult to read your tests :)

Gongreg:
Okay, I've made this execute as a default behaviour.
Also I've added check to see if user is inside some branch, because lerna publish needs to push to branch (shows error message if he is not in a branch).
Lerna updated works properly even when user simply checkouts older commit.

spudly:
Any movement on this? I'd really like to see this become the default on branches and it looks like the only thing blocking the travis build is a few linter issues...

Gongreg:
I'll fix the linter issues today/tomorrow. But I think this thing really needs tests if it is being done as default behaviour. I will inspect them more, I gave up too soon last time I looked into them. Also I've left the company I've worked in, I will contact them to allow me to update the code or I will reopen the PR from different fork.

Gongreg:
Created new pr.

PR #425, opened 2016-12-05
symlink packages specified by relative directory

### LINKED ISSUE #424: Packages specified as file relative are not symlinked
if package `b` is linked to package `a` as `file:../a` it will not be symlinked to package `b`. To me, this is a bug.

dmfenton:
Resolves #424. One benefit I'd like to point out is that relative dependencies are compatible with Yarn whereas non-published dependencies are not.

### DIFF
--- a/src/Package.js
+++ b/src/Package.js
@@ -97,7 +97,7 @@ export default class Package {
     if (semver.satisfies(actualVersion, expectedVersion)) {
       return true;
     // or the dependency is specified as a relative folder
-    } else if (/^file:\.\.\//.test(dependency.version)) {
+    } else if (/^file:/.test(dependency.version)) {
       return true;
     }
 
--- a/src/Package.js
+++ b/src/Package.js
@@ -96,6 +96,9 @@ export default class Package {
     // check if semantic versions are compatible
     if (semver.satisfies(actualVersion, expectedVersion)) {
       return true;
+    // or the dependency is specified as a relative folder
+    } else if (/^file:\.\.\//.test(dependency.version)) {
+      return true;
     }
 
     if (showWarning) {
--- a/test/Package.js
+++ b/test/Package.js
@@ -15,7 +15,10 @@ describe("Package", () => {
         bin: "bin.js",
         scripts: { "my-script": "echo 'hello world'" },
         dependencies: { "my-dependency": "^1.0.0" },
-        devDependencies: { "my-dev-dependency": "^1.0.0" },
+        devDependencies: {
+          "my-dev-dependency": "^1.0.0",
+          "my-relative-dependency": "file:../my-relative-dependency"
+        },
         peerDependencies: { "my-peer-dependency": "^1.0.0" }
       },
       "/path/to/package"
@@ -61,7 +64,10 @@ describe("Package", () => {
 
   describe("get .devDependencies", () => {
     it("should return the devDependencies", () => {
-      assert.deepEqual(pkg.devDependencies, { "my-dev-dependency": "^1.0.0" });
+      assert.deepEqual(pkg.devDependencies, {
+        "my-dev-dependency": "^1.0.0",
+        "my-relative-dependency": "file:../my-relative-dependency"
+      });
     });
   });
 
@@ -75,7 +81,8 @@ describe("Package", () => {
     it("should return the combined dependencies", () => {
       assert.deepEqual(pkg.allDependencies, {
         "my-dependency": "^1.0.0",
-        "my-dev-dependency": "^1.0.0"
+        "my-dev-dependency": "^1.0.0",
+        "my-relative-dependency": "file:../my-relative-dependency"
       });
     });
   });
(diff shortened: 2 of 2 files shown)


dmfenton:
the appveyor build fail was unrelated
```
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files (x86)\\nodejs\\\\node.exe" "C:\\Program Files (x86)\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
npm ERR! node v0.10.47
npm ERR! npm  v2.15.1
npm ERR! path C:\Users\appveyor\AppData\Roaming\npm-cache\lodash\4.17.2\package.tgz
npm ERR! code EPERM
npm ERR! errno 50
npm ERR! Error: EPERM, rename 'C:\Users\appveyor\AppData\Roaming\npm-cache\lodash\4.17.2\package.tgz'
npm ERR!  { [Error: EPERM, rename 'C:\Users\appveyor\AppData\Roaming\npm-cache\lodash\4.17.2\package.tgz']
npm ERR!   errno: 50,
npm ERR!   code: 'EPERM',
npm ERR!   path: 'C:\\Users\\appveyor\\AppData\\Roaming\\npm-cache\\lodash\\4.17.2\\package.tgz',
npm ERR!   parent: 'babel-cli' }
npm ERR! 
npm ERR! Please try running this command again as root/Administrat
```

gigabo:
@dmfenton Please rebase this when you get a chance.

dmfenton:
Rebased

src/Package.js
```diff
@@ -96,6 +96,9 @@ export default class Package {
     // check if semantic versions are compatible
     if (semver.satisfies(actualVersion, expectedVersion)) {
       return true;
+    // or the dependency is specified as a relative folder
+    } else if (/^file:\.\.\//.test(dependency.version)) {
```
gigabo: Seems not too fancy?  The [supported dependency forms](https://docs.npmjs.com/files/package.json#local-paths) are pretty limited.

I guess we can start with this, and I'll file a ticket for improving support.

src/Package.js
```diff
@@ -96,6 +96,9 @@ export default class Package {
...
+      return true;
```
gigabo: Hmm... so this also doesn't check whether the referenced directory actually exists.

What is the behavior if I have a typo in a relative dependency path?

dmfenton: bootstrapping would fail with an npm error as it should. I don't think it should be the job of this tool to detect whether someone has misconfigured their package.json

src/Package.js
dmfenton: that link shows that local directories will be serialized to `file:` in the package.json. I suppose we could loosen whats there a bit to look for only`file:`. That should work well.

xaka:
would be nice to get this in! we're facing the same use case...

gigabo:
@hzoo @rygine @evocateur I'd like to get some more eyes on this.

It _seems_ straightforward, but it it also seems like it could introduce unexpected behavior?

Really just want to make sure we've thought carefully about (and are fine with) the implications.

evocateur:
@dmfenton @gigabo I'm concerned that there are no tests validating the relative file symlinking during bootstrap, though I do appreciate the direct tests on the `Package` class.

Also, it concerns me that there is no restriction against publishing modules that have relative file `dependencies` (relative file `devDependencies` don't concern me, as they'll always be in the "right place" by virtue of the monorepo). I don't know offhand if `npm` itself throws an error in that case, so it's worth investigating.

doug-wade:
@evocateur I just created a tiny little repo with two modules, one which uses a relative import to import the other, and npm does _not_ emit a warning or error of any kind.   I agree that we're introducing a footgun here; I think I might want to see an error emitted if there is a relative import in a module not marked "private" before we merge.

akosyakov:
Any updates on it?

src/Package.js
```diff
@@ -96,6 +96,9 @@ export default class Package {
     // check if semantic versions are compatible
     if (semver.satisfies(actualVersion, expectedVersion)) {
       return true;
+    // or the dependency is specified as a relative folder
+    } else if (/^file:/.test(dependency.version)) {
```
akosyakov: i think `file:` is not actually necessary, one can do:
```json
{
  "dependencies": {
    "lodash": "../../node_modules/lodash"
  }
}
```

I would check npm repo how they test whether a version is a local path.

MiguelMadero:
@doug-wade  @dmfenton any update here? 
We have a mono where all packages are private and looks like this is the only reason why yarn doesn't work on its own, we can only install through lerna. I would be happy to look into the blocking issues if nobody is actively working on it. Looks like the only blocker if we do this for not private packages

dmfenton:
I've found a workaround and the devs weren't happy with my solution so I won't be carrying this forward. You're welcome to adopt this PR, however.

src/Package.js
airato: looks like both are valid
https://docs.npmjs.com/files/package.json#local-paths

CapitanRedBeard:
@MiguelMadero Our private repo would also benefit from the use of this PR. I think the only developer concern was 

> I think I want to see an error or warning emitted if there is a relative import in a module not marked "private" before we merge.

Have you looked into this?

Issue #359, opened 2016-10-03
Question: how to fix versions between dependencies?


SamyPesse:
I don't know if Lerna has some automation to manage this, and I couldn't find any related issues to this:

Basically, I understand that the "Fixed/Locked mode" uses one version for all modules (this is what I want), but is there a way to fix dependencies between modules?

For example, if A depends on B, and the `version` in `lerna.json` is `4.1.0`, I want `packages/A/package.json` to be updated with:

``` js
{
  "dependencies": {
    "B": "4.1.0"
  }
}
```

instead of `"B": "^4.0.0"` (it seems to be what Babel is doing).

SamyPesse:
Update: It doesn't look like something that can be done by lerna. I'll write a script in the repository to update all `package.json` files to fix versions according to the "master" version.

SamyPesse:
Here is the script I'm using:

```js
#! /usr/bin/env node

const fs = require('fs');
const path = require('path');

const lernaConfig = require('../lerna.json');

// List all the packages
const PACKAGES_DIR = path.resolve(__dirname, '../packages');
const packages = fs.readdirSync(PACKAGES_DIR);

function updateDependencies(dependencies) {
    Object.keys(dependencies).map((key) => {
        if (!packages.includes(key)) {
            return;
        }

        dependencies[key] = lernaConfig.version;
    });
}

packages.forEach((name) => {
    // Avoid .DS_Store
    if (name[0] === '.') {
        return;
    }

    const pkgPath = path.resolve(PACKAGES_DIR, name, 'package.json');
    const pkg = require(pkgPath);

    pkg.version = lernaConfig.version;
    updateDependencies(pkg.dependencies);
    updateDependencies(pkg.devDependencies);

    fs.writeFileSync(pkgPath, JSON.stringify(pkg, null, 3), 'utf-8');
});
```

Gongreg:
Hey, when do you run this script? One person in our team accidentally made a change in multiple packages, which looked like it didn't create a breaking change, but because inner package dependencies are not locked, it broke other place which used older version. This thing would help us a lot.

thejameskyle:
Merging into #259

gigabo:
Looking at this again, I think it may actually be different from #259.  If I'm reading this right, this issue is about pinning to exact versions _within_ the repo, while #259 is about managing _external_ deps.

@Gongreg pointed out that https://github.com/lerna/lerna/pull/426 seems to address _this_ issue.

[...]

PR #426, opened 2016-12-05
Add support for hidden '--exact' flag

L8D:
This allows for users to opt to use exact versions of cross-dependencies when publishing.

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -294,7 +294,7 @@ export default class PublishCommand extends Command {
       const version = this.updatesVersions[depName];
 
       if (deps[depName] && version) {
-        deps[depName] = "^" + version;
+        deps[depName] = this.flags.exact ? version : "^" + version;
       }
     });
   }
--- a/test/PublishCommand.js
+++ b/test/PublishCommand.js
@@ -783,4 +783,102 @@ describe("PublishCommand", () => {
       }));
     });
   });
+
+  /** =========================================================================
+   * NORMAL - EXACT REPO VERSION
+   * ======================================================================= */
+
+  describe("normal mode with --repo-version and --exact", () => {
+    let testDir;
+
+    beforeEach((done) => {
+      testDir = initFixture("PublishCommand/normal", done);
+    });
+
+    it("should publish the changed packages", (done) => {
+      const publishCommand = new PublishCommand([], {
+        repoVersion: "1.0.1",
+        exact: true
+      });
+
+      publishCommand.runValidations();
+      publishCommand.runPreparations();
+
+      assertStubbedCalls([
+       [ChildProcessUtilities, "execSync", {}, [
+         { args: ["git tag"] }
+       ]],
+       [PromptUtilities, "confirm", { valueCallback: true }, [
+         { args: ["Are you sure you want to publish the above changes?"], returns: true }
+       ]],
+       [ChildProcessUtilities, "execSync", {}, [
+         { args: ["git add " + escapeArgs(path.join(testDir, "lerna.json"))] },
+         { args: ["git add " + escapeArgs(path.join(testDir, "packages/package-1/package.json"))] },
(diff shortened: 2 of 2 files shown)


thejameskyle:
Want to add a test?

L8D:
Yeah...

L8D:
@thejameskyle added.

src/commands/PublishCommand.js
```diff
@@ -292,7 +292,7 @@ export default class PublishCommand extends Command {
       const version = this.updatesVersions[depName];
 
       if (deps[depName] && version) {
-        deps[depName] = "^" + version;
+        deps[depName] = this.flags.exact ? version : "^" + version;
```
gigabo: It's not obvious to me just looking at this that it works.

I mean, I know it _does_ (thanks to your tests), but I'm not confident about operator precedence between `?:` and `+`.  This would be a clearer for me with a set of parentheses.

gigabo:
@L8D Why is this a _hidden_ option?

Or, asked another way, what makes this option useful?  Could you write that down as documentation so other users could more easily benefit from it?

Gongreg:
Does it solve this issue #359 ?

src/commands/PublishCommand.js
lukebatchelor: I think you might have read it wrong. It's just a ternary operator

```
const a = someFlag ? 1 : 2;
```

Sets a to 1 if someFlag is true and 2 if it's false.

Gongreg: I think to understand properly you need to know arity of operators, so it can become confusing for some people. You can read it both as 
```
(this.flags.exact ? version : "^") + version;
```
and
```
this.flags.exact ? version : ("^" + version);
```

lukebatchelor: Oh *facepalm*

gigabo:
@Gongreg Yeah it looks like it does.  Thanks for making that connection!

nkbt:
We'd love to have this one in too, since we never use `^` versions anywhere and always save exact versions.

If Lerna supports `save-exact = true` from `.npmrc` (per package) that would be even nicer then extra flag.

jkimbo:
Just to add, this should be the default when using the `--canary` flag since otherwise cross dependencies get set to something like `^2.3.1-alpha.123456` which doesn't necessarily resolve to that specific version.

lewisdawson:
@gigabo @thejameskyle Any idea on when this may be merged into master? This feature definitely is useful when trying to manage multi-dependency builds across different environments.

hzoo:
> If Lerna supports save-exact = true from .npmrc (per package) that would be even nicer then extra flag.

I guess this could be a diff feature request.

Are we still making this hidden? if it solves #359 we such document in the readme?

L8D:
@hzoo I originally made it "hidden" because I wasn't gonna spend time writing documentation for it in this PR. If someone is up for documenting it that'd be great but I'm using this patched version of lerna for my production builds so I figure'd I drop a PR for it so I don't have to maintain my patched version.

dortzur:
Please god merge this. We're having issues publishing a fix from master branch when dev is ahead.

gigabo:
> I wasn't gonna spend time writing documentation

*sigh*

Okay, thanks @L8D.  Added https://github.com/lerna/lerna/issues/510 for documentation.

gigabo:
Also added https://github.com/lerna/lerna/issues/511 for durable configuration in `lerna.json`.

dortzur:
@gigabo thanks!!!

lewisdawson:
@gigabo Thanks for merging this in!

Brantron:
So if we have `package A` which relies on `package B` and a breaking change is introduced in `package B`, this allows us to `A` to a previously published version of `B`?

evocateur:
@Brantron This just affects the `save-prefix` of the dependency (which is `"^"` by default, yielding `^1.2.3`; `--exact` makes it `""`, yielding `1.2.3`)

Issue #402, opened 2016-11-09
[Feature] Display private packages in ls


dmfenton:
Is this by design?

before
```
$ lerna ls 
Lerna v2.0.0-beta.30
Independent Versioning Mode
- ember-arcgis-opendata
- ember-arcgis-page-layout
- ember-arcgis-portal-services
- ember-cli-opendata-itempicker
- ember-cli-opendata-pages
```

removed private from opendata-ui and opendata-admin

```
$ lerna ls
Lerna v2.0.0-beta.30
Independent Versioning Mode
- ember-arcgis-opendata
- ember-arcgis-page-layout
- ember-cli-opendata-itempicker
- ember-cli-opendata-pages
- opendata-admin
- opendata-ui
```

thejameskyle:
That was intentional I believe.

We can do a couple of things:

```js
$ lerna ls
- x (private)
- y
- z
```

```js
$ lerna ls --all
- x (private)
- y
- z
```

```js
$ lerna ls --all
- x
- y
- z
```

dmfenton:
The first option makes the most sense. I don't see the use cases for the other two.

hzoo:
The relevant code is 
https://github.com/lerna/lerna/blob/524377cdc767afe50f1e2b4dd4f91d6df8934974/src/commands/LsCommand.js#L11-L12

ben-eb:
Resolved in https://github.com/lerna/lerna/issues/427.

PR #427, opened 2016-12-07
issue #402: display private packages in lerna ls

chrishelgert:
as mentioned in #402, this pr will bring private repos to _lerna ls_.

* test for LsCommand is running
* eslint 0 erros 🙂

example


hzoo:
Nice!

gigabo:
Nice @chrishelgert!  The color makes it really stand out.

Want to give the output from `lerna updated` and `lerna publish` similar treatment?

PR #434, opened 2016-12-13
Use urlEncode to handle spaces in file names, closes #432

### LINKED ISSUE #432: import failure: "git diff header lacks filename information when removing 1 leading pathname component"
I'm trying to init a new lerna instance, and import some (very large) old repo's. And I am running into this error.

```
lerna import ../<some-large-repo>/
Lerna v2.0.0-beta.30
About to import 1947 commits into from ../<some-large-repo>/ into packages/<some-large-repo>
? Are you sure you want to import these commits onto the current branch? Yes
Errored while running ImportCommand.execute
Failed to apply commit a76cb3c23.
Error: Command failed: git am -3
error: git diff header lacks filename information when removing 1 leading pathname component (line 15046)
error: could not build fake ancestor

Rolling back to previous HEAD (commit 3ff03a0).
```

ericwooley:


ericwooley:
I added files in the test fixture that has spaces in them. Which broke the tests, then fixed them. Hopefully that is enough test coverage.

src/commands/ImportCommand.js
```diff
@@ -87,16 +87,52 @@ export default class ImportCommand extends Command {
...
+        }
+        if (a.length > b.length) {
+          return -1;
+        }
+        return 0;
+      })
+      // Map files to original and urlEncoded
+      .map(function (file) {
+        return {
+          encodedString: encodeURIComponent(file),
```
gigabo: Whoa, `git am` supports URL encoded filenames?  Nice find!

src/commands/ImportCommand.js
```diff
@@ -87,16 +87,52 @@ export default class ImportCommand extends Command {
...
+      patch = patch
         .replace(/^([-+]{3} [ab])/mg,     replacement)
         .replace(/^(diff --git a)/mg,     replacement)
         .replace(/^(diff --git \S+ b)/mg, replacement)
         .replace(/^(rename (from|to)) /mg, `\$1 ${this.targetDir}/`);
 
+      // Remove all url encoded file names from patch
+      patch = filesInPatch.reduce(function (partiallyReplacedPatch, file) {
+        return partiallyReplacedPatch.replace(new RegExp(file.encodedString, "mg"), file.originalString);
+      }, patch);
```
gigabo: Oh!  I see, you're un-doing the URL encoding.

I like this solution.  Much nicer than trying to make a regex that handles the spaces!

src/commands/ImportCommand.js
```diff
@@ -87,16 +87,52 @@ export default class ImportCommand extends Command {
...
 
       // Create a patch file for this commit and prepend the target directory
       // to all affected files.  This moves the git history for the entire
       // external repository into the package subdirectory, commit by commit.
-      const patch = this.externalExecSync(`git format-patch -1 ${sha} --stdout`)
+      let patch = this.externalExecSync(`git format-patch -1 ${sha} --stdout`);
+
+      // Replace all file names with url encoded file names
+      patch = filesInPatch.reduce(function (partiallyReplacedPatch, file) {
+        return partiallyReplacedPatch.replace(new RegExp(file.originalString, "mg"), file.encodedString);
```
gigabo: Is there any danger here of regex meta characters in a filename messing up the match?

Or will the URL encoding take care of that?

src/commands/ImportCommand.js
ericwooley: Thanks! Yea, I could not find a reasonable way at all to do regex. 

However this is still running into other problems I have yet to figure out on the larger repos.

I am using `git subtree` to import them, and it seems to have worked really well. I'm still trying to figure out if it worked correctly in terms of commit history.

ericwooley:
It seems like maybe this should be closed? It's been open for quite a while, and has since been conflicted.

PR #435, opened 2016-12-13
Use symlinks with relative paths instead of absolute on non-windows environments (fixes #423).

### LINKED ISSUE #423: Use symlinks with relative paths instead of absolute on non-windows environments
It seems that with [the great effort to make symlinks work on win32](https://github.com/lerna/lerna/pull/348) portability of a lerna repo has been reduced, as the use of "junction" to create a symlink requires [the use an absolute destination path](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback). What this means is that after running `lerna bootstrap` and then moving the project folder, all symlinks are invalid and no longer point to an existing directory.

Why care? When deploying a repo that uses Lerna to Heroku the app gets built in a temporary folder, to then be copied over to it's runtime destination. You *can* run `lerna bootstrap` before booting your server, however, this also means any other `prepublish` hooks will be triggered, causing much of the app to build again (as that's the main means of building node apps before they are deployed on Heroku).

JaapRood:
The use of absolute paths in symlinks breaks portability of the build, which causes issues in environments like Heroku, where an app is built in a temporary directory to then be moved later. See #423 for more information.

The changes proposed adapt `FileSystemUtilities` to write symlinks with relative paths and resolve them upon reading them. In windows absolute paths are maintained as relative paths are [not supported](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback).

I'm new to the codebase and don't have access to a windows machine with a development environment, so please feel free to pull my approach completely apart.

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -5,7 +5,8 @@ import rimraf from "rimraf";
 import fs from "fs";
 import cmdShim from "cmd-shim";
 import readCmdShim from "read-cmd-shim";
-import { resolve, dirname, relative, isAbsolute } from "path";
+import { resolve, dirname, relative } from "path";
+import pathIsAbsolute from "path-is-absolute";
 
 const ENDS_WITH_NEW_LINE = /\n$/;
 
@@ -63,7 +64,7 @@ export default class FileSystemUtilities {
       }
       type = "file";
     }
-    if (isAbsolute(src) && process.platform !== "win32") {
+    if (pathIsAbsolute(src) && process.platform !== "win32") {
       if (type === "junction") {
         type = "dir";
       }
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -5,7 +5,7 @@ import rimraf from "rimraf";
 import fs from "fs";
 import cmdShim from "cmd-shim";
 import readCmdShim from "read-cmd-shim";
-import { resolve, dirname } from "path";
+import { resolve, dirname, relative, isAbsolute } from "path";
 
 const ENDS_WITH_NEW_LINE = /\n$/;
 
@@ -62,6 +62,12 @@ export default class FileSystemUtilities {
         return;
       }
       type = "file";
+    }
+    if (isAbsolute(src) && process.platform !== "win32") {
+      if (type === "junction") {
+        type = "dir";
+      }
+      src = relative(dirname(dest), src);
     }
     fs.lstat(dest, (err) => {
       if (!err) {
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -63,7 +63,9 @@ export default class FileSystemUtilities {
       }
       type = "file";
     }
-    src = relative(dirname(dest), src);
+    if (process.platform === "win32") {
+      src = relative(dirname(dest), src);
+    }
     fs.lstat(dest, (err) => {
       if (!err) {
         // Something exists at `dest`.  Need to remove it first.
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -65,9 +65,6 @@ export default class FileSystemUtilities {
       type = "file";
     }
     if (pathIsAbsolute(src) && process.platform !== "win32") {
-      if (type === "junction") {
-        type = "dir";
-      }
       src = relative(dirname(dest), src);
     }
     fs.lstat(dest, (err) => {
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -6,7 +6,6 @@ import fs from "fs";
 import cmdShim from "cmd-shim";
 import readCmdShim from "read-cmd-shim";
 import { resolve, dirname, relative } from "path";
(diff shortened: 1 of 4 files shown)


src/FileSystemUtilities.js
```diff
@@ -63,6 +64,12 @@ export default class FileSystemUtilities {
       }
       type = "file";
     }
+    if (pathIsAbsolute(src) && process.platform !== "win32") {
+      if (type === "junction") {
+        type = "dir";
+      }
+      src = relative(dirname(dest), src);
```
gigabo: So this is taking a path that was _already_ absolute and making it relative?

Where are we making it absolute?  Can we just skip that part in non-windows environments?

My worry here is we might take a dep that was _explicitly_ absolute and make it relative.

gigabo:
Thanks for taking this on @JaapRood!

src/FileSystemUtilities.js
gigabo: Thanks for the explanation. 👍

src/FileSystemUtilities.js
```diff
@@ -63,6 +64,12 @@ export default class FileSystemUtilities {
       }
       type = "file";
     }
+    if (pathIsAbsolute(src) && process.platform !== "win32") {
+      if (type === "junction") {
+        type = "dir";
+      }
```
gigabo: I think this bit is unnecessary?  The `type` argument is [ignored outside of windows](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback).

> The type argument can be set to 'dir', 'file', or 'junction' (default is 'file') and is only available on Windows (ignored on other platforms).

JaapRood: You're totally right, I must have misread that. For some reason I had it in my head that only `'junction'` was ignored on other platforms, but it says all of them are. Makes total sense. I'll push a fix.

JaapRood:
Hmm, Travis seems to run this pull request against the tests of beta 31 instead of beta 30 that this is based off. As I've added a newer commit than master, we can also no longer just pull in changes through the Github UI. @gigabo want me to rebase on current master and force push it?

gigabo:
Yep, please rebase.

The travis build checks out via [`git fetch origin +refs/pull/435/merge:`](https://travis-ci.org/lerna/lerna/jobs/184096785#L128), which makes sure it's running against current master.  Good thing in this case!

gigabo:
Once you rebase you can run `npm test` to see the failures locally.

JaapRood:
And we're back in business :)

src/FileSystemUtilities.js
```diff
@@ -63,6 +64,9 @@ export default class FileSystemUtilities {
       }
       type = "file";
     }
+    if (pathIsAbsolute(src) && process.platform !== "win32") {
```
gigabo: Reading through [the docs](https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback) again it actually looks like a relative path would be fine on Windows, since it will be automatically converted to absolute?

> Note that Windows junction points require the destination path to be absolute. When using 'junction', the target argument will automatically be normalized to absolute path.

And if the path coming in here is always absolute can we just skip this whole conditional?

Then we could also avoid adding a dep on `pathIsAbsolute`.

test/BootstrapCommand.js
```diff
@@ -15,6 +15,10 @@ import assertStubbedCalls from "./_assertStubbedCalls";
 
 const STDIO_OPT = ["ignore", "ignore", "pipe"];
 
+const resolveSymlink = (symlinkLocation) => {
+  return path.resolve(path.dirname(symlinkLocation), fs.readlinkSync(symlinkLocation));
+};
```
gigabo: Oh, this is nice to have factored out.  Thanks!

src/FileSystemUtilities.js
JaapRood: Fair point, with the removing of the `type = 'dir'` earlier on, the point of having this around is getting weaker. I've gotten in the habit of checking for absolute paths as passing a relative path to `Path.relative` has always left me a bit confused:

```
$ node
> const path = require('path')
undefined
> path.relative('/wooo/aaa', '/wooo/bbb')
'../bbb'
> path.relative('/wooo/aaa', './bbb')
'../../Users/jaaprood/sites/lerna/bbb'
```

However, earlier on I made the argument that we can always expect an absolute path, so in this case those worries don't really mean much. I'll remove it, saving a dep 👍

JaapRood:
@gigabo Appveyor gives us an answer to the question whether we need the windows conditional: we do. Upon passing the relative path to `fs.symlink` it seems to normalize it with the current working directory as the base, instead of the location of symlink. That makes sense: you need a base to resolve a relative path against, there's no way for `fs.symlink` to know what that is.

gigabo:
Oh yeah that makes sense.  Sorry for leading you astray!

JaapRood:
Well, that commit was embarrassing! 😂

JaapRood:
@gigabo is there anything else I can do to make this happen?

gigabo:
Nope, I think it's ready.  Thanks again @JaapRood!

Issue #431, opened 2016-12-12
Highlight private packages in updated/publish output


gigabo:
@chrishelgert Added private package annotation to `ls` output in https://github.com/lerna/lerna/pull/427.

The `updated` command includes the `(private)` annotation, but without the red highlighting.

The `publish` command also includes private packages in its list of "changes", but it doesn't annotate them at all.

It would be nice to have consistent output across these three commands, with private packages highlighted in the same way in each.

gigabo:
Oh, whoops, the `publish` portion of this was already covered in https://github.com/lerna/lerna/issues/337.

I'll close that since this is more complete.

chrishelgert:
i can give it a try :simple_smile:

gigabo:
This landed in [`lerna@2.0.0-beta.31`](https://github.com/lerna/lerna/releases/tag/v2.0.0-beta.31)! 🎉

PR #436, opened 2016-12-13
Highlight private packages in updated/publish output

chrishelgert:
As requested in #431, this PR adds the styled private annotations for updated and publish-command.
i wasn´t really sure about the position of private in the publish-command, but i think, this solution looks a litte bit nicer.

If you like it otherway, i can change the position :)


chrishelgert:
the error in appveyor looks like a problem with the build-cache 😢

src/commands/PublishCommand.js
```diff
@@ -231,7 +232,8 @@ export default class PublishCommand extends Command {
     this.logger.newLine();
     this.logger.info("Changes:");
     this.logger.info(this.updates.map((update) => {
-      return `- ${update.package.name}: ${update.package.version} => ${this.updatesVersions[update.package.name]}`;
+      const pkg = update.package;
+      return `- ${pkg.name}: ${pkg.version} => ${this.updatesVersions[pkg.name]}${pkg.isPrivate() ? ` (${chalk.red("private")})` : ""}`;
```
gigabo: Yeah, sees nicer to have it at the end like this. 👍

chrishelgert:
my pleasure 😃

gigabo:
Heh, well... AppVeyor passed, but it looks like this has a conflict with https://github.com/lerna/lerna/pull/386 which just merged before it.

@chrishelgert Please rebase when you get a chance and we'll get this in!

chrishelgert:
@gigabo so no conflicts with base branch :)

Issue #167, opened 2016-05-26
[Feature] Expose a formal programatic API


jpnelson:
@thejameskyle asked me to compile this list of things we'd need to expose for a public API, for https://github.com/atlassian/lerna-semantic-release
- We need a way to know what all of the packages are (currently using `PackageUtilities.getPackages`). We need to be able to get the `package.json` data from the packages.

Other than that, lerna is mostly used through the CLI – you `lerna run` most of the semantic release steps. The only other things necessary have been added in PR's, such as https://github.com/lerna/lerna/pull/163

gigabo:
lerna-changelog is currently using:

- "lerna/lib/Repository" for `lernaJson` (config object) and `rootPath` (repo root)
- "lerna/lib/progressBar" just to present a lerna-like UI feel

Wish list:

- Would _like_ to get at a `Package` object, ideally given just a package name.  Doesn't look like such an interface exists even internally currently.  Something simple like: `lerna.getPackageByName(<name>)`.  Want to use this to [check `package.isPrivate()`](https://github.com/lerna/lerna-changelog/issues/15).

seansfkelley:
#174 might be a good forcing function for this hypothetical API.

ahfarmer:
I currently use the non-formal programmatic API to find out which packages have changed on CI, and only test those packages (for speed reasons).

Would love it if it was a formal API. What I'm doing is not ideal:

```
const logger = require('lerna/lib/logger');
const UpdatedPackagesCollector = require('lerna/lib/UpdatedPackagesCollector');
const Repository = require('lerna/lib/Repository');
logger.setLogLevel('silent');

const repo = new Repository();
const collector = new UpdatedPackagesCollector(repo, {}, {});
console.log(collector.collectUpdatedPackages()); // but I don't actually log them - I use this information
```

lukebatchelor:
@ahfarmer That is literally exactly what I'm about to do.

@gigabo FYI, I'm looking at setting this up in our CI and then pushing changes back to allow this. Is this still something you think we'd like in lerna? Do you have a preferred way of exposing functions?

Alxandr:
Any news on this? Will it be available for 2.0?

evocateur:
@Alxandr It's not really gonna happen until #174 picks up some steam. The subpath imports _technically_ work now, but I wouldn't recommend that as subpath imports are essentially abusing what amounts to a private API, and is thus not guaranteed by semver (that is, it could break at any patch or minor version).

PR #437, opened 2016-12-13
[Feature] Added Callback to Command Run

kwelch:
Exposes a callback to be passed to the internal runCommand.

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -20,7 +20,7 @@ export default class Command {
     this.concurrency = (!flags || flags.concurrency === undefined) ? DEFAULT_CONCURRENCY : Math.max(1, +flags.concurrency || DEFAULT_CONCURRENCY);
   }
 
-  run() {
+  run(callback) {
     this.logger.info("Lerna v" + this.lernaVersion);
 
     if (this.repository.isIndependent()) {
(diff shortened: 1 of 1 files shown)


gigabo:
Interesting... :thinking: 

@kwelch Could you provide some context here?  What does this patch provide?

kwelch:
I am using lerna to manage a multi-page web app so each package is actually a section of the web app.

Specifically this is for a script at the top level that calls the `RunCommand`. I am looking to run the `build` script in each package and after all packages are build consolidate the artifacts into a single directory. 

This may be anti-pattern as it seem there has not been much work on the node api side to allow for use within a script file.

gigabo:
I wouldn't call it an anti-pattern, but it's definitely not well supported right now.  I don't think it's part of the public API, either, so I worry a little bit about your relying on it. 😬 

Is there a way you could use the CLI instead in your workflow?

kwelch:
I am going to try a post script, it is less ideal but agreed it is safer.

kwelch:
Would a PR adding a public api be something that the lerna team would want?

Should be rather simple, just need a factory function around each of the classes that exposed commands. I would recommend also allowing for functionality from this PR of being able to add a callback to each command.

gigabo:
Yes we definitely want a [public API](https://github.com/lerna/lerna/issues/167)!

Not sure I understand exactly what you're suggesting with the factory functions.

evocateur:
A proper public API would go a long way toward disentangling the commands from their CLI particulars, for sure.

kwelch:
Closing as this has had no progress.

PR #440, opened 2016-12-14
Change testing NODE_ENV to "lerna-test" (Fixes #406)

### LINKED ISSUE #406: [Bug] Lerna swallows exit code when NODE_ENV is test
Due to this code: https://github.com/lerna/lerna/blob/v2.0.0-beta.30/src/Command.js#L156-L158

Lerna commands won't pass through their exit codes when `NODE_ENV` is `test`.

CircleCI automatically sets `NODE_ENV` to `test` for builds that use Node, so if you do something like `lerna run test`, your build won't fail when your tests fail.

ryb73:
fixes #406 

Previously, when running Lerna with `NODE_ENV` set to "test", it would output nothing to stdout and always exit with a code of `0`. Now the output and exit code will not be suppressed.

### DIFF
--- a/src/progressBar.js
+++ b/src/progressBar.js
@@ -18,7 +18,7 @@ class ProgressBarController {
     }
 
     // Don't do any of this while testing
-    if (process.env.NODE_ENV === "test") {
+    if (process.env.NODE_ENV === "lerna-test") {
       return;
     }
 
--- a/src/Command.js
+++ b/src/Command.js
@@ -68,7 +68,7 @@ export default class Command {
     }
 
     if (
-      process.env.NODE_ENV !== "test" &&
+      process.env.NODE_ENV !== "lerna-test" &&
       this.lernaVersion !== this.repository.lernaVersion
     ) {
       this.logger.warn(
@@ -164,7 +164,7 @@ export default class Command {
         callback(err, code);
       }
 
-      if (process.env.NODE_ENV !== "test") {
+      if (process.env.NODE_ENV !== "lerna-test") {
         process.exit(code);
       }
     };
--- a/src/logger.js
+++ b/src/logger.js
@@ -54,7 +54,7 @@ class Logger {
   }
 
   _emit(message) {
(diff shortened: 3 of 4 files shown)


src/progressBar.js
```diff
@@ -18,7 +18,7 @@ class ProgressBarController {
     }
 
     // Don't do any of this while testing
-    if (process.env.NODE_ENV === "test") {
+    if (process.env.NODE_ENV === "lerna-test") {
```
taion: the progress bar does look a bit wonky in CI logs. wonder if there's a better flag here... maybe `env.CONTINUOUS_INTEGRATION` or something

gigabo: Yeah, wonky whenever we're not running interactively.

Maybe a better check here would be whether we're connected to a tty?

```javascript
if (process.stdout.isTTY) {
    ...
}
```

taion: it's irrelevant to this PR anyway, just a quirk i noticed after i unset `NODE_ENV` on circle. probably best to add a `FIXME` and sort this out later, since it's not that important

gigabo: Filed https://github.com/lerna/lerna/issues/447.  Thanks for looking at this @taion.

petercasinelli:
Hey, thanks for catching this! Is there any way you can release this sooner rather than later? It's a blocker for people relying on tests to fail a build. Right now, a build can pass even if the tests fail because the exit codes are not propagated. I'm experiencing this right now with CircleCI.

I considered pulling in the master branch in, but in order to use it, you'll need to run `npm i` and install Lerna globally from the cloned repo. This is unrealistic for continuous integration because you'd need to clone for every build.

taion:
This ought not be a blocker with Circle. Just do

```yml
dependencies:
  override:
    # Circle sets NODE_ENV to test by default, which messes with both Lerna and
    # babel-plugin-dev-expression.
    - unset NODE_ENV
    - npm i
```

PR #441, opened 2016-12-14
Update README examples with current beta version

camacho:
small update to the examples to list the most recent beta version


hzoo:
dono if we want to just keep doing this every release (otherwise should leave it at a version)?

PR #443, opened 2016-12-14
Add back proxy method of bootstrap with BootstrapProxyCommand

hzoo:


gigabo:
I wonder if we can do this without creating a new command?

Like, with an option in `lerna.json`?

```js
{
  "lerna": "2.0.0-beta.31",
  "proxy": true,
  ...
}
```

That way the bulk of bootstrap logic could be shared instead of duplicated.

Bug fixes / enhancements wouldn't need to be done in two places.

What do you think?

evocateur:
I would think `bootstrapConfig` in `lerna.json` would be the natural place for such configuration, consulted in the existing BootstrapCommand when it decides how to "link" the sibling dependencies.

```json
{
  "lerna": "2.0.0-beta.31",
  "bootstrapConfig": {
    "proxy": true,
    "prefix": "/* @prefix etc */"
  }
}
```

I suppose you could make `bootstrapConfig.proxy` also accept `{ "prefix": "value" }` for slightly better config fluency (answering the question of "prefix for WHAT?" in the example above).

hzoo:
I only wanted to add it back since projects had issues with the non-proxy method. I wasn't planning on doing any fixes for it at the time. Don't know

hzoo:
Could make bootstrap take a flag as mentioned instead of a separate command if we still want this. closing

PR #452, opened 2016-12-17
Add `link` command. Fixes #341.

### LINKED ISSUE #341: [Feature] support method to /only/ link packages
Use case: I add a dependency to an existing package on an existing package. All I want to do is create that one symlink between the two, but `lerna bootstrap` has to go through the whole rigamarole of checking and installing everything, which can be slow (especially if the repo is in some kind of intermediate state due to adding/removing a bunch of stuff, which it often is when I want to do this operation).

Possible solutions:
1. `bootstrap` respects `--scope` so I can pick the packages I want and reduce the time waiting for installation
2. `bootstrap` gets `--skip-externals` so it only does the symlink/prepublish
3. `bootstrap` gets `--link-only` (but this seems sort of annoyingly narrow when the other two are somewhat more generally applicable)
4. Lerna gets a new `link` command that only symlinks

While (1) should probably exist for general consistency, I think the ideal solution to solve this particular problem is actually (4). Given that #358 adds _more_ stuff to bootstrap that I don't care about running when I am in the above situation, having a stripped-down command instead of a bunch of flags to turn off useful bootstrap behaviors seems preferable.

seansfkelley:
(This is branched from #412 to help avoid any potential merge conflicts it might have with it. Merge that first.)

This adds a `link` command per #341. Use case: I added a local package dependency, but I don't want to incur the full cost of `lerna bootstrap` (with attendant preinstall/install/postinstall/prepublish costs) cause I just need the symlink.

This will interact with #443 (though I don't think it will actually merge conflict) in that `link` should presumably respect the configured link type, assuming that method is eventually used in the PR, but it has no such configuration now.


gigabo:
This is interesting.

Going to hold off on looking closely at it until #412 is in, since the diff is inflated right now.

evocateur:
I am a big fan of the LinkUtilities abstraction.

I am a little concerned about the ambiguity of adding a subcommand that echoes `npm link`'s purpose, but actually does something very different. Would a flag like `npm install`'s `--ignore-scripts` (which omits running any applicable lifecycle scripts defined in package.json) be more appropriate here?

seansfkelley:
Good point re: the `link` name. I suppose that's why `bootstrap` is called that, rather than `install`. The goal of this command is _just_ to link, so any hypothetical flags to bootstrap would (for my use case, at least) also have to prevent installation. With 122 packages as of this writing, it takes while a while to npm to hit the filesystem and check that everything is in a good state. I figured that was a large enough behavioral difference to justify a new command, even if said command overlaps in behavior with `bootstrap`.

I'm open to suggestions; I just submitted this PR because this is the solution that would work for me and has a reasonable implementation right now.

test/LinkCommand.js
```diff
@@ -0,0 +1,51 @@
...
+  let packagesArg;
+  let packageGraphArg;
+
+  beforeEach((done) => {
+    let calledSymlinkPackages = false;
+    packagesArg = null;
+    packageGraphArg = null;
+    stub(LinkUtilities, "symlinkPackages", (packages, packageGraph, callback) => {
+      if (calledSymlinkPackages) {
+        throw new Error("already called symlinkPackages!");
```
seansfkelley: Note to self: this should call `callback`, not throw.

lukebatchelor:
Just wondering, is the advantage of this over `bootstrap --scope` that you don't attempt to install dependencies? 

I find that that doesnt really take too long, but maybe its a bigger issue for your repo.

seansfkelley:
@lukebatchelor that is precisely the issue, and yeah, I have a pretty large repo such that [npm slamming the file system becomes obnoxiously slow](https://github.com/palantir/yerna#motivation). I think bootstrap actually does a reasonable thing and would prefer that the link command didn't have to exist (indeed, while [I included a `link` command in Yerna](https://github.com/palantir/yerna#yerna-1), I almost never actually use it because install/bootstrap is fast and deterministic).

That said, I don't have the bandwidth to carry this through and since I don't use Lerna anymore, I don't want to add unnecessary complications/support burden, so I'm going to close this PR.

PR #454, opened 2016-12-18
Use close event to wait for spawned processes to finish

xaka:
As per https://nodejs.org/api/child_process.html#child_process_event_close, the close event is emitted when the stdio streams of a child process have been closed, which is important as there is a chance for race condition when process has exited (process's exit event), but collecting  stderr hasn't been finished yet (stderr stream's done event). Using close event seems to be pretty common if you look at implementations of child_process.exec, grunt.utils.spawn, etc.


gigabo:
Oh, interesting.  That's a subtle distinction.  Thanks for the improvement @xaka!

PR #458, opened 2016-12-20
use message passed as argument to the static method input() in PromptUtilities

btiwaree:
This commit uses message which is passed to the static method input() as an argument. Previously, string 'Enter a custom version' was being used instead of message passed from its caller.

See [src/commands/PublishCommand.js:222](https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L222) for usage of input()


gigabo:
Nice catch @btiwaree!

PR #459, opened 2016-12-21
Fix import logger info typo

sdgluck:


gigabo:
Heh, thanks @sdgluck. :)

PR #460, opened 2016-12-21
Add --message option for custom commit msgs when publishing

### LINKED ISSUE #395: [Feature] Add custom commit message option to publishConfig
Currently it looks like the commit messages for publishing are [hard](https://github.com/lerna/lerna/blob/cc9b48725f89ad94a2fa2b5cd7f984da199cd5a1/src/commands/PublishCommand.js#L311)-[coded](https://github.com/lerna/lerna/blob/cc9b48725f89ad94a2fa2b5cd7f984da199cd5a1/src/commands/PublishCommand.js#L325). This presents problems for us since we enforce a pre-defined commit message style in a git hook via [validate-commit-message](https://www.npmjs.com/package/validate-commit-message). 

Currently, we plan on getting around this by patching lerna's commit messages within the `prepare-commit-message` hook. However, it would be nice if lerna exposed some functionality to customize the `publish` commit messages so that repo maintainers with commit guidelines wouldn't have use workarounds like those described above. I'm happy to help contribute this feature if the team is open to adding it.

traviskaufman:
This commit introduces a new option for publish: --message. When given
this option, lerna will use the supplied message as the commit message
when it commits updates while publishing. If no message flag is
specified, the default hard-coded messages are used.

Resolves #395

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -310,13 +310,8 @@ export default class PublishCommand extends Command {
   }
 
   gitCommitAndTagVersionForUpdates() {
-    let message = "Publish" + EOL;
-
-    const tags = this.updates.map((update) => {
-      const tag = `${update.package.name}@${this.updatesVersions[update.package.name]}`;
-      message += `${EOL} - ${tag}`;
-      return tag;
-    });
+    const tags = this.updates.map((update) => `${update.package.name}@${this.updatesVersions[update.package.name]}`);
+    const message = this.flags.message || tags.reduce((msg, tag) => msg + `${EOL} - ${tag}`, `Publish${EOL}`);
 
     GitUtilities.commit(message);
     tags.forEach(GitUtilities.addTag);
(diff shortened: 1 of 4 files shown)


src/commands/PublishCommand.js
```diff
@@ -310,13 +310,8 @@ export default class PublishCommand extends Command {
   }
 
   gitCommitAndTagVersionForUpdates() {
-    let message = "Publish" + EOL;
```
emmenko: Shouldn't we only change the `message` part? The `tags` part should stay the same.

```bash
# commit message now
Publish

- package-1@1.0.1
- package-2@1.1.1

# commit message with custom message
chore: release :rocket:

- package-1@1.0.1
- package-2@1.1.1
```

doug-wade:
This looks ready to merge to me 👍

traviskaufman:
As @doug-wade mention this patch should only change the commit message, leaving the tags alone. I believe this is verified since the tests I wrote for this use the same tag names as other tests and pass. If there's anything else that needs to be done let me know. Otherwise, it would be great if we could merge this soon! Our codebase is in need of it :)

traviskaufman:
🔔 😄

traviskaufman:
🔔 😄

traviskaufman:
Merge conflicts fixed. Thanks for the review!

hzoo:
Good stuff @traviskaufman!

PR #475, opened 2017-01-03
Lerna checks changes between last tag in current branch

Gongreg:
Lerna by default now check changes between last tag in current branch (instead of current latest tag in whole repository), it allows to publish older versions of project. (Issue 89).

Checks that user is in a branch and not a detached commit, because lerna can't publish without git branch.

(Issue #89). Reopened pull request #415, because I couldn't work on it anymore.

### DIFF
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -63,12 +63,7 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static getLastTag() {
-    try {
-      return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
-    } catch (e) {
-      return "0.0.0";
-    }
-
+    return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
   }
 
   @logger.logifySync()
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -46,7 +46,8 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static getLastTaggedCommitInBranch() {
-    return ChildProcessUtilities.execSync("git rev-list -n 1 $(git describe --abbrev=0 --tags)");
+    const tagName = GitUtilities.getLastTag();
+    return ChildProcessUtilities.execSync("git rev-list -n 1 " + tagName);
   }
 
   @logger.logifySync()
@@ -62,7 +63,12 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static getLastTag() {
-    return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
+    try {
+      return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
+    } catch(e) {
+      return '0.0.0';
+    }
+
   }
 
   @logger.logifySync()
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -45,6 +45,11 @@ export default class GitUtilities {
   }
 
   @logger.logifySync()
+  static getLastTaggedCommitInBranch() {
+    return ChildProcessUtilities.execSync("git rev-list -n 1 $(git describe --abbrev=0 --tags)");
+  }
+
+  @logger.logifySync()
   static getFirstCommit() {
     return ChildProcessUtilities.execSync("git rev-list --max-parents=0 HEAD");
   }
@@ -53,6 +58,11 @@ export default class GitUtilities {
   static pushWithTags(remote, tags) {
     ChildProcessUtilities.execSync(`git push ${remote} ${GitUtilities.getCurrentBranch()}`);
     ChildProcessUtilities.execSync(`git push ${remote} ${tags.join(" ")}`);
+  }
+
+  @logger.logifySync()
+  static getLastTag() {
+    return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
   }
 
   @logger.logifySync()
@@ -83,6 +93,11 @@ export default class GitUtilities {
   @logger.logifySync()
   static getCurrentBranch() {
     return ChildProcessUtilities.execSync("git symbolic-ref --short HEAD");
+  }
+
+  @logger.logifySync()
+  static getCurrentBranchDescription() {
+    return ChildProcessUtilities.execSync("git branch | sed -n '/\* /s///p'");
   }
 
   @logger.logifySync()
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -65,8 +65,8 @@ export default class GitUtilities {
   static getLastTag() {
     try {
       return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
-    } catch(e) {
-      return '0.0.0';
+    } catch (e) {
+      return "0.0.0";
     }
 
   }
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -98,7 +98,7 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static getCurrentBranchDescription() {
-    return ChildProcessUtilities.execSync("git branch | sed -n '/\* /s///p'");
+    return ChildProcessUtilities.execSync("git symbolic-ref --short -q HEAD");
   }
 
   @logger.logifySync()
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -27,6 +27,15 @@ export default class UpdatedPackagesCollector {
 
   collectUpdatedPackages() {
     logger.info("Checking for updated packages...");
+
+    const currentBranch = GitUtilities.getCurrentBranchDescription();
+    if (currentBranch !== "master") {
+      logger.info("Working not on master! Current branch: " + currentBranch);
+    }
+
+    const tag = GitUtilities.getLastTag();
+    logger.info("Comparing with: " + tag);
+
     progressBar.init(this.packages.length);
 
     const hasTags = GitUtilities.hasTags();
@@ -43,7 +52,7 @@ export default class UpdatedPackagesCollector {
 
       commits = this.getAssociatedCommits(currentSHA);
     } else if (hasTags) {
-      commits = GitUtilities.describeTag(GitUtilities.getLastTaggedCommit());
+      commits = GitUtilities.describeTag(GitUtilities.getLastTaggedCommitInBranch());
     }
 
     const updatedPackages = {};
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -30,10 +30,6 @@ export default class UpdatedPackagesCollector {
 
     const currentBranch = GitUtilities.getCurrentBranchDescription();
 
-    if (currentBranch !== "master") {
-      logger.info("Working not on master! Current branch: " + currentBranch);
-    }
-
     const hasTags = GitUtilities.hasTags();
 
     if (hasTags) {
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -34,12 +34,17 @@ export default class UpdatedPackagesCollector {
       logger.info("Working not on master! Current branch: " + currentBranch);
     }
 
-    const tag = GitUtilities.getLastTag();
-    logger.info("Comparing with: " + tag);
+    const hasTags = GitUtilities.hasTags();
+
+    if (hasTags) {
+      const tag = GitUtilities.getLastTag();
+      logger.info("Comparing with: " + tag);
+    } else {
+      logger.info("No tags found! Comparing with initial commit.");
+    }
 
     progressBar.init(this.packages.length);
 
-    const hasTags = GitUtilities.hasTags();
     let commits;
 
     if (this.flags.canary) {
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -28,8 +28,6 @@ export default class UpdatedPackagesCollector {
   collectUpdatedPackages() {
     logger.info("Checking for updated packages...");
 
-    const currentBranch = GitUtilities.getCurrentBranchDescription();
-
     const hasTags = GitUtilities.hasTags();
 
     if (hasTags) {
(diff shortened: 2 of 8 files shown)


Gongreg:
I've fixed existing tests, but it definitely needs tests for this new behaviour.

doug-wade:
The tests failing on appveyor seem to be due to the `$(git ...)` command substitution at https://github.com/lerna/lerna/pull/475/files#diff-154eeef66f5acd021d530f6f5494e476R49.  [Afaict](http://stackoverflow.com/questions/434038/whats-the-cmd-powershell-equivalent-of-back-tick-on-bash), $() command substitution is supported in PowerShell, so I'm a little surprised.  Maybe you need `"`'s around it or something?  Certainly splitting the two commands into two separate `execSync`s would also solve it.

spudly:
Did some googling. It looks like in powershell you only use the '$' if it's inside of a string, so the following are both correct in windows:

* `git rev-list -n 1 (git describe --abbrev=0 --tags)`
* `git rev-list -n 1 "$(git describe --abbrev=0 --tags)"`

I think it makes sense to use the second one (just wrap it in quotes) as that should work in both bash and powershell. In fact, that's what is already being done in the `commit` function.

@Gongreg, let me know if you need any help. I'd really like to see this merged asap.

Gongreg:
This needs a test case, not sure where to place it:

* Git init with two packages
* Lerna publish v1.0.0
* Edit both packages
* Lerna publish v1.1.0
* Git checkout v1.0.0
* Git checkout -b v1-bugfix
* Edit one of packages

Expected result:
Lerna updated shows only one package.
Lerna publish suggests versions:
v1.0.1
v1.1.0
v2.0.0

Gongreg:
0.12v build failed, don't know what to do about it.

spudly:
Looks to me like someone on the lerna team needs to login to appveyor and restart the build. It's probably an intermittent npm issue that causes the build to fail sometimes.

spudly:
(or you could just add another commit, which would have the same effect)

Gongreg:
Didn't help.

doug-wade:
I, too, have been having trouble with the build failing on deprecated nodes.  I opened #484, which drops deprecated nodes, and have been building prs on that commit to side-step the issue.  It will mean that your pr merging will block on the deprecation, but it might be easier than playing dice with the build.

spudly:
Looks like this issue:
http://help.appveyor.com/discussions/problems/3147-error-eperm-operation-not-permitted-rename-on-npm-install
https://github.com/npm/npm/issues/9696

... which is fixed in the version of npm bundled with node 4

You may be able to fix it by adding `--force` to the `npm install` command in the appveyor.yml file. Others also had success with `appveyor-retry npm install`

spudly:
Anything I can do to help with this? I'm happy to contribute code, debug the failing build, etc. if you grant me commit access on the fork to add commits to the PR.

Gongreg:
@spudly I've invited you to repo. You are definitely welcome to help.

spudly:
Well, the build is passing with no changes.

In case anyone else is struggling with this issue, none of these solutions I found worked:
* change `npm install` to `appveyor-retry npm install`
* change `npm install` to `npm install --force`
* upgrade npm to v3
* `npm cache clean`

What did work? Just running the build over and over. It's a stubborn, intermittent issue, but I finally got a build where npm install didn't fail so the tests could run. All tests are passing on all supported node versions.

spudly:
@gigabo, @hzoo, @kittens, & @thejameskyle,

I think this one is ready to be merged.

Gongreg:
Congratulations on fixing it :D

Gongreg:
Is this going to be merged in sometime?

src/commands/PublishCommand.js
```diff
@@ -17,6 +17,11 @@ export default class PublishCommand extends Command {
       this.logger.info("Publishing canary build");
     }
 
+    const currentBranch = GitUtilities.getCurrentBranchDescription();
+    if (currentBranch && currentBranch.includes("detached at")) {
+      callback("You are working on detached branch, create new branch to publish changes.");
```
gigabo: Oh, good thinking handling this case.  Hadn't even considered it.

test/UpdatedCommand.js
```diff
@@ -64,7 +64,7 @@ describe("UpdatedCommand", () => {
       let calls = 0;
       stub(logger, "info", (message) => {
         if (calls === 0) assert.equal(message, "Checking for updated packages...");
-        if (calls === 2) assert.equal(message, `- package-1\n- package-2\n- package-3\n- package-4\n- package-5 (${chalk.red("private")}`);
+        if (calls === 2) assert.equal(message, `- package-1\n- package-2\n- package-3\n- package-4\n- package-5 (${chalk.red("private")})`);
```
gigabo: Huh?  Was this test just wrong?

src/GitUtilities.js
```diff
@@ -56,6 +62,16 @@ export default class GitUtilities {
   }
 
   @logger.logifySync()
+  static getLastTag() {
+    try {
+      return ChildProcessUtilities.execSync("git describe --tags --abbrev=0");
+    } catch (e) {
+      return "0.0.0";
```
gigabo: Tags start with `v`, right?  Should this be `v0.0.0`?

gigabo: Actually, come to think of it... should this use the version from `lerna.json`?

test/UpdatedCommand.js
Gongreg: It failed in my machine, iirc it failed in build too. Looking into test case it doesn't even make sense that it passed before.

src/GitUtilities.js
Gongreg: Removing out this try catch, added check case in updated packages collector. No need to handle it here then. It didn't even make sense to try catch it.

Gongreg:
Fixed your comments.

src/GitUtilities.js
```diff
@@ -86,6 +97,11 @@ export default class GitUtilities {
   }
 
   @logger.logifySync()
+  static getCurrentBranchDescription() {
+    return ChildProcessUtilities.execSync("git branch | sed -n '/\* /s///p'");
```
evocateur: This should be `git symbolic-ref --short -q HEAD`, not `sed` shenanigans. Thus, when an empty string, it represents a detached head state.

src/commands/PublishCommand.js
evocateur: We really shouldn't be [depending on the output](https://git-blame.blogspot.com/2013/06/checking-current-branch-programatically.html) of a porcelain command.

```bash
# correct example from the linked blog post
if branch=$(git symbolic-ref --short -q HEAD)
then
  echo on branch $branch
else
  echo not on any branch
fi
```

Determining if the branch is detached then becomes `if (currentBranch === '') {`

src/UpdatedPackagesCollector.js
```diff
@@ -28,9 +28,24 @@ export default class UpdatedPackagesCollector {
 
   collectUpdatedPackages() {
     logger.info("Checking for updated packages...");
-    progressBar.init(this.packages.length);
+
+    const currentBranch = GitUtilities.getCurrentBranchDescription();
+
+    if (currentBranch !== "master") {
```
evocateur: I don't think we should be prejudicing `master` here. It's also not Lerna's job to inform the user which branch they are on.

Gongreg:
I tried to use your command but something wasn't working with it then. I don't have a possibility to check it out more now.

spudly:
@Gongreg,

Since I've got commit access on your fork, I'm going to go ahead and make the changes requested so this pull request can move along. Let me know if that's a problem.

spudly:
Rebased onto master and fixed some tests that failed as a result.

One test is still failing - looks unrelated. Will look at it later.

spudly:
Seems like there's a timing and/or callback issue related to the hoisting tests that were recently added.

First it failed with this message:

> 1) BootstrapCommand lifecycle scripts should run preinstall, postinstall and prepublish scripts:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
      at Timeout.<anonymous> (C:\projects\lerna\node_modules\mocha\lib\runnable.js:226:19)

Then I added an empty commit to trigger a rebuild just in case appVeyor was just running slow, but the second time (with no changes) it failed 2 tests:

>  1) BootstrapCommand lifecycle scripts should run preinstall, postinstall and prepublish scripts:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
      at Timeout.<anonymous> (C:\projects\lerna\node_modules\mocha\lib\runnable.js:226:19)
>  2) BootstrapCommand with hoisting should hoist:
     Error: done() called multiple times
      at Test.Runnable (C:\projects\lerna\node_modules\mocha\lib\runnable.js:55:17)
      at new Test (C:\projects\lerna\node_modules\mocha\lib\test.js:22:12)
      at context.it.context.specify (C:\projects\lerna\node_modules\mocha\lib\interfaces\bdd.js:85:18)
      at Suite.<anonymous> (C:/projects/lerna/test/BootstrapCommand.js:60:5)
      at context.describe.context.context (C:\projects\lerna\node_modules\mocha\lib\interfaces\bdd.js:47:10)
      at Suite.<anonymous> (C:/projects/lerna/test/BootstrapCommand.js:53:3)
      at context.describe.context.context (C:\projects\lerna\node_modules\mocha\lib\interfaces\bdd.js:47:10)
      at Object.<anonymous> (C:/projects/lerna/test/BootstrapCommand.js:22:1)
      at loader (C:\projects\lerna\node_modules\babel-register\lib\node.js:144:5)
      at Object.require.extensions.(anonymous function) [as .js] (C:\projects\lerna\node_modules\babel-register\lib\node.js:154:7)
      at require (internal/module.js:20:19)
      at C:\projects\lerna\node_modules\mocha\lib\mocha.js:220:27
      at Array.forEach (native)
      at Mocha.loadFiles (C:\projects\lerna\node_modules\mocha\lib\mocha.js:217:14)
      at Mocha.run (C:\projects\lerna\node_modules\mocha\lib\mocha.js:469:10)
      at Object.<anonymous> (C:\projects\lerna\node_modules\mocha\bin\_mocha:404:18)
      at run (bootstrap_node.js:394:7)
      at startup (bootstrap_node.js:149:9)
      at bootstrap_node.js:509:3

It's only failing in appVeyor/node6.

I'm pretty sure these failing tests are unrelated to the code in this PR.

@gigabo and @seansfkelley, it looks like you two wrote the tests that are failing. Any ideas?

seansfkelley:
Never seen that before. Perhaps something changed such that the synchronous setup code before `runCommand` is throwing?

spudly:
Ok, something is definitely wrong with the build process here. Both travis and appveyor are failing tests that don't even exist in the code!

This test straight-up doesn't exist, but is somehow failing anyway:
```
PublishCommand normal mode with --registry should publish the changed packages
```

It exists on lerna master but does not exist on the fork because it's behind master.

Guessing there must be some cache issue? Or maybe the build servers are checking out the wrong revision? Any ideas here?

I guess I'll rebase on master again...

spudly:
Success! 🎉 🎉 🎉 

Finally got the tests to pass.

Y'all have got a seriously brittle test suite. It fails intermittently on appveyor and when you're behind master, it seems to run the wrong test suite.

Either way, the build is passing. Any chance this could get merged?

spudly:
@evocateur @gigabo

gaearon:
This is super handy, can't wait to see it released!

evocateur:
@gaearon Hopefully soon!

Guria:
Seems I have an issue with it: https://travis-ci.org/cerebral/cerebral#L993-L994
My canary publishing triggers when tag `next` is updated. And now I updated lerna it stopped to work.

Guria:
Switched to publish from push to master and still have the same issue here: https://travis-ci.org/cerebral/cerebral/builds/210713536#L1979-L2002

Deploy from cloud9 from the same branch is just fine. Could anyone help me with it?

evocateur:
I was going to recommend avoiding a `git tag -f next` before it was changed, but now I'm not sure why `git symbolic-ref --short -q HEAD` works every time lerna is called except the last time...

How are you installing `lerna` globally? Or is this shell script called from an npm script (the build log is unclear)?

Guria:
Lerna is declared in package.json and used to be installed with `npm i`

evocateur:
It's unclear where [this line](https://github.com/cerebral/cerebral/blob/master/.travis.yml#L30) and others like it have found `lerna` in the Travis shell. I would expect it to find the local `./node_modules/.bin/lerna` from `npm i`, but not in a general shell without an explicit `npm i -g lerna`.

Guria:
Shell script located right in .travis.yml: https://github.com/cerebral/cerebral/blob/6f1668acd432f9759d81f06ed8e70a65af7bdcfd/.travis.yml

I suppose Travis adds `./node_modules/.bin` to the `$PATH`. I am doing nothing special to make it available.

evocateur:
So as near as I can figure, Travis is doing _something_ to a file under `.git/` before the `after_success` phase that is causing an error to occur when executing this command.

https://github.com/travis-ci/travis-ci/issues/4745 seems to indicate as much. The [error](https://travis-ci.org/liddellj/node-aws-travis-playground/builds/78551189#L439) would be `fatal: ref HEAD is not a symbolic ref`, if we indeed logged it correctly...

Guria:
Thanks a bunch. So do you have any suggestions how to deal with it now?

evocateur:
I did a lot of digging, and it looks like we can use a different method to avoid the error (`git rev-parse --abbrev-ref HEAD`) and be a little more forgiving about operations on detached HEAD when doing a canary release (which doesn't care about committing changes anyway).

I should have something later today.

Guria:
So it happens right after clone in fact: https://travis-ci.org/cerebral/cerebral/builds/210950579#L120

Does it mean that lerna 2.0.0-beta.38 can't be used with travis at all?

Guria:
Ohh. missed a suggestion. Thanks, will wait for a new release.

evocateur:
If we intend to support automated publishing in CI (where most, if not all, services will do a detached checkout), we'll have to figure out how to recover (in most cases, checking out the branch indicated by a variety of env vars depending on the service...)

Guria:
Any updates here? Should we have some new issue to track a fix?

evocateur:
@Guria Sorry, I got distracted. A separate issue would be great. `GitUtilities.getCurrentBranchDescription()` should be using `git rev-parse --abbrev-ref HEAD`, and should not be blocking a `--canary` publish.

Guria:
See #699

Issue #478, opened 2017-01-04
Use `path.join`, not string concatenation


garthk:
Never do [this](https://github.com/lerna/lerna/blob/3a5cb2419fdaf8469c85aa2d0e5e4b73fe41f07d/src/commands/ImportCommand.js#L37):

```js
    this.targetDir = "packages/" + externalRepoBase;
```

doug-wade:
Fixed in https://github.com/lerna/lerna/pull/480; closing

PR #480, opened 2017-01-04
Use package paths to determine import target directory

### LINKED ISSUE #477: Feature: use first packages entry for lerna import, not "packages/
Good beginner issue: if `lerna.json` contains `"packages": ["pkg/*"]`, `lerna import /path/to/pkgname` should offer to import to `pkg/pkgname`, not `packages/pkgname`.

In more detail:

* Look at the first entry in `packages` (default `[]`) for which `path.basename` returns `"*"`
* If there is no such entry, use `packages` as the prefix to `targetDir`
* Otherwise, use `path.dirname` of the entry as the prefix

While you're in there, fix #478 if it hasn't been fixed already.

garthk:
Fixes #477 and #478. Let me know how to approach testing; I'd prefer not to go backwards on coverage.

### DIFF
--- a/test/ImportCommand.js
+++ b/test/ImportCommand.js
@@ -2,8 +2,8 @@ import pathExists from "path-exists";
 import assert from "assert";
 import path from "path";
 import fs from "fs";
-import normalize from "normalize-path";
 import escapeArgs from "command-join";
+import mkdirp from "mkdirp";
 
 import PromptUtilities from "../src/PromptUtilities";
 import ChildProcessUtilities from "../src/ChildProcessUtilities";
@@ -171,7 +171,33 @@ describe("ImportCommand", () => {
       importCommand.runPreparations();
 
       importCommand.runCommand(exitWithCode(1, (err) => {
-        const expect = `Target directory already exists "${normalize(targetDir)}"`;
+        const expect = `Target directory already exists "${targetDir}"`;
+        try {
+          assert.equal((err || {}).message, expect);
+          done();
+        } catch (err) {
+          done(err);
+        }
+      }));
+    });
+
+    it("should infer correct target directory given packages glob", (done) => {
+      const targetDir = path.relative(
+        process.cwd(),
+        path.join(testDir, "pkg", path.basename(externalDir))
+      );
+      mkdirp.sync(targetDir);
+
+      changeLernaConfig(testDir, (lernaJson) => {
+        lernaJson.packages = [ "pkg/*" ];
+      });
+
+      const importCommand = new ImportCommand([externalDir], {});
+      importCommand.runValidations();
+      importCommand.runPreparations();
+
+      importCommand.runCommand(exitWithCode(1, (err) => {
+        const expect = `Target directory already exists "${targetDir}"`;
         try {
           assert.equal((err || {}).message, expect);
           done();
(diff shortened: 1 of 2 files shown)


doug-wade:
The failing test is due to a difference between how [normalize-path](https://github.com/jonschlinkert/normalize-path) and `path.join` normalize paths on Windows (normalize-path, surprisingly, forces every path to use unix-like slashes, whereas path uses forward slashes on Windows).  Changing the assertion at https://github.com/lerna/lerna/blob/master/test/ImportCommand.js#L174 to use path.join instead should resolve your issue.

doug-wade:
As to adding tests, I think you'll want a new test case [here](https://github.com/lerna/lerna/blob/master/test/ImportCommand.js) that tests a new [import fixture](https://github.com/lerna/lerna/tree/master/test/fixtures/ImportCommand) that has packagePaths in their lerna.json

src/Repository.js
```diff
@@ -16,7 +16,6 @@ export default class Repository {
     this.rootPath = path.resolve(GitUtilities.getTopLevelDirectory());
     this.lernaJsonLocation = path.join(this.rootPath, "lerna.json");
     this.packageJsonLocation = path.join(this.rootPath, "package.json");
-    this.packagesLocation = path.join(this.rootPath, "packages"); // TODO: Kill this.
```
evocateur: I agree we need to remove this unused property, but I'm not sure this PR is the place for that.

garthk: Point.

garthk:
Rebased on current `master`, discarding the `packagesLocation` change per @evocateur feedback. Waiting for CI to tell me if there's still a problem on Windows.

garthk:
I'm not sure this AppVeyor build failure was caused by my changes:
```
 1)  "after all" hook:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test
```

In other news, @evocateur will need to eyeball that I've accommodated his requested change. GitHub doesn't seem to have figured it out.

test/ImportCommand.js
```diff
@@ -171,7 +171,33 @@ describe("ImportCommand", () => {
       importCommand.runPreparations();
 
       importCommand.runCommand(exitWithCode(1, (err) => {
-        const expect = `Target directory already exists "${normalize(targetDir)}"`;
```
doug-wade: ?

garthk: Which would you prefer?

* Output native paths in the error message
* Coerce to Unix-style even on Windows?

In `master` now, we hard-code UNIX style slashes in `ImportCommand.targetDir`:

```
    this.targetDir = "packages/" + externalRepoBase;
```

… so, in any test checking `targetDir` either explicitly or implicitly via an `Error`, we need to coerce the test's expectation to UNIX style paths.

In my change, I use `path.join` for `targetDir`, so we need to stick with whatever `path.relative` or `path.join` gave us.

garthk:
Righto, whatever was stuffing AppVeyor cleared. Remaining review work:

- [ ] @evocateur to reassure GitHub I restored the line he wanted restored
- [ ] @doug-wade to escalate his raised eyebrow or let us know he's OK given the explanation

src/Repository.js
evocateur: Thanks!

doug-wade:
lgtm :+1:

Issue #38, opened 2016-02-05
[Feature] lerna owners


jamiebuilds:
It's rather tedious to have to add npm owners to every project. It would be nice to have commands that could do the following operations in every package.

```
$ lerna owners add username
$ lerna owners rm username
```

Which would do the following actions.

```
$ npm add owner username package
$ npm rm owner username package
```

It'd be even nicer if it tracked the owners so that when you added a new package it would add them all.

Would work well with #42

hzoo:
a great idea. do we want to do the whole "OWNERS" file to track. How to we how to configure/store data (this would go with the changelog - github project name, token, etc, issue labels, etc)

Marak:
Is possible to include `npm` as dependency and run commands.

You can also spawn `npm` as child process.

thejameskyle:
I was wondering if we wanted to create a `lerna.json` file that could track information like "version", "owners", "independent" and anything else that might come up

hzoo:
yay more configs `.lernarc`

thejameskyle:
Could just be `package.json` too

thejameskyle:
Could just be `package.json` too. Better than having eight files to track different things.

thejameskyle:
If anyone wants to use scripts for this for now, you can use these: https://gist.github.com/thejameskyle/eba3a918a3db80ddc656

Just dump those in a `scripts/` directory and `chmod +x` them all. Then create an `NPM_OWNERS` file and add the current owners' npm usernames on individual lines and run `./scripts/npm-owner-update.sh`. The other two scripts are there if you ever want to grant or revoke ownership to users.

hzoo:
We can add this to `lerna.json` now

thejameskyle:
Yeah, I was working on a PR this weekend for it

MoOx:
Any plan to grad something like https://github.com/babel/babel/blob/master/scripts/grant-npm-owner.sh into the core?

hzoo:
Yeah, just not implemented https://github.com/lerna/lerna/pull/294, and if you want to make a pr feel free

Sawtaytoes:
Is the config file name `lerna.json` and optionally `.lernarc` or just `lerna.json`?

evocateur:
Currently just `lerna.json`. Probably will expand in the 3.0 alpha when `cosmiconfig` is used, dunno yet.

PR #481, opened 2017-01-05
Add owners command

doug-wade:
Addresses #38


bin/lerna.js
```diff
@@ -20,6 +20,7 @@ var cli = meow([
   "  run        Run npm script in each package",
   "  exec       Run a command in each package",
   "  ls         List all public packages",
+  "  owners     Add and remove package owners",
```
gigabo: Might be nice to make it `owner` to be consistent with the `npm owner` command.

src/ConfigUtilities.js
```diff
@@ -0,0 +1,18 @@
...
+  static read(rootPath) {
+    config = config || require(path.join(rootPath, "lerna.json"));
+    return config;
+  }
+
+  @logger.logifySync()
+  static write(rootPath, conf) {
+    config = conf;
+    FileSystemUtilities.writeFileSync(path.join(rootPath, "lerna.json"), JSON.stringify(config));
+  }
```
gigabo: Nice that these are encapsulated so they can be stubbed for tests.

src/commands/OwnersCommand.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  runCommandInPackage(pkg) {
+    const pkgName = pkg._package.name;
+    if (this.command === "add") {
+      NpmUtilities.addOwner(this.ownerName, pkgName);
+      this.logger.info(`Added owner ${this.ownerName} to package ${pkgName}`);
+    } else if (this.command === "rm") {
+      NpmUtilities.removeOwner(this.ownerName, pkgName);
+      this.logger.info(`Removed owner ${this.ownerName} to package ${pkgName}`);
```
gigabo: ... _from_ package?

src/commands/OwnersCommand.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+    config.owners = [...ownersSet];
+    ConfigUtilities.write(this.repository.rootPath, config);
+    async.parallelLimit(this.filteredPackages.map((pkg) => (cb) => {
+      this.runCommandInPackage(pkg, cb);
+    }), this.concurrency, callback);
+  }
+
+  runCommandInPackage(pkg) {
+    const pkgName = pkg._package.name;
```
gigabo: There's actually [a `name` getter](https://github.com/lerna/lerna/blob/master/src/Package.js#L13-L15), so you should be able to just do `pkg.name` here.

doug-wade:
Thanks for the quick response @gigabo.  I updated the docs and added tests, and did a little general cleanup and commenting.

doug-wade:
rebased onto the latest master, and got the tests running -- wanna take a look @gigabo?

CONTRIBUTING.md
```diff
@@ -19,6 +19,14 @@ Or the linter:
 $ npm run lint
 ```
 
+If you want to run a single unit test, you can globally install the test runner
+and point it at the file you're interested in
+
+```sh
+$ npm i -g mocha
+$ mocha -t 5000 path/to/your/test.js
```
gigabo: Oh, I've just been doing `$(npm bin)/mocha path/to/test.js`.  Guess that's the same.

Nice addition to the docs!

src/index.js
```diff
@@ -2,24 +2,26 @@ import BootstrapCommand from "./commands/BootstrapCommand";
...
   import: ImportCommand,
+  owner: OwnerCommand,
   clean: CleanCommand,
   diff: DiffCommand,
   init: InitCommand,
-  run: RunCommand,
   exec: ExecCommand,
-  ls: LsCommand
+  run: RunCommand,
+  ls: LsCommand,
```
gigabo: Heads up: This is going to conflict with the [automatic hoisting PR](https://github.com/lerna/lerna/pull/507/files#diff-1fdf421c05c1140f6d71444ea2b27638).

Might help the merge to trim it down to just the add of `owner` for now and handle the reorder separately.

src/commands/OwnerCommand.js
```diff
@@ -0,0 +1,78 @@
...
+    // won't be in the configured registry.
+    async.parallelLimit(this.filteredPackages.map((pkg) => (cb) => {
+      this.runCommandInPackage(pkg, cb);
+    }), this.concurrency, callback);
+  }
+
+  runCommandInPackage(pkg, callback) {
+    switch (this.command) {
+    case ADD_COMMAND_NAME:
+      NpmUtilities.addOwner(this.ownerName, pkg.name);
```
gigabo: No trouble with this if called with a package/owner pair that's already set up?  Same with remove, below?

CONTRIBUTING.md
xaka: It's kind of dangerous to install mocha in such way as you might end up with version different from what `lerna` uses? It also leads to pollution of global namespace. As `mocha` is already part of `devDependencies` and aliased via `test` script, you can do the following:

```bash
npm run test -- <path> # or any other cmd arg mocha understands
```

src/commands/OwnerCommand.js
```diff
@@ -0,0 +1,78 @@
...
+
+    switch (this.command) {
+    case ADD_COMMAND_NAME:
+      ownersSet.add(this.ownerName);
+      break;
+    case RM_COMMAND_NAME:
+      if (ownersSet.has(this.ownerName)) {
+        ownersSet.delete(this.ownerName);
+      } else {
+        return new callback(new Error(this.ownerName + OWNER_NOT_FOUND_ERROR_FRAGMENT));
```
gigabo: Should this warn and proceed?  Could be trying to remove an owner that was added before owners were stored in lerna config.

CONTRIBUTING.md
gigabo: Ooh, nice tip @xaka!

src/commands/OwnerCommand.js
doug-wade: not that I can tell -- when I add myself as an owner repeatedly or remove an owner from a package I own repeatedly, npm doesn't throw or log an error or anything.

src/commands/OwnerCommand.js
doug-wade: oooh good call 👍

gigabo:
Are the failing tests here just waiting for a rebase onto #557?

doug-wade:
@gigabo yes

Issue #55, opened 2016-02-13
[Feature] Check npm login and registry before publishing


jamiebuilds:
If you aren't logged in to npm, lerna should stop and ask you to login. It would be good if it logged you back out afterwards.

kittens:
If you're doing this for security then it's pointless.

thejameskyle:
I have to switch between npm profiles for different (public and private) registries

thejameskyle:
I'm going to just do a check at the beginning to make sure that you're logged in to the right registry.

```
$ npm whoami
$ npm config get registry
```

Logging in is overkill, but I think the more checks we can do at the beginning is good.

PR #482, opened 2017-01-05
Fix 55

doug-wade:
Addresses #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,7 +424,8 @@ Running `lerna` without arguments will show all commands/options.
   "bootstrapConfig": {
     "ignore": "component-*"
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "registry": "http://localhost:4873/"
 }
 ```
 
@@ -434,6 +435,7 @@ Running `lerna` without arguments will show all commands/options.
 - `bootstrapConfig.ignore`: an glob that won't be bootstrapped when running the `lerna bootstrap` command.
 - `bootstrapConfig.scope`: an glob that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
+- `registry`: The url of the registry to publish non-private packages to, and to read package metadata from.
 
 ### Common `devDependencies`
 
--- a/src/Command.js
+++ b/src/Command.js
@@ -1,5 +1,7 @@
 import ChildProcessUtilities from "./ChildProcessUtilities";
 import FileSystemUtilities from "./FileSystemUtilities";
+import ConfigUtilities from "./ConfigUtilities";
+import NpmUtilities from "./NpmUtilities";
 import ExitHandler from "./ExitHandler";
 import progressBar from "./progressBar";
 import Repository from "./Repository";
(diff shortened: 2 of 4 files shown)


doug-wade:
The build only failed on windows for deprecated nodes; if we merge #484 it should fix the appveyor build.

README.md
```diff
@@ -434,6 +435,7 @@ Running `lerna` without arguments will show all commands/options.
 - `bootstrapConfig.ignore`: an glob that won't be bootstrapped when running the `lerna bootstrap` command.
 - `bootstrapConfig.scope`: an glob that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
+- `registry`: The url of the registry to publish non-private packages to, and to read package metadata from.
```
evocateur: Why not just use a `.npmrc` file in the root of the repository with that config? We'd still need to check it (and pass `--registry <url>` if present) when running npm operations in packages, but that's no different than currently.

Using the `.npmrc` file means we can use `npm config get registry` from the root and get the correct value, simplifying the various commands that need this. It also keeps the `lerna.json` focused, instead of duplicating information already available elsewhere.

src/Command.js
```diff
@@ -33,6 +35,31 @@ export default class Command {
...
+    if (lernaRegistry) {
+      if (npmRegistry !== lernaRegistry) {
+        this.logger.warn(
+          "Your lerna.json specifies that you should use the registry " +
+          `${lernaRegistry}.\nTo change npm to use ${lernaRegistry} ` +
+          "instead, run `npm config set registry ${lernaRegistry}`."
+        );
+      }
+    } else {
+      if (npmRegistry !== "http://registry.npmjs.com/") {
```
evocateur: As mentioned in https://github.com/lerna/lerna/pull/485, this should be `https://registry.npmjs.org/`.

(it's very confusing to "layer" PRs like this. As far as I can tell only `NpmUtilities.getConfig()` is necessary in the other PR, and it's very noisy to have a bunch of duplicate changes in both PRs...)

src/NpmUtilities.js
```diff
@@ -30,6 +30,25 @@ export default class NpmUtilities {
...
+    }
+  }
+
+  @logger.logifySync()
+  static getConfig(key) {
+    return ChildProcessUtilities.execSync(`npm config get ${key}`);
+  }
+
+  @logger.logifySync()
+  static setConfig(key, value) {
```
evocateur: Is `setConfig()` meant to be used anywhere, or is this added just to make the API symmetrical?

README.md
doug-wade: I dig it!  It also means I can unlayer the prs, since they'll no longer need the config utilities

src/Command.js
doug-wade: Sorry!  I just got on a roll and got a little excited ;)  I'll unlayer them per https://github.com/lerna/lerna/pull/482#discussion_r95507881

src/NpmUtilities.js
doug-wade: Mostly symmetry, but also because [the inital comment](https://github.com/lerna/lerna/issues/55#issue-133396688) explicitly mentioned "logg[ing] you back out afterwards."

doug-wade:
I'm closing this pending #485 merging to prevent confusion.

Issue #418, opened 2016-11-29
End support for Node 0.10 and 0.12


thejameskyle:
Official support for Node 0.10 has already been dropped, 0.12 will be dropped at the end of December, we should remove support for these– especially since 0.10 keeps failing in CI

hzoo:
We kept in 0.10 since we were supporting it in Babel but if you want we can drop and babel will just use the older version until we drop it there as well.

gigabo:
@hzoo Is there a babel issue we can link to that tracks when support will be dropped there?

hzoo:
Yep, and just encountered this today via preset-env 😛  https://github.com/babel/babel/issues/4315

hzoo:
We're going to drop in Babel soon so we can do this in a beta version as well.

hzoo:
https://github.com/lerna/lerna/issues/484

PR #484, opened 2017-01-05
Drop support for node 0.10 and node 0.12

doug-wade:
Addresses #418


gigabo:
Will be nice to not have to worry about these anymore.  Thanks @doug-wade!

Issue #471, opened 2016-12-30
feature request: command to create package


tuchk4:
should work at existing lerna project.

```
lerna create foo
```

will create `/packages/foo/package.json`

with defined fields:

- name
- repository
- version (1.0.0 default)
- author

cmrd-senya:
This request has been implemented in #1326

PR #485, opened 2017-01-05
Add "create" command

doug-wade:
Addresses #471

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,18 @@ repo.  Each commit is modified to make changes relative to the package
 directory.  So, for example, the commit that added `package.json` will
 instead add `packages/<directory-name>/package.json`.
 
+### create
+
+```sh
+$ lerna create <package-name>
+```
+
+Creates a new package at `packages/<package-name>`, unless there is a configured
+package location, in which case the package is created in a new directory named
+`<package-name>` in the first configured location.
+
+This is useful for adding new packages to an existing lerna repository.
+
 ## Misc
 
 Lerna will log to a `lerna-debug.log` file (same as `npm-debug.log`) when it encounters an error running a command.
@@ -453,7 +465,8 @@ Running `lerna` without arguments will show all commands/options.
   "bootstrapConfig": {
     "ignore": "component-*"
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "registry": "https://registry.npmjs.org"
 }
 ```
 
@@ -463,6 +476,7 @@ Running `lerna` without arguments will show all commands/options.
 - `bootstrapConfig.ignore`: an glob that won't be bootstrapped when running the `lerna bootstrap` command.
 - `bootstrapConfig.scope`: an glob that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
+- `registry`: The registry to publish public modules into.
 
 
 ### Common `devDependencies`
--- a/src/Command.js
+++ b/src/Command.js
@@ -99,10 +99,10 @@ export default class Command {
     const ignore = this.flags.ignore || (this.configFlags && this.configFlags.ignore);
 
     if (scope) {
-      this.logger.info(`Scoping to packages that match '${scope}'`);
+      this.logger.info(`Scoping to packages that match "${scope}"`);
     }
     if (ignore) {
-      this.logger.info(`Ignoring packages that match '${ignore}'`);
+      this.logger.info(`Ignoring packages that match "${ignore}"`);
     }
     try {
       this.repository.buildPackageGraph();
--- a/src/commands/CreateCommand.js
+++ b/src/commands/CreateCommand.js
@@ -1,0 +1,38 @@
+import Command from "../Command";
+import ConfigUtilities from "../ConfigUtilities";
+import FileSystemUtilities from "../FileSystemUtilities";
+import GitUtilities from "../GitUtilities";
+import NpmUtilities from "../NpmUtilities";
+import path from "path";
+
+export default class CreateCommand extends Command {
+  initialize(callback) {
+    this.pkgName = this.input[0];
+
+    callback(null, true);
+  }
+
+  execute(callback) {
+    const author = NpmUtilities.getWhoIAm();
+    const name = this.pkgName;
+    const config = ConfigUtilities.readSync(this.repository.rootPath);
+    const {registry, packages} = config;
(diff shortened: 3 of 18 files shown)


src/commands/CreateCommand.js
```diff
@@ -0,0 +1,33 @@
+import Command from "../Command";
+import ConfigUtilities from "../ConfigUtilities";
+import FileSystemUtilities from "../FileSystemUtilities";
+import NpmUtilities from "../NpmUtilities";
+import path from "path";
+
+export default class LsCommand extends Command {
```
seansfkelley: In my repo, I use Yeoman to do package generation, which has been working great. Perhaps this command isn't really necessary?

doug-wade: I actually do the same thing: https://github.com/doug-wade/clefs/tree/master/packages/generator-clefs-plugin -- I suppose we could consider having `init` run [generator-gulp-plugin](https://github.com/hparra/generator-gulp-plugin) or something.  However, asking new users to learn about and maintain a yeoman generator seems like a pretty high barrier to entry; I'm inclined to support a simple command in lerna and add the yeoman generator suggestion to the docs.

gigabo: Don't forget to update the class name here (still `LsCommand`).

src/commands/CreateCommand.js
```diff
@@ -0,0 +1,33 @@
...
+    FileSystemUtilities.mkdirp(pkgRoot, () => {
+      const manifest = {
+        name,
+        author,
+        version: "1.0.0"
+      };
+      if (registry) {
+        manifest.publishConfig = { registry };
+      }
+      FileSystemUtilities.writeFileSync(path.join(pkgRoot, "package.json"), JSON.stringify(manifest));
```
gigabo: Should we `git add` it, too?

src/commands/CreateCommand.js
```diff
@@ -0,0 +1,33 @@
...
+    const pkgRoot = path.join(".", "packages", this.pkgName);
+    const registry = ConfigUtilities.read(this.repository.rootPath).registry;
+    FileSystemUtilities.mkdirp(pkgRoot, () => {
+      const manifest = {
+        name,
+        author,
+        version: "1.0.0"
+      };
+      if (registry) {
+        manifest.publishConfig = { registry };
```
gigabo: I wonder if a nicer way to handle a `registry` in Lerna config would be to specify it on the command line when we use npm commands.  That way a _change_ to the registry in `lerna.json` would have an immediate effect across packages.

gigabo:
@doug-wade This is cool, thanks for putting it together!

If you get a chance please do the README/test thing. ;)

src/Command.js
```diff
@@ -33,6 +35,31 @@ export default class Command {
...
+    if (lernaRegistry) {
+      if (npmRegistry !== lernaRegistry) {
+        this.logger.warn(
+          "Your lerna.json specifies that you should use the registry " +
+          `${lernaRegistry}.\nTo change npm to use ${lernaRegistry} ` +
+          "instead, run `npm config set registry ${lernaRegistry}`."
+        );
+      }
+    } else {
+      if (npmRegistry !== "http://registry.npmjs.com/") {
```
evocateur: The public [registry](https://docs.npmjs.com/misc/registry) is always `https://registry.npmjs.org/`

src/commands/CreateCommand.js
```diff
@@ -0,0 +1,33 @@
...
+  initialize(callback) {
+    this.pkgName = this.input[0];
+
+    callback(null, true);
+  }
+
+  execute(callback) {
+    const author = NpmUtilities.getWhoIAm();
+    const name = this.pkgName;
+    const pkgRoot = path.join(".", "packages", this.pkgName);
```
evocateur: This needs to support the configurable `packages` location(s), yes?

(presumably picking the first location in the event multiple exist)

src/commands/CreateCommand.js
doug-wade: Setting [publishConfig in package.json](https://docs.npmjs.com/files/package.json#publishconfig) is safer because you can forget to specify the correct registry when using npm or lerna on the command line.  I think we should probably _also_ specify it on the command line, so you catch the change in the registry immediately, but I would also like the added safety if someone uses npm directly.

src/commands/CreateCommand.js
doug-wade: I dig it -- I'm also dropping a commit; lemme know if that's too much

src/commands/CreateCommand.js
doug-wade: 👍

doug-wade:
Hi @evocateur!  Sorry for the long delay.  I've added a mess of tests, and tightened up the diff to only what's strictly necessary.  I think this one should be ready to merge.

README.md
```diff
@@ -453,7 +465,8 @@ Running `lerna` without arguments will show all commands/options.
   "bootstrapConfig": {
     "ignore": "component-*"
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "registry": "https://registry.npmjs.org"
```
doug-wade: I'm hoping I can expand on this to more fully address https://github.com/lerna/lerna/issues/515, though I'm saving that for another pr.  I think we should _also_ set it in package.json, just to be safe (in case someone runs `npm publish` in one of the node modules, for instance).

doug-wade:
🔔 @evocateur

README.md
evocateur: I'm definitely glad we're setting `publishConfig.registry` in the created package.json, that's key.

evocateur:
@doug-wade sorry for the delay, :shipit:

casesandberg:
Shouldn't this use the native `npm init`? Otherwise, it won't adhere to any custom `.npm-init.js` set and leave the end user with no way to customize it.

evocateur:
That's a good point. Always a good thing to build on top of existing functionality.

cilice:
@evocateur @doug-wade  Any updates on this whether it'll be merged? It's approved, but being a PR since January, there are now some conflicts.

anfedorov:
getting the conflicts resolved sounds like a good next step

wbhob:
I am upset this never got resolved.

wbhob:
But I do like this feature, great work!

Nufflee:
Any updates on this?

Brantron:
Whaaat? This is sweet and it's just sitting here open?

evocateur:
It's too conflicted to apply to the current codebase. I do plan on implementing it soon, however.

PR #486, opened 2017-01-05
Add export command

doug-wade:
This pr adds a new command, an export command.  This is conceptually similar to create-react-app's eject functionality, and is (somewhat paradoxically) intended to increase and ease adoption of lerna by making new users feel that they aren't "locked in" to the framework, and is the companion to the existing "import" command.

### DIFF
--- a/src/commands/ExportCommand.js
+++ b/src/commands/ExportCommand.js
@@ -1,0 +1,68 @@
+import Command from "../Command";
+import ChildProcessUtilities from "../ChildProcessUtilities";
+import FileSystemUtilities from "../FileSystemUtilities";
+import async from "async";
+import path from "path";
+import userHome from "user-home";
+
+export default class ExportCommand extends Command {
+  initialize(callback) {
+    this.dry = this.flags.dryRun;
+
+    this.pkg = this.input[0];
+    this.to = this.input[1] || userHome;
+
+    if (this.dry) {
+      this.logger.info("The following directories would be moved");
+    }
+
+    callback(null, true);
+  }
+
+  execute(callback) {
+    if (!this.to) {
+      // we actually know that this is an error earlier, during initialization,
+      // but calling the initialization errback with an error doesn't halt
+      // execution and print a nice error like the execute errback.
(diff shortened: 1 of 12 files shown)


src/commands/ExportCommand.js
```diff
@@ -0,0 +1,68 @@
...
+
+    if (this.dry) {
+      this.logger.info(`  - ${pkg._location} -> ${to}`);
+      return callback(null, true);
+    } else {
+      const branchName = `export-${pkg.name}`;
+      ChildProcessUtilities.execSync(`git subtree split -P ${pkg._location.replace(process.cwd() + "/", "")} -b ${branchName}`);
+      FileSystemUtilities.mkdirSync(to);
+      ChildProcessUtilities.execSync("git init", {cwd: to});
+      ChildProcessUtilities.execSync(`git pull ${process.cwd()} ${branchName}`, {cwd: to});
```
gigabo: Can we make this `master` in the newly created repo?

src/commands/ExportCommand.js
```diff
@@ -0,0 +1,68 @@
...
+      ChildProcessUtilities.execSync(`git rm -r ${pkg._location}`);
```
gigabo: Should we delete the temporary branch?

src/commands/ExportCommand.js
```diff
@@ -0,0 +1,68 @@
...
+      const pkg = this.packageGraph.get(this.pkg);
+
+      if (!pkg) {
+        if (this.to) {
+          return callback(`no such package to export ${this.pkg}`);
+        } else {
+          // they're not exporting a single package to their home directory,
+          // they're exporting everything to a provided directory!
+          this.to = this.pkg;
+          this.pkg = undefined; // http://jsperf.com/delete-vs-undefined-vs-null/16
```
gigabo: :squint: 16 million ops/sec not enough here? ;)

image

gigabo: I'm a little confused by the logic here, actually.  Looks like the outer if/else branch we're in here has already locked us into exporting a single `pkg`, and `this.pkg` actually isn't even used after this?

PR #499, opened 2017-01-15
[bootstrap] Support for "prepare" npm event-script #421

### LINKED ISSUE #421: [bootstrap] Support for "prepare" npm event-script
As of npm@4 `prepublish` is split  into `prepare` - that behaves exactly the same as `prepublish` did before - and `prepublishOnly` that is only executed before the actual publish. IMO if `prepublishOnly` script is present in a  bootstrapped package it should be run instead of `prepublish` since it makes more sense.
 
What is your opinion on this?
If this makes sense I can implement it and open a PR

wzrdzl:
Closes #421

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -44,8 +44,7 @@ export default class BootstrapCommand extends Command {
     ], callback);
   }
 
-
-  runScriptInPackages(scriptName, callback) {
+  forEachPackage(task, callback) {
     const packages = this.filteredPackages.slice();
     const batches = PackageUtilities.topologicallyBatchPackages(packages, this.logger);
 
@@ -55,7 +54,7 @@ export default class BootstrapCommand extends Command {
       const batch = batches.shift();
 
       async.parallelLimit(batch.map((pkg) => (done) => {
-        pkg.runScript(scriptName, (err) => {
+        task(pkg, (err) => {
           this.progressBar.tick(pkg.name);
           done(err);
         });
@@ -71,6 +70,12 @@ export default class BootstrapCommand extends Command {
 
     // Kick off the first batch.
     bootstrapBatch();
+  }
+
+  runScriptInPackages(scriptName, callback) {
+    this.forEachPackage((pkg, done) => {
+      pkg.runScript(scriptName, done);
+    }, callback);
   }
 
   /**
@@ -92,12 +97,32 @@ export default class BootstrapCommand extends Command {
   }
 
   /**
-   * Run the "prepublish" NPM script in all bootstrapped packages
+   * Run the "prepublish" or "prepare" (depending on the version of npm client)
+   * NPM script in all bootstrapped packages
    * @param callback
    */
   prepublishPackages(callback) {
     this.logger.info("Prepublishing packages");
-    this.runScriptInPackages("prepublish", callback);
+
+    const npmClientVersionString = NpmUtilities.getNpmClientVersion();
+    const npmClientMajorVersion = parseInt(npmClientVersionString, 10);
+
+    if (npmClientMajorVersion <= 3) {
+      this.runScriptInPackages("prepublish", callback);
+      return;
+    }
+
+    this.forEachPackage(
+      (pkg, done) => {
+        if (pkg.scripts["prepare"]) {
+          pkg.runScript("prepare", done);
+        } else {
+          // fallback to "prepublish" to imitate npm client's behavior
+          pkg.runScript("prepublish", done);
+        }
+      },
+      callback
+    );
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/commands/BootstrapCommand.js
```diff
@@ -92,12 +97,32 @@ export default class BootstrapCommand extends Command {
...
    * @param callback
    */
   prepublishPackages(callback) {
     this.logger.info("Prepublishing packages");
-    this.runScriptInPackages("prepublish", callback);
+
+    const npmClientVersionString = NpmUtilities.getNpmClientVersion();
+    const npmClientMajorVersion = parseInt(npmClientVersionString, 10);
+
+    if (npmClientMajorVersion <= 3) {
```
doug-wade: I'm just gonna point out that we don't actually _need_ to check the npm client version here -- `npm run prepare` will work just as well on 2 as on 4.  That being said, I think you made the right call: it's super important that users be able to fall back to `npm` if lerna isn't doing what they expect, and papering over the version mismatch would preclude that.

wizardzloy: The reason behind this check was discussed [here](https://github.com/lerna/lerna/issues/421#issuecomment-267119633). We, actually, just want to behave the same as your current instance of npm.

wizardzloy:
> Thanks so much for contributing @wizardzloy. I don't think we can merge this without the corresponding change to publish to call prepublishOnly, though.

do mean the changes to `lerna publish`?

evocateur:
We'll need this rebased with conflicts resolved before we can make headway on this.

Andarist:
@evocateur @wzrdzl any chance of this being pushed forward? If you need help, I can try also to help with this, just let me know what needs to be done :)

evocateur:
Ignoring the complexity of resolving the merge conflicts, the [prepublish/prepare/prepublishOnly](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare) [kerfuffle](https://github.com/npm/npm/issues/10074) has gotten _even more_ confusing. (For example, do we replicate npm5's `--ignore-prepublish` flag?)

> * **prepublish**: Run BEFORE the package is packed and published, as well as on local `npm install` without any arguments.
> * **prepare**: Run both BEFORE the package is packed and published, and on local `npm install` without any arguments. This is run AFTER **prepublish**, but BEFORE **prepublishOnly**.
> * **prepublishOnly**: Run BEFORE the package is prepared and packed, ONLY on `npm publish`.

So, at minimum, we're going to need two lifecycle event orders, one for `lerna bootstrap` and another for `lerna publish`. Oy. Digging around in the source of npm itself, here's the order as best I understand it (assume "run lifecycle if present" for all lifecycle events listed):

`lerna bootstrap`:
1. preinstall
2. `npm install --ignore-scripts`
3. postinstall
4. prepublish
5. prepare

`lerna publish`:
1. prepublish
2. prepare
3. prepublishOnly
4. `npm publish --ignore-scripts`
5. postpublish

I don't agree that we need to be prickly about the exact npm version being used. It ignores `yarn` entirely, for one thing, and it also severely increases the complexity of the commands for a very minimal gain. It's incumbent on the user to choose which lifecycles they support.

Andarist:
@evocateur this should be fairly easy, as both workflows (bootstrapping and publishing) are already in place and one would only need to add running additional scripts during both.

Should I just go for it and prepare a PR?

evocateur:
@Andarist Go for it!

Andarist:
This PR can be closed, other PR got merged in - https://github.com/lerna/lerna/pull/979#event-1226118295

pselden:
Does prepublishOnly get run? I don't see it anywhere in the code.

evocateur:
@pselden Not during `lerna bootstrap`, as `prepublishOnly` is meant to be run _only_ during `npm publish`. As `lerna publish` always calls `npm publish` under the hood, it should always be running `prepublishOnly` (if present) at the normal time that npm does so (per package, not repo-wide).

aichholzer:
@evocateur -Using the latest version of `lerna` and `prepublishOnly` is never run, not in my case at least:
```js
"scripts": {
  "prepublishOnly": "npm test",
  "pretest": "lerna bootstrap && eslint --quiet .",
  "test": "nyc --reporter=lcov _mocha -R spec --timeout 500 test/index.js",
  "posttest": "nyc report --reporter=text",
  "publish": "lerna publish"
  }
```

`npm run publish` and it goes straight into asking me which version I'd like to assign to the new release.

If I use `prepublish` instead of `prepublishOnly` it works, but this is causing issues with my CI build, since it's being run as well. Running `npm publish .` directly works as well with `prepublishOnly`.

Any ideas?

evocateur:
This is how npm scripts work. “prepublishOnly” _only_ runs during an execution of `npm publish`.

You should rename your custom “publish” script to “release” to avoid stomping on built-in npm script lifecyles. Then rename “prepublishOnly” to “prerelease” and it will work as you desire.

aichholzer:
Thank you @evocateur -Worked!
However, I thought `lerna publish` always runs `npm publish` under the hood, so `prepublishOnly` should also be run, no? -Which part did I miss..?

Issue #334, opened 2016-09-08
[Feature] Automatic hoisting of common dependencies to repo root


gigabo:
This was left out of #237 to reduce its scope, but is still a valuable improvement!

Hoisting external dependencies that are shared across packages:
- dramatically speeds up bootstrap
- reduces the storage footprint of a Lerna repo
- handles the singleton problem where a common dependency (e.g. React) must be shared

Should this be default behavior, or an option (`lerna bootstrap --hoist-external`)?

It definitely needs a way to opt-out at the dependency level (we have deps internally at Redfin that _can't_ be hoisted).  What does that look like?  An option like `--no-hoist=<externalDep>`?

seansfkelley:
What would be the semantics around different versions of the same dependency scattered around? If I have three packages that depend on `foo` at `^1.0.0`, `^1.1.0` and `^2.0.0`, how many versions of `foo` do I end up with and where are they? I'm assuming a copy of `1.1.0` since it satisfies the first two and one of `2.0.0`, both hoisted. This seems to demand the need for some kind of extra-nested fake-`node_modules` (`lerna_modules`?) like `lerna_modules/foo/1.1.0` and `lerna_modules/foo/2.0.0`.

gigabo:
I think what @rygine originally did in #237 made sense: Hoist the most common version.

Ideally packages within a repo will generally depend on compatible versions.

A [`lerna pin dep@version` command](https://github.com/lerna/lerna/issues/259#issuecomment-229507897) would help with that.

laurentdesmet:
Lerna is a library that is mostly used when you have a lot of packages to manage with a lot of dependencies between them. These kind of repos are the ones that need hoisted dependencies the most, so this is in favor of making it the default option. What are the technical reasons that you can't hoist the dependencies? I can think of dependencies that use the `bin` property of npm but what are other possible reasons?

rygine:
@laurentdesmet there are potentially a few downsides to hoisting:

- hoisted modules can conflict with existing modules at the repo root. so, where do the hoisted modules go? with #332 on the horizon, we can't necessarily guarantee that hoisted modules are located up the directory tree without conflicting with the root.
- `npm outdated` becomes broken. this may not be a deal breaker.
- `package.json` dependencies may become stale as a result of modules being installed "globally." this is more of a maintenance issue, but it has bitten me. 

the last 2 issues may be solved with https://github.com/maxogden/dependency-check or https://github.com/dylang/npm-check.

[...]

Issue #505, opened 2017-01-18
[Feature] Lerna bootstrap --skip-devdependencies


Reinmar:
This is a bit similar to https://github.com/lerna/lerna/issues/495, but a different reasoning.

We're using Lerna with a multi-repo as I described in https://github.com/lerna/lerna/issues/495#issuecomment-272258324. This means that we can't hoist repeating dev deps of our child packages to the main repository, because each child package is a separate repository and, hence, needs its own dependencies.

Threfore, when we run `lerna bootstrap` Lerna tries to install deps and dev deps for all packages. This is quite slow and error prone (I think that npm doesn't handle concurrency too well). Therefore, I'd like to avoid installing the dev dependencies of child packages when running `lerna bootstrap` on CI. They are not needed in such a case and we could speed up the build times significantly.

If such a feature would be accepted, we're willing to make a PR :)

PR #507, opened 2017-01-18
Automatic hoisting of common dependencies

gigabo:
The main thing preventing React Server from migrating back to Lerna is the lack of automatic hoisting.  When I started trying to port that back from Asini, though, I realized that it was built on top of some other improvements... so here's a stack of changes that culminates with an option to automatically hoist common dependencies to the repo root.

The original PRs from Asini for these are:

- [Add an option sieve for all commands](https://github.com/asini/asini/pull/15)
- [Validate and standardize exposure of command classes](https://github.com/asini/asini/pull/18)
- [Automatic hoisting of common dependencies to repo root](https://github.com/asini/asini/pull/36)

There was previously some discussion of hoisting in Lerna [here](https://github.com/lerna/lerna/issues/334).

When I originally did this for Asini I did some perf testing against Babel.  I also had a branch with experimental Yarn support, so I compared them.  Hoisting actually provided a _bigger_ perf improvement than Yarn.  But the best part is they can be _combined_.  Here are bootstrap times on my laptop for Babel with and without hoisting/Yarn:

```
              hoist

            no    yes
          +-----+-----+
        no| 30s | 12s |
yarn      |-----+-----+
       yes| 17s |  7s |
          +-----+-----+
```


lukebatchelor:
Sweet! Will this be opt-in, or automatic? I recall someone here spiking something like this and having issues, so hoping this wont stop us from upgrading in the future (more hoping it just works and we can have the sweet speed increases)

Reinmar:
Sounds great! I guess that this would help a lot with https://github.com/lerna/lerna/issues/505.

Reinmar:
I tested it for CKEditor 5 and the install time shortened from like 5 minutes to 35 seconds :)

gigabo:
@lukebatchelor It's opt-in with `--hoist` on the command-line or `"hoist"` in lerna.json.

Default behavior is unchanged.

The hoist option is boolean with no argument (hoist everything), but it takes an optional glob.  There is also a `--no-hoist`, which takes precedence, so you can easily hoist everything _except_ some subset.

@Reinmar That's awesome! 🚀

gigabo:
Reviewers please take a look at the original three PRs listed in the description above.  There are three distinct change sets here.  The main additional enhancement beyond hoisting is exposure of a `.getOptions()` method on command objects.  This provides values from both lerna.json and the command-line.  It allows for easy, consistent configuration of all flags at the global, command and invocation levels.

src/commands/BootstrapCommand.js
```diff
@@ -133,32 +134,251 @@ export default class BootstrapCommand extends Command {
...
+
+        // Only install deps that can't be hoisted in the leaves.
+        if (version === rootVersion) return;
+
+        dependents[version].forEach((pkg) => {
+
+          if (rootVersion) {
+            this.logger.warn(
+              `"${pkg}" package depends on ${name}@${version}, ` +
+              `which differs from the hoisted ${name}@${rootVersion}.`
```
evocateur: Is this really worth a warning? I can see an argument for warning about `devDependencies`, but warning for production `dependencies` seems overwrought.

src/commands/BootstrapCommand.js
```diff
@@ -133,32 +134,251 @@ export default class BootstrapCommand extends Command {
...
+
+      if (scope && PackageUtilities.getFilteredPackage({name}, {scope, ignore})) {
+
+        // Get the most common version.
+        const commonVersion = Object.keys(versions)
+          .reduce((a, b) => versions[a] > versions[b] ? a : b);
+
+        // Get the version required by the repo root (if any).
+        // If the root doesn't have a dependency on this package then we'll
+        // install the most common dependency there.
```
evocateur: Having used this previously with `asini`, I'm concerned about how hoisting _all_ packages, even ones not specified in the root package.json, _all the way_ to the root makes future `npm` commands (ls, outdated, shrinkwrap, etc) basically worthless due to "extraneous package" errors.

I am all for recognizing when the root package.json has a matching version already, and doing the appropriate binary linking if necessary. Could we perhaps install the "common among packages but not in root" dependencies in a sibling of the packages themselves? (In the case of multiple `packages` locations, the closest common ancestor would suffice)

For example, with the default `'packages/*'` glob, any common dependencies not present in the root would install into `./packages/node_modules`, thus preserving the usefulness of root `npm` commands while gaining the benefits of hoisting. Does that sound reasonable?

src/commands/BootstrapCommand.js
gigabo: It means that something you wanted to have hoisted couldn't be hoisted.  That's at best going to slow down your bootstrap, and at worst going to cause errors due to multiple instances of a dependency when you expected one to be shared.  This warning has saved us a few times in React Server!

src/commands/BootstrapCommand.js
gigabo: IMO it only really makes sense to put deps in the leaves or the root.  Anywhere else is going to either require special configuration or have weird behavior (or both).  Would a `lerna clean` help with the interference with npm commands at the root?

Actually come to think of it I don't think `lerna clean` will remove anything at the root. 🤔

rygine: @evocateur i experimented with trying to get commands such as `npm outdated` to work by symlinking the hoisted deps at the root to the `node_modules` of each package. this caused lots of issues with the way npm treats symlinks. IMO, it wasn't worth it.

one of the other problems i ran into was forgetting to add a dependency to a package's `package.json` file because it was hoisted from another package and everything worked as expected. i thought of creating a tool for checking these sorts of things, but never got around to it.

src/commands/BootstrapCommand.js
rygine: yes, this warning has helped me out in the past too, especially if you want to have the same dependency version across all packages.

evocateur:
I'm willing to forgo `npm outdated` for `packages/node_modules` (which satisfies all common dependencies from `packages/*/package.json`), no fancy re-symlinking necessary. I'm more concerned about `npm shrinkwrap` in the root always failing due to extraneous dependencies.

evocateur:
That's fair, as long as we *only* warn.

gigabo:
Updated to apply cleanly to current master.

Any 👍 👎 on this?  Would be nice to get React Server back onto Lerna.

gigabo:
>  I guess there's always --nohoist...

That's the default.  Nothing is hoisted without `--hoist`.

gigabo:
> I would prefer not to merge this until we've found a way to seed each modules' `node_modules/.bin`

The executable [is linked](https://github.com/lerna/lerna/pull/507/files#diff-0d9bb533ff943e8e57f7829afef76ab5R323) into each dependent package.  That is one of the key benefits provided by automatic vs manual hoisting of devDependencies.

This [broke recently in Asini](https://github.com/asini/asini/pull/80) (and horked React Server).  We haven't gotten to fixing it yet because we're hoping that we'll be able to migrate back to Lerna and halt development on Asini.  That regression is not part of this PR.

doug-wade:
@gigabo OOoooooOOOohhh.  Sorry, I misunderstood what was afoot.  Lgtm.

src/Command.js
```diff
@@ -21,6 +22,47 @@ export default class Command {
...
+  // Override this to inherit config from another command.
+  // For example `updated` inherits config from `publish`.
+  get otherCommandConfigs() {
+    return [];
+  }
+
+  getOptions(...objects) {
+
+    // Items lower down override items higher up.
+    return objectAssign(
```
doug-wade: I think an Object.assign here will get you past the appveyor failure

gigabo: Ooh, nice can we use real `Object.assign` now! 🤘

noherczeg:
@doug-wade When should we expect an npm release containing this incredible addition? :)

doug-wade:
iiuc @hzoo is [working on one now](https://lernajs.slack.com/archives/discussion/p1485901901000815). If that is the case, I doubt it'd be more than a couple of days.

doug-wade:
And it's out! https://twitter.com/left_pad/status/826857968930992128 @noherczeg

noherczeg:
@gigabo I'm currently at `2.0.0-beta.36`, ran `npm run lerna bootstrap --hoist`, and nothing changed. Every dependency is kept (dl-ed, and copied) at it's respective package. What am I missing?

gigabo:
Yeah, looking into it now.  The port from Asini seems not to have gone well. :(

gigabo:
@noherczeg Hmm... actually I tracked down the hoist _slowness_ to a poor interaction with https://github.com/lerna/lerna/pull/547.  Filed https://github.com/lerna/lerna/pull/568 as a patch for that.

That wouldn't affect where things are installed, though.  It just makes bootstrap _slower_ than it needs to be.
 
When you run `bootstrap` do you see warnings about version mismatches preventing hoisting?

noherczeg:
@gigabo I'm not seeing a single warning during bootstrap. There might be a misunderstanding btw, what I practically meant is that "hoisting does not work at all". Have I missed some prerequisite, or it should just work if I run `npm run lerna bootstrap --hoist`?

gigabo:
I wonder if npm is swallowing your option.  You could try terminating npm's options, like:

```bash
npm run lerna bootstrap -- --hoist
```

Or you could try putting `"hoist": true` into your lerna.json.

noherczeg:
@gigabo Yes, it was my bad, sorry :(

It does work, thank you!

PR #514, opened 2017-01-20
Update README.md (s/--exclude/--ignore/)

xaka:
Fixes typo introduced in https://github.com/lerna/lerna/pull/390.


xaka:
@lukebatchelor

gigabo:
Nice catch @xaka!

Issue #61, opened 2016-02-20
Should Lerna support Mercurial?


thejameskyle:
One of the concerns about monorepos is that git doesn't scale well for massive repositories, and while there is lots of work being done to improve this. Today mercurial has much better support for it, companies like [Facebook](https://code.facebook.com/posts/218678814984400/scaling-mercurial-at-facebook/) and Google both use mercurial for large scale projects (@kittens you probably have insight on this). 

Do we want to consider dual-supporting both git and mercurial? I haven't used mercurial before, but based on some initial reading I don't think it'd be hard to abstract out our git operations that could support both.

thejameskyle:
> We should wait for people to request since we aren't using it

Yeah that sounds good.

kittens:
Any large company that uses a monorepo will already have developed solutions that solve this problem, and most of them wont even have it because of the way the codebases are architectured (only one version, `master`).

thejameskyle:
Is it worth supporting otherwise then?

onigoetz:
Hello, I would love the Mercurial support in Lerna, was anything done in that direction ?

hzoo:
@onigoetz sorry no, we've already been pretty overloaded with other issues!

onigoetz:
@hzoo Sure, no problem. I will experiment with Lerna in a project or two, would you be interested by a pull request for this ? (I can't promise anything, but if it answers my needs, I could take the time to contribute)

PR #525, opened 2017-01-23
Initial Mercurial implementation

onigoetz:
Hello,

I needed mercurial support for Lerna, so I implemented it.

It mostly works, I could use `lerna diff` and `lerna publish` successfully.

This is an ongoing pull request as there are a few TODOs left.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ Let's use `babel` as an example.
 ```
 
 - Lerna checks if each dependency is also part of the Lerna repo.
-  - In this example, `babel-generator` is a dependency, while `source-map` is not.
+  - In this example, `babel-generator` is an internal dependency, while `source-map` is an external dependency.
   - `source-map` is `npm install`ed like normal.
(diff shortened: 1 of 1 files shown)


doug-wade:
Hi @onigoetz I really appreciate you chipping in, but I share the concerns expressed in https://github.com/lerna/lerna/issues/61.  I don't really know much about mercurial, personally, and this would mean any time I added a new git command, I'd have to add an equivalent mercurial command.  As much as I'd love to support every use case, I think this be more than we can handle right now. Maybe we could consider a pull request in the future if you run a fork of the project for a while with mercurial support, iron out the public api and demonstrate that there's a community to support the mercurial commands?

onigoetz:
Hello, sure I can maintain a fork on my side, can you maybe "advertise" it somewhere in the readme ? 

Something like "Lerna supports only Git at the moment, a community-maintained fork is available here" ?

doug-wade:
Sounds reasonable to me.

PR #535, opened 2017-01-26
Don't include unnecesary files in the npm package

### LINKED ISSUE #516: Audit and trim published files
Publishing more than necessary.  Should update `.npmignore`.

http://unpkg.com/lerna/

image

Thanks @mjackson for pointing this out!

gpittarelli:
Fixes #516 

Looks to me like only bin/lib are needed (well, and package.json, CHANGELOG.md, LICENSE, README.md which are included by default)--so I switched out the `.npmignore` for a 2 entry `files`.
Edit: ^ undid the switch

I wasn't 100% sure about the yarn files; looks like the answer is don't publish them: see the last line of https://yarnpkg.com/en/docs/yarn-lock

I sanity checked with:
```
npm pack
tar xvf lerna-2.0.0-beta.33.tgz
cd package
ls -a  # => bin  CHANGELOG.md  lib  LICENSE  package.json  README.md
./bin/lerna.js # => prints help message
```

Cuts the package size in half :) 
```
$ du -b --apparent-size lerna-2.0.0-beta.33.tgz after.tgz             
72884   lerna-2.0.0-beta.33.tgz
33347   after.tgz
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,0 @@
-src
-test
-tmp
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,14 @@
+appveyor.yml
+.babelrc
+CHANGELOG.md
+CODE_OF_CONDUCT.md
+.editorconfig
+.eslintrc
+.gitignore
+lerna.json
+.travis.yml
+yarn.lock
+src
+test
+tmp
+.npmignore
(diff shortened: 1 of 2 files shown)


doug-wade:
Hi @gpittarelli!  I really dig saving 40K of disk space, but I'd really prefer that we use .npmignore rather than files/  With `files` if we add a new file and forget to update files, `lerna` will be broken for clients; if we use `.npmignore`, instead the risk is we might include files clients don't need.

PR #541, opened 2017-01-28
Upgrade `inquirer` dependency

wtgtybhertgeghgtwtg:
Reduces dependencies by one.
I didn't see it specified in `engines`, but since [Travis only runs on `node>=4`](https://github.com/lerna/lerna/blob/master/.travis.yml#L10-L13), I figured `node@0.12.0` support was dropped.  If that's wrong, I'll close this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "command-join": "^1.1.1",
     "cross-spawn": "^4.0.0",
     "glob": "^7.0.6",
-    "inquirer": "^0.12.0",
+    "inquirer": "^3.0.1",
     "lodash.find": "^4.3.0",
     "lodash.unionwith": "^4.2.0",
     "meow": "^3.7.0",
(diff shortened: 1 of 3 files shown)


gigabo:
Thanks @wtgtybhertgeghgtwtg.  IIRC the interface for `inquirer` changed between these two versions.  We may need to update our usage.

wtgtybhertgeghgtwtg:
I think you're right, they moved from callbacks to promises.  I'll update that in `PromptUtilities`.

wtgtybhertgeghgtwtg:
It should be fixed.  However, `PromptUtilities` can't be tested.

evocateur:
Parenthetically: I looked into inquirer's tests, and they're solid (and complicated, yikes). We do enough validation of internal `PromptUtilities` usage that I'm confident these changes are safe.

PR #543, opened 2017-01-28
[clean] Support --include-filtered-dependencies flag

### LINKED ISSUE #540: lerna clean --scope should respect --include-filtered-dependencies
This is phrased as a statement, but it's more of a question. In our internal repo, I just attempted to run:

```
➜  main git:(master) ✗ lerna clean --scope corvair --include-filtered-dependencies
Lerna v2.0.0-beta.34
Scoping to packages that match 'corvair'
About to remove the following directories:
- /home/roblg/code/main/corvair/node_modules
? Proceed? (ynH)
```

It picks up the one package, but not any of its dependencies. If I run `lerna bootstrap` with the same arguments, it finds all the relevant transitive dependencies (all 29):

```
➜  main git:(master) ✗ lerna bootstrap --scope corvair --include-filtered-dependencies
Lerna v2.0.0-beta.34
Scoping to packages that match 'corvair'
Bootstrapping 29 packages
...
```

Seems like `clean` should have the same behavior?

My setup:
```
lerna: v2.0.0-beta.34
node: 6.9.1
npm: 3.10.8
```

roblg:
Sorry about the horkage in the CleanCommand test. I moved the existing tests into their own `describe`, which caused whitespace changes, which... you can see the results.

### DIFF
--- a/src/commands/CleanCommand.js
+++ b/src/commands/CleanCommand.js
@@ -1,16 +1,22 @@
 import async from "async";
 import Command from "../Command";
 import FileSystemUtilities from "../FileSystemUtilities";
+import PackageUtilities from "../PackageUtilities";
 import PromptUtilities from "../PromptUtilities";
 import progressBar from "../progressBar";
 
 export default class CleanCommand extends Command {
   initialize(callback) {
+    this.packagesToClean = this.filteredPackages;
+    if (this.flags.includeFilteredDependencies) {
+      this.packagesToClean = PackageUtilities.addDependencies(this.packagesToClean, this.packageGraph);
+    }
+
     if (this.flags.yes) {
       callback(null, true);
     } else {
       this.logger.info(`About to remove the following directories:\n${
-        this.filteredPackages.map((pkg) => "- " + pkg.nodeModulesLocation).join("\n")
+        this.packagesToClean.map((pkg) => "- " + pkg.nodeModulesLocation).join("\n")
       }`);
       PromptUtilities.confirm("Proceed?", (confirmed) => {
         if (confirmed) {
@@ -24,7 +30,7 @@ export default class CleanCommand extends Command {
   }
 
   execute(callback) {
-    progressBar.init(this.filteredPackages.length);
+    progressBar.init(this.packagesToClean.length);
     this.rimrafNodeModulesInPackages((err) => {
       progressBar.terminate();
       if (err) {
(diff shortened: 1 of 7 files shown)


test/CleanCommand.js
```diff
@@ -10,59 +10,103 @@ import stub from "./_stub";
...
-        testDir, "packages/package-" + curPkg, "node_modules"
-      );
+      cleanCommand.runValidations();
+      cleanCommand.runPreparations();
 
-      assert.equal(actualDir, expectedDir);
+      let curPkg = 1;
+      stub(FileSystemUtilities, "rimraf", (actualDir, callback) => {
+        const expectedDir = path.join(
+          testDir, "packages/package-" + curPkg, "node_modules"
```
doug-wade: I think if you used path.join here it'd fix the appveyor failures

roblg: 🤦‍♂️ I suspect it would. Good call.

roblg: Added `path.join` down around line 128. I also realized I oopsidentally checked in a `describe.only`, so I removed that too :grin:

src/commands/CleanCommand.js
```diff
@@ -1,16 +1,22 @@
...
 import FileSystemUtilities from "../FileSystemUtilities";
+import PackageUtilities from "../PackageUtilities";
 import PromptUtilities from "../PromptUtilities";
 import progressBar from "../progressBar";
 
 export default class CleanCommand extends Command {
   initialize(callback) {
+    this.packagesToClean = this.filteredPackages;
+    if (this.flags.includeFilteredDependencies) {
+      this.packagesToClean = PackageUtilities.addDependencies(this.packagesToClean, this.packageGraph);
```
evocateur: I was about to ask why not just mutate `this.filteredPackages` in place, but then I realized this is a pattern used elsewhere and it makes it more readable to boot. 👓

roblg: Yep, I mostly just blindly copied the pattern. I think if this were to be extended other commands in the future, it could make sense to pull the check up into Command.js, and modify the `filteredPackages` there, maybe? I haven't looked closely at that though.

evocateur:
Yeah, no worries for now, mostly just recording for myself that this was in evidence elsewhere.

We'll likely have to do a fuller analysis of all the flags and such, not sure _all_ the commands need such filtering...

PR #546, opened 2017-01-29
Drop `object-assign`.

### LINKED ISSUE #544: Remove usage of object-assign package
Since we have dropped node 0.10/0.12 we can just use the native Object.assign now.

- [ ] remove from package.json
- [ ] remove import statements and use `Object.assign` instead.

wtgtybhertgeghgtwtg:
Closes #544.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "minimatch": "^3.0.0",
     "mkdirp": "^0.5.1",
     "normalize-path": "^2.0.1",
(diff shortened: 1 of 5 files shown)


thejameskyle:
Question @wtgtybhertgeghgtwtg, how the heck do you remember that username?

wtgtybhertgeghgtwtg:
I have it on autocomplete.

PR #547, opened 2017-01-29
Proposal: Spawn child process for rimraf

roblg:
Improves performance of `lerna clean` on large projects.

I fully recognize that this could be contentious, but I also didn't see an issue/conversation about it, so I figured it couldn't hurt to submit. This shaves ~19s (21% improvement!) off our monorepo build with `lerna`:

```
# external rimraf (this PR)
lerna clean --concurrency 6 --loglevel debug --yes  81.95s user 114.90s system 281% cpu 1:10.00 total
lerna clean --concurrency 6 --loglevel debug --yes  78.54s user 117.52s system 282% cpu 1:09.41 total

lerna clean --concurrency 3 --loglevel debug --yes  71.90s user 108.78s system 249% cpu 1:12.42 total


#internal rimraf
lerna clean --concurrency 6 --loglevel debug --yes  55.34s user 85.40s system 156% cpu 1:29.84 total

lerna clean --concurrency 1 --loglevel debug --yes  59.14s user 94.69s system 142% cpu 1:48.14 total
```

For the purposes of conversation, I did the simplest thing I could think of, but I'm also open to exploring alternative solutions. I suspect that shelling out to `rm` and/or the windows equivalent would be even faster still, but I'll resist the urge to do that. :grin:

Update: One downside to this approach is that `rimraf` isn't a require-time dependency anymore, which might cause static analysis (or a human with a restrictive regex) to erroneously conclude that it's not actually being used.

Update 2: Added a couple of additional timings for posterity (`--concurrency 3` for external, `--concurrency 1` for internal).

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -1,11 +1,11 @@
 import pathExists from "path-exists";
 import logger from "./logger";
 import mkdirp from "mkdirp";
-import rimraf from "rimraf";
 import fs from "fs";
 import cmdShim from "cmd-shim";
 import readCmdShim from "read-cmd-shim";
-import { resolve, dirname, relative } from "path";
+import { resolve, dirname, relative, join } from "path";
+import ChildProcessUtilities from "./ChildProcessUtilities";
 
 const ENDS_WITH_NEW_LINE = /\n$/;
 
@@ -51,7 +51,10 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    rimraf(filePath, callback);
+    ChildProcessUtilities.exec("npm bin", {cwd: __dirname}, (err, npmBinPath) => {
+      if (err) return callback(err);
+      ChildProcessUtilities.spawn(join(npmBinPath.trim(), "rimraf"), [filePath], {}, callback);
+    });
   }
 
   @logger.logifyAsync()


src/FileSystemUtilities.js
```diff
@@ -51,7 +51,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    rimraf(filePath, callback);
+    ChildProcessUtilities.spawn(join(__dirname, "../node_modules/.bin/rimraf"), [filePath], {}, callback);
```
doug-wade: `$(npm bin)/rimraf ${filePath}`?

doug-wade: $(npm bin)${path.sep}rimraf ${filePath}? I'm not sure this'll work as written on Windows.

doug-wade: on second thought, it passed on appveyor, so...

doug-wade: nit: I think we should use `$(npm bin)` here

roblg: Can do. Something like this?

```
const npmBin = ChildProcessUtilities.execSync("npm bin");
ChildProcessUtilities.spawn(join(npmBin, "rimraf"), [filePath], {}, callback)
```

roblg:
@doug-wade updated to use `npm bin`. I waffled on `exec` vs. `execSync` for that call, and ended up settling on `exec`. The perf is basically the same, and I think they're both roughly equally readable (surprisingly), but waiting for a synchronous exec in a function that's called in loops seems... wrong somehow.

src/FileSystemUtilities.js
```diff
@@ -51,7 +51,10 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    rimraf(filePath, callback);
+    ChildProcessUtilities.exec("npm bin", {cwd: __dirname}, (err, npmBinPath) => {
```
gigabo: Can/should we cache this?  Seems wasteful to shell out for it every time when it's constant.

roblg: I had that thought too, and then decided that this was probably roughly equivalent to a shell script that did something like `$(npm bin)/rimraf /some/dir/node_modules`, so I didn't bother fixing it. It seems like the cost of running `rimraf` (node-based filesystem walking and unlinking) would dwarf the `npm bin` exec for most projects anyway?

If we want to cache it, I guess we'd do that in a module-level variable in FileSystemUtilities? Wouldn't be hard to add.

gigabo: Yeah, maybe premature optimization.

gigabo:
Thanks @roblg!

gigabo:
@roblg I've edited the title of the PR as prep for inclusion in the changlog.  Please tweak as you see fit.

PR #551, opened 2017-01-30
Fixes commit messages for windows

diogofcunha:
Commit message are broken on windows machines as `$(echo)` is a unix variable.

This is a blocker.


src/GitUtilities.js
```diff
@@ -20,8 +20,7 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static commit(message) {
-    // Use echo to allow multi\nline strings.
-    ChildProcessUtilities.execSync("git commit -m \"$(echo \"" + message + "\")\"");
+    ChildProcessUtilities.execSync("git commit -m \"" + message + "\"");
```
gigabo: So do multi-line strings still work with this?

evocateur: As long as there are no escaped double quotes inside `message`, yes. Perhaps we should `JSON.stringify(message)`?

olegskl: We have tested it on Windows 7 and it doesn't work. The `EOL`s terminate the command on windows and we end up with just a "Publish" message:

publish

Using `JSON.stringify` doesn't work either (throws an error, see below).

error_json

olegskl: We had more success with this change: https://github.com/florianorpeliere/lerna/commit/86d53258ac084deb954daffca8a0081a593d7d0f

diogofcunha: @olegskl that one looks good to me, is there a PR open with that change?

warash:
Hi Guys,
this is pretty urgent for my project. I would be super gladful if this fix will be merged to new release.
thanks

gigabo:
Yeah, a test that actually runs through the shell would be nice.

If nothing else the existing tests will need to be updated.

image

orkisz:
This is a blocker for me as well. Including a fix in next release will be much appreciated. Thanks!

pleunv:
Would really love to see this merged as well... Anything us mere mortals can do to speed this up? :)

pleunv:
I had a bit of a look at this. The tests are failing because the EOL is hardcoded and differs on Windows and Unix. One possible fix is [this one](https://github.com/CogniStreamer/lerna/commit/89faf553d504681b79922d7e4c32c8468882e1fd).

There is an underlying issue with converting the EOLs to a `git commit` with each line as a separate `-m` however, as this seems to introduce multiple new lines, and changes the commit from 

```
Publish

 - @org/app@0.0.2
 - @org/module-one@0.0.2
 - @org/module-one@0.0.3
```
to
```
Publish

 - @org/app@0.0.2

 - @org/module-one@0.0.2

 - @org/module-one@0.0.3
```

This is of course undesired. I really don't know how to get around this behavior.
There's an alternative approach for multiline messages by prepending the message with a dollar sign, like `git commit -m $"My multiline\nmessage"`, but this only works reliably on *nix shells and will print the `$` in the message when executed in a DOS prompt.

I really would love to get this fixed as this appears to be the only blocking issue for Lerna support on Windows, however, I don't see a way to correctly implement the commit messages, except by resorting to writing it to a temp file. Anyone else an idea?

pleunv:
Using [temp files](https://github.com/CogniStreamer/lerna/commit/e7149abfaa33da1a2ea121ebd5276b7038ea61c1) for commit messages with `git commit -F` seems to work consistently across systems. Could clean this up if this is an approach that would be considered.

evocateur:
I think the temp file approach is an acceptable compromise, given the other complications. (Many thanks for the deep-dive investigation!)

billyzkid:
Any idea on when when we can expect this fairly simple fix to be merged? It looked like it was close for a minute and is back in limbo again. I have "$(echo ' Publish" showing up in commit logs which is not very helpful.

pleunv:
This pull request is changing the formatting of the commit messages and has issues with cross-env compatibility so would need to be looked at again.
I've got a branch with a fix [here](https://github.com/CogniStreamer/lerna) and a [temporary npm package](http://npmjs.com/@cognistreamer/lerna-windows-compat) for personal use but I think this approach might have to be discussed further because of the temp files that get created and the fact they are currently also breaking the tests. Right now I'm swamped however so I won't have time to look at this further the coming week. But feel free to play with it and give input :)

evocateur:
The tempfile approach has the most promise, for a variety of reasons (beyond Windows support alone). We would definitely welcome a PR for this problem based on that.

evocateur:
Fixed in #758

simondel:
Awesome, thanks!

orkisz:
Great news. Thanks!

PR #557, opened 2017-01-31
Fix broken hoisting tests

doug-wade:
Looks like the merge of https://github.com/lerna/lerna/pull/507 was a little rocky; my apologies.  Here's a fix to the broken tests.


gigabo:
Ugh, mid-air collision. Thanks @doug-wade!

Issue #511, opened 2017-01-19
Add support for `exact` flag in lerna.json


gigabo:
https://github.com/lerna/lerna/pull/426 added an `--exact` flag, but only on the command line.

It would be nice to allow durable configuration via `lerna.json`.  Should be a one-liner after https://github.com/lerna/lerna/pull/507 lands since that PR standardizes configuration and exposes a `.getOptions()` method on command objects.

Will need to update docs, though.  And tests would be nice. 😁

dortzur:
I can try and take this on.

gigabo:
Thanks @dortzur!

gigabo:
@dortzur #507 has landed in master, so `this.getOptions()` is available in command objects! 🚀

YasserKaddour:
@dortzur I hope you don't mind me doing it, Thanks

PR #558, opened 2017-01-31
Updating README: `clean` respects --include-filtered-dependencies

roblg:
Forgot to update the README when I added this a couple days back.


gigabo:
Thanks for fixing this up @roblg.

BTW - Now that #507 has landed this could be easily updated to support configuration via `includeFilteredDeps` in lerna.json.  Command objects now have a `this.getOptions()` which retrieves merged options from config and the CLI.  Similar to https://github.com/lerna/lerna/issues/511.

Issue #515, opened 2017-01-20
publish using private registry argument


jogelin:
I read many discussions about the possibility to publish to a private registry but is there a way to specify a private registry as an argument in the publish command ?

Using npm we can do `npm publish --registry http://my-private-registry`

I can use the publishConfig configuration in each package.json but it's a duplication which is something that I want to avoid by using lerna ;-)

Do you think it's a good idea to have : `lerna publish --registry=http://my-private-registry` ?

rtsao:
You can do `npm_config_registry=https://my.npm.registry.org lerna <command>` for all commands.

hzoo:
* [#599](https://github.com/lerna/lerna/pull/599) Explicit registry flag feature. ([@noherczeg](https://github.com/noherczeg))

```sh
$ lerna publish --registry https://my-private-registry
```

PR #563, opened 2017-02-01
Support for explicit registry argument

noherczeg:
In response to #515

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,13 +293,26 @@ dependencies, without committing, tagging, pushing or publishing.
 
 ```sh
 $ lerna publish --registry https://my-private-registry
+$ lerna bootstrap --registry https://my-private-registry
 ```
 
-When run with this flag, `publish` will use the specified registry for your package(es).
+When run with this flag, the above commands will use the specified registry for your package(s).
 
 This is useful if you do not want to explicitly set up your registry
 configuration in all of your package.json files individually when e.g. using
 private registries.
+
+> The `registry` flag can also be set in lerna.json.
+
+There could also be scenarios where your private registry is split up into an upstream
+and a downstream part (like in [Nexus3](https://books.sonatype.com/nexus-book/3.0/reference/npm.html)).
+
+In this case if you would like to configure each in lerna.json you should use the following properties:
+* upstreamRegistry
+* downstreamRegistry
+
+> If you defined these properties then the bootstrap and publish commands will each use the one suited for the job.
+
 
 #### --force-publish [packages]
 
--- a/README.md
+++ b/README.md
@@ -289,6 +289,18 @@ dependencies, without committing, tagging, pushing or publishing.
 
 > Only update versions and dependencies; don't actually publish (this only affects publish).
 
+#### --registry [registry]
+
+```sh
+$ lerna publish --registry https://my-private-registry
+```
+
+When run with this flag, `publish` will use the specified registry for your package(es).
+
+This is useful if you do not want to explicitly set up your registry
+configuration in all of your package.json files individually when e.g. using
+private registries.
+
 #### --force-publish [packages]
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -289,31 +289,6 @@ dependencies, without committing, tagging, pushing or publishing.
 
 > Only update versions and dependencies; don't actually publish (this only affects publish).
 
-#### --registry [registry]
-
-```sh
-$ lerna publish --registry https://my-private-registry
-$ lerna bootstrap --registry https://my-private-registry
-```
-
-When run with this flag, the above commands will use the specified registry for your package(s).
-
-This is useful if you do not want to explicitly set up your registry
-configuration in all of your package.json files individually when e.g. using
-private registries.
-
-> The `registry` flag can also be set in lerna.json.
-
-There could also be scenarios where your private registry is split up into an upstream
-and a downstream part (like in [Nexus3](https://books.sonatype.com/nexus-book/3.0/reference/npm.html)).
-
-In this case if you would like to configure each in lerna.json you should use the following properties:
-* upstreamRegistry
-* downstreamRegistry
-
-> If you defined these properties then the bootstrap and publish commands will each use the one suited for the job.
-
-
 #### --force-publish [packages]
 
 ```sh
@@ -667,6 +642,19 @@ can be used to opt out of hoisting for certain dependencies.
 ```sh
 $ lerna bootstrap --hoist --nohoist=babel-*
 ```
+
+#### --registry [registry]
+
+```sh
+$ lerna publish --registry https://my-private-registry
+$ lerna bootstrap --registry https://my-private-registry
+```
+
+When run with this flag, the above commands will use the specified registry for your package(s).
+
+This is useful if you do not want to explicitly set up your registry
+configuration in all of your package.json files individually when e.g. using
+private registries.
 
 ### Wizard
 
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -21,7 +21,7 @@ export default class NpmUtilities {
     };
 
     if (registry) {
-      opts.env = {npm_config_registry: registry}
+      opts.env = {npm_config_registry: registry};
     }
 
     ChildProcessUtilities.spawn("npm", args, opts, callback);
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -97,8 +97,15 @@ export default class NpmUtilities {
   }
 
   @logger.logifyAsync()
-  static publishTaggedInDir(tag, directory, callback) {
-    ChildProcessUtilities.exec("cd " + escapeArgs(directory) + " && npm publish --tag " + tag, null, callback);
+  static publishTaggedInDir(opts, directory, callback) {
+    const validOpts = ["tag", "registry"];
+    const args = Object.keys(opts).map((argKey) => {
+      if (validOpts.indexOf(argKey) > -1) {
+        return `--${escapeArgs(argKey)} ${escapeArgs(opts[argKey])}`;
+      }
+    });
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -289,6 +289,18 @@ dependencies, without committing, tagging, pushing or publishing.
 
 > Only update versions and dependencies; don't actually publish (this only affects publish).
 
+#### --registry [registry]
+
+```sh
+$ lerna publish --registry https://my-private-registry
```
evocateur: I'm not sure `publish` is the only command that would need this; certainly `bootstrap` could also potentially need it (for example, dependencies on private modules that aren't co-located in the monorepo).

We should also allow this setting in `lerna.json`, accessed via the `Command##getOptions()` API.

src/NpmUtilities.js
```diff
@@ -47,8 +47,15 @@ export default class NpmUtilities {
...
-    ChildProcessUtilities.exec("cd " + escapeArgs(directory) + " && npm publish --tag " + tag, null, callback);
+  static publishTaggedInDir(opts, directory, callback) {
+    const validOpts = ["tag", "registry"];
+    const args = Object.keys(opts).map((argKey) => {
+      if (validOpts.indexOf(argKey) > -1) {
+        return `--${escapeArgs(argKey)} ${escapeArgs(opts[argKey])}`;
+      }
+    });
+    let command = ("npm publish " + args.join(" ")).trim();
+    ChildProcessUtilities.exec("cd " + escapeArgs(directory) + " && " + command, null, callback);
```
evocateur: I'm not a huge fan of this argument whitelisting, though I understand the purpose.

Instead of direct flags, I would rather provide config like this via [env vars](https://docs.npmjs.com/misc/config#environment-variables) passed in the options argument.

```
const command = ("npm publish --tag " + tag).trim()
const env = {
  npm_config_registry: opts.registry
};
ChildProcessUtilities.exec("cd " + escapeArgs(directory) + " && " + command, { env }, callback);
```

src/commands/PublishCommand.js
```diff
@@ -345,7 +345,13 @@ export default class PublishCommand extends Command {
       const run = (cb) => {
         this.logger.verbose("Publishing " + pkg.name + "...");
 
-        NpmUtilities.publishTaggedInDir("lerna-temp", pkg.location, (err) => {
+        let args = {
+          "tag": "lerna-temp"
+        };
+        if (this.flags.registry) {
+          args.registry = this.flags.registry;
```
evocateur: this should use `getOptions()`

README.md
doug-wade: I might even suggest that _all_ npm operations provide the registry flag -- I think that's what I would expect if I provided a registry key in lerna.json.

src/commands/PublishCommand.js
noherczeg: I just noticed that that exists after the PR, will swap it out!

README.md
noherczeg: I'll look into your suggestions, and work on them!

noherczeg:
@evocateur You mentioned `dist-tag` for this to apply to as well, but I fail to see what does that have to do with registries. Could you please elaborate on this?

evocateur:
What happens when you try to query a dist-tag from a registry where the package doesn't exist?

noherczeg:
@evocateur My bad. I checked the docs before my question, but failed to realize that this is not a git related "tag" (never made tags as aliases before). My last addition still does not include dist tag support, but I'll make it happen next evening.

src/NpmUtilities.js
```diff
@@ -6,18 +6,22 @@ import semver from "semver";
...
-  static installInDir(directory, dependencies, callback) {
+  static installInDir(directory, dependencies, registry, callback) {
     let args = ["install"];
 
     if (dependencies) {
       args = args.concat(dependencies);
     }
 
-    const opts = {
+    let opts = {
```
evocateur: `const` doesn't prohibit object mutation, only reassignment to the identifier.

README.md
```diff
@@ -289,6 +289,31 @@ dependencies, without committing, tagging, pushing or publishing.
...
+When run with this flag, the above commands will use the specified registry for your package(s).
+
+This is useful if you do not want to explicitly set up your registry
+configuration in all of your package.json files individually when e.g. using
+private registries.
+
+> The `registry` flag can also be set in lerna.json.
+
+There could also be scenarios where your private registry is split up into an upstream
+and a downstream part (like in [Nexus3](https://books.sonatype.com/nexus-book/3.0/reference/npm.html)).
```
evocateur: Let's keep this initial implementation as simple as possible. Is there a great need for Nexus3 support from lerna users?

noherczeg: We are using Nexus, that's where the example came from. I understand your point about keeping it simple, therefore I will remove every other part besides registry from my PR. Mostly because nothing prevents us from keeping our upstream publishConfigs in the packages like they are currently :)

doug-wade: I agree with @evocateur that the initial implementation should be as simple as possible, but I'd still like to see nexus support in a subsequent pr.

rtsao:
This may not be necessary. `npm_config_registry=https://my.npm.registry.org lerna <command>` works.

noherczeg:
Hey guys!

I'm curious if you find my PR okay or is there anything I can make it better?

README.md
```diff
@@ -642,6 +642,19 @@ can be used to opt out of hoisting for certain dependencies.
...
 ```
 
+#### --registry [registry]
+
+```sh
+$ lerna publish --registry https://my-private-registry
+$ lerna bootstrap --registry https://my-private-registry
+```
+
+When run with this flag, the above commands will use the specified registry for your package(s).
```
doug-wade: not just the above commands anymore, right?

src/NpmUtilities.js
```diff
@@ -18,22 +18,29 @@ export default class NpmUtilities {
...
+
     ChildProcessUtilities.spawn("npm", args, opts, callback);
   }
 
   @logger.logifySync()
-  static addDistTag(packageName, version, tag) {
-    ChildProcessUtilities.execSync(`npm dist-tag add ${packageName}@${version} ${tag}`);
+  static addDistTag(packageName, version, tag, registry) {
+    const opts = NpmUtilities.getTagOpts(registry);
+    ChildProcessUtilities.execSync(`npm dist-tag add ${packageName}@${version} ${tag}`, opts);
```
doug-wade: making this part of the public api like this makes me think you're shelling out to npm every time.  I feel like 

```
ChildProcessUtilities.execSync(`npm dist-tag add ${packageName}@${version} ${tag}`, {registry});
```

would be better, as would making getTagOpts a function

```
ChildProcessUtilities.execSync(`npm dist-tag add ${packageName}@${version} ${tag}`, getTagOpts(registry));
```

README.md
noherczeg: Bad wording, will change it in the evening!

src/NpmUtilities.js
noherczeg: I put it there because I thought that since NPM is handling the publish process it'd be a good enough fit there. By "making getTagOpts a function" do you mean a global distinct function?

README.md
noherczeg: I checked it, and found that the `run` command does not get decorated yet. Could you guys give me some wisdom on this? Do we need it to be present on run-s as well? I'm not sure about it.

src/commands/PublishCommand.js
```diff
@@ -338,14 +338,15 @@ export default class PublishCommand extends Command {
     });
 
     this.progressBar.init(this.packagesToPublish.length);
+    const registry = this.getOptions().registry;
```
evocateur: Let's set this as an instance prop in the `initialize` method, like `this.npmRegistry = this.getOptions().registry;`

test/PublishCommand.js
```diff
@@ -691,6 +691,155 @@ describe("PublishCommand", () => {
...
+        ]],
+        [ChildProcessUtilities, "exec", { nodeCallback: true }, [
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-1")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-2")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-3")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-4")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] }
+          // No package-5.  It's private.
+        ], true],
+        [ChildProcessUtilities, "execSync", {}, [
+          { args: ["npm dist-tag ls package-1"], returns: "lerna-temp: 1.0.1\nstable: 1.0.0" },
```
evocateur: I thought the dist-tag commands passed the env var too?

noherczeg: @evocateur Sorry, but I can not make the tests error out on me. Seems like I misunderstood how to do them properly so I need to evaluate again :(

noherczeg:
Webstorm  totally messed up my repo during conflict resolution, will create a proper one soon. Sorry for the inconvenience!

evocateur:
Let's stay on this PR to maintain comment history. Feel free to force push into your branch to correct the errors, no need for a "fresh" branch.

We'll squash merge anyway, so no worries about noisy commits. :)

noherczeg:
@evocateur I messed it up really bad, so I didn't wanted to risk losing anything, or creating corrupt commits (because barely managed to keep track what happened where).

PR #566, opened 2017-02-01
Fix rimraf bin resolution

rtsao:
On NPM3, `rimraf` is being hoisted to the parent node_modules, so `exec("npm bin", {cwd: __dirname},` produces the wrong bin path. So in the current version of lerna, the clean command is broken.

I've simplified the logic to use `require.resolve`, which will work regardless of where rimraf is. This should also be faster as well!

Fixes bug introduced in: https://github.com/lerna/lerna/pull/547

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -4,7 +4,7 @@ import mkdirp from "mkdirp";
 import fs from "fs";
 import cmdShim from "cmd-shim";
 import readCmdShim from "read-cmd-shim";
-import { resolve, dirname, relative, join } from "path";
+import { resolve, dirname, relative } from "path";
 import ChildProcessUtilities from "./ChildProcessUtilities";
 
 const ENDS_WITH_NEW_LINE = /\n$/;
@@ -51,10 +51,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    ChildProcessUtilities.exec("npm bin", {cwd: __dirname}, (err, npmBinPath) => {
-      if (err) return callback(err);
-      ChildProcessUtilities.spawn(join(npmBinPath.trim(), "rimraf"), [filePath], {}, callback);
-    });
+    ChildProcessUtilities.spawn(require.resolve("rimraf/bin"), [filePath], {}, callback);
   }
 
   @logger.logifyAsync()
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -51,6 +51,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
+    // Note: if rimraf moves the location of its executable, this will need to be updated
     ChildProcessUtilities.spawn(require.resolve("rimraf/bin"), [filePath], {}, callback);
   }
 


src/FileSystemUtilities.js
```diff
@@ -51,10 +51,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    ChildProcessUtilities.exec("npm bin", {cwd: __dirname}, (err, npmBinPath) => {
-      if (err) return callback(err);
-      ChildProcessUtilities.spawn(join(npmBinPath.trim(), "rimraf"), [filePath], {}, callback);
-    });
+    ChildProcessUtilities.spawn(require.resolve('rimraf/bin'), [filePath], {}, callback);
```
gigabo: Linter's choking on the single-quotes here.

rtsao: Yep, just realized :)

roblg:
Nice! 👍

src/FileSystemUtilities.js
```diff
@@ -51,10 +51,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    ChildProcessUtilities.exec("npm bin", {cwd: __dirname}, (err, npmBinPath) => {
-      if (err) return callback(err);
-      ChildProcessUtilities.spawn(join(npmBinPath.trim(), "rimraf"), [filePath], {}, callback);
-    });
+    ChildProcessUtilities.spawn(require.resolve("rimraf/bin"), [filePath], {}, callback);
```
gigabo: This relies on the internal structure of the `rimraf` package not changing.

Could we combine this with a walk up to `.bin` so we can use the public interface?

Something like:

```javascript
path.join(require.resolve("rimraf"), "../.bin/rimraf")
```

rtsao: Yeah I thought about this but figured rimraf was pretty stable. Alternatively https://github.com/thlorenz/resolve-bin could be used, but I didn't want to include another dependency.

rtsao: To your idea, is `./.bin/rimraf` guaranteed to be set in the same node_modules folder where `rimraf` is installed? If so, then yeah I like your idea. Would we need to use `path.resolve` in that case?
```js
path.resolve(require.resolve("rimraf"), "../.bin/rimraf")
```

xaka: I'm using https://github.com/thlorenz/resolve-bin in multiple projects to get proper path to bin file.

rtsao: @gigabo thoughts?

gigabo: Yeah, I guess relying on the relative location of the `.bin/` directory isn't much different from relying on the location of `bin.js` in rimraf. 😆 

I'd like to avoid adding a dep if possible.

Maybe a comment to the effect of "this will break if the location of the executable changes within rimraf" would be enough to help us fix things up quickly in the unlikely case that it actually moves.

rtsao: 👍 sounds good. I'll add a comment.

rtsao: @gigabo @hzoo If everything looks good it'd be nice to get this released since `lerna clean` is basically broken right now.

gigabo:
LGTM.  Thanks @rtsao!

hzoo:
awesome

Reinmar:
I think it affected us too (https://travis-ci.org/ckeditor/ckeditor5/builds/197615503#L1201) so we're waiting for a new release :)

hzoo:
Ok released beta.36!

PR #568, opened 2017-02-02
Check if directories exist before removing during hoist

gigabo:
These noop rimrafs were cheap before #547, but they're expensive now.

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -51,8 +51,16 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static rimraf(filePath, callback) {
-    // Note: if rimraf moves the location of its executable, this will need to be updated
-    ChildProcessUtilities.spawn(require.resolve("rimraf/bin"), [filePath], {}, callback);
+
+    // Shelling out to a child process for a noop is expensive.
+    // Checking if `filePath` exists to be removed is cheap.
+    // This lets us short-circuit if we don't have anything to do.
+    pathExists(filePath).then((exists) => {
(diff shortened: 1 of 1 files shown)


src/commands/BootstrapCommand.js
```diff
@@ -346,7 +346,9 @@ export default class BootstrapCommand extends Command {
         async.series(root.map(({name, dependents}) => (cb) => {
           async.series(dependents.map(({nodeModulesLocation: dir}) => (cb) => {
             if (dir === this.repository.nodeModulesLocation) return cb();
-            FileSystemUtilities.rimraf(path.join(dir, name), cb);
+            const filePath = path.join(dir, name);
+            if (!FileSystemUtilities.existsSync(filePath)) return cb();
+            FileSystemUtilities.rimraf(filePath, cb);
```
xaka: nit: as the same might happen in other places, would it be more reasonable to let `FileSystemUtilities.rimraf` handle the case?

gigabo: Yeah, that might be cleaner.  Will be an unnecessary `stat` in most cases, but that's cheap.

xaka: btw, i'd probably file a bug against `rimraf` as it claims to be an equivalent to `rm -rf` and it's not because `-f` makes `rm` to swallow the error while `rimraf` fails.

gigabo: It's not that it _fails_, it's just _slow_ compared to calling `rimraf` in-process.

xaka: got it, thx!

PR #574, opened 2017-02-03
Use correct logger method in Package method.

evocateur:
This fixes #573

PR #575, opened 2017-02-03
Add coverage report

doug-wade:
In response to https://github.com/lerna/lerna/pull/574#issuecomment-277347146

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ node_modules
 lib
 tmp
 .changelog
+coverage/
+.nyc_output/
--- a/.npmignore
+++ b/.npmignore
@@ -12,3 +12,5 @@ src
 test
 tmp
 .npmignore
+.nyc_output
+coverage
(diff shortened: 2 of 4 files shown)


gigabo:
Whoa, speedy @doug-wade is speedy. 🚀

gigabo:
What would we have to do to get a GitHub integration for this and see it in PRs?

doug-wade:
I use [codecov](https://codecov.io/) for that, which adds a [summary to pull requests](https://github.com/doug-wade/idd/pull/64#issuecomment-261154176), but not a full coverage report.  I can set that up over the weekend if that's what we want.  The other provider I've used in [coveralls](https://coveralls.io/), but codecov is newer and newer is always better :wink:.

hzoo:
Babel uses codecov https://github.com/babel/babel/commit/db67d7fdce4feed8c2a1aaa4ee353eb4fd60a5df also does es6 coverage

PR #576, opened 2017-02-03
Install with no arguments

gigabo:
Use temporary package.json files.

This is groundwork for:

- Pluggable npm clients (yarn).
- A `lerna upgrade` command.

This is a port of https://github.com/asini/asini/pull/45.  Please see original PR commentary during review.

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -39,6 +39,16 @@ export default class FileSystemUtilities {
     fs.writeFile(filePath, ensureEndsWithNewLine(fileContents), callback);
   }
 
+  @logger.logifyAsync()
+  static rename(from, to, callback) {
+    fs.rename(from, to, callback);
+  }
+
+  @logger.logifySync()
+  static renameSync(from, to) {
+    fs.renameSync(from, to);
+  }
+
   @logger.logifySync()
   static writeFileSync(filePath, fileContents) {
     fs.writeFileSync(filePath, ensureEndsWithNewLine(fileContents));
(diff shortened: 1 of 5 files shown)


src/NpmUtilities.js
```diff
@@ -11,18 +13,62 @@ export default class NpmUtilities {
...
+        }
+      });
+    });
+  }
+
+  // Take a dep like "foo@^1.0.0".
+  // Return a tuple like ["foo", "^1.0.0"].
+  // Handles scoped packages.
+  // Returns undefined for version if none specified.
+  static splitVersion(dep) {
```
doug-wade: I share @rygine's [enthusiasm for regexes](https://github.com/asini/asini/pull/45/files#r85001506), but it seems that you could write tests for splitVersion directly that verified that it works as intended, which would make me feel much better.

src/FileSystemUtilities.js
```diff
@@ -39,6 +39,16 @@ export default class FileSystemUtilities {
     fs.writeFile(filePath, ensureEndsWithNewLine(fileContents), callback);
   }
 
+  @logger.logifyAsync()
```
doug-wade: It seems like most [FileSystemUtilities have tests](https://github.com/lerna/lerna/blob/master/test/FileSystemUtilities.js); seems worthwhile to add some for these as well.

src/NpmUtilities.js
gigabo: Good call.  Added [some tests](https://github.com/lerna/lerna/pull/576/files#diff-b4cd8e152e8c3ce22a41f4bae7d04939).

src/FileSystemUtilities.js
gigabo: Another good call.  Added [some more tests](https://github.com/lerna/lerna/pull/576/files#diff-c2bcab92ef3ab638c3bb52c8a6cdf9e9).

PR #581, opened 2017-02-07
Improve support for --include-filtered-dependencies

roblg:
* Adds support to `ls`,`exec`,`run`
* Removes need for additional instance variable on BootstrapCommand, CleanCommand
* Adds tests for ls, exec, run

This PR looks huge because I had to change whitespace in a couple of tests, and added new fixtures, but the "actual" work is mostly in `Command.js`, with some find-and-replace in `BootstrapCommand.js`, and `CleanCommand.js`.

### DIFF
--- a/test/RunCommand.js
+++ b/test/RunCommand.js
@@ -10,40 +10,100 @@ import stub from "./_stub";
 import FakeChild from "./_fakeChild";
 
 describe("RunCommand", () => {
-  let testDir;
 
-  beforeEach((done) => {
-    testDir = initFixture("RunCommand/basic", done);
-  });
+  describe("in a basic repo", () => {
+    let testDir;
 
-  it("should run a command", (done) => {
-    const runCommand = new RunCommand(["my-script"], {});
-
-    runCommand.runValidations();
-    runCommand.runPreparations();
-
-    let calls = 0;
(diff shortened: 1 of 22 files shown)


evocateur:
Add a `?w=1` querystring to the file diff to ignore whitespace: https://github.com/lerna/lerna/pull/581/files?w=1

Liking this so far!

evocateur:
(Oddly enough, ignoring whitespace made the additions in the ExecCommand test to disappear, so it's not rocket surgery...)

test/RunCommand.js
```diff
@@ -10,82 +10,116 @@ import stub from "./_stub";
...
+    });
 
-    const children = [];
-    stub(ChildProcessUtilities, "exec", (command, options, callback) => {
-      children.unshift(new FakeChild);
-      ChildProcessUtilities.registerChild(children[0]);
-      callback();
+  });
+
+  describe.only("with --include-filtered-dependencies", () => {
```
hzoo: I guess we need to lint for this?

evocateur: doh, yeah :/

evocateur: Pretty sure that's no longer there, though?

roblg: Ugh. I swear I checked for this this time :(

PR #584, opened 2017-02-08
Bump `command-join`.

wtgtybhertgeghgtwtg:
Removes `array-from` and `repeat-string` from the dependency tree.
Interesting side-note: `command-join` originally did not require `array-from` and `repeat-string`, but they were added [so it could be used by `lerna`](https://github.com/seangenabe/command-join/pull/1).


hzoo:
lol that's funny

PR #595, opened 2017-02-10
Publish npm packages in topological order.

loganfsmyth:
Expanding on the work done in https://github.com/lerna/lerna/pull/412 by @seansfkelley.

This makes the `publish` command publish packages to npm in topological order so that there is never a time where npm could have a package whose dependencies do not yet exist in the registry. There's already some `dist-tags` toggling going on, but that doesn't have any effect on users who already have dependencies on a version range in their deps.

Feedback welcome. Main questions:

If you have dependency cycles in your repo, this means Lerna will start logging `"Encountered a cycle in the dependency graph. This may cause instability!"` on every call to `lerna publish` where it wouldn't have before, is that cool?

### DIFF
--- a/bin/lerna.js
+++ b/bin/lerna.js
@@ -40,7 +40,7 @@ var cli = meow([
   "  --repo-version          Specify repo version to publish",
   "  --concurrency           How many threads to use if lerna parallelises the tasks (defaults to 4)",
   "  --loglevel              What level of logs to report (defaults to \"info\").  On failure, all logs are written to lerna-debug.log in the current working directory.",
-  "  --no-sort            When executing tasks, ignore the dependency ordering of packages (only affects run, exec and bootstrap)",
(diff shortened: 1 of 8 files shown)


gigabo:
Thanks @loganfsmyth, this is an important improvement.  I actually wonder if it should even be an option.  Is there ever a time when publishing _out_ of dependency order is appropriate?

loganfsmyth:
Happy to remove the option usage, I only included it to be consistent with the other PR.

gigabo:
Yeah, I guess there's value in just being consistent across commands.  At least the _default_ is to sort.

Incidentally noticed that the sort option is only supported on the command-line.  Added https://github.com/lerna/lerna/pull/596 for durable config via `lerna.json`.

loganfsmyth:
Cool. Let me know if there's anything else I need to do.

hzoo:
Tested in the Babel 6.23 (npm link) release so, seems like it worked!! (and not much celebration since we had a different regression)

PR #596, opened 2017-02-10
Support `sort` option in lerna.json

gigabo:
Can be configured at the top level or per-command.

Example:

```js
{
  ...
  "commands": {
    "run": {
      "sort": false
    }
  }
}
```

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -18,7 +18,11 @@ export default class Command {
     this.repository = new Repository();
     this.progressBar = progressBar;
     this.concurrency = (!flags || flags.concurrency === undefined) ? DEFAULT_CONCURRENCY : Math.max(1, +flags.concurrency || DEFAULT_CONCURRENCY);
-    this.toposort = !flags || (flags.sort == null ? true : flags.sort);
+
+    const {sort} = this.getOptions();
+
+    // If the option isn't present then the default is to sort.
+    this.toposort = sort == null || sort;
   }
 
   get name() {


src/Command.js
```diff
@@ -18,7 +18,11 @@ export default class Command {
     this.repository = new Repository();
     this.progressBar = progressBar;
     this.concurrency = (!flags || flags.concurrency === undefined) ? DEFAULT_CONCURRENCY : Math.max(1, +flags.concurrency || DEFAULT_CONCURRENCY);
-    this.toposort = !flags || (flags.sort == null ? true : flags.sort);
+
+    const {sort} = this.getOptions();
+
+    // If the option isn't present then the default is to sort.
+    this.toposort = sort == null || sort;
```
evocateur: Since `this.toposort` is never referenced until the `initialize` method of subclasses, it seems like we could wait to determine this in `runPreparations()` or something. My concern (perhaps unfounded?) is that calling getOptions() in the constructor might "miss" the `otherCommandConfigs` override? Do I just suck at class inheritance?

gigabo: @evocateur Good thought!  The constructor name is actually pulled from the object bound to `this` rather than the class where the method is defined, so it should work as intended in this case.

Here's a small script to demonstrate:

```js
class Foo {
  constructor() {
    console.log(this.constructor.name);
  }
}

class Bar extends Foo {}

new Bar;
```

This prints `Bar` rather than `Foo`, since that's what is instantiated.

Thanks for thinking through this so carefully!

PR #597, opened 2017-02-10
Support command config in either "commands" or "command"

gigabo:
Noticed that the docs in the README say it's under "commands", but the code
was only looking in "command".  This patch supports either/or.

Example:

```js
{
  ...
  "commands": {
    "bootstrap": {
      "includeFilteredDependencies": true
    }
  }
}


hzoo:
We can make the breaking change but ok +1

PR #599, opened 2017-02-11
Explicit registry flag feature

noherczeg:
Corrected version of #563

Again, sorry for messing up the previous PR.

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -338,7 +338,6 @@ export default class PublishCommand extends Command {
     });
 
     this.progressBar.init(this.packagesToPublish.length);
-    this.npmRegistry = this.getOptions().registry;
 
     async.parallelLimit(this.packagesToPublish.map((pkg) => {
       let attempts = 0;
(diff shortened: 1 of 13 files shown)


src/commands/PublishCommand.js
```diff
@@ -338,14 +338,15 @@ export default class PublishCommand extends Command {
     });
 
     this.progressBar.init(this.packagesToPublish.length);
+    this.npmRegistry = this.getOptions().registry;
```
evocateur: This should be done once, in the initialize method.

noherczeg: Moved to runPreparations() so every command has access to it. I hope it does not count as an overkill :)

evocateur:
Sounds good!

evocateur:
Once @doug-wade has a chance to review this, we'll be good to go.

noherczeg:
Thanks @evocateur ! :)

gigabo:
Looks like the master build [failed](https://travis-ci.org/lerna/lerna/jobs/202430796) after this landed.

image

doug-wade:
ugh.  must have collided.  I'll hop on it.

PR #601, opened 2017-02-15
Fix --ignore flag when globs are expanded to an array

rtsao:
Suppose we have:
- `package-a`
- `package-b`
- `package-c`

<hr/>

The following works as expected:
```
lerna bootstrap --ignore=package-@(a|b)
```

The message is:
```
Ignoring packages that match 'package-@(a|b)'
```

And the following packages are bootstrapped:
- `package-c`

<hr/>

The following does *not* work as expected:
```
lerna bootstrap --ignore=package-{a,b}
```

The message is:
```
Ignoring packages that match 'package-a,package-b'
```

And the following packages are bootstrapped:
- `package-a` (incorrect)
- `package-b` (incorrect)
- `package-c`

The latter glob is expanded to an array even before reaching `filterPackage` in `PackageUtilities`. (Notice the difference in messages). However, the logic regarding negation (i.e. `maybeNegate`) was only *within* `glob.some`, which would not work in the case that `glob.length > 1`.

This PR changes the boolean reducer depending on the desired negation (i.e. `Array.prototype.some` for the standard case, and `Array.prototype.every` for the negated case.

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -140,10 +140,11 @@ export default class PackageUtilities {
 
     if (!Array.isArray(glob)) glob = [glob];
 
-    const maybeNegate = negate ? (v) => !v : (v) => v;
-    const boolReducer = negate ? Array.prototype.every : Array.prototype.some;
-
-    return boolReducer.call(glob, (glob) => maybeNegate(minimatch(pkg.name, glob)));
+    if (negate) {
+      return glob.every((glob) => !minimatch(pkg.name, glob));
+    } else {
+      return glob.some((glob) => minimatch(pkg.name, glob));
+    }
   }
 
   static getFilteredPackage(pkg, {scope, ignore}) {
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -141,8 +141,9 @@ export default class PackageUtilities {
     if (!Array.isArray(glob)) glob = [glob];
 
     const maybeNegate = negate ? (v) => !v : (v) => v;
+    const boolReducer = negate ? Array.prototype.every : Array.prototype.some;
 
-    return glob.some((glob) => maybeNegate(minimatch(pkg.name, glob)));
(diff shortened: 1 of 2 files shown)


rtsao:
At a high-level, it seems somewhat weird to me that `package-{a,b}` gets expanded to an array but `package-@(a|b)` does not. I'm not sure why this is the case.

That said, the existing logic was clearly designed to handle arrays (it just so happens it didn't quite handle arrays correctly in the negation case and there were no tests for the array case) so I figured fixing this bug in this way seemed appropriate.

gigabo:
Thanks for the patch @rtsao!

> it seems somewhat weird to me that `package-{a,b}` gets expanded

That _is_ weird.  Is the shell expanding it?

Are you in a directory where `./package-a/` and `./package-b/` are direct children?

> didn't quite handle arrays correctly in the negation case

Nice catch!  Thanks for adding a test for it, too. 🎉

src/PackageUtilities.js
```diff
@@ -141,8 +141,9 @@ export default class PackageUtilities {
     if (!Array.isArray(glob)) glob = [glob];
 
     const maybeNegate = negate ? (v) => !v : (v) => v;
+    const boolReducer = negate ? Array.prototype.every : Array.prototype.some;
 
-    return glob.some((glob) => maybeNegate(minimatch(pkg.name, glob)));
+    return boolReducer.call(glob, (glob) => maybeNegate(minimatch(pkg.name, glob)));
```
gigabo: Hmmm... this is getting a little roundabout.  Maybe simpler at this point to just have a top-level if/else?

Like:

```js
if (negate) {
  return glob.every((glob) => !minimatch(pkg.name, glob));
} else {
  return glob.some((glob) => minimatch(pkg.name, glob));
}
```

rtsao: I thought about doing that but didn't want to rock the boat in terms of the existing code. Agreed, that's much simpler, so I'll update the PR with that refactor. 👍

PR #604, opened 2017-02-16
Fix midair collision

doug-wade:
I broke the build :cry:.  This fixes it.  Thanks to @gigabo for [catching it](https://github.com/lerna/lerna/pull/599#issuecomment-280487962)

### DIFF
--- a/test/PublishCommand.js
+++ b/test/PublishCommand.js
@@ -729,19 +729,15 @@ describe("PublishCommand", () => {
         ]],
         [ChildProcessUtilities, "exec", { nodeCallback: true }, [
           { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-1")) + " && npm publish --tag lerna-temp"] },
-          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-2")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
           { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-3")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
-          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-4")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] }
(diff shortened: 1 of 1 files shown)


test/PublishCommand.js
```diff
@@ -729,27 +729,27 @@ describe("PublishCommand", () => {
         ]],
         [ChildProcessUtilities, "exec", { nodeCallback: true }, [
           { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-1")) + " && npm publish --tag lerna-temp"] },
-          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-2")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
           { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-3")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
-          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-4")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] }
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-4")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] },
+          { args: ["cd " + escapeArgs(path.join(testDir, "packages/package-2")) + " && npm publish --tag lerna-temp", {env: {"npm_config_registry":"https://my-private-registry"}}] }
```
gigabo: :squint: `[1,3,4,2]`?

Is that a consistent order, or are these just all racing against each other?

gigabo: Whelp, the build passed, so... fixed forever! 😁

doug-wade: I'm p sure its a consistent order.  2 and 4 have a dependency on 1, and 1 and 3 have no dependencies, so there are two batches: `[[1,3], [2,4]]`.  As to whether 1 always goes before 3 and 4 always goes before 2, I'm less certain, but in my ten or so runs they are consistent, and all the ci suites passed, so it seems so?

gigabo: That makes sense.  Thanks for checking!

Issue #371, opened 2016-10-11
[Feature] Switch to Yarn


cpojer:
https://github.com/yarnpkg/yarn

please! I want to switch to Yarn on Jest :)

joshblack:
Would this require just adding configuration for an `installer` for this command: https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L19

Or something a little more verbose?

BenGale:
From a real quick look:

Could be worth having a YarnUtilities and instead of importing the NPM one, import something that knows how to read the config (not read enough of the source yet to know how config is managed) and give back the correct implementation. Shouldn't be too bad as the commands needed map pretty well to each other from the looks of it.

hzoo:
cc @kittens @thejameskyle

thejameskyle:
I'm considering just using yarn instead of npm client since it's compatible and fixes a lot of issues we've had

thejameskyle:
That being blocked on supporting private packages etc

laurentdesmet:
This would close #334, like @thejameskyle [already mentioned](https://github.com/lerna/lerna/issues/334#issuecomment-246478045), if it's really ultrafast according to [its marketing](https://yarnpkg.com/). I didn't test yarn yet, but I'm very excited to try it out 😃 .

marvinhagemeister:
@laurentdesmet Just gave it a go and the hype is real! It's day and night between yarn and npm install times. I'm rooting for them to fix the remaining issues, most notably private and scoped package dependencies. They really did an amazing job!

laurentdesmet:
Just tried it and I was blown away with the performance improvement. An `npm install` in the root of our company's monorepo managed by lerna went from 1m51s to 8s. This is amazing! I can't wait for scoped dependency support, together with support for yarn within lerna. Awesome job and lots of respect for the creators!

tomdale:
I would love to be able to use Yarn in one of my monorepo projects today, because it doesn't have any scoped or private packages. Not sure when the ETA for that landing in Yarn is, but if it's a ways out it would be a bummer to block Lerna support on it. I am losing days of my life to npm at present. 😭

tomdale:
Yarn and Lerna are fantastic by the way. Y'all are killing it.

natew:
We've added it and are using it now in our branch :)

https://github.com/motion/lerna/pull/4

iamchenxin:
Maybe should add an option in lerna.json for choosing npm or yarn?

[...]

PR #605, opened 2017-02-17
Add support for pluggable npm clients

gigabo:
Allows bootstrap to use `yarn` (for example).

Command-line: `lerna bootstrap --npm-client=yarn`

In lerna.json: `"npmClient": "yarn"`

Relates to https://github.com/lerna/lerna/issues/371.

Pretty small change with the groundwork laid by https://github.com/lerna/lerna/pull/576.


README.md
```diff
@@ -643,6 +643,24 @@ can be used to opt out of hoisting for certain dependencies.
 $ lerna bootstrap --hoist --nohoist=babel-*
 ```
 
+#### --npm-client [client]
+
+Install external dependencies using `[client] install`.  Must be an executable
+that knows how to install npm dependencies.
```
doug-wade: why only install?

wtgtybhertgeghgtwtg:
Shouldn't it be called `package-manager` instead of `npm-client` or something?

gigabo:
> Shouldn't it be called `package-manager` instead of `npm-client` or something?

It's the client that Lerna uses to connect to the npm registry.

NPM yarn and the like ["third-party registry clients"](http://blog.npmjs.org/post/151660845210/hello-yarn).

README.md
gigabo: Because that's the only command it's used for. 😁 

Maybe better to say "Must be a compatible npm registry client"?

wtgtybhertgeghgtwtg:
Does `lerna` specify the `npm` registry?  I think `yarn` tries to pull from its own registry, first.

timdp:
Wasn't there talk of folding monorepo functionality into yarn at one point? That'd make this PR obsolete, right?

karlhorky:
> Wasn't there talk of folding monorepo functionality into yarn at one point?

Yep, here https://github.com/lerna/lerna/issues/371#issuecomment-279339181

hzoo:
@timdp Yep! That requires a rewrite of Lerna into Yarn and would likely take more than a little while. For everyone that wants to continue using Lerna in the meantime there's is and why we continue to work on in the first place

timdp:
Interesting, thanks! :+1:

rexxars:
Want. What's the blocker for getting this merged?

neeschit:
This is awesome!! Thanks a ton!

hulkish:
+1

Issue #567, opened 2017-02-02
force publish for prereleased packages


spudly:
My packages have to go through a QA process before published versions are deployed to production, so I create prerelease versions which are used during the QA process (for example, 1.1.0-rc.0, 1.1.0-rc.1. 1.1.0-rc.2, etc.). When I am satisfied that version is bug-free, I want to publish it with a non-prerelease version before deployment to production. When lerna publishes the packages, I want each prereleased package to be updated to the same non-prerelease version. Unfortunately, lerna only looks at the changes between tags, so lerna thinks there is nothing to update and doesn't update the package versions.

It'd be nice if there was a `--force-publish-prereleased` flag so that I could force all prerelease versions to be bumped to non-prereleased versions.

For example, assuming I had this setup and there were no changes since the last release:
```
lerna.json: 1.0.0-rc.1
├── pkg-a@1.0.0
├── pkg-b@1.0.1-rc.1
├── pkg-c@1.0.1-rc.0
├── pkg-d@1.0.0
```

I would expect it to bump the versions as follows:
```
lerna.json: 1.0.1 <-- updated
├── pkg-a@1.0.0 <-- not updated
├── pkg-b@1.0.1 <-- updated
├── pkg-c@1.0.1 <-- updated
├── pkg-d@1.0.0 <-- not updated
```

Right now I can workaround by using `--force-publish` but that republishes lots of packages unnecessarily, so it's not ideal.

I should also note that I'm publishing from a CI server (jenkins) so my TTY is not interactive; I can't selectively choose versions for each package. I'm publishing using `lerna publish --repo-version $VERSION --yes`

I'm willing to work on a pull request if you can give me some assurance that a feature like this would be accepted. Also, I'm not married to `--force-publish-prereleased` so feel free to bikeshed all you want.

evocateur:
I too will soon enter the brave new world of auto-publishing from Jenkins, so I am sympathetic to this request. #607 is close to ready, but it doesn't completely cover your case here (since you don't want to increment _all_ versions by a given semver keyword).

It sounds like we need a flag similar to `--scope` (although ugh the name) that is also complementary to `--cd-version`....

cif:
I'd suggest simply supporting more --cd-version values a la https://github.com/sindresorhus/np#usage

e.g. `lerna publish --yes --cd-version=prerelease`

spudly:
Perhaps named scope groups?

For example,
```bash
[...]

Issue #461, opened 2016-12-22
publish to latest using CI 


bramthedev:
Hello,
as far as I understand it is impossible to publish to npm as the latest version using a CI script.
I'm currently using this script, which is run on the server:

lerna publish -c --yes --repo-version `grep -Po '(?<="version":\s")[^"]*' package.json`

this takes the version of package.json. But npm automatically sets it default to a previous version, even when the version is incremented by 0.0.1

npm install 'package' does not return the one that was succesfully pushed, even though it exists in npm when asking for that specific version.

I'm sorry if this is not the correct place to discuss this issue, I've never opened an issue on github before, this is my second month employed in IT :)

gigabo:
Yeah, it's not really straightforward to automatically publish a new non-canary version right now.

Something like [what `np` does](https://github.com/sindresorhus/np#usage) might be nice:

```
    Version can be:
      patch | minor | major | prepatch | preminor | premajor | prerelease | 1.2.3
```

It _looks_ like the grep you're using to produce your `--repo-version` is actually returning the _current version_?  It may be that you need to increment the patch level on that to get what you're looking for.  Also the `-c` (`--canary`) option you're including prohibits Lerna from tagging your newly published version as "latest", so npm won't provide it by default.

gigabo:
> I've never opened an issue on github before, this is my second month employed in IT :)

Welcome! 😀

kristianmandrup:
Would be nice with a short paragraph/recipe on a typical setup for CI/CD. I guess it would just be publish without pushing to npm, triggering the CI from the git repo? or is there a smarter way?

Rowno:
I got CI publishing working on CircleCI using `lerna publish --skip-npm`, this script:
```js
#!/usr/bin/env node
'use strict'
const path = require('path')
const fs = require('fs')
const execa = require('execa')

const TAG_REGEX = /[^@]+@[0-9]+(\.[0-9]+)*/

const tag = process.argv[2]

if (typeof tag !== 'string') {
  console.error('Tag required')
  process.exit(1)
}

if (!TAG_REGEX.test(tag)) {
  console.error('Invalid tag')
  process.exit(1)
}

const [packageName, packageVersion] = tag.split('@')

const packagePath = path.join(__dirname, '..', 'packages', packageName)
const packageJsonPath = path.join(packagePath, 'package.json')

if (!fs.existsSync(packageJsonPath)) {
  console.error('Package doesn՚t exist')
  process.exit(1)
}

[...]

PR #607, opened 2017-02-17
Increment version by semver keyword with --cd-version flag

cif:
We do automated code analysis in our deploy pipelines and have a new Lerna repo which needs to operate in independent version mode. Currently the only way to automate publish (no prompts) from build server CL is to specify `--repo-version=x.x.x --yes` (or canary), however this forces all modules into that version. 

This PR allows setting up any changed "next" version(s) appropriately via root package.json / publish script `--repo-version=[patch|minor|major]`, or using `--repo-version=current` and pre setting the next value(s) in each module.

When code is approved and subsequently merged, our build pipelines can use the right versions and deploy to our private npm repository.


cif:
FYI Tests on appveyor are failing due to differences in NT/*nix. e.g. \r\n  vs \n
```
Call 6 was expected to be: ChildProcessUtilities.execSync("git commit -m \"$(echo \"Publish\n\n - package-1@1.0.1\n - package-2@2.0.1\n - package-3@3.0.1\n - package-4@4.0.1\")\"", ...), but was actually: ChildProcessUtilities.execSync("git commit -m \"$(echo \"Publish\r\n\r\n - package-1@1.0.1\r\n - package-2@2.0.1\r\n - package-3@3.0.1\r\n - package-4@4.0.1\")\"", ...)
```
Haven't encountered that before, passing locally

test/PublishCommand.js
```diff
@@ -204,6 +204,85 @@ describe("PublishCommand", () => {
         }
       }));
     });
+
+    it("should use semver increments when passed to repoVersion flag", (done) => {
```
cif: Just a suggestion, but feel it would be worth refactoring these tests to smaller assertion blocks / calls rather than the larger fixture > result chunks. CLI tools with lots of arg permutations is hard to test all paths

gigabo:
Oh, wow.  This is great!  Thanks @cif!

I think this addresses https://github.com/lerna/lerna/issues/461.

src/commands/PublishCommand.js
```diff
@@ -151,6 +151,20 @@ export default class PublishCommand extends Command {
...
+          this.flags.repoVersion === "current"
+      ) {
+        const versions = {};
+        this.updates.forEach((update) => {
+          versions[update.package.name] = this.flags.repoVersion === "current" ?
+            update.package.version : semver.inc(update.package.version, this.flags.repoVersion);
+        });
+
+        return callback(null, { versions });
+      }
```
gigabo: Hmm... actually I wonder if `--repo-version` is the right option for this.  That's really meant specifically for non-independent versioning.  And actually looking at this again it will only work correctly for independently versioned repos.  It won't update the `version` in lerna.json unless the return value from `getVersionsForUpdates` includes a `version` (singular) value.

It might actually be better to add a _new_ option that works for both independent and non-independent repos.

test/PublishCommand.js
gigabo: Sounds interesting, but not sure I understand clearly.  Can you give an example of how you'd like to see it organized?

src/commands/PublishCommand.js
cif: I had the same thoughts :) This was the easiest point of interception to get this working locally without introducing much new functionality. Stoked maintainers are willing to help guide this one, we really need it. Any suggestions on independent flag name? I feel it should be fairly straight forward within the same method to intercept and alter the test. I'll get back to you with more suggestions on that first thing next week. As I stated in the comment, definitely tough to get all paths (especially the inverse coverage) when you have many options

cif:
I made all the changes requested and made the corresponding tests match more closely with your current pattern. Unfortunately, I can't figure out why Appveyor is upset. I changed the `ci` script to remove all the verbose logging to see if I could isolate any one problematic test. All tests are working locally, Travis is happy but Appveyor is timing out: 
```
182 passing (38s)
  1 failing
  1)  "after all" hook:
     Error: timeout of 5000ms exceeded. Ensure the done() callback is being called in this test.
      at done (C:\projects\lerna\node_modules\mocha\lib\runnable.js:285:13)
      at callFn (C:\projects\lerna\node_modules\mocha\lib\runnable.js:344:7)
      at Hook.Runnable.run (C:\projects\lerna\node_modules\mocha\lib\runnable.js:319:7)
      at next (C:\projects\lerna\node_modules\mocha\lib\runner.js:298:10)
      at Immediate.<anonymous> (C:\projects\lerna\node_modules\mocha\lib\runner.js:320:5)
```

What's interesting, is that there are 182 tests currently, so the mysterious fail doesn't seem to exist at all?

package.json
```diff
@@ -9,8 +9,8 @@
     "dev": "babel -w src -d lib",
     "lint": "eslint src test",
     "fix": "eslint src test --fix",
-    "test": "mocha -t 5000",
-    "ci": "npm run lint && cross-env DEBUG_CALLS=true npm run test",
+    "test": "mocha -t 20000",
+    "ci": "npm run lint && cross-env npm run test",
```
cif: Had to increase timeout for appveyor tests to pass (see main conversation thread for error there). Also removed DEBUG_CALLS to make test output readable. If you want it restored, happy to do as part of this PR

gigabo: Seems fine to me.  We removed it in Asini, too.

We could actually also remove the `cross-env`, too, since that was just for getting the environment variable set consistently across environments.  Then we could drop the dep, I think.

@hzoo @evocateur @doug-wade Any objection to dropping this?

src/commands/PublishCommand.js
```diff
@@ -150,12 +150,38 @@ export default class PublishCommand extends Command {
...
+        const versions = {};
+        this.updates.forEach((update) => {
+          versions[update.package.name] = this.flags.cdVersion === "current" ?
+            update.package.version : semver.inc(update.package.version, this.flags.cdVersion);
+        });
+
+        // Use the cdVersion flag to bump the global version as well
+        const version = this.flags.cdVersion === "current" ?
+          this.globalVersion : semver.inc(this.globalVersion, this.flags.cdVersion);
+        return callback(null, { versions, version });
```
gigabo: Looking at the caller again I think in independent mode `version` (singular) is expected _not_ to be set.  So this should _either_ set `versions` (independent) _or_ set `version` (not).  Kind of a weird interface. 😖

cif: I was hesitant to send both to callback in the deconstructed object, however in independent mode `updateVersionInLernaJson` method isn't called due to execute:
```
if (!this.repository.isIndependent() && !this.flags.canary) {
        this.updateVersionInLernaJson();
      }
``` 
I did local tests switching between both and publishing to our local npm repo. I also added both normal and independent modes to the test suite.

cif: If it's more clear, we can use another `!this.repository.isIndependent()` check here and send either-or to callback

cif: I went ahead and made the change to `getVersionsForUpdates()` so the object passed is explicitly `version` or `versions`

src/NpmUtilities.js
```diff
@@ -76,21 +76,21 @@ export default class NpmUtilities {
   }
 
   @logger.logifySync()
-  static addDistTag(packageName, version, tag, registry) {
+  static addDistTag(directory, packageName, version, tag, registry) {
```
cif: I was getting odd NPM ERR on the final publish step in my build and realized because our modules are all @scoped/module in the package.json but simply `module` in the directory that tagging would only work if you cd into the module before tagging

evocateur: How about specifying the [`cwd` option](https://nodejs.org/dist/latest-v6.x/docs/api/child_process.html#child_process_child_process_execsync_command_options) for `execSync` rather than the manual `cd`?

src/commands/PublishCommand.js
```diff
@@ -155,6 +155,34 @@ export default class PublishCommand extends Command {
...
+        // Warn about possible problems with git and current
+        if (this.flags.cdVersion === "current" && !this.flags.skipGit) {
+          this.logger.info("WARNING! Using --cd-version=current without --skip-git can result in commits without changes and cause errors");
+        }
+        // If the version is independent then send versions
+        if (this.repository.isIndependent()) {
+          const versions = {};
+          this.updates.forEach((update) => {
+            versions[update.package.name] = this.flags.cdVersion === "current" ?
+              update.package.version : semver.inc(update.package.version, this.flags.cdVersion);
```
evocateur: Can you put the ternary operators first, please? It is much easier to read, in my experience.

```
versions[update.package.name] = (this.flags.cdVersion === "current")
  ? update.package.version
  : semver.inc(update.package.version, this.flags.cdVersion);
```

src/commands/PublishCommand.js
```diff
@@ -155,6 +155,34 @@ export default class PublishCommand extends Command {
...
 
   getVersionsForUpdates(callback) {
+    if (this.flags.cdVersion) {
+      // Allows automatic bumping to next semver via cdVersion flag
+      if (this.flags.cdVersion === "patch" ||
+          this.flags.cdVersion === "minor" ||
+          this.flags.cdVersion === "major" ||
+          this.flags.cdVersion === "current"
+      ) {
+        // Warn about possible problems with git and current
```
evocateur: What is the use-case for `--cd-version current`?

cif: Kind of a long story having to do with our deployment pipeline but we came to the conclusion it was a better strategy to always version. Removing the option in latest changes

src/commands/PublishCommand.js
cif: Ternaries no longer needed due to current version removed

src/NpmUtilities.js
cif: I changed `getTagOpts` signature to accept directory in addition to registry. `publishTaggedInDir` was already using the same `cd ... && npm ...` method. I updated this so that all methods are now using your suggestion

src/commands/PublishCommand.js
```diff
@@ -155,6 +155,34 @@ export default class PublishCommand extends Command {
...
+        if (this.flags.cdVersion === "current" && !this.flags.skipGit) {
```
cif: Warning goes away due to current removal as well.

package.json
```diff
@@ -10,7 +10,7 @@
     "lint": "eslint src test",
     "fix": "eslint src test --fix",
     "test": "mocha -t 20000",
-    "ci": "npm run lint && cross-env DEBUG_CALLS=true npm run test",
+    "ci": "npm run lint && cross-env npm run test",
```
evocateur: I don't think we want to remove `DEBUG_CALLS=true`, here

gigabo: Previous thread: https://github.com/lerna/lerna/pull/607#discussion_r102284105

evocateur: Whoops, nevermind then. I don't have a dog in the hunt, just was concerned about PR scope.

nutgaard:
Is there anything stopping this from getting merged at this point? Anything I can help with? 
I would really really like to see this feature in lerna. :)

cif:
There's a bit of talk in #567 about how to support the other semvers e.g. `prerelease` as well as release phase/scope `(rc | beta | ...)`. Should all be fairly straight forward updates to PublishCommand.js, IMO wouldn't have to be part of this PR but that's up to the maintainers :)

omerxx:
Hey @cif / @evocateur,
This feature would help us a lot, saving time and effort, and solve the problem very neatly.
Is there a forecast on when this should be ready and merged into lerna?

Thank you guys for the hard work!

evocateur:
@gigabo Are you satisfied with this now?

I'll look into the conflicts today.

gigabo:
Yep, looks good.  Thanks @cif!

cif:
@evocateur Just took a swing at resolving them, waiting on tests will update if they fail master was still assuming that call exec were using `"cd ..."` whereas the changes you suggested eliminated that need (use cwd opt)

evocateur:
@cif thanks, I appreciate it. I'm also in the process of rebasing, and so far I've found that upstream changes make it really tricksy...

evocateur:
in `test/PublishCommand.js`, the `expectedOpts` stuff should really be `expectedEnv`, allowing the `cwd` to change correctly in the stub assertions. I'm almost there.

omerxx:
Wow! You guys are awesome... Thank you for this!

cif:
@evocateur sounds like we may have overlapped but I just pushed up fixes for those tests

evocateur:
Thanks @cif !

bradbarrow:
I'd love to better understand how this is used in your CD process @cif 

If I understand, you'd have to run `lerna publish --yes --cd-version=minor` for instance which would then force all of the packages to be bumped a minor bump?

Is that the behaviour? 

What happens if in a single CI build I want one package to be published in a major increment and one in a minor?

cif:
@bradbarrow We only increment the same value for each package.  In our case we only use independent versioning because our starting points were already different which is why I created the PR.  

To handle the use case you describe it would require code changes as well as a specification for each increment on a package level (perhaps in each `package.json`) which seems tedious to maintain especially if you have a lot of packages.

On a higher/philosophical level, it's worth asking the question - if your packages are versioning semantically independent of each other, are they really fit to be contained in a mono-repo project or are they independent modules at that point? 

Generally, most of the teams I see who have the choice of lockstep vs independent (i.e. Babel https://github.com/babel/babel/blob/7.0/lerna.json) will choose lockstep if possible.  

Hope that helps!

PR #608, opened 2017-02-18
Add a --stream option to the run command

### LINKED ISSUE #446: Show immediate log output of npm scripts
Logs are currently *only* displayed after a successful command (like `lerna run test`). This means scripts like `test:watch` or `build:watch` won't work well.

@gigabo mentioned:
> The reason it all comes out at once at the end is to avoid having output from different packages interleaved.
> Might be nice to have an option to turn on immediate output, maybe with a prefix of package name so the interleaving isn’t so inscrutable?

Is this something worth exploring? I'd be happy to help.

gigabo:
Emit output from child processes immediately.

- Each line is prefixed with the name of the originating package.
- Output from stdout goes to stdout, and stderr goes to stderr.

```sh
$ lerna run watch --stream
```

This closes #446.

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -46,19 +46,7 @@ export default class ChildProcessUtilities {
   static spawn(command, args, opts, callback) {
     let output = "";
 
-    const childProcess = ChildProcessUtilities.registerChild(
-      spawn(command, args, Object.assign({
-        stdio: "inherit"
-      }, opts))
-        .on("error", () => {})
-        .on("close", (code) => {
-          if (code) {
-            callback(`Command exited with status ${code}: ${command} ${args.join(" ")}`, output);
-          } else {
-            callback(null, output);
-          }
-        })
-    );
+    const childProcess = _spawn(command, args, opts, (err) => callback(err, output));
 
     // By default stderr, stdout are inherited from us (just sent to _our_ output).
     // If the caller overrode that to "pipe", then we'll gather that up and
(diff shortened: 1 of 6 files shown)


leonardorb:
That's great! This will be very helpful.

src/ChildProcessUtilities.js
```diff
@@ -74,6 +62,35 @@ export default class ChildProcessUtilities {
     }
   }
 
+  static spawnStreaming(command, args, opts, prefix, callback) {
+    opts = Object.assign({}, opts, {
+      stdio: ["ignore", "pipe", "pipe"],
```
MoOx: Hi. Sorry to bother you on such an old PR but I am curious about the motivation for ignoring stdin? I understand that it could not really make sense in some case (that I don't have in mind - I am not an stdin/out professional), but I am facing [an issue](https://github.com/lerna/lerna/issues/1745) and switching `ignore` with `pipe` or `inherit` seems to change nothing in most case but could help me with my issue.

evocateur: I think the biggest problem with `"inherit"` is in concurrent execution: which process "gets" to have `stdin`, should it be necessary? Imagine 8 concurrent `npm` executions demanding a one-time password, for example.

I _am_ sympathetic to this, but I think we would need an option, like `--inherit-stdin`, to explicitly enable this in both `lerna exec` and `lerna run`. Open questions:

* Should `--inherit-stdin` imply `--stream`?
* Should `--inherit-stdin` be allowed if `--parallel` is enabled?

There are two options, as I see it, to avoid the interleaved/conflicting `stdin` problem:

1. `--inherit-stdin` implicitly restricts concurrency to `1`
2. Recommend `--concurrency 1` in docs, warn loudly when `this.concurrency !== 1`

Issue #592, opened 2017-02-09
`lerna run --scope my-package fail` prints same error three times


nfcampos:
With this in `packages/my-package/package.json`
```
{
  "scripts": {
    "fail": "exit 1"
  }
}
```
running `lerna run --scope my-package fail` outputs
```
Lerna v2.0.0-beta.36
Scoping to packages that match 'trainline-desktop'
NpmUtilities.execInDir        ("run fail", [], "./packages/trainline-desktop")
Error: Command failed: npm run fail 

npm ERR! Darwin 16.4.0
npm ERR! argv "/Users/nunoc/.nvm/versions/node/v6.9.1/bin/node" "/Users/nunoc/.nvm/versions/node/v6.9.1/bin/npm" "run" "fail"
npm ERR! node v6.9.1
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! trainline-desktop@0.0.0 fail: `exit 1`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the trainline-desktop@0.0.0 fail script 'exit 1'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the trainline-desktop package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     exit 1
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs trainline-desktop
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls trainline-desktop
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/nunoc/dev/trainline-desktop/packages/trainline-desktop/npm-debug.log

    at ChildProcess.exithandler (child_process.js:206:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Socket.<anonymous> (internal/child_process.js:334:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at Pipe._handle.close [as _onclose] (net.js:498:12)

> trainline-desktop@0.0.0 fail /Users/nunoc/dev/trainline-desktop/packages/trainline-desktop
> exit 1


NpmUtilities.runScriptInDir   ("fail", [], "./packages/trainline-desktop")
[...]

PR #614, opened 2017-02-21
Fix (some) error repetition when running failing `npm run` script

nfcampos:
see discussion at #592 
see commit messages for details of changes

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -90,13 +90,8 @@ export default class NpmUtilities {
   }
 
   @logger.logifyAsync()
-  static execInDir(command, args, directory, callback) {
-    ChildProcessUtilities.exec(`npm ${command} ${escapeArgs(args)}`, { cwd: directory, env: process.env }, callback);
-  }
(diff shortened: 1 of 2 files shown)


nfcampos:
@evocateur  done

evocateur:
@nfcampos Thanks! Now to investigate why the Travis build failed. -_-

nfcampos:
@evocateur yeah im really not sure, it might be because of a unhappy combination of
  - jest running test files in parallel
  - jest re-using workers to do so
  - lerna tests relying on `process.pid` for temp path names
making it so on rare cases they somehow conflict with each other and two concurrent tests end up touching the same temp path

evocateur:
Yeah, it probably was naïve of me to use the `process.pid`. I'll see what I can do to make that more robust. ([`yarn`'s method](https://github.com/yarnpkg/yarn/blob/master/src/util/fs.js#L761) seems pretty safe)

nfcampos:
yes it does seem pretty safe

Issue #293, opened 2016-08-02
[Support] npm dist-tag not available on Nexus, looking for workaround


MaartenVerbaarschot:
Hi,

We have a private NPM registry at our company, using Nexus. They don't support npm dist-tag yet ([NEXUS-9862](https://issues.sonatype.org/browse/NEXUS-9862)). Since Nexus is pretty common for private NPM registries I imagine other people are also getting blocked by this. 

Of course, ideally Nexus introduces npm dist-tag soon and this issue would be resolved. But in the meantime, I would like to check if there is a workaround :). 

When I run `lerna publish` I get the following error:

```
Error updating version as latest
Error: Command failed: npm dist-tag ls mypackagename
npm ERR! registry error parsing json
npm ERR! dist-tag ls Couldn't get dist-tag data for mypackagename
```

I've looked a bit at the [usages of npm dist-tag in the source code](https://github.com/lerna/lerna/blob/49e2ade19cc76b1fc65badfdbd87aef151edd297/src/commands/PublishCommand.js#L394-L404), and because they're in a try/catch block, the publishing still succeeds. However, my Jenkins build will be marked as 'broken' because there are errors.

I was wondering if the npm dist-tag is strictly needed to work with Lerna? It only marks the published packages as the latest, right? So, am I correct that we could work around this error by working with exact version numbers in all our package.json files (thus not depending on what is marked as the 'latest')?

And if the above is correct, would you be open to introducing an option for lerna to skip the use of dist-tag? (e.g. `lerna publish --skip-npm-dist-tag`)

dliu120:
Try putting a .npmrc file pointing to your private NPM repo -- this solved it for me.

MaartenVerbaarschot:
dliu120: The private repo in the .npmrc is what's causing the error in the first place, as it is our private repo that is not supporting the dist-tag.

Which software are you using for your private repo? (Not Nexus I presume? :))

dliu120:
We set up our own private NPM hosted on Azure.

It says [here](http://larrymyers.com/tech/2016/06/05/private-npm-with-nexus.html) that Nexus provides full functionality of NPM, so I'm surprised that dist-tag does not work.

timche:
Nexus is working really well with Lerna except the `npm dist-tag` thing :/ Looking for a workaround as well.

timche:
I've just tried out Lerna with [sinopia](https://github.com/rlidwka/sinopia) and everything is working fine 👍

[...]

PR #620, opened 2017-02-24
Feature: skip-temp-tag

noherczeg:
I have no idea if this is a proper solution, therefore I'd like to have some guidance on how to properly test this.

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -351,7 +351,8 @@ export default class PublishCommand extends Command {
   npmPublishAsPrerelease(callback) {
     const {skipTempTag} = this.getOptions();
     // if we skip temp tags we should tag with the proper value immediately therefore no updates will be needed
-    const tag = !skipTempTag ? "lerna-temp" : this.getDistTag();
+    const tag = skipTempTag ? this.getDistTag() : "lerna-temp";
+
     this.updates.forEach((update) => {
       this.execScript(update.package, "prepublish");
     });
@@ -398,6 +399,11 @@ export default class PublishCommand extends Command {
 
   npmUpdateAsLatest(callback) {
     const {skipTempTag} = this.getOptions();
+
+    if (skipTempTag) {
+      return callback();
+    }
+
     this.progressBar.init(this.packageToPublishCount);
 
     PackageUtilities.runParallelBatches(this.batchedPackagesToPublish, (pkg) => (cb) => {
@@ -407,10 +413,7 @@ export default class PublishCommand extends Command {
         attempts++;
 
         try {
-          if (!skipTempTag) {
-            this.updateTag(pkg);
-          }
-
+          this.updateTag(pkg);
           this.progressBar.tick(pkg.name);
           cb();
           break;
@@ -431,7 +434,6 @@ export default class PublishCommand extends Command {
   }
 
   updateTag(pkg) {
-    // we didn't skip temp tagging so we need to handle the situation accordingly
     const distTag = this.getDistTag();
 
     if (NpmUtilities.checkDistTag(pkg.name, "lerna-temp", this.npmRegistry)) {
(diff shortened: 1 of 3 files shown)


noherczeg:
In response to #293

src/commands/PublishCommand.js
```diff
@@ -361,29 +362,35 @@ export default class PublishCommand extends Command {
...
-            this.logger.error("Attempting to retry publishing " + pkg.name + "...", err);
-            run(cb);
-          } else {
-            this.logger.error("Ran out of retries while publishing " + pkg.name, err);
-            cb(err);
-          }
-        });
+        if (skipTempTag) {
+          this.progressBar.tick(pkg.name);
+          this.execScript(pkg, "postpublish");
```
gigabo: Where is the actual publish happening with `--skip-temp-tag`?

noherczeg:
So the idea is that
If we ignore temp tagging then:
- `npmPublishAsPrerelease()` should immediately set the proper tag value
- and `npmUpdateAsLatest()` should skip the `updateTag()` procedure which handles temp tag removal, and addition of the proper tags

If we did not ignore temp tagging then:
- `npmPublishAsPrerelease()` will publish with the "lerna-temp" tag
- while `npmUpdateAsLatest()` will trigger the `updateTag()` procedure...

noherczeg:
I also introduced the `getDistTag()` method which might be an overkill, so if you guys feel like it I could remove it. I made it because as it stands in the code I kind of needed something like that in multiple places.

src/commands/PublishCommand.js
```diff
@@ -394,6 +397,7 @@ export default class PublishCommand extends Command {
   }
 
   npmUpdateAsLatest(callback) {
+    const {skipTempTag} = this.getOptions();
```
evocateur: If we've skipped temp-tagging, why not just avoid the `runParallelBatches` call entirely?

```
const {skipTempTag} = this.getOptions();

if (skipTempTag) {
  return callback();
}

this.progressBar.init(...)
```

src/commands/PublishCommand.js
```diff
@@ -433,4 +429,26 @@ export default class PublishCommand extends Command {
...
+    } else if (this.flags.canary) {
+      NpmUtilities.addDistTag(pkg.name, pkg.version, distTag, this.npmRegistry);
+    } else {
+      NpmUtilities.addDistTag(pkg.name, this.updatesVersions[pkg.name], distTag, this.npmRegistry);
+    }
+  }
+
+  getDistTag() {
+    const {npmTag, canary} = this.getOptions();
+    return npmTag ? npmTag : canary ? "canary" : "latest";
```
evocateur: I really dislike nested ternaries, especially all on one line. Wouldn't boolean defaults work here?

```
return npmTag || (canary && "canary") || "latest";
```

src/commands/PublishCommand.js
```diff
@@ -349,6 +349,9 @@ export default class PublishCommand extends Command {
   }
 
   npmPublishAsPrerelease(callback) {
+    const {skipTempTag} = this.getOptions();
+    // if we skip temp tags we should tag with the proper value immediately therefore no updates will be needed
+    const tag = !skipTempTag ? "lerna-temp" : this.getDistTag();
```
evocateur: Let's invert the ternary (`const tag = skipTempTag ? this.getDistTag() : "lerna-temp";`) and make sure there's a newline between this assignment and the next `forEach` block.

src/commands/PublishCommand.js
noherczeg: Yeah, I missed that. Corrected int #0c6ac7e

evocateur:
LGTM, thanks for your responsiveness @noherczeg!

If @gigabo has no objections, I will merge this tomorrow.

PR #625, opened 2017-02-28
QOL, nvm descriptor

noherczeg:
We've been using [nvm](https://github.com/creationix/nvm) for quite a while now. If you guys deem it worthy this might come handy for others as well :)

This allows the developers to not care about tracking down what version does Lerna require. We just need to type `nvm use` in the repo root and if it's available it'll set it up, while if it's missing it tells us what to install with `nvm install ${version}`.

That's all.


noherczeg:
@evocateur  I see. You should reject this then! :)

evocateur:
@noherczeg I appreciate the suggestion!

PR #628, opened 2017-03-01
changelog [skip ci]

hzoo:
Want to we want to put here to explain the issue/how it's solved?

cc @Gongreg, @spudly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,12 @@ $ lerna bootstrap --npm-client=yarn
 
 Very important fix for Babel that we used in the last release. This prevents a timing issue when publishing where a module will try to download a package that isn't published yet because it is published before it's own dependency is published itself. We used to get many issues from users on non-public npm about "babel-types" not being found.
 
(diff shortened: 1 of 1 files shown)


spudly:
I would just add the following:

This prevents unchanged packages from being republished when running `lerna publish` from a branch, which you might do if you're publishing a prerelease, or publishing a bug fix in an earlier release.

Gongreg:
@spudly Added missing information. Don't have what else to add.

Also I didn't have a chance to thank you @spudly. So, Thank you! :)

evocateur:
I rewrote the commit message when I merged it, I'm not sure why the changelog generator didn't pick that up: https://github.com/lerna/lerna/commit/d246667b9b8513d2c3ac14a796d9dc04126a1294

> We now check for changes since the most recent tag in the current branch, instead of the most recent tag in entire repository. This allows publishing older versions of a project in maintenance branches, as well as nightly releases from a feature branch.
>
> Additionally, we now ensure that the user is in a non-detached branch because we can't publish without a valid git branch.

hzoo:
Yeah probably a bug - free free to edit/merge

evocateur:
whoops, forgot `[skip ci]` 😿

PR #632, opened 2017-03-01
add --packageName parametr. Now we can pass package name to deploy

Nik5036350:
Patch to have an ability to specify package what we want to push.

To publish specific package you need to pass --packageName paramentr. For instance: lerna publish --packageName @platform/carousel-gallery


evocateur:
This is already possible by using a combination of `--force-publish` and `--ignore`:
```
lerna publish --force-publish=@platform/carousel-gallery --ignore='**'
```

Lerna is designed to automate the publish decision, and the complicated nature of supporting your use case is intentional.

Issue #449, opened 2016-12-15
feature request: watching


jeffwhelpley:
There are various watching tools like the one for TypeScript (i.e. `tsc -w`). This doesn't seem to work with `lerna run`. It would be awesome if you could essentially do `tsc -w` (or any other similar watching tool) on each package at the same time.

gigabo:
I think https://github.com/lerna/lerna/issues/446 is the main thing standing in the way of this.

trotzig:
I'm not sure #446 solved it completely. I have a repo with a few packages, all with a `build:watch` script. Most packages depend on the `core` package. When I run `lerna run --stream build:watch` I see output for the core package, but none of the others. This is likely because they are waiting for the `core` package to finish, which it never will in watch mode. 

EDIT: I just found the `--no-sort` option, which resolves the dependency lock that I ran into. Posting this here anyway in case others find it useful.

evocateur:
@trotzig This could definitely use some README documentation, agreed.

scottmas:
--no-sort doesn't seem to fix every case. For instance I have 5 packages in my repo that implement a `dev` script. Only 2 of them run with `--stream` only but only 4 of the 5 run even with `--stream --no-sort`. Not sure what else I should do. The command in the offending package executes just fine when I run it manually.

gigabo:
@scottmas Lerna has a default concurrency of 4.  You can increase that with the `--concurrency` option.

scottmas:
Okay that makes sense. It seems a bit sub-optimal that to get sane "watch" behavior in any fair sized mono repo we need to implement three flags: `--stream --no-sort --concurrency 9999`. Having to specify concurrency limits doesn't make sense in a `--watch` context since the concurrency will always have to be equal or greater to the number of scripts. 

Seems to me like it would be a good idea to roll all this behavior into a `--watch` flag or something of the like.

scottmas:
Or make `--stream` set concurrency to number of threads and set `--no-sort`? Also an option.

evocateur:
I like the idea of a synthetic `--watch` flag more than overloading `--stream`.

reggi:
What's the current status with this issue? I want to watch multiple lerna packages at once to build each individually with each packages `npm run build:watch` script.

evocateur:
`lerna run --parallel build:watch`

Dombo:
[...]

PR #634, opened 2017-03-02
Add documentation about "watch" commands next to --no-sort

trotzig:
I ran into an issue where `lerna run build:watch` ("build:watch" being a
babel command with the `--watch` flag) would only run for one package
(called "core"). It turns out to be caused by lerna running commands in
a topologically sorted order, executing packages that others depend on
first before moving on.

The nature of "watch" commands is usually to never end. In my case that
meant that lerna was waiting indefinitely for the "core" package command
to finish.

By reading the documentation I came across the `--no-sort` flag, which
helped resolve my issue. It wasn't super clear in the README that this
flag would help out with my problem, so I decided to add a little bit
about "watch" commands right next to it.

Relevant discussion:
https://github.com/lerna/lerna/issues/449#issuecomment-283463490


scottmas:
This bit me hard until I found this. The --stream flag seems almost worthless with the --no-sort option not set.

PR #635, opened 2017-03-02
Switch to Jest

evocateur:
Jest rocks my socks. It's faster than `mocha`+`nyc`, and more accurate in its coverage report.

The `--watch` mode is a legit dream. (`yarn test -- --watch`)

Non-scientific comparisons:

| | mocha | jest |
|-| -----: | ---: |
| `yarn test`                | 15s | 7s |
| `yarn run coverage` | 20s | 8s |

image

You'll notice files we've completely forgotten to cover, which are entirely absent from the `nyc` output. Have I mentioned it's fast? 🚀


evocateur:
balls, CI all failed for a variety of reasons 😛

spudly:
+1 on this. When my team switched from AVA to jest our test execution time for a single (large) package went from 5 minutes to 7 seconds.

bcoe:
@evocateur this sounds great. Jest uses the same underlying instrumentation logic as nyc, but doesn't rely on source-maps (we've never quite hammered all the kinks out of the source-maps in babel codebases).

evocateur:
@bcoe Indeed! I'm a big fan of `nyc` and its underlying utilities.

appveyor.yml
```diff
@@ -23,7 +23,7 @@ test_script:
   - node --version
   - npm --version
   - yarn --version
-  - yarn run ci
+  - yarn run ci -- --runInBand
```
evocateur: Without this, the Windows builds have all sorts of impossible-to-debug race conditions with the filesystem operations (fixture copy, etc). _With_ it, it runs super fast (relative to "normal" multi-process Jest runs on Windows).

bin/lerna.js
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env node
+"use strict";
```
evocateur: I got tired of my editor yelling at me with lint errors whenever I opened this file. Easy enough to start linting it automatically, as we no longer support node < v4.

package.json
```diff
@@ -5,12 +5,13 @@
...
-    "coverage": "nyc --reporter=lcov npm run test",
+    "coverage": "jest --coverage",
     "dev": "babel -w src -d lib",
-    "lint": "eslint src test",
-    "fix": "eslint src test --fix",
-    "test": "mocha -t 20000",
-    "ci": "npm run lint && npm run test",
+    "lint": "eslint . --ignore-path .gitignore",
+    "fix": "npm run lint -- --fix",
+    "pretest": "npm run lint -- --cache",
```
evocateur: This is pretty big. Contributors that run tests locally shouldn't have to remember to _also_ `yarn run lint` to avoid silly lint errors breaking their PR builds.

package.json
```diff
@@ -51,6 +53,7 @@
   "devDependencies": {
     "babel-cli": "^6.7.5",
     "babel-eslint": "^6.0.2",
+    "babel-jest": "^19.0.0",
```
evocateur: Now that we're using `yarn` in CI, this is less important. Without it, however, npm 2 (the default for node v4) would fail to run tests because babel-jest wasn't a sibling of jest.

package.json
```diff
@@ -59,10 +62,29 @@
...
-    "nyc": "^10.1.2"
+    "pify": "^2.3.0",
+    "recursive-copy": "^2.0.6"
+  },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.js"
+    ],
+    "modulePathIgnorePatterns": [
+      "<rootDir>/test/fixtures"
```
evocateur: Jest does aggressive pre-caching of package.json files, apparently, and this was seriously harshing the mellow of a CI process when booting up the tests. So we ignore it. 😎

src/FileSystemUtilities.js
```diff
@@ -21,7 +21,7 @@ export default class FileSystemUtilities {
 
   @logger.logifyAsync()
   static mkdirp(filePath, callback) {
-    mkdirp(filePath, callback);
+    mkdirp(filePath, { fs }, callback);
```
evocateur: Passing `graceful-fs` to `mkdirp` is an effort to minimize race conditions when conducting tons of parallel filesystem operations.

(These were especially noticeable because jest by default tries to run as many tests in parallel processes as possible)

src/commands/ImportCommand.js
```diff
@@ -38,7 +38,7 @@ export default class ImportCommand extends Command {
     this.targetDir = path.join(targetBase, externalRepoBase);
 
     try {
-      if (fs.statSync(this.targetDir)) {
+      if (FileSystemUtilities.existsSync(this.targetDir)) {
```
evocateur: This didn't need to be a `statSync` since the result was never used.

test/BootstrapCommand.js
```diff
@@ -35,7 +35,7 @@ describe("BootstrapCommand", () => {
       bootstrapCommand.runPreparations();
 
       bootstrapCommand.runCommand(exitWithCode(0, (err) => {
-        if (err) return done(err);
+        if (err) return done.fail(err);
```
evocateur: Apparently Jasmine uses this pattern, contra Mocha.

test/ImportCommand.js
```diff
@@ -1,30 +1,35 @@
...
-    });
-
-    beforeEach((done) => {
-      externalDir = initExternalFixture("ImportCommand/external", done);
-    });
+    let testDir;
+    let externalDir;
+
+    beforeEach(() =>
+      Promise.resolve()
```
evocateur: This started out as a parallel `Promise.all()`, but the synchronous `process.chdir()` inside `initFixture()`'s promise chain made that non-deterministic, hence the ordering.

test/PackageUtilities.js
```diff
@@ -4,134 +4,186 @@ import path from "path";
...
-  describe(".getPackagesPath()", () => {
-    it("should append the packages path to the repo path given", () => {
-      assert.equal(
-        PackageUtilities.getPackagesPath("/path/to/repo"),
-        path.join("/path/to/repo/packages")
-      );
+  describe("public API", () => {
+    let testDir;
+
+    beforeEach(() => initFixture("PackageUtilities/basic").then((dir) => {
```
evocateur: This fixture used to be initialized for _every_ test in this file, despite over half of them doing another fixture init (for filtering) and ignoring this. Seemed wasteful.

test/PackageUtilities.js
```diff
@@ -180,55 +231,4 @@ describe("PackageUtilities", () => {
       });
     });
   });
-
-  describe(".filterPackages()", () => {
```
evocateur: Somehow we got two blocks of `filterPackages` tests. I condensed them, using the public API instead of the "private" underscore-prefixed method.

test/PublishCommand.js
```diff
@@ -800,18 +797,18 @@ describe("PublishCommand", () => {
...
     };
 
-    beforeEach((done) => {
-      testDir = initFixture("PublishCommand/normal", done);
-    });
+    beforeEach(() => initFixture("PublishCommand/normal").then((dir) => {
+      testDir = dir;
+
+      // mock out the ENV to a simpler version for testing
+      process.env = mockEnv;
```
evocateur: I moved the `process.env` mocking into the tail-end of `beforeEach` because now (thanks to promises) I can be assured that all fixture initialization that needs the "real" env will not be harmed by this mock.

test/helpers/fixtureUtils.js
```diff
@@ -0,0 +1,71 @@
...
+    process.pid,
+    CURRENT_SHA,
+  ];
+
+  if (suffix) {
+    // e.g., "lerna-12345-deadbeef-external"
+    prefix.push(suffix);
+  }
+
+  const tmpDir = path.join(os.tmpdir() || "/tmp", prefix.join("-"));
```
evocateur: Using a real temp directory helps avoid lots of subtle bugs in the tests due to "fall through" into the parent git repo. Also, it's just cleaner, and in some cases even faster (since tempdirs are often located on ramdisks).

test/helpers/initFixture.js
```diff
@@ -0,0 +1,38 @@
+import path from "path";
+import cp from "recursive-copy";
+import {
+  fixtureNamer,
+  getTempDir,
+  gitInit,
+  removeAll,
+} from "./fixtureUtils";
```
evocateur: look ma, code reuse!

evocateur:
@hzoo @gigabo @doug-wade I would appreciate a review, when you have a chance. I'm antsy to get this merged so testing further improvements is more robust.

I realize it's a giant changelist, and I apologize for that in advance. I've gone through the whole thing and put comments on most of the significant bits. The vast majority of the changes were motivated by the multi-process nature of `jest` (when run locally or on Travis).

I _really_ wanted to codemod `assert` into jest's `expect()` syntax, but I'll save that for another time.

I'm prepared to help rebase open PRs should conflicts arise.

evocateur:
I'm getting really tired of flapping tests on AppVeyor like [this one](https://ci.appveyor.com/project/lerna/lerna/build/1.0.704/job/rlhy1yfsjj4j0lin#L428). Running Jest sequentially for Windows builds really helps avoid this dumb environmental problem (while we address the underlying architectural race conditions with a better testing experience).

cpojer:
Hey, wow, thanks for doing this.

evocateur:
One step closer to dogfooding the whole stack :D

JaKXz:
@evocateur this is great, well done! I'm curious about one statement though: 

> more accurate in its coverage report [than `nyc` + `mocha`]

AFAIK `jest` uses a lot of `istanbul`'s libs under the hood for collecting coverage information, especially `babel-plugin-istanbul`. Could you expand / provide examples on how it's more accurate, and I'm hoping that will help us make `nyc` more accurate :) thanks!

evocateur:
@JaKXz Indeed, `jest` does use `istanbul`'s libs! I didn't mean to disparage `nyc` at all, but I do recall significant differences in the coverage display, even _without_ `--all` (or `jest`'s equivalent).

When I spoke to @bcoe about this disparity, I recall he said something about the source maps were causing problems with dynamically-transpiled ES6, or something to that effect. It was a few weeks ago, and I can't remember offhand where the thread was.

More than anything, I think the `jest` transition yielded more accurate coverage reporting because we are no longer hacking `babel-register` into `mocha`, and benefitting from `jest`'s "batteries-included" approach.

bcoe:
@evocateur @JaKXz would love to get to the bottom of the source-map discrepancy; @JaKXz we shoudl make an effort to capture discrepancies when possible, and start to get some failing tests in place on nyc.

Issue #630, opened 2017-03-01
--hoist with yarn removes root packages


jquense:
Running into an issue trying out the new `npmClient` option. From a clean repo (`lerna clean && rm -rf node_modules`) running `yarn install && lerna bootstrap --hoist` results in a broken configuration. From what I can tell, when the packages are hoisted to the root directory all the root dependencies are bulldozed over and removed. As a consequence binaries (like eslint) are missing/unavailable to packages. This seems to be yarn's install behavior

I'm not sure how to debug further but I'd be happy to do what I can with some direction

evocateur:
This is one of the reasons I've never felt especially comfortable with `--hoist`'s behavior of installing the dependencies at the root of the repo (instead of the closest common directory, e.g. `./packages/node_modules` for the default `packages/*` location).

I'm assuming your `lerna.json` looks similar to this:
```json
{
  "lerna": "2.0.0-beta.38",
  "packages": [
    "packages/*"
  ],
  "npmClient": "yarn",
  "version": "0.0.0"
}
```

I've had good luck with a yarn root, `lerna --hoist` (_without_ the new `npmClient` option), **and** fully-specified `dependencies` and `devDependencies` of those hoisted sub-packages in the root. I don't have to use `--npm-client yarn` because `npm install` is never called during bootstrap (which I call from the `postinstall` lifecycle script, so `yarn` has already installed everything in the root).

My approach is obviously a laborious and error-prone process for projects with many (or conflicting) sub-package dependencies. I did it because of my dislike of `--hoist` just dumping stuff in the root regardless of what the root `package.json` specifies. I am in favor of changing the target of `--hoist` to the closest common directory, but also open to other solutions.

gigabo:
>  binaries (like eslint) are missing/unavailable to packages.

Oh, they sure are! 🤦‍♂️ 

That's... unfortunate.

>  I am in favor of changing the target of --hoist to the closest common directory, but also open to other solutions.

What if the closest common directory is the root?

gigabo:
Minimal repro of the problematic Yarn behavior:

```sh
$ cat package.json
{
  "dependencies": {
    "eslint": "^3.16.1"
  }
}
$ npm i rimraf
$ ls node_modules/.bin/
rimraf
$ yarn install
$ ls node_modules/.bin/
eslint
```

It doesn't remove `node_modules/rimraf/`.  It just removes the link in `node_modules/.bin`.

evocateur:
[...]

PR #638, opened 2017-03-02
Install with all dependencies when installing

gigabo:
This makes install all-or-nothing.  If a single dependency is missing then
Lerna will install all dependencies for that location.  This is mostly a noop,
but it's necessary for consistent behavior across npm clients such as Yarn.

It preserves the _true_ noop when there's nothing to install.

This addresses https://github.com/lerna/lerna/issues/630.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -271,9 +271,8 @@ export default class BootstrapCommand extends Command {
           name,
           dependents: (dependents[rootVersion] || [])
             .map((dep) => this.packageGraph.get(dep).package),
-          dependency: this.repository.hasDependencyInstalled(name, rootVersion)
-            ? null // Don't re-install if it's already there.
-            : `${name}@${rootVersion}`,
+          dependency: `${name}@${rootVersion}`,
+          isSatisfied: this.repository.hasDependencyInstalled(name, rootVersion),
         });
       }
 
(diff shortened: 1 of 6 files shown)


src/commands/BootstrapCommand.js
```diff
@@ -312,9 +312,17 @@ export default class BootstrapCommand extends Command {
 
     // Start root install first, if any, since it's likely to take the longest.
     if (Object.keys(root).length) {
+
+      // If we have anything to install in the root then we'll install
+      // _everything_ that needs to go there.  This is important for
+      // consistent behavior across npm clients.
+      const depsToInstallInRoot = root.filter(({isSatisfied}) => !isSatisfied).length
```
evocateur: Isn't this what `Array.prototype.some()` is for? (it short-circuits, too)

```
// reusable utility for both checks
function hasUnsatisfiedDependencies(deps) {
  return deps.some(dep => dep.isSatisfied !== true);
}

const depsToInstallInRoot = hasUnsatisfiedDependencies(root)
  // ...

// further down
if (hasUnsatisfiedDependencies(deps)) {
  actions.push(
  //...
```

gigabo: Ooh, nice!  Thanks, updated. 👍

gigabo:
image

Oh, nearly missed this.  Summary comments in the review box _still_ aren't picked up by GitHub's slack integration, and they appear _above_ the line-level comments with which they're associated, so they're really easy to miss.  I avoid them for this reason.

> Is this impossible to test?

That seems unlikely. 😐

PR #641, opened 2017-03-02
Add prompts for prerelease versions

rtsao:
Added prompts for `prepatch`, `preminor`, `premajor`, and `prerelease` which correspond to the same functionality in `npm version`

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -216,7 +216,7 @@ export default class PublishCommand extends Command {
     const prepatch = semver.inc(currentVersion, "prepatch");
     const preminor = semver.inc(currentVersion, "preminor");
     const premajor = semver.inc(currentVersion, "premajor");
-    const prerelease = semver.inc(currentVersion, "prerelease");
+
 
     let message = "Select a new version";
     if (packageName) message += ` for ${packageName}`;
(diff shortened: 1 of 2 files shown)


src/commands/PublishCommand.js
```diff
@@ -213,6 +213,10 @@ export default class PublishCommand extends Command {
     const patch = semver.inc(currentVersion, "patch");
     const minor = semver.inc(currentVersion, "minor");
     const major = semver.inc(currentVersion, "major");
+    const prepatch = semver.inc(currentVersion, "prepatch");
+    const preminor = semver.inc(currentVersion, "preminor");
+    const premajor = semver.inc(currentVersion, "premajor");
+    const prerelease = semver.inc(currentVersion, "prerelease");
```
evocateur: We should probably allow customization of the `identifier` argument to `semver.inc`, yes?

Otherwise the output is a little [terse](https://runkit.com/58b8a5649a21f300136b74aa/58b8a5649a21f300136b74ab):
```js
semver.inc("1.0.0", "prerelease")       // => 1.0.1-0
semver.inc("1.0.0", "prerelease", "rc") // => 1.0.1-rc.0
// etc
```

rtsao: Should it prompt you for an identifier if one already exists? Or maybe it prompts you but hitting enter will use the existing identifier.

rtsao:
I briefly looked at the PublishCommand tests but it wasn't immediately apparent to me how I could have easily added tests for this. But I'll check again.

evocateur:
@rtsao Yeah, I apologize, the PublishCommand tests are a bit verbose and hard to grok for newcomers. In fact, I'm not even sure if we're _currently_ testing the other prompt values...

src/commands/PublishCommand.js
evocateur: I think your last suggestion hits the right balance of complexity and usefulness. (I just deleted a long-ish paragraph over-ruminating on all the cases)

rtsao: I've updated the PR with this functionality

rtsao:
Thoughts on tests for this? After digging a little deeper, there's many layers of abstraction in terms of how the commands are tested, so I'm not sure about how best to add tests for this.

evocateur:
https://github.com/lerna/lerna/blob/master/test/PublishCommand.js#L143 is one place where you could add the new keyword options, just to make sure it's doing the right thing.

(I have no idea why the test is telling various packages to go _backward_, ugh)

(ugh there are no tests of the _existing_ "custom" version prompt)

Screw it, I've put you through enough torment already.

PR #642, opened 2017-03-03
add markdown magic for auto docs

camacho:
I added [markdown magic](https://github.com/DavidWells/markdown-magic) to try and help with documentation (hopefully without being too presumptuous).

In this case, I was noticing that the version of Lerna was always wrong in the documentation. I thought it might be helpful if we could have that always up to date with a simple script. I also updated the install command to have the version included.

Adding in markdown-magic also gives more flexibility for any additional functionality the project might want in the future ([some plugins here](https://github.com/DavidWells/markdown-magic#plugins)).

I added the necessary files under a `scripts` directory to keep it separate from all the other run and test code, and updated the linting commands to cover this new dir.


hzoo:
That's really cool, although I figured we should just remove that from the readme either that line or just use 2.0.0?

camacho:
No strong opinions on my end. I've found it to be a useful tool on larger
projects with complex docs, code examples, etc. but it is additional
overhead.

camacho:
@hzoo would you like to close this PR and open a new one with updated examples?

PR #646, opened 2017-03-06
Update vague text in readme.md

jaredrada:
I think this copy more clearly explains the relationship of the two example dependencies (`babel-generator` and `source-map`). The old copy is somewhat vague. Do this change make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ Let's use `babel` as an example.
 ```
 
 - Lerna checks if each dependency is also part of the Lerna repo.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -189,7 +189,7 @@ Let's use `babel` as an example.
 ```
 
 - Lerna checks if each dependency is also part of the Lerna repo.
-  - In this example, `babel-generator` is a dependency, while `source-map` is not.
+  - In this example, `babel-generator` is an interal dependency, while `source-map` is an external dependency.
```
hzoo: /s/interal/internal

Issue #652, opened 2017-03-08
lerna publish with --registry flag errors with "env: node: No such file or directory"


YellowKirby:
When using lerna publish with the `--registry` CLI flag, I'm getting the following error:

```sh
> lerna publish --registry=https://some-custom-registry.com

...

? Are you sure you want to publish the above changes? Yes

Publishing packages to npm...
NpmUtilities.publishTaggedInDir("lerna-temp", "./packages/base-types", "https://some-custom-registry.com")
Error: Command failed: cd /Users/username/dev/project/packages/base-types && npm publish --tag lerna-temp
env: node: No such file or directory

    at ChildProcess.exithandler (child_process.js:210:12)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:194:7)
    at maybeClose (internal/child_process.js:899:16)
    at Socket.<anonymous> (internal/child_process.js:342:11)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:191:7)
    at Pipe._handle.close [as _onclose] (net.js:504:12)
```

The problem seems to be that [`NpmUtilities.getTagOpts()`](https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L137) sets up an `env` property that gets passed to `child_process` which _only_ contains the `npm_config_registry` value. When playing around with `child_process` interactively, it seems that passing `null` (or an empty object) for the 2nd parameter will pass through any environment variables, but this is not the case when an explicit `env` property is set:

When no `env` option is provided to `child_process`:
```sh
> child.exec('env', { }, (err, stdout, stderr) => console.log(stdout)) && false
false
> TERM=xterm-256color
SHELL=/usr/local/bin/fish
USER=username
PATH=/Library/Java/JavaVirtualMachines/jdk1.8.0_102.jdk/Contents/Home/bin:/Users/username/.yarn/bin:/Users/username/.fzf/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
PWD=/Users/username/dev/project
_=/usr/bin/env
...
```

But if an `{ env: { npm_config_registry: "foo" } }` is provided:
```sh
> child.exec('env', { env: { npm_config_registry: 'foo' } }, (err, stdout, stderr) => console.log(stdout)) && false
false
> npm_config_registry=foo
PWD=/Users/username/dev/project
SHLVL=1
_=/usr/bin/env
```

In the latter case, since `PATH` isn't set, the `node` executable cannot be found.

[...]

PR #654, opened 2017-03-08
Merge current process.env when using --registry flag

noherczeg:
In response to #652

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -134,6 +134,11 @@ export default class NpmUtilities {
   }
 
   static getTagOpts(registry) {
-    return registry ? {env: {npm_config_registry: registry}} : null;
+    const opts = {};
+    if (registry) {
+      opts.env = Object.assign({}, process.env, {npm_config_registry: registry});
+      return opts;
+    }
+    return null;
   }
 }
--- a/test/PublishCommand.js
+++ b/test/PublishCommand.js
@@ -954,7 +954,6 @@ describe("PublishCommand", () => {
         done();
       }));
     });
-
   });
 
   /** =========================================================================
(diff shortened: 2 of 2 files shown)


src/NpmUtilities.js
```diff
@@ -134,6 +134,11 @@ export default class NpmUtilities {
   }
 
   static getTagOpts(registry) {
-    return registry ? {env: {npm_config_registry: registry}} : null;
+    const opts = {};
+    if (registry) {
+      opts.env = Object.assign({}, process.env, {npm_config_registry: registry});
```
noherczeg: I think creating a clone of the env as a base is a good thing in this case since we only need the extended params for tagging. If it's an overkill, or in fact a bad thing, please let me know!

evocateur: That's totally the right thing to do, since _any_ non-`null` value overwrites everything, and we definitely want to transfer everything else.

test/PublishCommand.js
```diff
@@ -964,49 +983,49 @@ describe("PublishCommand", () => {
         [ChildProcessUtilities, "execSync", {}, [
           { args: ["git tag"] }
         ]],
-       [PromptUtilities, "confirm", { valueCallback: true }, [
```
evocateur: Important as these whitespace changes are, let's fix this separately (I'll approve it and merge immediately)

noherczeg: @evocateur If I do that eslint won't pass.

evocateur: How in the world did it pass otherwise? If you `npm run lint` in master right now, it makes absolutely no complaint about this weird indentation.

(the consistency apparently is trying to validate that the closing square bracket is on the same column as the opening square bracket, not the general indent, which is frustrating)

evocateur:
appveyor is soooo slooooowwwww :(

evocateur:
Thanks @noherczeg !

noherczeg:
@evocateur Hope it'll be fine now :)

evocateur:
@noherczeg Oops, I found another place we're losing the rest of `process.env`: [`NpmUtilities.installInDir()`](https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L26)

Minimal fix seems straightforward:
```diff
    if (registry) {
-      opts.env = {npm_config_registry: registry};
+      opts.env = Object.assign({}, process.env, {
+        npm_config_registry: registry,
+      });
    }
```

noherczeg:
@evocateur Sorry for the late response, I'll look into it!

PR #655, opened 2017-03-08
Actually warn when a matching dependency version is not satisfied

evocateur:
Passing in the logger works just as well as a boolean flag (just like the topological sorting method).

However, in both instances that `Package##hasMatchingDependency()` is actually called, neither of them ever passed `true` in the second arg, so it was basically dead code.

If it was intentional to omit logging in all use cases, then we can delete the dead code.

### DIFF
--- a/src/Package.js
+++ b/src/Package.js
@@ -1,7 +1,6 @@
 import path from "path";
 import semver from "semver";
 import NpmUtilities from "./NpmUtilities";
-import logger from "./logger";
 
 export default class Package {
   constructor(pkg, location) {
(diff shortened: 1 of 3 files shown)


jquense:
👍  I'd also like the ability to make this throw in certain situations. As it is I wrote up a little script that uses hasMatchingDependency for each package as a lint step, to make sure we don't accidently get out of sync on package versions.

evocateur:
Do you see that throwability as an extra flag (--strict) or a subcommand (`lerna check`)?

jquense:
I'd be happy with either I think. whatever makes more sense for the project. The only time we generally don't want that behavior when bootstraping/publishing is when a package depends on itself (in which case i'd need to use an older published version)

PR #661, opened 2017-03-10
Add support for `exact` flag in lerna.json

### LINKED ISSUE #511: Add support for `exact` flag in lerna.json

YasserKaddour:
Fixes #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,15 +235,6 @@ When run with this flag, `publish` will specify updated dependencies in updated 
 
 For more information, see the package.json [dependencies](https://docs.npmjs.com/files/package.json#dependencies) documentation.
 
-May also be configured in `lerna.json`:
-
-```js
-{
-  ...
-  "exact": true
-}
-```
-
 #### --npm-tag [tagname]
 
 ```sh
--- a/README.md
+++ b/README.md
@@ -235,6 +235,15 @@ When run with this flag, `publish` will specify updated dependencies in updated 
 
 For more information, see the package.json [dependencies](https://docs.npmjs.com/files/package.json#dependencies) documentation.
 
+May also be configured in `lerna.json`:
+
+```js
+{
+  ...
+  "exact": true
+}
+```
+
 #### --npm-tag [tagname]
 
 ```sh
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -235,6 +235,15 @@ When run with this flag, `publish` will specify updated dependencies in updated
...
 
+May also be configured in `lerna.json`:
+
+```js
+{
+  ...
+  "exact": true
+}
+```
+
```
YasserKaddour: I am not very sure where I am supposed to add this.

evocateur: We already describe the `lerna.json` "durable" config pattern elsewhere, I think this is implied (much like #598, which only made a single line change and no docs updated).

src/commands/PublishCommand.js
```diff
@@ -362,7 +362,7 @@ export default class PublishCommand extends Command {
       const version = this.updatesVersions[depName];
 
       if (deps[depName] && version) {
-        deps[depName] = this.flags.exact ? version : "^" + version;
+        deps[depName] = this.getOptions().exact ? version : ("^" + version);
```
evocateur: I think we should cache this `getOptions()` result outside of the `forEach`, since it will never change. I appreciate the disambiguating parentheses on the concatenation ternary.

test/PublishCommand.js
```diff
@@ -1170,6 +1170,107 @@ describe("PublishCommand", () => {
   });
 
   /** =========================================================================
+   * NORMAL - EXACT CONFIG
```
evocateur: I appreciate test cases, although I think in this case we've already covered the ability for "durable" `getOptions()` config to bubble up through the various levels (`lerna.json` vs CLI flags, etc).

I won't say "no" to more tests, though. ^_^

src/commands/PublishCommand.js
YasserKaddour: :sweat_smile:

test/PublishCommand.js
YasserKaddour: gigabo said in #426 test would be nice, so I did just that :smile:

YasserKaddour:
Thanks for merging

gigabo:
Thanks @YasserKaddour!

PR #663, opened 2017-03-11
add tests for NpmUtilities.getExecOpts()

noherczeg:


src/NpmUtilities.js
```diff
@@ -23,7 +23,9 @@ export default class NpmUtilities {
     };
 
     if (registry) {
-      opts.env = {npm_config_registry: registry};
+      opts.env = Object.assign({}, process.env, {
+        npm_config_registry: registry,
+      });
```
evocateur: I was playing with `NpmUtilities.getExecOpts()` locally during the contentious rebasing process yesterday, and I could have sworn I added it here, too. (technically `ChildProcessUtilities.spawn()` in this method, but same diff as far as the `opts` go). I think we should take the opportunity to reuse the new `getExecOpts()` method:

```
const opts = NpmUtilities.getExecOpts(directory, registry);
opts.stdio = ["ignore", "pipe", "pipe"];
```

Like all good refactorings, the rest of your (much appreciated!) tests won't need to be further modified. :)

(we can leave the _other_ exec opts `cwd` and `env` in this file alone for now, they're not aware of the registry for the moment)

test/NpmUtilities.js
```diff
@@ -21,4 +21,28 @@ describe("NpmUtilities", () => {
...
+      NODE_ENV: "lerna-test",
+    };
+
+    afterEach(() => {
+      process.env = originalEnv;
+    });
+
+    it("should handle environment variables properly", () => {
+      process.env = mockEnv;
+      const want = {cwd: "test_dir", env: Object.assign({}, mockEnv, {npm_config_registry: "https://my-secure-registry/npm"})};
```
evocateur: Let's spread this across multiple lines for readability, please.

src/NpmUtilities.js
noherczeg: Sure, will do in the evening!

Issue #645, opened 2017-03-06
`lerna init` on an empty folder don't generate any file


gre:
the prerequisite is to have a git repo but the command says it generated files successfully but didn't.

```
$ mkdir foo
$ cd foo
$ lerna init
Lerna v2.0.0-beta.38
Updating package.json.
Updating lerna.json.
Successfully created Lerna files
$ ls
# // nothing
```

evocateur:
I get something subtly different:
```bash
[~/github]$ mkdir foo
[~/github]$ cd foo/
[~/github/foo]$ lerna init
fatal: Not a git repository (or any of the parent directories): .git
Initializing Git repository.
Lerna v2.0.0-beta.38
Creating package.json.
Updating lerna.json.
Successfully created Lerna files
[~/github/foo] (master)$ ls -l
total 16
drwxr-xr-x  9    306B Mar  7 13:24 .git/
-rw-r--r--  1     91B Mar  7 13:24 lerna.json
-rw-r--r--  1     60B Mar  7 13:24 package.json
```

That fact that your logging said "Updating <file>" instead of "Creating <file>" is suspicious.

What are the versions of node, npm, and git in your shell?
```
node --version
npm --version
git --version
```

khaled:
same issue here.

node version:  v6.9.1
npm version: 3.10.8
git version: 2.2.1

khaled:
Looks like `lerna init` was creating initialization files in the parent directory, which happened to be a git repo.  After running `git init` in the child directory, `lerna init` did the right thing.

evocateur:
I think you've correctly identified the issue, @khaled. More and more, I think our "assume the git root is _always_ the place we want" is a bit naïve. Certainly in the case of `lerna init`, we should be using `process.cwd()` instead of `git rev-parse --show-toplevel`.

PR #664, opened 2017-03-11
fix: lerna init should always use current working directory

noherczeg:


noherczeg:
In response to #645

noherczeg:
I'm not sure about the current implementation though. What I was thinking about was leaving GitUtilities as it was, but in InitCommand:
```
constructor(input, flags) {
    super(input, flags);
    this.repository.rootPath = process.cwd();
}
```
My problem was though that the parent's constructor already did tons of stuff with the original value at that point. So some insight for this would be very much appreciated!

src/GitUtilities.js
```diff
@@ -83,7 +83,7 @@ export default class GitUtilities {
 
   @logger.logifySync()
   static getTopLevelDirectory() {
-    return ChildProcessUtilities.execSync("git rev-parse --show-toplevel");
+    return process.cwd();
```
evocateur: I think this is the wrong place to change the evaluation of a given Command's `rootPath` property. In fact, I don't think `Repository` should _ever_ be calling `git init`:

```diff
diff --git a/src/Repository.js b/src/Repository.js
index c9870f1..9ab637c 100644
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -1,24 +1,20 @@
-import GitUtilities from "./GitUtilities";
+import path from "path";
+import findUp from "find-up";
 import FileSystemUtilities from "./FileSystemUtilities";
 import PackageUtilities from "./PackageUtilities";
 import Package from "./Package";
 import NpmUtilities from "./NpmUtilities";
-import path from "path";
-import logger from "./logger";
 
 const DEFAULT_PACKAGE_GLOB = "packages/*";
 
 export default class Repository {
   constructor() {
-    if (!GitUtilities.isInitialized()) {
-      logger.info("Initializing Git repository.");
-      GitUtilities.init();
-    }
+    // findUp returns null when not found, and path.resolve starts from process.cwd()
+    const lernaJsonLocation = findUp.sync("lerna.json") || path.resolve("lerna.json");
 
-    this.rootPath = path.resolve(GitUtilities.getTopLevelDirectory());
-    this.lernaJsonLocation = path.join(this.rootPath, "lerna.json");
+    this.rootPath = path.dirname(lernaJsonLocation);
+    this.lernaJsonLocation = lernaJsonLocation;
     this.packageJsonLocation = path.join(this.rootPath, "package.json");
-    this.packagesLocation = path.join(this.rootPath, "packages"); // TODO: Kill this.
 
     // Legacy
     this.versionLocation = path.join(this.rootPath, "VERSION");
```

Instead, `lerna init` should have that responsibility, as befits an "init" command. Normal commands don't need that baggage/overhead:

```diff
diff --git a/src/commands/InitCommand.js b/src/commands/InitCommand.js
index b41c462..f6c04d0 100644
--- a/src/commands/InitCommand.js
+++ b/src/commands/InitCommand.js
@@ -1,3 +1,4 @@
+import GitUtilities from "./GitUtilities";
 import FileSystemUtilities from "../FileSystemUtilities";
 import Command from "../Command";
 import objectAssignSorted from "object-assign-sorted";
@@ -8,7 +9,11 @@ export default class InitCommand extends Command {
   runPreparations() {}
 
   initialize(callback) {
-    // Nothing to do...
+    if (!GitUtilities.isInitialized()) {
+      this.logger.info("Initializing Git repository.");
+      GitUtilities.init();
+    }
+
     callback(null, true);
   }
```

noherczeg: @evocateur Sure, np! :) Glad I could help!

noherczeg: @evocateur You can close this if it is no longer necessary!

Issue #662, opened 2017-03-10
add support for conventional-changelog format commit messages


bcoe:
I've become a convert to _this whole monorepo thing_. I see it as a great tool for simplifying the development, and perhaps more importantly, the management of two of the ecosystems I maintain ([conventional-changelog was recently ported over to lerna](https://github.com/conventional-changelog/conventional-changelog), and we're in the process of [moving over Istanbul](https://github.com/istanbuljs/istanbuljs)).

_.... my problem._

## Problem Statement

I've become fairly dependent on [conventional-commit-messages](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md), I adhere to a few simple rules when squashing commits:

* prefix with `fix:` for patches.
* prefix with `feat:` for new features.
* add `BREAKING CHANGE:` on a newline to describe a breaking API change.

By doing this I'm able to:

* automatically generate a CHANGELOG for repos; this is [really valuable for providing users with contextual information](https://github.com/yargs/yargs/pull/740#issuecomment-270508120).
* I also use this CHANGELOG format to recommend the next version # (if a release is taking a month to land, it's sometimes easy to forget if it's a `patch`, `minor`, or `major`).

When running `lerna` in independent mode, I lose this valued functionality.

## Proposal

I've been hard at work adding support to conventional-changelog for a lerna-style-git-history:

https://github.com/conventional-changelog/conventional-changelog/issues/168

I would like to add an additional flag for publishing in `independent` mode (`--conventional-commits`?) when running in this mode, two things would happen:

1. we would not prompt the user for version increments, we'd detect this based on the commit history (instead having a confirm step).
2. we would auto-generate the CHANGELOG in each `./pacakge/:name`.

I think adding this functionality would be pretty straightforward, and could be facilitated by pulling in a couple of the `conventional-changelog` core libraries, and augmenting the functionality of `PublishCommand.js`.

-----

thoughts, comments, concerns (CC: @hzoo, @evocateur).

evocateur:
I am generally in favor of this (also p excited that you're getting excited about `lerna` :D). As I mentioned in #643, this feels like a plugin system, but I am not opposed to a more focused solution to avoid larger, more complicated refactoring.

`--conventional-commits` seems like a reasonable flag. It would also be configurable in `lerna.json` as `"conventionalCommits": true`, either at the root or `command.publish` namespace.

bcoe:
[...]

PR #665, opened 2017-03-11
feat: adds support for conventional-commits (for CHANGELOG generation and verison bump)

bcoe:
## Problem

Independent mode forces you to manually bump each package; it can be easy to lose track of whether versions are `major`, `minor`, `patch`.

## Solution

Using a [conventional-commit-format](https://github.com/bcoe/conventional-changelog-standard/blob/master/convention.md) the git history can be parsed, and used to:

* determine the appropriate version bump.
* generate a lovely CHANGELOG, so  that users who depend on your library have good forensic information (_why the heck was this a major bump?_).

## Notes

Please note that this functionality is optional, and is turned on by the flag `--conventional-commits`.

## Todo

- [x] add tests for `ConventionalCommitUtilties.js`. 

addresses #662 (CC: @hzoo, @evocateur).


bcoe:
@Tapppi, @stevemao, this pulls together everything being discussed in https://github.com/conventional-changelog/conventional-changelog/issues/168 -- validating the week of noodling I've been doing in our `conventional-changelog` repos.

src/commands/PublishCommand.js
```diff
@@ -342,6 +352,13 @@ export default class PublishCommand extends Command {
       // write new package
       FileSystemUtilities.writeFileSync(packageJsonLocation, pkg.toJsonString());
 
+      // we can now generate the Chnagelog, based on the
```
stevemao: typo: Changelog

src/ConventionalCommitUtilties.js
```diff
@@ -0,0 +1,61 @@
...
+      const pkgPath = `./packages/${name}/`;
+      let changelogContents = "";
+      if (FileSystemUtilities.existsSync(changelogLocation)) {
+        changelogContents = FileSystemUtilities.readFileSync(changelogLocation);
+      }
+
+      // run conventional-changelog-cli to generate the markdown
+      // for the upcoming release.
+      const newEntry = ChildProcessUtilities.execSync(
+        `${require.resolve("conventional-changelog-cli/cli.js")} -l ${name} --commit-path=${pkgPath} --pkg=${pkgPath} -p angular`
```
stevemao: Is there any reason we don't use the api module, `conventional-changelog`, here?

src/ConventionalCommitUtilties.js
```diff
@@ -0,0 +1,61 @@
...
+      );
+
+      // CHANGELOG entries start with <a name=, we remove
+      // the header if it exists by starting at the first entry.
+      if (changelogContents.indexOf("<a name=") !== -1) {
```
stevemao: I think we could tweak the [template](https://github.com/conventional-changelog-archived-repos/conventional-changelog-writer#headerpartial) here.

src/ConventionalCommitUtilties.js
bcoe: It was my feeling that relying on the `conventional-recommended-bump` cli, and `conventional-changelog-cli` fit well with the existing patterns in the codebase; simply shelling out to `npm` with `NpmUtilities.js` , or git with `GitUtilities.js`:

* this minimized the number of modules from the conventional CHANGELOG ecosystem that I would need to pull in to support this functionality.
* and allows me, while adding tests, to use the same helpers/conventions used when adding tests to `npm`/`git`:
  * it's my plan to mock `ChildProcess`, using the same approach as here: https://github.com/lerna/lerna/blob/master/test/PublishCommand.js#L1707

src/ConventionalCommitUtilties.js
bcoe: @stevemao mind going into details don't completely follow?

bcoe: @stevemao I think we have slightly different use-cases in mind; I like the idea of adding in a header that explains the conventional-changelog format to curious parties -- I've found this is very useful for evangelizing the approach -- 

the annoying thing is it means we need to avoid adding the header multiple times as the CHANGELOG is updated; is there an existing module that solves this use-case?

src/commands/PublishCommand.js
bcoe: good catch 👍

src/ConventionalCommitUtilties.js
stevemao: We could change the handlebars template for the header. It's easy to do it with the api module 

```js
conventionalChangelog(options, context, gitRawCommitsOpts, parserOpts, {
    headerPartial: '{{#if isPatch~}}...' // without '<a name='
})
```

But with the cli at the moment we need to provide a `config` file to change this option. The [presets are good example of overwriting the default template partials](https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-angular/index.js#L121)

bcoe: @stevemao I think this is the header for specific commits? I'm talking about the use-case:

```md
# Overall Title

Header that introduces conventional commits and doesn't change.

* commit.
* commit.
* commit

```

Why don't we add the actual CHANGELOG merging logic to `conventional-changelog-cli`; then we could use the same logic for `lerna` and for `standard-version`; I think this would be a good solution because it DRYs up both codebases logic.

stevemao: The header of `conventional-changelog-writer` is per release. I'm not talking about the `CHANGELOG_HEADER` here and I think this is done right. I'm just saying stripping `<a name=` manually shouldn't be needed since we could just overwrite the `headerPartial` template.

bcoe: @stevemao gotcha; I think my comment is just confusing.

I don't want to strip the generated header; I just want to insert the new commit message between CHANGELOG_HEADER and the old commit messages; this is the same logic used by standard-version., and might be worth DRYing up.

stevemao: Oops sorry, misread the code :)

Tapppi:
@bcoe This is awesome, I think enabling a workflow as easy as [`standard-version`](https://github.com/conventional-changelog/standard-version) is for regular repos for individual mode would be a big thing. E.g. this is a blocker in our organisations move to lerna repos.. I'll definitely try to help with reviews / PR's as much as I have time (or at least lurk around leaving comments 😊 ).

src/commands/PublishCommand.js
```diff
@@ -342,6 +352,13 @@ export default class PublishCommand extends Command {
       // write new package
       FileSystemUtilities.writeFileSync(packageJsonLocation, pkg.toJsonString());
 
+      // we can now generate the Changelog, based on the
+      // the updated version that we're about to release.
+      if (this.flags.conventionalCommits) {
+        ConventionalCommitUtilties.updateChangelog(update.package.name, changelogLocation);
+        changedFiles.push(changelogLocation);
```
evocateur: I would prefer to encapsulate the `changelogLocation` assignment inside `updateChangelog()`. Perhaps return the changed location for pushing into `changedFiles`, then?

In fact, changelogs being important enough, regardless of how one generates them, I can justify adding a `changelogLocation` getter to the [Package](https://github.com/lerna/lerna/blob/master/src/Package.js) class, thus removing the need for a return value.

```diff
diff --git a/src/Package.js b/src/Package.js
index 76a88da..d4d4bee 100644
--- a/src/Package.js
+++ b/src/Package.js
@@ -17,6 +17,10 @@ export default class Package {
     return this._location;
   }
 
+  get changelogLocation() {
+    return path.join(this._location, "CHANGELOG.md");
+  }
+
   get nodeModulesLocation() {
     return path.join(this._location, "node_modules");
   }
```

thejameskyle:
Hi @bcoe, thanks for opening this.

I just wanted to make you aware of some plans for the future. We're planning on merging Lerna into Yarn through "workspaces" (sounds like npm has some related plans). I'm not sure a feature like this makes sense to be built directly into Yarn or npm, but definitely on top of it.

I'm wondering how we can enable you to build features like this on top of Lerna and later Yarn so that you don't have to rethink everything in the future.

One thing I've always wanted in Lerna which is part of the plan for Yarn is `lerna version`. This would match `npm/yarn version` but for Lerna/Yarn workspaces. This would let you stage changes before publishing them.

I think this functionality could be built on top of that and give you a lot more flexibility in the future.

evocateur:
@thejameskyle Is there a planning wiki/issue for "workspaces" that we can follow?

bcoe:
@thejameskyle @evocateur I was thinking events resembling package lifecycle scripts would get you pretty far ... _but_, one killer feature of a conventional changelog I couldn't think of an elegant way to facilitate, was recommending the version bump:

If you're using a conventional changelog format, we know when the release should be a `patch`, `minor` or `major` based on commit format -- it would be nice to be able to delegate to an external script for deciding on the semver bump that lerna is applying.

I've been talking to @iarna about my requirements around a monorepo, and was aware of the plans of rolling this workflow (at least in some form) into yarn -- my two cents, I think that the tagging and publication workflow that lerna facilitates in an abstract way is quite powerful, it would be cool to continue having a library that lives separately from your npm client of choice to facilitate the advanced workflows I rely on as an open-source maintainer: git-tagging, CHANGELOG generation, etc.

evocateur:
I'm not sure `yarn` or `npm` should be gaining a bunch of `git`-related abilities (counting commits, measuring tags, diffing a branch, etc), but anything to support easier parallel operations (publishing, managing dist-tags, and perhaps even hoisting) would be welcome.

thejameskyle:
I think Lerna should as much as possible try to model what would be built into Yarn or npm. People have and are going to continue building a lot of tooling on top of this functionality and it would be nice if that tooling didn't lock people in place.

There's a lot we can build into Lerna as an independent tool, but what's maybe more interesting is using Lerna as an experiment into how we can build features like workspaces into our package managers while finding the right extension points.

If we can build conventional-changelog tooling on top of Lerna, that tool can be easily reused if people want to switch to Yarn or npm.

bcoe:
@thejameskyle @evocateur here's how I think I could offer this functionality in a more abstract way:

1. have a `pretag` lifecycle script, if this script returns a semver valid tag on `stdout` use this tag `in lieu` of displaying a prompt in independent mode.
2. have a `precommit` hook; I would then be able to stage the updated CHANGELOG here.

With regards to this pull request, should @evocateur and I continue moving this feature forward without this abstraction (it feels like it will take longer to build out this lifecycle abstraction, and I sure wouldn't complain if I had `conventional-changelog` support int he interim).

src/ConventionalCommitUtilties.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      FileSystemUtilities.writeFileSync(
+        changelogLocation,
+         // only allow 1 \n at end of content.
+        `${CHANGELOG_HEADER}
+
+${newEntry}
+
+${changelogContents}`.replace(/\n+$/, "\n")
```
bcoe: @evocateur that's a really smart idea.

src/ConventionalCommitUtilties.js
```diff
@@ -0,0 +1,61 @@
...
+const CHANGELOG_HEADER = `# Change Log
+
+All notable changes to this project will be documented in this file.
+See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.`;
+
+export default class ConventionalCommitUtilties {
+  @logger.logifySync()
+  static recommendVersion(name, version) {
+    try {
+      const pkgPath = `./packages/${name}/`;
```
bcoe: @evocateur yeah, I noticed this when I started writing tests; wasn't sure how complex `pkg.location` could become.

It feels like `changelog.location` should then be something like `path.resolve(pkg.location, 'changelog.md')`?

evocateur:
Yes, exactly!

test/ConventionalCommitUtilities.js
```diff
@@ -0,0 +1,114 @@
...
+        ]],
+        [ChildProcessUtilities, "execSync", {}, [
+          {
+            args: [`${require.resolve("conventional-changelog-cli/cli.js")} -l bar --commit-path=/foo/bar --pkg=/foo/bar/package.json -p angular`],
+            returns: "<a name='change' />feat: I should be placed in the CHANGELOG"
+          }
+        ]],
+        [FileSystemUtilities, "writeFileSync", {}, [
+          {
+            args: ["/foo/bar/CHANGELOG.md", dedent(`
```
evocateur: The path here needs to use `path.normalize` so windows doesn't whine

bcoe:
@evocateur I've addressed your code review; I think all the suggestions were great, and things feel cleaner post refactoring.

I've also added tests for the `ConventionalCommitUtility`; which helped cleanup the API further.

I've been testing against [this repo](https://github.com/bcoe/standard-version-lerna-test). I'm really happy with how simple the workflow feels.

test/ConventionalCommitUtilities.js
bcoe: @evocateur how did I not know about `path.normalize`? I've been using `path.sep` all over the place in a ton of my repos.

evocateur:
I'm satisfied, let's hope Windows is too! Pretty excited for this, it's gonna make independent mode _so_ much smoother.

test/ConventionalCommitUtilities.js
evocateur: It's implicitly called behind the scenes when you use `path.join()` or `path.resolve()`, too, though for this purpose a direct `normalize` call seems appropriate.

evocateur:
I'm hoping #635 helps iron out some of those flapping tests. Thanks for your patience with the path issues (try saying that six times fast)!

bcoe:
@evocateur awesome 💃 thanks for your help on this, I can't wait to use this with `conventional-changelog` and `istanbul` ... at which point I can get back to actually fixing really hard to figure out instrumentation bugs ...

evocateur:
Whoops, we forgot to update `yarn.lock` 😨

evocateur:
follow up in https://github.com/lerna/lerna/pull/676

bestander:
FYI, I have created a task in Yarn where I aggregate the progress on Yarn workspaces and related PRs, RFC and discussions https://github.com/yarnpkg/yarn/issues/3294

evocateur:
Thanks!

Issue #277, opened 2016-07-13
[Feature] Canary should have incrementing version numbers.


cpojer:
The canary feature is fantastic, however because it uses commit hashes instead of numbers, publishing a canary release with the same basic version number might require the wrong versions when using ^ for all the other package dependencies. Can we either overwrite the dependencies to require explicit canary versions OR properly version canaries so that newer canaries will always be installed over older ones? :)

hzoo:
Ah good point.. didn't really think of that at the time..

cpojer:
Canary is *super* useful btw.

Also, it would be nice if I could select the next version. Right now it uses lerna.json which is the current version but I'd prefer to publish a canary for the next minor or patch release most of the time, so I have to manually edit lerna.json.

thejameskyle:
Can we just follow the normal prompting for versions flow for canary?

marcins:
I just ran into this issue, where the "latest" version of a dep was chosen over the canary because the canary version was based on the last released version number not the "next" one.

If you're automatically building a canary from CI you don't want the prompting flow, so as long as there's still a way to publish a canary with the "next" version number without interactivity then that'd work for my use case.

tdreyno:
Bumping this again. This is so broken it hurts.

Guria:
Guys, we should either use incrementing part in canary version before hash, or substitute exact version for controlled packages without caret in front.
Now our canary builds are definetely broken, because users will get wrong dependencies versions :(

L8D:
Can we _please_ at least add a flag for using exact versions in published packages? It would even be useful for non-canary releases.

KyleAMathews:
Is there anything holding up a PR fixing these issues? I haven't looked at the Lerna codebase much but would be happy to make a PR a) supporting incrementing number releases and b) add support for specifying a next version.

I'm using Lerna + canary releases to great effect with Gatsby and am planning on automating canary releases on build passes (anyone doing this btw?) and these two changes would make that much nicer.

tdreyno:
@KyleAMathews That'd be great. I hacked it in to my local build: https://github.com/tdreyno/lerna/commit/5f743ca6e347347d6ebf477017f6cebc992c99f7

evocateur:
@tdreyno I'd say #426 covers your hacked-in change, since beta 33.

tdreyno:
Thanks @evocateur, I'll try it out!

ivosabev:
Same problem with `2.0.0-rc.0`

[...]

PR #54, opened 2016-02-12
Push version tags explicitly

gigabo:
Somehow `--follow-tags` doesn't always work.

https://github.com/kittens/lerna/issues/52

Issue #52, opened 2016-02-12
Version tags not automatically pushed


gigabo:
Hi,

After each `lerna publish` I've found that I need to `git push origin <tag>`.

It _looks_ like this should happen automatically.  That's what the `--follow-tags` option is for, right?

I'm using the master branch of `lerna` (since `v1.1.0` doesn't actually work for me).

I'm on git version 2.6.2.

Thanks,

Bo

thejameskyle:
> That's what the `--follow-tags` option is for, right?

It should, can you test out the `git push origin --follow-tags` on a test repo to see if it works for you?

> since v1.1.0 doesn't actually work for me

Why doesn't v1.1.0 work for you?

hzoo:
The fact that master works and 1.1.0 doesn't is weird

gigabo:
> can you test out the git push origin --follow-tags on a test repo to see if it works for you?

Will do.

> Why doesn't v1.1.0 work for you?

Fails thusly:

```
Checking packages...
fatal: No annotated tags can describe 'a3a40d6877e075527f3c069923dd7ecec06f24f6'.
However, there were unannotated tags: try --tags.

There was a problem publishing.
Error: Command failed: git describe a3a40d6877e075527f3c069923dd7ecec06f24f6
fatal: No annotated tags can describe 'a3a40d6877e075527f3c069923dd7ecec06f24f6'.
However, there were unannotated tags: try --tags.

    at checkExecSyncError (child_process.js:441:13)
    at Object.execSync (child_process.js:481:13)
    at execSync (/Users/bo/.nvm/versions/node/v4.2.3/lib/node_modules/lerna/lib/commands/updated.js:63:16)
    at exports.checkUpdatedPackages (/Users/bo/.nvm/versions/node/v4.2.3/lib/node_modules/lerna/lib/commands/updated.js:30:23)
    at Object.exports.execute (/Users/bo/.nvm/versions/node/v4.2.3/lib/node_modules/lerna/lib/commands/publish.js:27:23)
    at Object.<anonymous> (/Users/bo/.nvm/versions/node/v4.2.3/lib/node_modules/lerna/bin/lerna.js:41:9)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
```

hzoo:
[...]

PR #666, opened 2017-03-11
Create annotated git tags instead of lightweight tags.

AlexLandau:
Lerna uses its git tags to determine which packages have been updated since the last publication, so it's important that tags be committed to the repository. Git's annotated tags are designed for this in a way that lightweight tags are not. For example, only annotated tags are pushed by 'git push --follow-tags'.


AlexLandau:
Some more notes on the motivation here:

* We use a setup in which the publish happens on CI machines, not developer machines, and git push of the relevant tags is handled manually. This is much easier with annotated tags, where --follow-tags can be used. (In particular, git can be configured to use this option automatically on every push, a feature we already take advantage of for our single-package repositories.)

* According to the git-tag documentation, "Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels".

* #52 and #54 suggest that --follow-tags was intended to be used for pushing tags, but they didn't notice that it only applied to annotated tags. It may be possible to return to that approach after this commit, but I'd rather not be responsible for testing that, given our non-standard workflows.

evocateur:
This would be a prerequisite for a useful `git describe` to solve #277

Issue #292, opened 2016-07-30
[Feature] lerna-scripts


ianwremmel:
I'm not entirely sure what this would look like, but I'd like to be able to define something like

```
lerna-scripts: {
  test: "mocha ."
}
```

at the top level of my project. Then, if I run `lerna run test`, it should execute `mocha .` in each package directory with the top-level's set of node_modules and .bins. Further if a given package defines its own script via npm scripts, that script should be executed instead.

hzoo:
Does either `lerna run` or `lerna exec` not cover it?

https://github.com/lerna/lerna#run `lerna run test` would already run `npm test` in each package that has it

It sounds like instead you are saying you want something like a global `lerna run test` so you don't have to define it in each package?

ianwremmel:
right, it's the top level default that i'm looking for; i'd like e.g. `lerna run test` to use something defined at the top level unless there's defined to the specific package being tested.

Additionally, i believe the current implementation would require e.g. mocha to be installed for each package; ideally, `lerna run test` would use the mocha installed at the top level rather than needing to define one in every package (unless of course the package needs a special version, in which case that one should be preferred).

gigabo:
I'm not sure generic "tests" are the best example here.  Sure, you can omit `"test": "mocha ."` from your package.json, but you still need _actual tests_ in each package directory. ;)

But `eslint`?  `nsp`?  Yeah, might be nice.

Either way, using a binary from a dependency at the repo root would be great!

ianwremmel:
True, each directory would still need to contain test files, but, the majority of them should need the same command. In my example, it's `mocha .`, but in practice, it's something more like `mocha -r babel -R junit -O outDir=./reports ./test/unit/spec/** && mocha -r babel -R junit -O outDir=./reports ./test/integration/spec/**`.

In one case, though, I've got a yeoman generator and yeoman provides its own set of tools for running test, so, for that directory, i want to specify a test command in package.json which invokes the appropriate script.

At the moment, I'm currently solving this problem two Gruntfiles (one of which is largely unreadable).

ziflex:
+1 for this.
For example,  I want to use Babel for all packages, but don't want to write build tasks for each project, I just want to write one generic and apply it for each package.

evocateur:
[...]

PR #668, opened 2017-03-13
Add script command

ikatyang:
Related #292

For me, this is a way to solve problem like this:

package.json

```sh
{
  "scripts": {
    "test": "lerna exec -- mocha ./lib/test/**/*.test.js"
  }
}
```

while using `npm run test`, I cannot pass any arguments to it, since it would be a wrong order.

For example:

```sh
npm run test -- --scope some-package
```

will be

```sh
lerna exec -- mocha ./lib/test/**/*.test.js --scope some-package
```

instead of

```sh
lerna exec --scope some-package -- mocha ./lib/test/**/*.test.js
```

Additionally, I think it would be better to have hook scripts (pre/post) and nested script, so I made this command.

This command is separated from original lerna lifecycle scripts (`./scripts/prepublish.js`, etc.), they won't affect each other.

## Command

```sh
lerna script <script-name> [-- <args>...]
```
Support flags

- `--stream`
- `--scope`
- `--ignore`
- `--include-filtered-dependencies`

## Configs

lerna.json
```json
{
  "scripts": {
    "prebuild": "rm -rf ./lib",
    "build": "babel",
    "postbuild": "lerna script test && lerna script lint",
    "build-watch": "lerna script build -- --watch",
    "test": "mocha",
    "lint": "eslint"
  }
}
```

Support nested scripts

- it will run a command in the package who executed the nested script.

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -35,6 +35,12 @@ export default class ChildProcessUtilities {
     );
   }
 
+  static execStreaming(command, opts, prefix, callback) {
+    opts = this.getStreamingOpts(opts);
+    const childProcess = this.exec(command, opts, callback);
+    this.streaming(childProcess, prefix);
+  }
+
   static execSync(command, opts) {
     const mergedOpts = Object.assign({
       encoding: "utf8",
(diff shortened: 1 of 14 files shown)


evocateur:
As indicated in [this comment](https://github.com/lerna/lerna/issues/292#issuecomment-298370120), this is largely implemented in recent release candidates. We appreciate the effort, but don't have any intention of re-implementing the npm's script lifecycle in lerna.

PR #676, opened 2017-03-15
chore: update lockfile with changes from #665

evocateur:
Not sure how I missed this, perhaps future validation of some sort is in order?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


bcoe:
@evocateur 👍 

image

☝️ my knowledge of yarn.lock.

evocateur:
@bcoe We definitely have some opportunity to improve our contributor guide (such as, I dunno, _mentioning_ yarn? 🤓 )

PR #686, opened 2017-03-17
expose getUpdatedPackages for programmatic API

spudly:
I have a need to generate zip files (from Jenkins) for each updated package after they are published, but in order to do that, I need to know which packages have been updated.

This quick and dirty change makes this possible. Instead of trying to parse the list of packages from the updated command (which is made for human, not machine consumption), I can just create a script that calls the getUpdatedPackages function and puts the resulting list somewhere (maybe a file or env var). Then I'll run lerna publish. After a successful publish, I'll create a zip file for each package in the env var.

I assume you'll want some changes to this, if you don't just reject it outright, but thought I'd go ahead and create the PR to get the conversation started.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,8 @@ import RunCommand from "./commands/RunCommand";
 import ExecCommand from "./commands/ExecCommand";
 import LsCommand from "./commands/LsCommand";
 import {exposeCommands} from "./Command";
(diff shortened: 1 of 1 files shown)


spudly:
`--json` would be even better.

spudly:
Closing in favor of #824

Issue #555, opened 2017-01-31
lerna assumes it manages git root


joscha:
When using lerna within a folder of a repository, e.g.
```
/ <-- here is the git root
/maven-project-a/ <-- calling `lerna init` here, this folder also has a `package.json`
```

it assumes that the `lerna.json` and `package.json` file will live in the root folder, e.g. creating a structure like this:
```
/lerna.json
/package.json (empty with only the lerna package)
/project-a/ (package.json here is unaltered)
```
if I move `lerna.json` into the `project-a` folder and add lerna to the dependencies of that `package.json` I get
```
lerna bootstrap
Lerna v2.0.0-beta.34
`package.json` does not exist, have you run `lerna init`?
```
when running `lerna bootstrap` from within `project-a`.

Would you accept a PR that teaches lerna to stop searching for a lerna config in the git root if it is invoked from a subdirectory which has a valid `package.json` (e.g. `"dependencies": {    "lerna": "2.0.0-beta.34" }`) and a valid `lerna.json`?

doug-wade:
You can use the [`packages` option in lerna.json](https://github.com/lerna/lerna#lernajson) to configure `maven-project-a/packages/` to be where your lerna modules are located, which should allow you to structure your project however is most convenient for maven.  Does that cover your use case, or is there something more I'm missing?

joscha:
Thanks for the answer @doug-wade, the packages location is actually not the problem, adapting that path works fine. The problem lies in lerna thinking its not been `lerna init`ed when the lerna managed package is not equal to the git root. I can create a gist to demonstrate the problem if that is easier?

doug-wade:
@joscha I think in this case a repro repository would be best, since then I can clone it and use it locally -- unless there's some cool gist feature I don't know about, to use it locally would take a lot of copy-paste.

joscha:
sure, repo is here: https://github.com/joscha/lerna-broken-nesting-example

joscha:
@doug-wade were you able to verify the problem or is there anything else I can provide?

evocateur:
@joscha I understand this problem, and I could've sworn there were a couple other issues recently with _basically_ the same problem. In any case...

Right now, we are indeed making the assumption that `lerna.json` lives in the repo root, which we obtain in the normal git plumbing fashion (`git rev-parse --show-toplevel`). It's robust because it doesn't care what your `CWD` is, and so on.

[...]

PR #687, opened 2017-03-17
Support lerna execution from subdirectories of repo root

### LINKED ISSUE #645: `lerna init` on an empty folder don't generate any file

evocateur:
A few high-level goals, here:

* `lerna init` should be the only time we (maybe) run `git init`.
* `lerna` commands should ensure they are run in a git repo during the validation phase, not instantiation.
* Use more "lazy" getters to avoid wasted work and streamline the command startup path.
* Use high-quality libraries ([find-up][], [load-json-file][]) instead of rolling our own.

fixes #645, #555

[find-up]: https://www.npmjs.com/package/find-up
[load-json-file]: https://www.npmjs.com/package/load-json-file

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -40,7 +40,15 @@ export default class ChildProcessUtilities {
       encoding: "utf8",
       maxBuffer: MAX_BUFFER
     }, opts);
-    return child.execSync(command, mergedOpts).trim();
+
+    let stdout = child.execSync(command, mergedOpts);
+    if (stdout) {
+      // stdout is undefined when stdio[1] is anything other than "pipe"
+      // and there's no point trimming an empty string (no piped stdout)
+      stdout = stdout.trim();
+    }
+
+    return stdout;
   }
 
   static spawn(command, args, opts, callback) {
--- a/src/Command.js
+++ b/src/Command.js
@@ -17,13 +17,25 @@ export default class Command {
     this.logger = logger;
     this.repository = new Repository();
     this.progressBar = progressBar;
+  }
 
-    const {sort, concurrency} = this.getOptions();
+  get concurrency() {
+    if (!this._concurrency) {
+      const { concurrency } = this.getOptions();
+      this._concurrency = Math.max(1, +concurrency || DEFAULT_CONCURRENCY);
+    }
 
-    this.concurrency = Math.max(1, +concurrency || DEFAULT_CONCURRENCY);
+    return this._concurrency;
+  }
 
-    // If the option isn't present then the default is to sort.
-    this.toposort = sort == null || sort;
+  get toposort() {
+    if (!this._toposort) {
+      const { sort } = this.getOptions();
+      // If the option isn't present then the default is to sort.
+      this._toposort = sort == null || sort;
+    }
+
+    return this._toposort;
   }
 
   get name() {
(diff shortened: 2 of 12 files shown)


evocateur:
This PR is based on #664. Many thanks @noherczeg!

src/ChildProcessUtilities.js
```diff
@@ -40,7 +40,15 @@ export default class ChildProcessUtilities {
       encoding: "utf8",
       maxBuffer: MAX_BUFFER
     }, opts);
-    return child.execSync(command, mergedOpts).trim();
+
+    let stdout = child.execSync(command, mergedOpts);
+    if (stdout) {
+      // stdout is undefined when stdio[1] is anything other than "pipe"
+      // and there's no point trimming an empty string (no piped stdout)
```
evocateur: I encountered this issue when I ignored `stderr` in `GitUtilities.isInitialized()`. That was a fun debugging session.

src/Command.js
```diff
@@ -83,8 +96,8 @@ export default class Command {
   }
 
   runValidations() {
-    if (this.concurrency < 1) {
```
evocateur: Turns out that this block could _never_ be entered because of the way we currently interpret the `sort` option.

src/Repository.js
```diff
@@ -1,81 +1,94 @@
...
   get lernaVersion() {
-    return this.lernaJson && this.lernaJson.lerna;
+    return this.lernaJson.lerna;
   }
 
   get version() {
-    return this.lernaJson && this.lernaJson.version;
-  }
-
-  get publishConfig() {
```
evocateur: The `publishConfig` and `bootstrapConfig` properties are already handled by `Command.getOptions()`, so it seemed safe to remove these.

src/commands/PublishCommand.js
```diff
@@ -374,9 +375,8 @@ export default class PublishCommand extends Command {
     }
   }
 
-  updatePackageDepsObject(pkg, depsKey) {
+  updatePackageDepsObject(pkg, depsKey, exact) {
     const deps = pkg[depsKey];
-    const {exact} = this.getOptions();
```
evocateur: It was bugging me that we were calling this every time, even if there were no deps.

src/Command.js
```diff
@@ -83,8 +96,8 @@ export default class Command {
...
-      this.logger.warn("command must be run with at least one thread.");
+    if (!GitUtilities.isInitialized()) {
```
peter-mouland: Hi @evocateur,

Just testing the upgrade path from v1-beta.36 to v2-rc.4 and come across this check.  We noticed it as it started breaking on CI where it copies a previously cloned repo to another area to perform more tasks (rather than cloning again).

I'm sure we can change our build setup easy enough - but I was just wondering what the reason is as to why Lerna needs to be executed in a git repo or if this check could be dropped?

Thanks for your help

evocateur:
We recently (last week) eased this restriction to only affect commands that actually need git (publish, diff, etc). It doesn't solve it for rc4, but I hope to release it soon.

PR #692, opened 2017-03-19
Fix #677, possible approach for transitive deps during hoist

### LINKED ISSUE #677: transitive dependencies of non-hoisted dependencies replace hoisted versions
Say I have a package that depends on `lodash@4` and another module `uses-old-lodash` with a transitive dependency on `lodash@3`.  If I do a regular `lerna bootstrap`, I end up with the following in my `node_modules` folder:

```
.
└── my-package
    └── node-modules
        ├── lodash@4
        └── uses-old-lodash
            └── node-modules
                └── lodash@3
```

Normally, using `lerna bootstrap --hoist` will pull the dependencies up to the top level without a problem.  But, if for some reason we can't hoist `uses-old-lodash` (either because of a version conflict or `--nohoist=uses-old-lodash`), then we end up with the following structure:

```
.
├── node_modules
│   └── lodash@4
└── my-package
    └── node-modules
        ├── lodash@3
        └── uses-old-lodash
```

Because of the way npm is flattening dependencies (npm@3 and above), `lodash@3` comes up to the root of `my-package/node-modules` when `lodash@4` gets hoisted.  So if I `require('lodash')` in a script in `my-package`, I'm going to get the wrong version of `lodash`.

It would be nice if the transitive dependencies did not get flattened in the case where a different version of the same dependency has been hoisted.

jppurcell9:
Here's a possible approach to fixing #677.  Instead of installing the leaf dependencies directly, we install them in a "shadow" directory `package-name/node_modules/_node_modules`.  Once all the leaves for `package-name` are installed, we go back and move each dep from `package-name/node_modules/_node_modules/dep` to `package-name/node_modules/dep` and symlink the shadow modules back into dep (`package-name/node_modules/dep/node_modules` -> `package-name/node_modules/_node_modules`).  This approach should work for both npm and yarn clients.

This isn't ready to be merged - tests are failing and more complex cases need to be addressed (version collisions in the flattened shadow dir, for example).  Just wanted to get some feedback on the approach to see if this is worth pursuing.

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -377,14 +377,56 @@ export default class BootstrapCommand extends Command {
         // If we have any unsatisfied deps then we need to install everything.
         // This is important for consistent behavior across npm clients.
         if (deps.some(({isSatisfied}) => !isSatisfied)) {
-          actions.push(
-            (cb) => NpmUtilities.installInDir(
-              pkg.location, deps.map(({dependency}) => dependency), this.npmConfig, (err) => {
-                this.progressBar.tick(pkg.name);
-                cb(err);
-              }
-            )
-          );
+          if (root.length === 0) {
+            actions.push(
+              (cb) => NpmUtilities.installInDir(
+                pkg.location, deps.map(({dependency}) => dependency), this.npmConfig, (err) => {
+                  this.progressBar.tick(pkg.name);
+                  cb(err);
+                }
+              )
+            );
+          } else {
+            const targetLocation = path.join(pkg.location, "node_modules", "_node_modules");
+            const depNames = deps.map(({ dependency }) => dependency.match(/(.*)@/)[1]);
+            actions.push(
+              (cb) => async.series([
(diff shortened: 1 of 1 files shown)


evocateur:
I appreciate the proof-of-concept!

However, it looks suuuuper-complicated, and I'd rather not engage in _too_ much hackery (aside from our current `yarn` bait-n-switch) with npm clients while work is [underway](https://github.com/npm/npm/pull/15900) to specify "native" behaviour that should solve this (yarn is [coordinating](https://github.com/yarnpkg/rfcs/pull/34) with npm folks, too).

In the mean time, I'd be much happier with a solution that simply used `npm install --global-style` (regardless of `npmClient` option) to deal with non-hoisted package dependencies. It hits the sweet spot of "already tested elsewhere" and "acceptable compromise of functionality", for me.

jppurcell9:
Agreed - and getting this right (not just this proof of concept) is going to be even more complicated.  I'll close this and try again with `--global-style`.

Issue #591, opened 2017-02-09
Init doesn't create a packages folder


xtuc:
In the README, the getting started section says that the `lerna init` command:

> [...] will create a lerna.json configuration file as well as a packages folder.

Looking at the source code I think this behavior doesn't exists anymore, see [InitCommand](https://github.com/lerna/lerna/blob/master/src/commands/InitCommand.js#L16-L18)

Tested on `lerna@2.0.0-beta.37`

gigabo:
Nice catch @xtuc.  That was removed from the documentation when https://github.com/lerna/lerna/pull/365 stopped creating the folder, but looks it was added back into the documentation in https://github.com/lerna/lerna/pull/517.  I guess the resolution there should have been to update the _website_ rather than the README.  Should be solved by building the website from the README, which I think is underway in https://github.com/lerna/website/pull/14.

jwb:
I wonder: if the website generation isn't making progress, should we prune out the details and just link to the README, so we aren't providing bad info to users?

PR #703, opened 2017-03-21
Update README.md

dnalborczyk:
**lerna init** does not create a **packages** folder anymore.

compare also:
https://github.com/lerna/lerna/issues/591

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,12 +85,11 @@ Your repository should now look like this:
 
 ```
 lerna-repo/
-  packages/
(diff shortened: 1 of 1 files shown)


evocateur:
We appreciate the PR! I think we need to have a discussion in #704 before we move further on this, since it is kinda awkward to simply remove an important part of the whole pattern.

evocateur:
This was finally fixed in #970, with docs updated in #985.

Issue #706, opened 2017-03-21
--only-explicit-updates should be renamed to discourage use


thejameskyle:
Most people should not be using `--only-explicit-updates`. It causes the silent updating of local packages in git that are not published to npm until later on.

This has caused bad releases for some people using it without fully understanding what it does.

I think these names are better:

- `--dangerously-only-publish-explicit-updates`
- `--update-dependencies-without-publish`

jwineman:
We ran into this exact issue at Cloudflare with CF-UI.  Im in favor of renaming it.

evocateur:
I've used it when independently versioning to reduce the tag noise for changes that I _know_ are compatible (using caret versions). I can appreciate that I'm an edge case, and I was blatantly abusing the flag just so my git tag list was a little less bloated.

I see even less reason for a fixed versioning monorepo to use this, considering the drawbacks. I lean toward `--update-dependencies-without-publish` for renaming.

hzoo:
Yeah it was only intended for backwards-compat in Babel

thejameskyle:
lol, we could rename it `--custom-magical-flag-for-babel-and-no-one-else-do-not-use-or-ill-bite-you`

tajo:
Can you please come up with some specific examples? 

Two libs A and B where B is a dep of A: 
1. What happens for minor bumps of B with the flag?
2. What happens for minor bumps of B without the flag?
3. What happens for major bumps of B with the flag? 
4. What happens for minor bumps of B without the flag? 

More I read about this, more confused I am.

hzoo:
Like mentioned it wasn't meant for anyone to use, just for Babel since we had 100 pkgs and lots of dependencies where all of them were already using `^` so there was no reason to update all packages.

thejameskyle:
The only reason that's okay for Babel to do is because they have a good understanding of Lerna and almost never make major version bumps.

thejameskyle:
Moving to #707

PR #707, opened 2017-03-21
Remove --only-explicit-updates

thejameskyle:
Re: https://github.com/lerna/lerna/issues/706

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,21 +649,6 @@ $ lerna bootstrap --scope "package-*" --ignore "package-util-*" --include-filter
 # package matched by "package-*"
 ```
 
-#### --only-explicit-updates
-
-> **Note:** You will rarely want to do this. It will update the versions in dependencies without publishing new versions of them. This means your packages will silently be updated without asking you for permission while publishing.
-
-Only will bump versions for packages that have been updated explicitly rather than cross-dependencies.
-
-> This may not make sense for a major version bump since other packages that depend on the updated packages wouldn't be updated.
-
-```sh
-$ lerna updated --only-explicit-updates
-$ lerna publish --only-explicit-updates
-```
-
-Ex: in Babel, `babel-types` is depended upon by all packages in the monorepo (over 100). However, Babel uses `^` for most of its dependencies so it isn't necessary to bump the versions of all packages if only `babel-types` is updated. This option allows only the packages that have been explicitly updated to make a new version.
-
 #### --loglevel [silent|error|warn|success|info|verbose|silly]
 
 What level of logs to report.  On failure, all logs are written to lerna-debug.log in the current working directory.
--- a/src/Command.js
+++ b/src/Command.js
@@ -156,7 +156,7 @@ export default class Command {
     }
 
     if (this.flags.onlyExplicitUpdates) {
-      this.logger.warn("`--only-explicit-updates` has been removed. This flag was only ever added for Babel and we never should have exposed it to everyone.")
+      this.logger.warn("`--only-explicit-updates` has been removed. This flag was only ever added for Babel and we never should have exposed it to everyone.");
       this._complete(null, 1);
       return;
     }
--- a/src/Command.js
+++ b/src/Command.js
@@ -154,6 +154,12 @@ export default class Command {
       this._complete(null, 1);
       return;
     }
+
+    if (this.flags.onlyExplicitUpdates) {
+      this.logger.warn("`--only-explicit-updates` has been removed. This flag was only ever added for Babel and we never should have exposed it to everyone.")
+      this._complete(null, 1);
+      return;
+    }
   }
 
   runPreparations() {
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -131,7 +131,7 @@ export default class UpdatedPackagesCollector {
 
   collectUpdates() {
     return this.packages.filter((pkg) => {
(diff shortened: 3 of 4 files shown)


evocateur:
Pretty sure `yarn run fix` would make the tests pass

wyuenho:
Hold on

wyuenho:
In cf-ui, the packages are versioned independently. There's a graph of dependencies where updating one package often leads to whatever requires that to update, and whatever else requires that thing that requires that package you've changed and so on. Not using --only-explicit-updates often leads to more publishing errors in the case where someone accidentally bumped to a wrong version number on a dependee (what's the right word?) package, publishing that package, only then later find out that package wasn't ready to be published. So now 20 git tags and a few commits later, everyone's moaning about it's a PITA to roll back to the previous git state and still can't unpublish.

This is actually a very useful flag to prevents a lot of headache.

thejameskyle:
@wyuenho Using `--only-explicit-updates` is exactly what got you in that scenario. You would not have had problems if you hadn't been using it because you would have been alerted to the fact that you were updating more packages than you intended.

wyuenho:
> alerted to the fact that you were updating more packages than you intended

I'm not sure what this means. Correct me if I'm wrong, but a major version bump causes a cascade of dependee (sic) updates and I don't recall a way to tell lerna to skip those I really don't want to publish. Is there a way in lerna to NOT insist on publishing a package other than `--only-explicit-updates` that I'm not aware of?

thejameskyle:
When you go to publish a set of updated packages, Lerna looks for the dependents because it needs to ensure that the packages stay in sync. Otherwise your repo ends up in a "dirty" state where Lerna is unsure if you intended to pin things to older versions. It will still link them, but they will fall behind (especially in independent mode).

If you didn't update dependents when making a major version bump, Lerna's linking would silently turn off and you'd have to go through and manually update everything.

The main reason that Babel wants to turn this off is because it exists in a world where npm2 is still in use by many of the project's users and deduping doesn't work as well as it should. The less versions of a dependency that exists the better deduping ends up working.

When you go to publish your updated packages `lerna publish` will let you know about all the packages being updated, including unchanged dependents. In independent mode it will also ask you to provide a new version for each updated package/dependent to be published.

The only time you should ever disable this behavior is when you want to manually pin a dependency to an older version of a package. The _only_ time you should want to do that is when you've made a breaking change and don't want to update everywhere yet. `--scope` and `--ignore` exist for this purpose. You can also manually pin a version range (which is what I would generally do).

Publishing new versions of packages and creating git tags are both really cheap to do. If you are using npm3 or Yarn you should never notice any problems with just publishing new packages all the time.

wyuenho:
AFAICT `--scope` and `--ignore` do not work for `lerna updated` and `lerna publish`. They only work for run, exec, clean, ls and bootstrap.

> The only time you should ever disable this behavior is when you want to manually pin a dependency to an older version of a package. The only time you should want to do that is when you've made a breaking change and don't want to update everywhere yet.

That. This actually happens quite a lot when there are a lot of independent development going on where team members all need changes on different packages quickly. An all-or-nothing approach does not work for this style of development. The only way to publish a set of packages minus a few is to manually go in to the individual packages and type `npm publish`. This will of course still mess up lerna the next time around. There doesn't exist anything else that works for us except this wonderful feature.

thejameskyle:
Just pin your versions, it works just fine.

wyuenho:
This case come up from time to time:

- Module a1.0 depends on c^1.0
- Module b in your app depends on a^1.0 and c==1.0
- c introduce a breaking change for coworker foo, bumps to v2.0
- lerna publish bumps version of module a, the code that fixes the breaking change only qualifies as a minor update, so the module a1.1 and c2.0 go out
- There are now 2 versions of module c, one for the app, another for module a, dedupe in any npm or yarn won't work.

The only way to deal with this is really to pin to a@1.0 or update the app code to use c^2.0. But if module a also has something a release needs, the only option is to go fix the code to use c^2.0, which may push back the release for some time.

It would be nice if this feature is preserved, but we can try to adjust if you don't.

hzoo:
Also this was only made for the fixed versioning scheme (Babel), and also means that if you make a breaking change in one package that is a dependency you make a breaking change in all of them (why we haven't made a breaking change in babel).

wyuenho:
> if you make a breaking change in one package that is a dependency you make a breaking change in all of them 

So if the dep graph is A->B->C, if I bump C's major, the assumption in Lerna is B *also* might break, and we should bump B's major, and all packages that depend on B, and A, and all packages in the repo that depend on A?

I'm very confused. What's the point of allowing independent versions at all if we can't publish independently?

hzoo:
I'm talking about the fixed versioning mode, never intended this flag to be used with independent mode (didn't think about it). I think the issue is that we don't know that if you bump a package A that is a dependency in B whether it is a major version or minor version.

wyuenho:
It seems to me the problem is as James said, the symlinks and the fact that Lerna doesn't have any notions of "state snapshots" across publish events. Any invariant assumption violation confuses Lerna and makes recovery even more of a headache.

Just a last question, what's the exact behavior of `--force-publish`? Using the case [here](https://github.com/lerna/lerna/pull/707#issuecomment-288271305), if I force publish the C2.0 only, what would happen the next time I run `lerna publish` ?

thejameskyle:
`--force-publish` adds packages to the publish, it does not exclude ones that will already be in it

thejameskyle:
I'm going to merge this, if there needs to be a separate feature request for something to aid with pinning dependents to specific versions then that is a separate issue. `--only-explicit-updates` was not designed for that purpose and using it that way is incorrect.

PR #719, opened 2017-03-29
Use yargs to handle CLI args and subcommands

noherczeg:
## Description
I'm proposing this PR as a possible solution to the problem of the lerna cli tool help maintenance hassle, and the Command options handling in Lerna.

The current implementation should be considered as a draft, or a conversation starter. It could be changed, redesigned, or used as a starting point!

## Motivation and Context
I found it really hard to maintain the list of flags and when to use them in the README.md file manually. I also found the allocation of said flags problematic in the Command implementations as a contributor.

My main goals were the following:
- Provide an in-place configuration for each Command were developers can find and define all the flags they want to provide for each Command.
- Make the Commands and their respective options descriptors available statically so e.g. [yargs](https://github.com/yargs/yargs) or any other cli util could access them without the need of instantiation.
- Provide an API method in the `Command.js` class besides `.getOptions()` called `.getAvailableOptions()` which filters all the options present in the system, and only returns the flags which are applicable to the Command it is called on (this IMHO should not replace the existing `getOptions()` method, because we might need it for flexibility).

All child Commands inherit options defined on the base `Comand` class.

## How Has This Been Tested?
Anyone can test how the new help system works by using the already familiar `--help` flag on the `lerna` command it self, or now with any Lerna command, e.g.:
```
$ lerna bootstrap --help
```

## Types of changes

## Checklist:

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -18,20 +18,24 @@ export const builder = {
     coerce: (val) => Math.max(1, val)
   },
   "ignore": {
-    describe: "Ignores packages with names matching the given glob (Works only in combination with the 'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).",
+    describe: `Ignores packages with names matching the given glob (Works only in combination with the 
+               'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).`,
     type: "string",
     requiresArg: true
   },
   "include-filtered-dependencies": {
-    describe: "Flag to force lerna to include all dependencies and transitive dependencies when running 'bootstrap', even if they should not be included by the scope or ignore flags"
+    describe: `Flag to force lerna to include all dependencies and transitive dependencies when running 
+               'bootstrap', even if they should not be included by the scope or ignore flags`
   },
   "registry": {
-    describe: "When run with this flag, forwarded npm commands will use the specified registry for your package(s).",
+    describe: `When run with this flag, forwarded npm commands will use the specified registry for your 
+               package(s).`,
     type: "string",
     requiresArg: true
   },
   "scope": {
-    describe: "Restricts the scope to package names matching the given glob (Works only in combination with the 'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).",
+    describe: `Restricts the scope to package names matching the given glob (Works only in combination 
+               with the 'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).`,
     type: "string",
     requiresArg: true
   },
@@ -138,7 +142,7 @@ export default class Command {
   }
 
   runValidations() {
-    const {independent, onlyExplicitUpdates} = this.getOptions();
+    const { independent, onlyExplicitUpdates } = this.getOptions();
     if (!GitUtilities.isInitialized(this.repository.rootPath)) {
       this.logger.warn("This is not a git repository, did you already run `git init` or `lerna init`?");
       this._complete(null, 1);
--- a/src/Command.js
+++ b/src/Command.js
@@ -9,6 +9,38 @@ import logger from "./logger";
 
 const DEFAULT_CONCURRENCY = 4;
 
+export const builder = {
+  "concurrency": {
+    describe: "How many threads to use if lerna parallelises the tasks.",
+    type: "number",
+    requiresArg: true,
+    default: DEFAULT_CONCURRENCY,
+    coerce: (val) => Math.max(1, val)
+  },
+  "ignore": {
+    describe: "Ignores packages with names matching the given glob (Works only in combination with the 'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).",
+    type: "string",
+    requiresArg: true
+  },
+  "include-filtered-dependencies": {
+    describe: "Flag to force lerna to include all dependencies and transitive dependencies when running 'bootstrap', even if they should not be included by the scope or ignore flags"
+  },
+  "registry": {
+    describe: "When run with this flag, forwarded npm commands will use the specified registry for your package(s).",
+    type: "string",
+    requiresArg: true
+  },
+  "scope": {
+    describe: "Restricts the scope to package names matching the given glob (Works only in combination with the 'run', 'exec', 'clean', 'ls' and 'bootstrap' commands).",
+    type: "string",
+    requiresArg: true
+  },
+  "sort": {
+    describe: "Sort packages topologically",
+    default: true
+  }
+};
+
 export default class Command {
   constructor(input, flags, cwd) {
     this.input = input;
@@ -106,6 +138,7 @@ export default class Command {
   }
 
   runValidations() {
+    const {independent, onlyExplicitUpdates} = this.getOptions();
     if (!GitUtilities.isInitialized(this.repository.rootPath)) {
       this.logger.warn("This is not a git repository, did you already run `git init` or `lerna init`?");
       this._complete(null, 1);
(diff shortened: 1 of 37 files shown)


noherczeg:
Possible improvement could be the [commandDir](https://github.com/yargs/yargs#example-command-hierarchy-using-commanddir) approach where the current `bin/lerna.js` mechanics could be abstracted away into an `src/cli.js` file.

noherczeg:
It's also worth noting that `.getAvailableOptions()` used with destructuring does not provide any build in checks so if someone requests an option which is not "registered", then said option will be `undefined`!

I would love to make the retrieval of options as strict as possible, so I welcome any suggestions regarding this matter!

src/Command.js
```diff
@@ -83,6 +117,11 @@ export default class Command {
     );
   }
 
+  getAvailableOptions() {
```
evocateur: I think we can stick with `getOptions()` if we're using command modules. The only command that currently uses the `otherCommandConfigs` getter (to share `PublishCommand` options) can just as easily import the `builder` object from `PublishCommand` itself and spread it into its own `builder` export.

```js
// ...
import { builder as publishOptions } from "./PublishCommand";

// command export

// describe export

export const builder = {
  ...publishOptions,
};

// handler export

export default class UpdatedCommand extends Command {
// ...
```

(the `otherCommandConfigs` bit is still necessary to read `lerna.json` stuff)

src/commands/CleanCommand.js
```diff
@@ -1,11 +1,26 @@
-import async from "async";
-import Command from "../Command";
 import FileSystemUtilities from "../FileSystemUtilities";
 import PromptUtilities from "../PromptUtilities";
+import Command from "../Command";
+import async from "async";
```
evocateur: Please don't reorder the imports for cosmetic reasons, it's a historical artifact that I'd rather not propagate.

src/commands/BootstrapCommand.js
```diff
@@ -3,15 +3,40 @@ import NpmUtilities from "../NpmUtilities";
...
+    type: "string"
+  }
+};
+
 export default class BootstrapCommand extends Command {
+  static getSupportedOptions() {
+    return Object.assign({}, Command.getSupportedOptions(), SUPPORTED_OPTS);
+  }
+
+  static get describe() {
```
evocateur: As I've shown elsewhere, this should just be a named export, not a static property of the default export.

src/commands/InitCommand.js
```diff
@@ -4,18 +4,37 @@ import GitUtilities from "../GitUtilities";
 import Command from "../Command";
 import objectAssignSorted from "object-assign-sorted";
 
+const SUPPORTED_OPTS = {
+  exact: {
+    describe: "Specify cross-dependency version numbers exactly rather than with a caret (^)"
```
evocateur: In the context of `InitCommand`, `--exact` is actually intended to pin the version of `lerna` that was used to initialize in `package.json` without a caret. `"Specify lerna dependency version in package.json without a caret (^)"` is probably better, here.

src/commands/PublishCommand.js
```diff
@@ -12,11 +12,69 @@ import chalk from "chalk";
 import path from "path";
 import { EOL } from "os";
 
+const SUPPORTED_OPTS = {
```
evocateur: We should probably alphabetize the keys of this options object, just for ease of maintenance. (I think yargs alphabetizes them when listing `--help`?)

noherczeg:
@evocateur Thanks for the review, I'll go through all of them today!

noherczeg:
This should not be merged until I provided integration tests for the lerna cli.

test/integration/lerna-publish.test.js
```diff
@@ -57,4 +60,15 @@ describe("lerna publish", () => {
       });
     });
   }));
+
+  test("updates independent versions by npm", () => {
+    return initFixture("PublishCommand/integration").then((cwd) => {
+      return Promise.resolve()
+        .then(() => installInDir(cwd))
+        .then(() => execa("npm", ["run", "lp"], { cwd }))
```
noherczeg: @evocateur Could this (and the other new integration tests) be considered as a "proof" that the new Command and arg mechanics don't break existing functionality?

noherczeg:
I've spent hours trying to figure out what is this folder mismatch thing going on without any results. I do like the new testing format, but to be honest I kind of feel like it has a bit more black magic to it than I'd like it to.

@evocateur Could you give me some directions or suggestions what I'm doing wrong?

test/integration/lerna-bootstrap.test.js
```diff
@@ -21,6 +21,10 @@ describe("lerna bootstrap", () => {
           .then((result) => {
             expect(result.stdout).toMatchSnapshot("stdout: simple");
           })
+          .then(() => execa(LERNA_BIN, ["bootstrap", "--ignore", "@integration/package-1"], { cwd }))
```
evocateur: This needs to run in a seperate `test.concurrent()` block. It's not isolated enough, here. (which is testing the "defaults")

test/integration/lerna-publish.test.js
evocateur: We've already got the bootstrap test installing locally and running from an npm script, I don't think we need another one here (just do the `execa(LERNA_BIN...)` part). They take so long to run :P

But yes, I would consider these validation that the same CLI flags work the same way with yargs as they did with meow.

test/integration/lerna-run.test.js
```diff
@@ -0,0 +1,20 @@
...
+const installInDir = (cwd) =>
+  execa("npm", ["install", "--cache-min=99999"], { cwd });
+
+describe("lerna run", () => {
+
+  test.concurrent("can run script in packages", () => {
+    return initFixture("RunCommand/basic").then((cwd) => {
+      return Promise.resolve()
+        .then(() => installInDir(cwd))
+        .then(() => execa(LERNA_BIN, ["run", "my-script", "--scope=package-1"], { cwd }))
```
evocateur: The snapshots need to be "stable", that is, they can't have the temp directory in them, because that changes every run. To pass the `--silent` flag to `npm` (which omits the logging that contains the unstable tmpdir), add `--`, `--silent` to the args here. It's how I got the bootstrap integration `lerna run test` to have stable snapshots.

I agree it's a bit weird, especially how we handle the "extra args" (the ones that follow `--` aren't parsed by yargs or meow, but passed on to `npm run`, _as args to `npm`, not the script_.

evocateur:
@noherczeg Thanks for sticking with this, it's looking great so far!

noherczeg:
```
Snapshot Summary
 › 1 obsolete snapshot found, run with `npm run integration -- -u` to remove them.
Test Suites: 1 failed, 1 passed, 2 of 4 total
Tests:       1 failed, 4 passed, 5 total
Snapshots:   8 passed, 8 total
Time:        4.766s
Ran all test suites
```

Thats a new one. Strange since everything is ok at my end. Run `-- -u` locally but it didn't change a thing.

test/fixtures/RunCommand/integration-lifecycle/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "independent",
+  "scripts": {
+    "postinstall": "lerna run my-script --scope=package-1"
+  }
```
evocateur: I really don't think we need a "lerna run" from npm script integration test, we're already doing that in `lerna-bootstrap.test.js`.

In any case, the reason your tests are failing right now is that you're missing a local dependency on lerna here, like in the PublishCommand/integration fixture:
```json
  "devDependencies": {
    "lerna": "__TEST_PKG_URL__"
  }
```

Issue #721, opened 2017-03-30
`NODE_ENV=production` is not honored when installing packages


kevinpumpup:
Background: I want to use `lerna bootstrap` in a production setting, but there seems no way to add a `--production` flag to the underlying npm or yarn. Therefore I am trying the `NODE_ENV` env which are used by `npm` and `yarn`. See
- https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-production
- https://docs.npmjs.com/cli/install

However, `lerna bootstrap` does not seem to honor `NODE_ENV` either.

## Expected Behavior
`lerna bootstarp` should honor `NODE_ENV` by exposing it to underlying `npm install` or `yarn install` command

## Current Behavior
`NODE_ENV` is not honored.

## Possible Solution
TBD

## Steps to Reproduce (for bugs)
https://github.com/kevinpumpup/lerna-example

- `export NODE_ENV=production`
- `yarn` or `npm install` under `packages/foo` will not install dev dependencies
- `lerna bootstrap` will install dev depedencies

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.0.0-beta.38",
  "packages": [
    "packages/*"
  ],
  "version": "0.0.0"
}
```
</p></details>

## Context
Want to use `lerna bootstrap` in a production context.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.0.0-beta.38 |
| `npm --version`  | 3.10.3 |
| `yarn --version` | 0.21.3 |
| `node --version` | v6.5.0 |

| OS | Version |
| --- | --- |
| macOS | 10.11.6 |

evocateur:
Thanks for the great description!

Indeed, when we are [installing dependencies](https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L11) in a given directory, we are not [explicitly propagating `process.env`](https://github.com/lerna/lerna/blob/master/src/NpmUtilities.js#L125) to the spawned subprocess (unless you've passed `--registry` config, which is then mixed in). I _thought_ that meant it defaulted to `process.env` anyway, but this clearly indicates the wrongness of that assumption.

Does this repro if you try this with `lerna@canary`?

kevinpumpup:
@evocateur Yes it is reproduced with `lerna@canary`.

evocateur:
Thanks for checking. We would love a PR, if you're interested.

arcticShadow:
[...]

PR #727, opened 2017-04-03
fix: always honour env vars in spawned processes

noherczeg:
Environment variables are only passed down to spawned processes when someone provides a registry flag. This is problematic (e.g. #721). I'm trying to fix this problem in this PR.

If this is totally not the right way to solve it please let me know!

## Description
`.getExecOpts()` has been modified to always pass down env vars.

## How Has This Been Tested?
Unit tests are updated, and an additional test has been added as well for the Bootstrap Command ("should always honour env vars").

## Types of changes

## Checklist:

### DIFF
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -170,12 +170,15 @@ describe("BootstrapCommand", () => {
         gotRimraf[path] = true;
         callback();
       });
+      const wantOpts = { cwd: testDir, stdio: STDIO_OPT, env: process.env };
+      const wantPkg3Opts = { cwd: path.join(testDir, "packages" ,"package-3"), stdio: STDIO_OPT, env: process.env };
+      const wantPkg4Opts = { cwd: path.join(testDir, "packages", "package-4"), stdio: STDIO_OPT, env: process.env };
 
       assertStubbedCalls([
         [ChildProcessUtilities, "spawn", { nodeCallback: true }, [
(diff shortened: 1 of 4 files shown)


arcticShadow:
Note as per my investigation (documented in 721), the default is to pass down all env vars. (and it actually works)  So this change is not nessasary.

test/BootstrapCommand.js
```diff
@@ -789,4 +794,41 @@ describe("BootstrapCommand", () => {
...
+      process.env = mockEnv;
+    }));
+
+    afterEach(() => {
+      process.env = originalEnv;
+    });
+
+    it("should always honour env vars", (done) => {
+      const bootstrapCommand = new BootstrapCommand([], {});
+      const wantPgkOpts = { cwd: path.join(testDir, "packages"), stdio: STDIO_OPT, env: process.env };
```
evocateur: If we're mocking the env here, shouldn't the `env` key here be `env: mockEnv`?

evocateur:
I've got a ton of changes coming soon in another branch that affects this. As @arcticShadow notes, we're not currently breaking the env in non-registry config cases anyway.

test/BootstrapCommand.js
noherczeg: @evocateur I did that on purpose, since beforEach replaces process.env's value with the mock. It's just the syntax though, so you are right as well, doesn't matter that much.

evocateur: Hah, you're right! My brain is a bit mushy from my deep-diving into the tests

PR #743, opened 2017-04-07
Publish: Add a `diff` flag

koddsson:
## Description

This patch basically takes some lines from `DiffCommand` and executes
them if the `--diff` flag is set.

## Motivation and Context

When using lerna sometimes I want to know the code that lerna is about
to publish to help me figure out the version I would like for it to be
published as.

## How Has This Been Tested?

I some manual testing of this after looking at the tests for publish and it worked fine for me when independent publishing. I proceeded to add some tests for normal and independent mode but they are sadly both failing and I'm not sure on what's wrong with them. Any pointers here would be great 😊

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,14 @@ $ lerna publish --conventional-commits
 
 When run with this flag, `publish` will use angular conventional-commit format to [determine version bump](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-recommended-bump) and [generate CHANGELOG](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli)
 
+#### --diff
+
+```sh
+$ lerna publish --diff
+```
+
+When run with this flag, `publish` will present you with a simple `git diff` of each package before the version selection.
+
 #### --git-remote [remote]
 
 ```sh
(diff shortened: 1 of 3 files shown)


evocateur:
I don't believe saddling `lerna publish` with yet another flag is worth the added complexity, in this case. I'd rather improve the existing `lerna updated` and `lerna diff` commands to suit this purpose.

koddsson:
@evocateur No worries!

Is there a way you can think of that I can achieve what I want in lerna (which is just getting the package diff when selecting the version) or should I just hack together a bash command?

koddsson:
I'm actually not sure I can do this in a bash script. There's stuff I'd need to do like `--no-pager` so I can print the git diff output but maybe I can add those options to the diff command.

But maybe I'm the only one that wants this flag 😅

evocateur:
@koddsson Would `lerna diff <packageName>` in a separate shell work? If you pass `--no-pager`, that seems like a bad experience when you've got more than a page worth of diff...

koddsson:
Yeah I think that's probably the most simple solution. I'll close this, thanks for your time!

evocateur:
Thanks for understanding!

PR #746, opened 2017-04-07
npm init starts at 1.0.0 and so should we

krainboltgreene:
## Description

The `npm init` starts at version 1.0.0. It's better for the package environment to not ship 0.X.Y packages. So for lerna's initialization we should match their practice.

## Motivation and Context

Matching peer tool's best practice.

## Types of changes

This is a patch level change.

## Checklist:

### DIFF
--- a/src/commands/InitCommand.js
+++ b/src/commands/InitCommand.js
@@ -69,7 +69,7 @@ export default class InitCommand extends Command {
     } else if (this.repository.version) {
       version = this.repository.version;
     } else {
-      version = "0.0.0";
+      version = "1.0.0";
     }
 
     if (!this.repository.initVersion) {


evocateur:
You'll need to update the test assertions [here](https://github.com/lerna/lerna/blob/master/test/InitCommand.js#L82) and [here](https://github.com/lerna/lerna/blob/master/test/InitCommand.js#L214).

That being said, the behaviour of an inaugural `lerna publish` is a little unclear when the version is _already_ 1.0.0. Ideally, we'd have an integration test for that.

mAAdhaTTah:
Wouldn't this force the first publish to be `1.0.0`, if it increments from what's currently there?

krainboltgreene:
Your first publish should be at a minimum `1.0.0`.

mAAdhaTTah:
Sorry, typo on my end, I meant force the first publish to be `1.0.1` minimum, since it increments. My b.

krainboltgreene:
Ah, gotcha. I have no idea.

krainboltgreene:
I now see what you mean @mAAdhaTTah.

Honestly publish should *not* trigger a version as that's not how npm works and npm is a very relatable tool.

krainboltgreene:
A good example of this @evocateur is if npm fails to publish (Like with scoped packages, as they require an extra `--access=public` flag).

PR #759, opened 2017-04-13
Use execa for child_process calls

evocateur:
## Description
`execa` gives us much more robust interface for `child_process`, with a Promise API that will make async/await much easier to migrate to.

### Notables
* `execa` makes calling dependency binaries a cinch.
* Registering children takes advantage of the rich Promise API.
* Streamed output now has prefixing via `strong-log-transformer` pipes (timestamps available with an option).

## How Has This Been Tested?
Updated unit and integration tests, as well as local testing via `yarn link`.

## Types of changes
- [x] Internal

There are cleanups to the recent CLI refactoring; unfortunately they're too tightly coupled with the other changes in the PR.

## Checklist:

### DIFF
--- a/src/FileSystemUtilities.js
+++ b/src/FileSystemUtilities.js
@@ -12,6 +12,11 @@ function ensureEndsWithNewLine(string) {
   return ENDS_WITH_NEW_LINE.test(string) ? string : string + "\n";
 }
 
+// globs only return directories with a trailing slash
+function trailingSlash(filePath) {
+  return path.normalize(`${filePath}/`);
+}
+
 export default class FileSystemUtilities {
   @logger.logifyAsync()
   static mkdirp(filePath, callback) {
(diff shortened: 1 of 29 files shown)


src/FileSystemUtilities.js
```diff
@@ -59,16 +64,14 @@ export default class FileSystemUtilities {
...
-    // Checking if `filePath` exists to be removed is cheap.
+    // Checking if `dirPath` exists to be removed is cheap.
     // This lets us short-circuit if we don't have anything to do.
-    pathExists(filePath).then((exists) => {
+    pathExists(dirPath).then((exists) => {
       if (!exists) return callback();
 
-      // Note: if rimraf moves the location of its executable, this will need to be updated
-      ChildProcessUtilities.spawn(require.resolve("rimraf/bin"), [filePath], {}, callback);
+      ChildProcessUtilities.spawn("rimraf", ["--no-glob", trailingSlash(dirPath)], {}, callback);
```
dylang: This lost the `require.resolve`, so it doesn't seem to work if `rimraf` is not in your path or a top-level dependency.

evocateur:
Yeah, if you're using yarn to install lerna, it breaks compatibility with npm and does not link transitive binaries. I'll have a fix out soon.

PR #760, opened 2017-04-13
fix: pad has drop node <4, so use 1.0.x

popomore:
## Description

pad has drop node <4 [from this commit](https://github.com/adaltas/node-pad/commit/74d415db344786079710d0d20beb1b754cd207cb)

> Starting with version 1.1.0, Node pad rely on Node.js 4.0.0 or more recent. Stick to version 1.0.x if using an older version of Node.js.

but lerna 1.x can support 0.10 and 0.12, so this chnage will break. This PR will stick to version 1.0.x.

## Types of changes

## Checklist:


hzoo:
@popomore we have dropped node 0.10/0.12 as well so it's ok

popomore:
But [some project](https://github.com/twada/power-assert-runtime/pull/17) use lerna for initializing in their test, it's ok for lerna@2 to drop the version, but can you support on 1.x?

popomore:
Is travis disabled on 1.x?

evocateur:
We don't have the resources to commit to maintaining the 1.x release. It's likely broken in several other subtle ways, so this is as good a reason as any to stop promising that it will work.

Thanks for taking the time to make the pull request, @popomore!

PR #761, opened 2017-04-13
Use shell option when spawning `lerna exec`

jwb:
## Motivation and Context
The README example in the *exec* section would not supply any argument to npm view.

## How Has This Been Tested?
When I invoked the command properly, I discovered that environment variables are not expanded at all, so I enabled the shell option in the call to `spawn`.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -38,6 +38,7 @@ export default class ExecCommand extends Command {
   runCommandInPackage(pkg, callback) {
     ChildProcessUtilities.spawn(this.command, this.args, {
       cwd: pkg.location,
+      shell: true,
       env: Object.assign({}, process.env, { LERNA_PACKAGE_NAME: pkg.name })
     }, (code) => {
       if (code) {
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -3,10 +3,10 @@ import PackageUtilities from "../PackageUtilities";
 import Command from "../Command";
 
 export function handler(argv) {
-  return new ExecCommand([argv.command, ...argv.arguments], argv).run();
+  return new ExecCommand([argv.command, ...argv.args], argv).run();
 }
 
-export const command = "exec <command> [arguments..]";
+export const command = "exec <command> [args..]";
 
 export const describe = "Run an arbitrary command in each package.";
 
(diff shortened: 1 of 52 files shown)


evocateur:
Can you add an [integration test](https://github.com/lerna/lerna/blob/master/test/integration/lerna-exec.test.js) for this?

src/commands/ExecCommand.js
```diff
@@ -38,6 +38,7 @@ export default class ExecCommand extends Command {
   getOpts(pkg) {
     return {
       cwd: pkg.location,
+      shell: true,
```
evocateur: I'm not sure this property is necessary, as [execa](https://github.com/sindresorhus/execa) is handling this for us.

evocateur: `lerna exec -- npm view \$LERNA_PACKAGE_NAME` works perfectly fine as-is on `master`, for example.

jwb: It was broken with spawn...

jwb: `npm view` reports on the package in the current directory.

evocateur: Ah yes, I just reproduced it locally.

```js
  test.concurrent("$LERNA_PACKAGE_NAME", () => {
    return initFixture("ExecCommand/basic").then((cwd) => {
      const args = [
        "exec",
        "echo",
        "\$LERNA_PACKAGE_NAME",
        "--concurrency=1",
      ];

      return execa(LERNA_BIN, args, { cwd }).then((result) => {
        expect(result.stdout).toMatchSnapshot("echo: $LERNA_PACKAGE_NAME");
      });
    });
  });
```

jwb:
Added an integration test that fails on the master branch and passes with the fix.

evocateur:
oh noes Windows :P

jwb:
How should I make the test use a different value on WinDies?

evocateur:
Actually, my windows env var syntax was wrong, it should be `%LERNA_PACKAGE_NAME%` apparently

src/commands/ExecCommand.js
jwb: The integration test isn't evaluating the variables before sending the arguments to lerna.

jwb:
Thanks for all the help @evocateur!

PR #770, opened 2017-04-18
Pass multiple directories to rimraf

evocateur:
## Description
Instead of one-at-a-time, pass chunked lists of directories to `rimraf --no-glob`.

## Motivation and Context
`--no-glob` is somewhat faster, and we might as well let `rimraf` burn through a concurrency-based, exists-filtered chunk as fast as it can.

This is building up to fixing the require.resolve() stuff I broke last week, and records my attempts to get a freaking `yarn` local tarball install to bloody work.

## How Has This Been Tested?
Unit tests, new integration test, local runs.

## Types of changes
- [x] Refactoring

## Checklist:
- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #768, opened 2017-04-15
Hoisting broken in lerna@2.0.0-rc.2 ???


brentertz:
Greetings.

I am updating a project from `lerna@2.0.0-beta.38` to `lerna@2.0.0-rc.2` and have noticed that `lerna bootstrap --hoist` is taking much longer.  

The following commands were run immediately following an `npm install` at the repo root;

Using `lerna@2.0.0-beta.38` - very fast
![master](https://cloud.githubusercontent.com/assets/202773/25067027/a1d2d944-21f4-11e7-90c3-9bbd97a34216.png)

Using `lerna@2.0.0-rc.2` - very slow
![update-lerna](https://cloud.githubusercontent.com/assets/202773/25067024/9c49ba42-21f4-11e7-8159-763fd66fb6dc.png)

Using `lerna@2.0.0-rc.2`, shared dependencies for the individual packages do not appear to be getting hoisted at all, but rather each package's node_modules is full of shared dependencies.  

This is contrast to using `lerna@2.0.0-beta.38`, in which case the node_modules folder for the individual packages just contains symlinks.

## Expected Behavior

I would expect shared dependencies from individual packages to be hoisted to the main package root.

## Current Behavior

Shared dependencies from individual packages are not getting hoisted, but instead are duplicated in each package's node_modules.

## Steps to Reproduce (for bugs)

1.  Clone my repo. `git clone https://github.com/mineral-ui/mineral-ui.git`
2.  Switch to the "update-lerna" branch. `git checkout update-lerna`
3.  Install dependencies. `npm install`
4.  Run bootstrap script. `npm run bootstrap`
5.  Note how long it takes.  Also, note the contents of the `node_modules` folders for any of the packages.

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.0.0-rc.2",
  "packages": [
    "packages/*",
    "packages/components/*"
  ],
  "version": "independent"
}

```
</p></details>

## Context

Attempting to update lerna in order to take advantage of new features, such as `--conventional-commits` flag and automated changelog generation.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.0.0-rc.2 |
| `npm --version`  | 3.10.10 |
| `node --version` | v6.10.2 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS El Capitan | 10.11.6 |

evocateur:
[...]

PR #776, opened 2017-04-21
Ignore node_modules when traversing nested package locations

evocateur:
## Description
When recursing (via `**`, or "globstar") in our search for `package.json` files, we should ignore any files found under a `node_modules` directory.

## Motivation and Context
Any support we currently have for nesting package locations is accidental, and bootstrapping a "dirty" tree will definitely cause issues. I happened to encounter this on an internal repo where I was trying to bootstrap a nested tree of private packages with a valid `src/**` glob.

This also happens to validate the "managed root package" case inherited from `asini`.

## How Has This Been Tested?
A new `globstar` fixture, as well as local `lerna ls` etc.

## Types of changes

## Checklist:

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -49,6 +49,11 @@ export default class PackageUtilities {
       cwd: rootPath,
       strict: true,
       absolute: true,
+      ignore: [
+        // allow globs like "packages/**",
+        // but avoid picking up node_modules/**/package.json
+        "**/node_modules/**",
+      ],
     };
 
     packageConfigs.forEach((globPath) => {
--- a/test/PackageUtilities.js
+++ b/test/PackageUtilities.js
@@ -12,13 +12,8 @@ import PackageUtilities from "../src/PackageUtilities";
 
 describe("PackageUtilities", () => {
   describe(".getPackages()", () => {
-    let testDir;
-
-    beforeEach(() => initFixture("PackageUtilities/basic").then((dir) => {
-      testDir = dir;
-    }));
-
-    it("should collect all the packages from the given packages directory", () => {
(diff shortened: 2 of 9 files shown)


noherczeg:
Could this be a perf. improvement over the lately reported problem regarding hoisted bootstrap?

evocateur:
I don't think so, since the default package location (`packages/*`) doesn't have a globstar, and we're definitely not adding one to any provided config.

But wait! #768 has _some_ nesting (`"packages/*", "packages/components/*"`), but no globstar. So I'm not sure. Gonna go play with the example repo and check...

evocateur:
Short answer: nope. I've got some better logging in a side branch, though, and so far it seems to be totally skipping `--hoist` when I run `lerna bootstrap`...

evocateur:
@noherczeg Figured it out!

```diff
   "hoist": {
     group: "Command Options:",
     describe: "Install external dependencies matching [glob] to the repo root",
-    type: "string",
     defaultDescription: "'**'",
+    coerce: (arg) => {
+      // `--hoist` is equivalent to `--hoist=**`.
+      return arg === true ? "**" : arg;
+    },
   },
   "nohoist": {
     group: "Command Options:",
```

It was previously always resolving `argv.hoist` as `""` when `--hoist` with no argument was passed (which, being "falsey", always failed the conditional [here](https://github.com/lerna/lerna/blob/master/src/commands/BootstrapCommand.js#L305)). Would have been nice to have an integration test for hoisting before this. Ah well.

I'll PR this (mostly reverting #770, btw) after my monster logging changes.

evocateur:
@noherczeg Opened #778 to fix the `--hoist` bug.

PR #777, opened 2017-04-22
move from --skip-temp-tag to --temp-tag

### LINKED ISSUE #742: Remove `--skip-temp-tag` (or make it the default)
> --skip-temp-tag

> When activated, this flag will alter the default publish process by not creating a temporary tag and handling the process accordingly. Instead it will immediately publish with the proper dist-tag as npm it self would. 

This isn't really necessary anymore since packages are published in topological order now. As a separate issue we do need to handle publish issues, and for others that still want it, we should make the opposite flag to turn it back on. Since you may publish 1 package but the others fail so you are safe from it being on `latest`.

noherczeg:
Explicit skipping of a temp tag is no longer necessary since packages are published in topological order. Fixes #742

## Description
I removed the `--skip-temp-tag` flag from the codebase and replaced it with `--temp-tag` with a default value of `false`.

This means that if nothing is provided Lerna will not create a temp-tag.

If someone would like to create a temp-tag in the process he/she could do it by providing the `--temp-tag` flag.

## How Has This Been Tested?
Inverted the test case where temp tag creation was disabled and tested if it indeed is created when requested.

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -731,11 +731,15 @@ This is useful if you do not want to explicitly set up your registry
 configuration in all of your package.json files individually when e.g. using
 private registries.
 
-#### --skip-temp-tag
+#### --skip-temp-tag (DEPRECATED)
 
-When activated, this flag will alter the default publish process by not creating
-a temporary tag and handling the process accordingly. Instead it will immediately
-publish with the proper dist-tag as [npm it self would](https://docs.npmjs.com/cli/dist-tag).
+This flag is no longer active due to the fact that the default behavior of 
+Lerna is to publish "as-is" without a temp-tag.
+
+#### --temp-tag
+
+When activated, this flag will alter the default publish process by creating
+a temporary tag in the first place to check for [...] and later removing it.
 
 ### Wizard
 
--- a/README.md
+++ b/README.md
@@ -731,15 +731,14 @@ This is useful if you do not want to explicitly set up your registry
 configuration in all of your package.json files individually when e.g. using
 private registries.
 
-#### --skip-temp-tag (DEPRECATED)
-
-This flag is no longer active due to the fact that the default behavior of 
-Lerna is to publish "as-is" without a temp-tag.
-
 #### --temp-tag
 
-When activated, this flag will alter the default publish process by creating
-a temporary tag in the first place to check for [...] and later removing it.
+When passed, this flag will alter the default publish process by first publishing 
+all changed packages to a temporary dist-tag (`lerna-temp`) and then moving the 
+new version(s) to the default [dist-tag](https://docs.npmjs.com/cli/dist-tag) (`latest`).
+
+This is not generally necessary, as Lerna will publish packages in topological 
+order (all dependencies before dependents) by default.
 
 ### Wizard
 
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -86,9 +86,11 @@ export const builder = {
     group: "Command Options:",
     describe: "Stop before actually publishing change to npm."
   },
-  "skip-temp-tag": {
+  "temp-tag": {
     group: "Command Options:",
-    describe: "Do not create a temporary tag while publishing."
+    describe: "Create a temporary tag while publishing.",
+    type: "boolean",
+    default: false
   }
 };
 
@@ -521,7 +523,7 @@ export default class PublishCommand extends Command {
   npmPublishAsPrerelease(callback) {
     // if we skip temp tags we should tag with the proper value immediately
     // therefore no updates will be needed
-    const tag = this.options.skipTempTag ? this.getDistTag() : "lerna-temp";
+    const tag = !this.options.tempTag ? this.getDistTag() : "lerna-temp";
 
     this.updates.forEach((update) => {
       this.execScript(update.package, "prepublish");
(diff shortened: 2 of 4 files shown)


noherczeg:
I need help with the wording of the new flag in the docs. I think @hzoo or @evocateur could help out with this. Not sure about keeping the old flag with a deprecation warning is needed either. I put it in just to not confuse devs already using said feature, but it could be removed.

Also need someone to check the updated test snapshots because I'm not 100% certain that everything is okay.

README.md
```diff
@@ -731,11 +731,15 @@ This is useful if you do not want to explicitly set up your registry
 configuration in all of your package.json files individually when e.g. using
 private registries.
 
-#### --skip-temp-tag
+#### --skip-temp-tag (DEPRECATED)
```
evocateur: I think we can just replace `--skip-temp-tag` with `--use-temp-tag`, since we're breaking the default behaviour anyway. Even if `--skip-temp-tag` is passed, it won't matter because that's what we'll be doing anyway.

The description of the flag would be similar to the existing docs, but slightly tweaked:
----
#### --use-temp-tag

When passed, this flag will alter the default publish process by first publishing all changed packages to a temporary dist-tag (`lerna-temp`) and then moving the new version(s) to the default [dist-tag](https://docs.npmjs.com/cli/dist-tag) (`latest`).

This is not generally necessary, as Lerna will publish packages in topological order (all dependencies before dependents) by default.

src/commands/PublishCommand.js
```diff
@@ -86,9 +86,11 @@ export const builder = {
     group: "Command Options:",
     describe: "Stop before actually publishing change to npm."
   },
-  "skip-temp-tag": {
+  "temp-tag": {
```
evocateur: `--use-temp-tag` scans a little better, here. Since it's a boolean option, you don't need to provide an explicit `default: false` since that's what happens anyway when the option is omitted.

src/commands/PublishCommand.js
```diff
@@ -521,7 +523,7 @@ export default class PublishCommand extends Command {
   npmPublishAsPrerelease(callback) {
     // if we skip temp tags we should tag with the proper value immediately
     // therefore no updates will be needed
-    const tag = this.options.skipTempTag ? this.getDistTag() : "lerna-temp";
+    const tag = !this.options.tempTag ? this.getDistTag() : "lerna-temp";
```
evocateur: I'd prefer to reverse this ternary, for clarity.
```js
const tag = this.options.useTempTag ? "lerna-temp" : this.getDistTag();
```

test/PublishCommand.js
```diff
@@ -536,12 +536,31 @@ describe("PublishCommand", () => {
...
+          const packageName = path.basename(directory);
+
           expect(publishedTagInDirectories(testDir))
-            .toMatchSnapshot("[normal --skip-temp-tag] npm publish --tag");
+            .toMatchSnapshot("[normal --temp-tag] npm publish --tag");
 
-          expect(NpmUtilities.checkDistTag).not.toBeCalled();
-          expect(NpmUtilities.removeDistTag).not.toBeCalled();
-          expect(NpmUtilities.addDistTag).not.toBeCalled();
+          expect(NpmUtilities.checkDistTag).lastCalledWith(
```
evocateur: These `lastCalledWith()`s should be replaced by a `expect(addedDistTagInDirectories(testDir)).toMatchSnapshot("[normal --use-temp-tag] npm dist-tag add");` assertion.

test/__snapshots__/PublishCommand.js.snap
```diff
@@ -1,13 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`[independent --canary --npm-tag=next --yes --exact] npm dist-tag add 1`] = `
```
evocateur: This snapshot should be replaced by 
```
expect(publishedTagInDirectories(testDir)).toMatchSnapshot("[independent --canary --npm-tag=next --yes --exact] npm publish --tag");
```

evocateur: Likewise with the other snapshots that have turned into empty objects, those aren't very useful and should be replaced by `publishedTagInDirectories()` where it makes sense.

README.md
noherczeg: I wanted to use `--temp-tag` because yargs has a [negate-fields](https://github.com/yargs/yargs#negate-fields) feature where one could say `--no-temp-tag` and that would work better than `--no-use-temp-tag`.

If this isn't importand I can switch to the one you requested.

test/__snapshots__/PublishCommand.js.snap
noherczeg: Think I updated everything in [#4ed5845](https://github.com/lerna/lerna/pull/777/commits/4ed58452bdd92813a21ba31ae1a6d9526f3d9525), dunno why github isn't detecting it.

evocateur:
Thanks @noherczeg!

cary-smith:
Thank you for making this the default behavior! Just started using lerna and ran into this issue with our own internal npm via Nexus.

PR #779, opened 2017-04-22
Log with npmlog

evocateur:
## Description
Use battle-tested `npmlog` instead of custom decorator-based `logger` and `isTTY`-ignorant `progressBar`.

## Motivation and Context
* Lots of recent bug reports have included `lerna-debug.log` files, but unfortunately they're largely useless as the quality of logs written is less than ideal. By using `npmlog` we can turn `--loglevel=silly` into a firehose of non-truncated debugging information.
* `npmlog`'s progress bar automatically respects non-interactive terminal sessions, making progress garbage in CI logs a thing of the past.
* `npmlog` (properly) outputs to `stderr`, enabling us to provide parsimonious `stdout` (beneficial for `lerna ls` and `lerna updated`, at the very least).

## How Has This Been Tested?
Lots of updated snapshots, and copious local testing across a variety of monorepos.

## Types of changes

## Checklist:


test/integration/lerna-exec.test.js
```diff
@@ -1,70 +1,67 @@
...
-      const args = [
-        "exec",
-        "--ignore=package-1",
-        "ls",
-        "--concurrency=1",
-        "--",
-        // args to ls
-        "-1",
-      ];
+  test.concurrent("--ignore <pkg> ls -- -1", async () => {
```
evocateur: While I was at it, I converted all the integration tests to async/await.

test/integration/__snapshots__/lerna-publish.test.js.snap
```diff
@@ -83,33 +83,41 @@ Array [
...
-"Lerna __TEST_VERSION__
-Current version: 1.0.0
-Checking for updated packages...
-No tags found! Comparing with initial commit.
+exports[`stderr: updates fixed versions 1`] = `
+"lerna info version __TEST_VERSION__
+lerna info current version 1.0.0
+lerna info Checking for updated packages...
+lerna WARN No tags found!
+lerna info Comparing with initial commit.
```
evocateur: I made the `No tags found!` part of the message an actual warning, so it jumps out. No functional change.

src/Command.js
```diff
@@ -1,19 +1,23 @@
...
 import FileSystemUtilities from "./FileSystemUtilities";
 import GitUtilities from "./GitUtilities";
-import ExitHandler from "./ExitHandler";
-import progressBar from "./progressBar";
-import Repository from "./Repository";
 import PackageUtilities from "./PackageUtilities";
-import logger from "./logger";
+import Repository from "./Repository";
+import filterFlags from "./utils/filterFlags";
+import writeLogFile from "./utils/writeLogFile";
```
evocateur: I tried to resist it, but I organized all of the imports in every file, first grouped by source (dependency or local), then alphabetized(ish, in this case).

src/Command.js
```diff
@@ -1,19 +1,23 @@
...
+
+// handle log.success()
+log.addLevel("success", 3001, { fg: "green", bold: true });
 
 const DEFAULT_CONCURRENCY = 4;
 
 export const builder = {
   "loglevel": {
-    default: "info",
+    defaultDescription: "info",
```
evocateur: This is the `npmlog` default, and I didn't want to squash it during tests (where it is globally silenced).

src/Command.js
```diff
@@ -153,25 +167,26 @@ export default class Command {
 
   runValidations() {
     if (!GitUtilities.isInitialized(this.repository.rootPath)) {
-      this.logger.warn("This is not a git repository, did you already run `git init` or `lerna init`?");
+      log.error("ENOGIT", "This is not a git repository, did you already run `git init` or `lerna init`?");
```
evocateur: These are actually errors, and should be logged as such.

Issue #479, opened 2017-01-04
Error: spawnSync /bin/sh ENOBUFS during ImportCommand.execute


garthk:
```
lerna(info)    About to import 2 commits into from /tmp/xxx into packages/xxx
lerna(verbose) Successfully ran ImportCommand.initialize
lerna(verbose) Attempting running ImportCommand.execute
lerna(error)   Errored while running ImportCommand.execute
    Error: spawnSync /bin/sh ENOBUFS
        at exports._errnoException (util.js:1026:11)
        at spawnSync (child_process.js:461:20)
        at Object.execSync (child_process.js:517:13)
        at execSync (/usr/local/lib/node_modules/lerna/lib/ChildProcessUtilities.js:76:40)
        at ImportCommand.externalExecSync (/usr/local/lib/node_modules/lerna/lib/commands/ImportCommand.js:128:50)
        at /usr/local/lib/node_modules/lerna/lib/commands/ImportCommand.js:146:30
        at /usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:718:13
        at iterate (/usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:262:13)
        at async.forEachOfSeries.async.eachOfSeries (/usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:281:9)
        at _parallel (/usr/local/lib/node_modules/lerna/node_modules/async/lib/async.js:717:9)
```

Node v9.6.1 on macOS 10.12.2.

sstone1:
I hit this problem, I worked around it by modifying a value in this file:
`/usr/local/lib/node_modules/lerna/lib/ChildProcessUtilities.js`

```javascript
// maxBuffer value for running exec
var MAX_BUFFER = 500 * 1024;
```

I changed MAX_BUFFER to `500 * 1024 * 1024`.

toddself:
Also happening on Arch Linux (kernel 4.8.13-1) using node 6.9.2 but only when trying to import large (200+ commits) repos

garthk:
272 commits in this repo.

jvivs:
Seeing the same error w/ importing a package with 173 commits into a fresh lerna repo. Relevant node, npm, and git versions:
```
$ npm version
{ npm: '3.10.10',
  ares: '1.10.1-DEV',
  http_parser: '2.7.0',
  icu: '56.1',
  modules: '46',
  node: '4.6.2',
  openssl: '1.0.2j',
[...]

PR #782, opened 2017-04-23
Add --max-buffer flag

noherczeg:
Add a new flag `---max-buffer` to solve spawn issues.

Fixes: #479

## Description
Added flag with docs, and a new test. I did not add a new test to every command because the buffer issue only surfaced while devs were using the `import` command.

## How Has This Been Tested?
With a new unit test.

## Types of changes

## Checklist:

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -58,6 +58,11 @@ export const builder = {
     type: "boolean",
     default: true,
   },
+  "max-buffer": {
+    describe: "Set max-buffer(bytes) for Command execution",
+    type: "number",
+    requiresArg: true
+  }
 };
 
 export default class Command {
@@ -115,6 +120,9 @@ export default class Command {
       this._execOpts = {
         cwd: this.repository.rootPath,
       };
+      if (this.options.maxBuffer) {
+        this._execOpts.maxBuffer = this.options.maxBuffer;
+      }
     }
 
     return this._execOpts;
@@ -166,7 +174,10 @@ export default class Command {
   }
 
   runValidations() {
-    if (!GitUtilities.isInitialized(this.repository.rootPath)) {
+    const opts = Object.assign({}, this.execOpts, {
+      cwd: this.repository.rootPath
+    });
+    if (!GitUtilities.isInitialized(opts)) {
       log.error("ENOGIT", "This is not a git repository, did you already run `git init` or `lerna init`?");
       this._complete(null, 1);
       return;
--- a/src/Command.js
+++ b/src/Command.js
@@ -120,6 +120,7 @@ export default class Command {
       this._execOpts = {
         cwd: this.repository.rootPath,
       };
+
       if (this.options.maxBuffer) {
         this._execOpts.maxBuffer = this.options.maxBuffer;
       }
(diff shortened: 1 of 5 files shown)


src/Command.js
```diff
@@ -166,7 +174,10 @@ export default class Command {
   }
 
   runValidations() {
-    if (!GitUtilities.isInitialized(this.repository.rootPath)) {
+    const opts = Object.assign({}, this.execOpts, {
+      cwd: this.repository.rootPath
```
evocateur: Actually, `GitUtilities.isInitialized()` should _already_ be passed `this.execOpts`, since `this.repository.rootPath === this._execOpts.cwd` 😭 

(so, uh, that's a gaping hole in our unit tests that I caused)

test/ImportCommand.js
```diff
@@ -69,6 +75,28 @@ describe("ImportCommand", () => {
       }));
     });
 
+    it("works with --max-buffer", (done) => {
```
evocateur: How about a few simple tests for `get .execOpts` in [`test/Command.js`](https://github.com/lerna/lerna/blob/master/test/Command.js#L81) instead? The `.toposort` tests are a good model, and they would be more contextually relevant in that file.

Basically, adding this test to `ImportCommand` is implying that _only_ `ImportCommand` respects it, which is clearly not the case.

evocateur: That being said, it is clearly important to ensure that `ImportCommand` is mixing `this.execOpts` into `this.externalExecOpts` correctly. (no mocking of `ChildProcessUtilities.execSync()` required)

```js
const ONE_HUNDRED_MEGABYTES = 1000 * 1000 * 100;
const importCommand = new ImportCommand([externalDir], {
  maxBuffer: ONE_HUNDRED_MEGABYTES,
}, testDir);

// runCommand() then...
expect(importCommand.execOpts).toHaveProperty("maxBuffer", ONE_HUNDRED_MEGABYTES);
expect(importCommand.externalExecOpts).toHaveProperty("maxBuffer", ONE_HUNDRED_MEGABYTES);
```

src/Command.js
```diff
@@ -115,6 +120,9 @@ export default class Command {
       this._execOpts = {
         cwd: this.repository.rootPath,
       };
+      if (this.options.maxBuffer) {
```
evocateur: Minor nit, but I wouldn't mind a newline above this inner conditional block, it helps readability

src/Command.js
noherczeg: I'm not sure I get what you are referring to :/ `GitUtilities` only has static methods, how would it have execopts preloaded into it?

test/ImportCommand.js
noherczeg: All valid points, will update the PR ASAP!

noherczeg: I think I'm done with all your requests.

noherczeg:
Well, I can't see a thing in the logs for the failing AppVeyor build... https://ci.appveyor.com/project/lerna/lerna/build/1.0.950/job/f4po1kc9kobxojtx

PR #791, opened 2017-04-26
Fix --stream run scripts in each package in parallel

douglasduteil:
## Description

Shouldn't `runParallelBatches` runs `batches` in *parallel* ? :smirk:

## Motivation and Context

In the `--stream` context,  `lerna run watch --stream` should run `npm run watch` on each packages and wait for the user :skull: .
How ever it appers that lerna in only running the first of them waiting for it to end before running the others... but as each `watch` are blocking script, they just never end :neckbeard: 

Looking at the name of the function `runParallelBatches`, I'm guessing that the `async.serial` here is somekind of typo in favor of  `async.parallel` , is it ?

## Types of changes

## Checklist:

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -193,7 +193,7 @@ export default class PackageUtilities {
   }
 
   static runParallelBatches(batches, makeTask, concurrency, callback) {
-    async.series(batches.map((batch) => (cb) => {
+    async.parallel(batches.map((batch) => (cb) => {
       async.parallelLimit(batch.map(makeTask), concurrency, cb);
     }), callback);
   }


evocateur:
The batches are required for the topological sorting (and not _just_ in `RunCommand`, which is why this patch failed all the tests).

What _should_ happen: constrain the changes to `RunCommand`, which overrides `this.toposort` if `this.options.stream` is true. The assumption being, if you want to stream, you're probably running a watch script of some sort.

You'll still have the problem of concurrency, however, which isn't solved as easily. You can't start a dozen subprocess watchers at the same time governing hundreds of files total because you'll hit the EMFILE limit very quickly (as well as a lot of process thrashing). That's one reason I recommend "hoisting" all watcher processes to the root, so it's a single process that can juggle file handles appropriately.

evocateur:
This has been superseded by the `--parallel` flag, added in #796

douglasduteil:
Oh Great !

I will test it.
Sorry I didn't have the time to update my PR

 Thanks @evocateur  :+1:

evocateur:
@douglasduteil No worries, thanks for volunteering :)

PR #796, opened 2017-05-01
Add --parallel flag to `lerna run`

evocateur:
## Description
Provide a new flag, `--parallel`, to facilitate easier and less error-prone spawning of long-running watchers from npm scripts (e.g., `lerna run watch`).

## Motivation and Context
This allows simpler invocation of `watch` scripts, with the caveat that concurrency and topological sorting are _completely_ ignored. This is generally the intention when calling `lerna run watch` and other similar script targets, hence the additional flag.

```sh
# the following commands are equivalent
$ lerna run watch --concurrency=1000 --stream
$ lerna run watch --parallel
```

Package filtering (`--scope` and `--ignore`) is still available when this new flag is being used, and it is advised to narrow the scope of parallel execution when you have more than a dozen packages or so (YMMV).

## How Has This Been Tested?
local monorepo dev and copious unit and integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -1,3 +1,5 @@
+import async from "async";
+
 import Command from "../Command";
 import NpmUtilities from "../NpmUtilities";
 import output from "../utils/output";
@@ -14,8 +16,14 @@ export const describe = "Run an npm script in each package that contains that sc
 export const builder = {
   "stream": {
     group: "Command Options:",
-    describe: "Stream output with lines prefixed by package."
-  }
+    describe: "Stream output with lines prefixed by package.",
+    type: "boolean",
+  },
+  "parallel": {
+    group: "Command Options:",
+    describe: "Run script in all packages with unlimited concurrency, streaming prefixed output",
+    type: "boolean",
+  },
 };
 
 export default class RunCommand extends Command {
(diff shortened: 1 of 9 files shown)


src/commands/RunCommand.js
```diff
@@ -14,8 +16,14 @@ export const describe = "Run an npm script in each package that contains that sc
 export const builder = {
   "stream": {
     group: "Command Options:",
-    describe: "Stream output with lines prefixed by package."
-  }
+    describe: "Stream output with lines prefixed by package.",
+    type: "boolean",
```
evocateur: Discovered during the writing of new integration tests, without this explicit `type` one could cause a mangling of the arguments passed in by innocently re-ordering them. Obviously not very ideal.

evocateur:
Argh, I'm getting bitten by [this line](https://github.com/npm/npm/blob/d46015256941ddfff1463338e3e2f8f77624a1ff/lib/run-script.js#L139) on Windows. 😡Ⓜ️

arekkas:
Awesome, exactly what I'm looking for! Would it be possible to release this change @evocateur ? :)

evocateur:
My goal is to release 2.0.0 this week.

Issue #792, opened 2017-04-27
Lerna does not find packages in non-default locations as of rc.4 - rc.3 works fine


remotezygote:
## Expected Behavior
The expected behavior is for lerna to find/manage packages as it did before rc.4. Previous output of `lerna ls`:

```
Lerna v2.0.0-rc.3
Independent Versioning Mode
@<redacted>/<redacted>         v1.0.41 
node-readme                    v0.1.10 
@<redacted>/<redacted>         v1.0.41 
@<redacted>/<redacted>         v1.1.6  
@<redacted>/<redacted>         v1.0.55 
@<redacted>/<redacted>         v1.0.60 
@<redacted>/<redacted>         v0.0.12 
@<redacted>/<redacted>         v1.0.18 
@<redacted>/<redacted>         v1.0.25 
@<redacted>/<redacted>         v1.0.9  
@<redacted>/<redacted>         v1.0.8  
@<redacted>/<redacted>         v1.0.19 
```

## Current Behavior
Current output of `lerna ls`:

```
lerna info version 2.0.0-rc.4
lerna info versioning independent
@<redacted>/<redacted> v1.0.41 
```

## Steps to Reproduce (for bugs)
1. Create a lerna-managed repo.
2. Put your packages in a folder other than the default `packages` folder.
3. Run `lerna ls`

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.0.0-beta.38",
  "packages": [
    "./",
    "packages/node_modules/*",
    "packages/node_modules/@opallabs/*"
  ],
  "version": "independent"
}
```
</p></details>

## Context
This makes all commands work only on the "main" package that it still finds. It should perform all commands against the full list of packages.

## Your Environment
As you can see from the `lerna.json` file above, I have my packages inside a `node_modules` directory inside the `packages` directory. This was something suggested by a posting from the PouchDB team, and works well. Worked well, until rc.4, I should say :) .

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.0.0-rc.4 |
| `npm --version`  | 4.5.0 |
| `node --version` | v7.8.0 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Sierra | 10.12.3 |

evocateur:
It's not exactly accurate to say "non-default packages locations are broken", because clearly non-default packages locations that _don't_ contain the string `node_modules` are just fine (I use it literally every day, so I'd know).

[...]

PR #797, opened 2017-05-01
Add bootstrap performance section to FAQ.

noherczeg:
People using Lerna could easily miss out features like hoisting and may think that it just isn't as great as it can be. I think mentioning hoisting in as many places as we could is a must. Since it saves time for humans and machines as well.

## Description
Added hoisting section to FAQ.

## Motivation and Context
#792

## How Has This Been Tested?
Change contains only docs.

## Types of changes

## Checklist:


evocateur:
Thanks @noherczeg!

Issue #438, opened 2016-12-13
[Bootstrap] Creates .git directory


marekcirkos:
Version: 2.0.0-beta.30

Every time I call `learn bootstrap` it creates `.git` directory. Is this expected behaviour? It is a bit frustrating when working with other repos.

Is it possible to disable git integration completely? `--skip-git` does not help.

Best regards

gigabo:
Hi @marekcirkos good question!  It seems like for `bootstrap` it shouldn't be necessary.

Interested in tracking down what's causing it? 😁

marekcirkos:
I already did :). 
You initialise it `Repository` constructor https://github.com/lerna/lerna/blob/d707386f1327e09a9e8f8530fff470e9fb24838c/src/Repository.js#L11

gigabo:
Oh!  Nice find!

Well... maybe that could be more targeted.  What commands actually _need_ to initialize it? :thinking:

marekcirkos:
Lol, I guess I edited my comment during your post :)

It gets allocated in BaseCommand for every single command https://github.com/lerna/lerna/blob/master/src/Command.js#L18

natew:
Any update here? I'm trying to run lerna from Dockerfile, running into this which would require adding unnecessary git dep to docker.

noherczeg:
@natew I'm looking into this right now.

malyw:
Hey @noherczeg , thanks for revisiting this.
In my case we use Git for development, but then deliver project for the customers, which can use anything.
In that case we want to avoid "lerna botstrap" to throw an error "This is not a git repository" as it still may be fully functional without it.
I tried you solution with version 2.0.0-rc4, but seems still there is a problem:
```bash
> lerna "bootstrap" "--skip-git-check"

lerna info version 2.0.0-rc.4
lerna ERR! ENOGIT This is not a git repository, did you already run `git init` or `lerna init`?
```

evocateur:
This is fixed by #803

PR #803, opened 2017-05-03
Skip git repo check by default in Commands which do not rely on git

noherczeg:
According to Bootstrap Command's description it is not necessarily true that we need to check if the command was run in a project where there is a pre-initialized git repository. This PR tries to resolve this issue.

## Description
I added a `--skip-git-check` flag to the Bootstrap Command which is on by default resulting in a behavior described above

## Motivation and Context
#438

## How Has This Been Tested?
I added integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -129,6 +129,10 @@ export default class Command {
     return this._execOpts;
   }
 
+  get requiresGit() {
+    return true;
+  }
+
   // Override this to inherit config from another command.
   // For example `updated` inherits config from `publish`.
   get otherCommandConfigs() {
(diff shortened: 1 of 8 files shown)


src/Command.js
```diff
@@ -175,7 +175,7 @@ export default class Command {
   }
 
   runValidations() {
-    if (!GitUtilities.isInitialized(this.execOpts)) {
+    if (!this.options.skipGitCheck && !GitUtilities.isInitialized(this.execOpts)) {
```
evocateur: I'd rather this was some sort of class property optionally overridden (kinda like the `otherCommandConfigs()` getter) when a command requires git. The base class has `get requiresGit() { return true; }`, and various commands that don't need git can opt-out by overriding the getter method.

noherczeg: I'll update the PR in the evening. Thanks for the feedback!

malyw:
Hey @noherczeg , thanks for revisiting this.
In my case we use Git for development, but then deliver project for the customers, which can use anything.
In that case we want to avoid "lerna botstrap" to throw an error "This is not a git repository" as it still may be fully functional without it.
I tried you solution with version 2.0.0-rc4, but seems still there is a problem:

```
> lerna "bootstrap" "--skip-git-check"

lerna info version 2.0.0-rc.4
lerna ERR! ENOGIT This is not a git repository, did you already run `git init` or `lerna init`?
```

evocateur:
Yep, this has not been released yet.

Issue #789, opened 2017-04-26
Lerna bootstrap fails due to concurrency (with npm)


Reinmar:
The latest RC version fails when running `lerna bootstrap`. I'm not sure when it was introduced, though, because only today builds started failing on Travis while RC3 was released 8 days ago. So, I may be actually running into two separate issues. Both turned to have the same solution, though, switching off concurrency.

The issue, at least now, can be reproduced in the following way:

```
p@m /workspace/test> git clone git@github.com:ckeditor/ckeditor5.git
Cloning into 'ckeditor5'...
remote: Counting objects: 6484, done.
remote: Compressing objects: 100% (71/71), done.
remote: Total 6484 (delta 32), reused 0 (delta 0), pack-reused 6407
Receiving objects: 100% (6484/6484), 1.16 MiB | 336.00 KiB/s, done.
Resolving deltas: 100% (3672/3672), done.
p@m /workspace/test> cd ckeditor5/
(master cbe80fd) p@m /workspace/test/ckeditor5> lerna bootstrap
Lerna v2.0.0-rc.3
Bootstrapping 1 packages
Preinstalling packages
Installing external dependencies
Installing hoisted dependencies into root
Pruning hoisted dependencies
FileSystemUtilities.rename    ("./package.json", "./package.json.lerna_backup")
Error: ENOENT: no such file or directory, rename '/workspace/test/ckeditor5/package.json' -> '/workspace/test/ckeditor5/package.json.lerna_backup'
    at Error (native)
NpmUtilities.installInDir     (".", ["guppy-pre-commit@^0.4.0"], {})
Error: ENOENT: no such file or directory, rename '/workspace/test/ckeditor5/package.json' -> '/workspace/test/ckeditor5/package.json.lerna_backup'
    at Error (native)
Errored while running BootstrapCommand.execute
Error: ENOENT: no such file or directory, rename '/workspace/test/ckeditor5/package.json' -> '/workspace/test/ckeditor5/package.json.lerna_backup'
    at Error (native)
```

When I try with `--concurrency=1` the issue doesn't seem to appear:

```
(master cbe80fd) p@m /workspace/test/ckeditor5> lerna bootstrap --concurrency=1
Lerna v2.0.0-rc.3
Bootstrapping 1 packages
Preinstalling packages
Installing external dependencies
Installing hoisted dependencies into root
Pruning hoisted dependencies
Symlinking packages and binaries
Postinstalling packages
Prepublishing packages
[...]

PR #806, opened 2017-05-05
Allow custom yarn mutex from lerna.json config

ChristopheVandePoel:
** hold on, getting this explanation in order :)
This small change adds the possibility for the user to add his own mutex setting.

## Description
The current rc of lerna defaults lerna to the use of `--mutex network:424242`. 

With this commit the user can add a 'mutex' setting to lerna.json to choose for himself what type of setting he prefers.

```
{
  "lerna": "2.0.0-rc.4",
  "packages": [
    "packages/*"
  ],
  "mutex": "file:/path/to/file",
  "npmClient": "yarn",
  "version": "independent"
}
```

A limitation with this approach seems to be that the mutex has to be called with a path that is located outside the main packages, if not, the lock is created in de package dir itself and so for every package, which off course defeats the purpose. The path either needs to be absolute or located in the parent-packages directory (or higher).

## Motivation and Context
For some people (including me) working with the default network mutex does not work on every environment after a `yarn cache clean && lerna clean`. 

Ran into trouble when trying to use it on WSL and it seems that there are some docker-users that are also affected by this problem. It's not entirely clear why these environments ignore the network mutex, but they do. WSL does not seem to ignore it if you change the mutex to use a file in stead. This commit gives the user the possibility to do so.

Issue reference:
https://github.com/lerna/lerna/issues/789
and I guess this one: 
https://github.com/yarnpkg/yarn/issues/3327

## How Has This Been Tested?
Built with the changes and if a file is set, the issue seems resolved. Ran both normal and integration tests without the issue occurring. On that note, If I run the integration tests on WSL with the current rc.4, they do not pass.

Further than that, this has not been tested in-depth with a lengthy use-case. It seems that the change I made is pretty much confined to the bootstrap command, although I cannot be entirely sure if nothing else is affected here.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -46,18 +46,19 @@ export default class BootstrapCommand extends Command {
   }
 
   initialize(callback) {
-    const { registry, npmClient } = this.options;
+    const { registry, npmClient, mutex } = this.options;
 
     this.npmConfig = {
       registry,
       npmClient,
+      mutex
     };
 
     this.batchedPackages = this.toposort
       ? PackageUtilities.topologicallyBatchPackages(this.filteredPackages)
       : [this.filteredPackages];
 
-    if (npmClient === "yarn") {
+    if (npmClient === "yarn" && !mutex) {
       return getPort(42424).then((port) => {
         this.npmConfig.mutex = `network:${port}`;
         callback(null, true);
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -430,6 +430,31 @@ describe("BootstrapCommand", () => {
         }
       }));
     });
+
+    it("gets user defined mutex when --npm-client=yarn", (done) => {
+      const bootstrapCommand = new BootstrapCommand([], {
+        npmClient: "yarn",
+        mutex: "file:/test/this/path"
+      }, testDir);
+
+      bootstrapCommand.runValidations();
+      bootstrapCommand.runPreparations();
+
+      bootstrapCommand.runCommand(exitWithCode(0, (err) => {
+        if (err) return done.fail(err);
+
+        try {
+          expect(NpmUtilities.installInDir.mock.calls[0][2]).toMatchObject({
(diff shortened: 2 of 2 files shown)


evocateur:
Please fill out the relevant template parts and add a test, thanks!

ChristopheVandePoel:
Sorry, I hit ctrl+enter accidentally.

I'll add a test asap

ChristopheVandePoel:
I added a test, but did not change the integration test, since there seems to be an issue with yarn supporting file:// as in the existing comment block there.

evocateur:
I've kicked a new AppVeyor build, the failure looks oddly unrelated.

It's okay to omit the integration test for now, the unit test is plenty for this sort of thing.

evocateur:
Not that it changes the fact you've experienced this issue with the network mutex, but the mutex creation currently only _prefers_ `42424`, which `get-port` attempts to provide, falling back to a random port.

evocateur:
Would you mind adding a blurb about the `mutex` config in `lerna.json` in the [`--npm-client` block](https://github.com/lerna/lerna#--npm-client-client) of the README? I would do it myself, but I can't make edits to your clone's `master` branch. Thanks!

ChristopheVandePoel:
I'll do that as soon as I can.

FTR, I ran get-port with 42424 straight in node and it gave me back the port without issue or error. IIRC it also showed me that the yarn install --mutex network:42424 failed from the log, and the  command has to come after get-port has found an open port. 

I have some more ideas how I can test this and will try to do them later today. But what I seem to see here, is that yarn (not lerna) ignores the mutex when it is set as a network in some cases.

jwb:
Without looking at the code for this, I wonder if the problem results from get-port allocating a random port not properly passing that port to descendants.

evocateur:
No, `get-port` returns a Promise, which when resolved calls BootstrapCommand's `initialize` callback appropriately. I agree that it must be wonkiness on yarn's side. Just need some documentation of the flag and we can get this merged.

PR #726, opened 2017-04-03
Add --only-updated option to exec and run subcommands

jameslnewell:
## Description

I have added a new flag that only `run`s/`exec`s packages that have been updated. 

Works with `scope` so that only packages that match the `scope` AND that have been updated will be `run`/`exec`.

## Motivation and Context

I have a situation where package B depends on package A. I've made a breaking change to package A and want to release package A, but I don't want/have-time to update package B to work with these changes yet. When `lerna run typescript` runs on package B, it will fail because the version of package A hasn't been bumped yet and lerna is symlinking the latest package A to package B.

`lerna run --only-updated typescript` will allow me to publish package A, re-run `lerna bootstrap` to install the previous version of package A in package B (instead of a symlink) and enable me to continue developing package B independently of package A.

## How Has This Been Tested?

Minimal manual tests checking the `run` and `exec` commands, with and without the `scope` flag.

e.g.
```
lerna run --scope "my-component-*" --only-updated -- build
lerna exec --scope "my-component-*" --only-updated -- echo
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes. (there were no existing tests to copy from for `filteredPackages` - will happily add some with guidance)
- [x] All new and existing tests passed.

PR #822, opened 2017-05-18
Add --since flag to all commands that accept --scope

### LINKED ISSUE #847: `lerna bootstrap` should support --only-updated flag
## Expected Behavior
`lerna bootstrap` should respect the new `--only-updated` flag for CI purposes. `lerna bootstrap --only-updated` should only symlink packages that have updates, and install the others as external dependencies.

## Current Behavior
`lerna bootstrap` does not respect the new `--only-updated` flag

## Possible Solution
Like I mentioned above, update `lerna bootstrap` to respect the `--only-updated` flag and only symlink modules if they've been updated. Otherwise, install them as external dependencies.

## Context
We would love to run our build/test/lint on only updated packages as this would speed up our CI builds significantly. Since as of today `lerna bootstrap` does not respect the new `--only-updated` flag, when we run `lerna exec --only-updated -- yarn build` it only builds out updated packages (yay). However; when we run `lerna exec --only-updated -- yarn test` now, tests that depend on modules that did not change fail because they were not built (boo). So currently, we can only get away with building _all_ packages and running lint/test on only updated packages.

### LINKED ISSUE #855: `lerna diff` against arbitrary gitref
## Expected Behavior

`lerna diff` could accept an extra param pointing to an arbitrary gitref instead of targetting its own created semver tags.
I understand this might modify the original semantics of the command, which state that it diffs "since the last release", however the additional flag could have documentation around this change of meaning.

## Current Behavior

`lerna diff` currently targets conventionally-named semver tags it creates as part of a `publish` operation, either a single one when using unified versioning or multiple tags when using independent versioning mode, when determining which packages have been changed.

## Context

My specific use case is `--scope`-ing the running of tests and some costly operations in a CI server to the set of packages modified in a PR.
With this extra parameter it would be way easier to use lerna to determine which packages have actually been changed just for that set of changes that rarely go back all the way to the previous release.

treshugart:
## Description

The `--since` option applies to all commands that accept `--scope`.

By default `--since` will scope the command to all packages that have been updated since the latest tag. If a ref is passed (i.e. `--since ref`), then it will scope the command to all packages that have been updated since the specified ref.

## Motivation and Context

Currently we have a fairly large repo of packages (currently 68) and the builds take quite awhile to complete because when CI runs, it uses `lerna run` and that executes on all packages. This means our build only gets slower as we add more packages.

Since we only care about packages that have actually changed, we can theoretically use `--scope` to filter the packages that should be run. However, in practice, this isn't very simple because we have to figure out which packages have been updated in the branch and then pass them off to this argument. This means we have to create a separate command outside of Lerna that then proxies `lerna run --scope`.

The problem with creating a separate script means that we're no longer directly using Lerna. Any new scripts we add to packages must now be run through this separate script and it creates what seems like unnecessary indirection when Lerna already has the facilities to do this internally. We felt the most elegant way would be to create a flag like `--scope` that follows the same naming convention and can even be used in conjunction with it, but that allows the caller to also specify a ref to scope the filtered packages to. If a ref isn't provided, it defaults to the latest tag, as it already does internally.

## How Has This Been Tested?

The changes were tested at the main `Command` level in the `.runValidations()` test by adding a section for `.filteredPackages`. I felt this was the best place for a few reasons:

- It can apply to most commands and packages are filtered in the base `Command` class.
- It seemed like there were other places the behaviour of `--scope` was integration tested, but I'm not sure if that would have been too much duplication for the testing of this since it would be testing pretty much the same thing every time. This also keeps the surface area of the changes smaller.

There was one change to the `Command` test file which was required in order to get the expected packages after updating the fixtures. It appears that whole modules were being mocked via Jest but not actually being used because the mocks that were being used were being manually done later in the file in `*Each()` blocks. I ended up just removing the global mocks and everything worked.

## Types of changes

## Checklist:

## Related issues

No original issue was provided but this also:

- fixes #847
- fixes #855

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,19 +687,27 @@ $ lerna run --scope toolbar-* test
 
 #### --since [ref]
 
-When executing a script or command, only run the script or command on packages that have been updated since the specified `ref`. If `ref` is not specified, it defaults to the latest tag.
+When executing a script or command, scope the operation to packages that have been updated since the specified `ref`. If `ref` is not specified, it defaults to the latest tag.
+
+List the contents of packages that have changed since the latest tag:
 
 ```sh
 $ lerna exec --since -- ls -la
 ```
 
+Run the tests for all packages that have changed since `master`:
+
 ```
 $ lerna run test --since master
 ```
 
+List all packages that have changed since `some-branch`:
+
 ```
 $ lerna ls --since some-branch
 ```
+
+*This can be particularly useful when used in CI, if you can obtain the target branch a PR will be going into, because you can use that as the `ref` to the `--since` option. This works well for PRs going into master as well as feature branches.*
 
 #### --ignore [glob]
 
--- a/README.md
+++ b/README.md
@@ -675,6 +675,22 @@ $ lerna exec --scope my-component -- ls -la
 $ lerna run --scope toolbar-* test
 ```
 
+#### --since [ref]
+
+When executing a script or command, only run the script or command on packages that have been updated since the specified `ref`. If `ref` is not specified, it defaults to the latest tag.
+
+```sh
+$ lerna exec --since -- ls -la
+```
+
+```
+$ lerna run test --since master
+```
+
+```
+$ lerna ls --since some-branch
+```
+
 #### --ignore [glob]
 
 Excludes a subset of packages when running a command.
@@ -720,20 +736,6 @@ $ lerna bootstrap --scope my-component --include-filtered-dependencies
 $ lerna bootstrap --scope "package-*" --ignore "package-util-*" --include-filtered-dependencies
 # all package-util's will be ignored unless they are depended upon by a
 # package matched by "package-*"
-```
-
-#### --only-updated
-
-When executing a script or command, only run the script or command on packages that have been updated since the last release.
-
-A package is considered "updated" using the same rules as `lerna updated`.
-
-```sh
-$ lerna exec --only-updated -- ls -la
-```
-
-```
-$ lerna run --only-updated test
 ```
 
 #### --loglevel [silent|error|warn|success|info|verbose|silly]
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -52,34 +52,27 @@ export default class UpdatedPackagesCollector {
   collectUpdatedPackages() {
     this.logger.info("", "Checking for updated packages...");
 
-    const hasTags = GitUtilities.hasTags(this.execOpts);
+    const { execOpts, options } = this;
+    const { canary } = options;
+    let { since } = options;
 
-    if (hasTags) {
-      const tag = GitUtilities.getLastTag(this.execOpts);
-      this.logger.info("", "Comparing with tag " + tag);
-    } else {
-      this.logger.warn("", "No tags found!");
-      this.logger.info("", "Comparing with initial commit.");
-    }
+    if (GitUtilities.hasTags(execOpts)) {
+      if (canary) {
+        let currentSHA;
 
-    let commits;
+        if (canary !== true) {
+          currentSHA = canary;
+        } else {
+          currentSHA = GitUtilities.getCurrentSHA(execOpts);
+        }
 
-    if (this.options.canary) {
-      let currentSHA;
-
-      if (this.options.canary !== true) {
-        currentSHA = this.options.canary;
-      } else {
-        currentSHA = GitUtilities.getCurrentSHA(this.execOpts);
+        since = this.getAssociatedCommits(currentSHA);
+      } else if (!since) {
+        since = GitUtilities.getLastTag(execOpts);
       }
-
-      commits = this.getAssociatedCommits(currentSHA);
-    } else if (hasTags) {
-      commits = GitUtilities.describeTag(
-        GitUtilities.getLastTaggedCommitInBranch(this.execOpts),
-        this.execOpts
-      );
     }
+
+    this.logger.info("", `Comparing with ${since || "initial commit"}.`);
 
     const updatedPackages = {};
 
@@ -88,16 +81,16 @@ export default class UpdatedPackagesCollector {
       updatedPackages[pkg.name] = pkg;
     };
 
-    const forced = getForcedPackages(this.options);
+    const forced = getForcedPackages(options);
 
-    if (!hasTags || forced.has("*")) {
+    if (!since || forced.has("*")) {
       this.packages.forEach(registerUpdated);
     } else {
       this.packages.filter((pkg) => {
         if (forced.has(pkg.name)) {
           return true;
         } else {
-          return this.hasDiffSinceThatIsntIgnored(pkg, commits);
+          return this.hasDiffSinceThatIsntIgnored(pkg, since);
         }
       }).forEach(registerUpdated);
     }
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -3,7 +3,6 @@ import async from "async";
 import ChildProcessUtilities from "../ChildProcessUtilities";
 import Command from "../Command";
 import PackageUtilities from "../PackageUtilities";
-import UpdatedPackagesCollector from "../UpdatedPackagesCollector";
 
 export function handler(argv) {
   return new ExecCommand([argv.command, ...argv.args], argv).run();
@@ -19,11 +18,6 @@ export const builder = {
     describe: "Bail on exec execution when the command fails within a package",
     type: "boolean",
     default: true,
-  },
-  "only-updated": {
-    group: "Command Options:",
-    describe: "Run command in packages that have been updated since the last release only",
-    type: "boolean",
   },
   "parallel": {
     group: "Command Options:",
@@ -49,19 +43,11 @@ export default class ExecCommand extends Command {
     // don't interrupt spawned or streaming stdio
     this.logger.disableProgress();
 
-    let filteredPackages = this.filteredPackages;
-    if (this.flags.onlyUpdated) {
-      const updatedPackagesCollector = new UpdatedPackagesCollector(this);
-      const packageUpdates = updatedPackagesCollector.getUpdates();
-      filteredPackages = PackageUtilities.filterPackagesThatAreNotUpdated(
-        filteredPackages,
-        packageUpdates
-      );
-    }
+    const { filteredPackages } = this;
 
     this.batchedPackages = this.toposort
       ? PackageUtilities.topologicallyBatchPackages(filteredPackages)
-      : [this.filteredPackages];
+      : [filteredPackages];
 
     callback(null, true);
   }
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -4,7 +4,6 @@ import Command from "../Command";
 import NpmUtilities from "../NpmUtilities";
 import output from "../utils/output";
 import PackageUtilities from "../PackageUtilities";
-import UpdatedPackagesCollector from "../UpdatedPackagesCollector";
 
 export function handler(argv) {
   return new RunCommand([argv.script, ...argv.args], argv).run();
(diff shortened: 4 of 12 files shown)


evocateur:
I just merged #726, does that cover your use case?

test/Command.js
```diff
@@ -187,7 +184,70 @@ describe("Command", () => {
...
+    describe(".filteredPackages", () => {
+      let testDir;
+
+      beforeEach(() => initFixture("UpdatedCommand/basic").then((dir) => {
+        testDir = dir;
+      }));
+
+      class TestCommand extends Command {}
+
+      function cli(cmd, ...args) {
```
treshugart: I saw similar helpers in the tests for the `updated` command, however, in order to reduce the amount of changes, I decided to write a few extra lines of code rather than refactoring them out into a common lib.

test/fixtures/UpdatedCommand/basic/packages/package-5/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "package-5",
   "version": "1.0.0",
-  "private": "true",
+  "private": true,
```
treshugart: Good samaritan fix.

test/Command.js
```diff
@@ -187,7 +184,70 @@ describe("Command", () => {
...
+        cli("git", "add", ".");
+        cli("git", "commit", "-m", "test");
+
+        const { filteredPackages } = run({ scopeUpdated: "master" });
+        expect(filteredPackages.length).toEqual(2);
+        expect(filteredPackages[0].name).toEqual("package-2");
+        expect(filteredPackages[1].name).toEqual("package-3");
+      });
+
+      // it("should respect --scope and --scoped-updated when used together", () => {
```
treshugart: TODO

test/Command.js
```diff
@@ -187,7 +184,75 @@ describe("Command", () => {
...
+        return execa.sync(cmd, args, { cwd: testDir }).stdout.toString("utf-8");
+      }
+
+      function run(opts) {
+        const cmd = new TestCommand([], opts, testDir);
+        cmd.run();
+        return cmd;
+      }
+
+      it("--scope should filter packages", () => {
```
treshugart: This isn't part of the functionality that I'm adding, but I wanted to make sure this behaviour was tested in this manner, especially since I'm testing the combination of `--scope` and `--scope-updated`.

test/Command.js
```diff
@@ -187,7 +184,75 @@ describe("Command", () => {
...
+      it("--scope-updated should return packages updated since the last tag", () => {
+        cli("git", "tag", "1.0.0");
+        cli("touch", "packages/package-2/random-file");
+        cli("git", "add", ".");
+        cli("git", "commit", "-m", "test");
+
+        const { filteredPackages } = run({ scopeUpdated: "" });
+        expect(filteredPackages.length).toEqual(2);
+        expect(filteredPackages[0].name).toEqual("package-2");
+        expect(filteredPackages[1].name).toEqual("package-3");
```
treshugart: Why is it that `package-3` appears in the list when `package-2` has changed? This is the default behaviour of the `UpdatedPackagesCollector`.

It would appear that packages that have `devDependencies` on packages that have changed get added to the list. This is more of a question than anything else, I guess, so that I understand it better.

treshugart:
@evocateur 

This is the same as `only-updated` in that it acts as a flag if no ref is provided. However, it differs in that it allows you to provide a ref. For our use case, we'd probably just use `master` as the target since we can't currently get the target branch for a PR in BitBucket Pipelines. However, Travis does allow this, so one would probably run it against `TRAVIS_BRANCH` in a PR and leave it off if run from `master`.

With the current functionality, it would be very easy to have a build run against packages that actually haven't changed if you don't release on every commit or push, for whatever reason be it feature branches that eventually go into master or if you accumulate commits in master and release every so often.

treshugart:
I spoke to @jameslnewell about this internally, and he seemed to think that being able to specify a `ref` would be useful, too. Maybe we could work on merging the two ideas into this?

evocateur:
I'm cool with `--only-updated=[sinceRef]` or a more explicit `--since=ref`, sure.

We definitely want to change the `this.flags.onlyUpdated` references to `this.options.onlyUpdated` to support "durable" config settings in lerna.json, I missed this during the previous PR.

treshugart:
Okay, cool. Would you prefer I bring in latest master and rework the `--only-updated` flag in with this functionality (so we don't lose the extra tests with the first PR)? Or do you want to revert it and then use only this one?

Either way, I'll do the following:

1. Rename `--scope-updated` to `--since`
2. Squash back to a single commit
3. Add docs (I intentionally kept this out until everything was hashed out)

evocateur:
Rebase and squash is fine, I like new tests :)

treshugart:
Merged and squashed commits after merge. Will annotate updates.

treshugart:
Hmm, I don't think that rebase went as expected. I'll fix up.

README.md
```diff
@@ -675,6 +675,22 @@ $ lerna exec --scope my-component -- ls -la
 $ lerna run --scope toolbar-* test
 ```
 
+#### --since [ref]
```
treshugart: I moved this to just after the `--scope` arg since it relates very closely to it.

README.md
```diff
@@ -675,6 +675,22 @@ $ lerna exec --scope my-component -- ls -la
...
+```sh
+$ lerna exec --since -- ls -la
+```
+
+```
+$ lerna run test --since master
+```
+
+```
+$ lerna ls --since some-branch
```
treshugart: I noticed while doing my updates that `--since` when used with `lerna ls` seems to make `lerna updated` redundant (except that one shows versions and one doesn't).

src/UpdatedPackagesCollector.js
```diff
@@ -52,52 +52,45 @@ export default class UpdatedPackagesCollector {
   collectUpdatedPackages() {
     this.logger.info("", "Checking for updated packages...");
 
-    const hasTags = GitUtilities.hasTags(this.execOpts);
+    const { execOpts, options } = this;
```
treshugart: Boyscout: Instead of accessing on `this` or `options` multiple times, I destructured here.

src/UpdatedPackagesCollector.js
```diff
@@ -52,52 +52,45 @@ export default class UpdatedPackagesCollector {
...
+    let { since } = options;
 
-    if (hasTags) {
-      const tag = GitUtilities.getLastTag(this.execOpts);
-      this.logger.info("", "Comparing with tag " + tag);
-    } else {
-      this.logger.warn("", "No tags found!");
-      this.logger.info("", "Comparing with initial commit.");
-    }
+    if (GitUtilities.hasTags(execOpts)) {
```
treshugart: While this is a boyscout, it's also a fix for the lines I added to the test to make sure that `since` overrides the latest tag, if there are tags present.

src/UpdatedPackagesCollector.js
```diff
@@ -52,52 +52,45 @@ export default class UpdatedPackagesCollector {
...
-
-      commits = this.getAssociatedCommits(currentSHA);
-    } else if (hasTags) {
-      commits = GitUtilities.describeTag(
-        GitUtilities.getLastTaggedCommitInBranch(this.execOpts),
-        this.execOpts
-      );
     }
 
+    this.logger.info("", `Comparing with ${since || "initial commit"}.`);
```
treshugart: @evocateur (mentioning since I want to place special emphasis here), I removed the warning if no tags exist because it doesn't really feel like a warning, especially with the addition of `--since` as an empty arg is implicitly the latest tag (or initial commit).

If there's any contention here, or it's a breaking change, I'm more than happy to add it back in.

test/Command.js
```diff
@@ -187,7 +184,83 @@ describe("Command", () => {
...
+        cli("git", "commit", "-m", "test");
+
+        const { filteredPackages } = run({ since: "" });
+        expect(filteredPackages.length).toEqual(2);
+        expect(filteredPackages[0].name).toEqual("package-2");
+        expect(filteredPackages[1].name).toEqual("package-3");
+      });
+
+      it("--since \"ref\" should return packages updated since the specified ref", () => {
+        // We first tag, then modify master to ensure that specifying --since will override checking against
```
treshugart: This was the addition that caught the bug where the latest tag would override the specified `ref` in `--since`. I annotated the spot above that fixes it.

src/commands/ExecCommand.js
```diff
@@ -49,19 +43,11 @@ export default class ExecCommand extends Command {
...
-    let filteredPackages = this.filteredPackages;
-    if (this.flags.onlyUpdated) {
-      const updatedPackagesCollector = new UpdatedPackagesCollector(this);
-      const packageUpdates = updatedPackagesCollector.getUpdates();
-      filteredPackages = PackageUtilities.filterPackagesThatAreNotUpdated(
-        filteredPackages,
-        packageUpdates
-      );
-    }
+    const { filteredPackages } = this;
```
treshugart: Boyscout: kept assignment but destructured and converted to `const` instead.

src/commands/ExecCommand.js
```diff
@@ -49,19 +43,11 @@ export default class ExecCommand extends Command {
...
 
     this.batchedPackages = this.toposort
       ? PackageUtilities.topologicallyBatchPackages(filteredPackages)
-      : [this.filteredPackages];
+      : [filteredPackages];
```
treshugart: Since we're not filtering anymore, we can use the `const`.

src/commands/RunCommand.js
```diff
@@ -46,15 +40,7 @@ export default class RunCommand extends Command {
...
-    let filteredPackages = this.filteredPackages;
-    if (this.flags.onlyUpdated) {
-      const updatedPackagesCollector = new UpdatedPackagesCollector(this);
-      const packageUpdates = updatedPackagesCollector.getUpdates();
-      filteredPackages = PackageUtilities.filterPackagesThatAreNotUpdated(
-        filteredPackages,
-        packageUpdates
-      );
-    }
+    const { filteredPackages } = this;
```
treshugart: Boyscout: kept assignment but destructured and converted to `const` instead.

evocateur:
@treshugart Just a quick note to say I haven't forgotten you, I'm just swamped at work and will try to carve out some time soon!

treshugart:
@evocateur ready for re-review. Not sure why AppVeyor is failing. I can't seem to spot the error, but it looks like something might have timed out? I ran the tests locally and everything seemed fine. I'll also have someone from my team review, too.

chrisui:
This looks great! I just opened an issue requesting this elsewhere. I also mentioned the use of a `--until` option (or `--git-to` in that issue) which might be good to add here as well?

See https://github.com/lerna/lerna/issues/876

treshugart:
@chrisui I personally don't have a use case for it (yet?), but it sounds like it could be useful. I would vote for keeping this PR as is with a plan to implement `--until` (or whatever) after it lands.

treshugart:
Hey @evocateur, I hate to be a pest, but I was hoping I could get a review / merge soon. We're currently running off of my fork and it seems that quite a few others need this as well. Cheers for the time you've spent already.

test/Command.js
evocateur: Ideally, this would be the approach used for all unit tests of commands. The `.run()` method really needs to return a Promise, already... (out of scope for this PR, clearly)

README.md
```diff
@@ -681,6 +681,30 @@ $ lerna exec --scope my-component -- ls -la
...
 ```
 
+#### --since [ref]
+
+When executing a script or command, scope the operation to packages that have been updated since the specified `ref`. If `ref` is not specified, it defaults to the latest tag.
+
+List the contents of packages that have changed since the latest tag:
+
+```sh
+$ lerna exec --since -- ls -la
```
evocateur: My only complaint with this usage is that it isn't as readable as `lerna exec --only-updated`. The nature of the beast, I suppose. (Aliases that are distinct long options are a bit awkward, and probably better to omit entirely).

treshugart:
No problem. Totally understand. Thanks for the review and merge!

chrisui:
This is great 👍 good work!

PR #824, opened 2017-05-19
Add json output to `ls` and `updated` commands

ricky:
## Description
Adds a `--json` flag, which is currently used only by `lerna ls` and `lerna updated` to produce machine-readable output.

## Motivation and Context
Inspired by the discussion in https://github.com/lerna/lerna/pull/686 and my own desire to not have our artisanally-crafted script be reliant on the plain text formatting of `lerna updated` moving forward.

## How Has This Been Tested?
Manually via invocations of the command with and without the `--json` flag, as well as new test cases.

## Types of changes

## Checklist:

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -62,6 +62,14 @@ export const builder = {
     describe: "Set max-buffer(bytes) for Command execution",
     type: "number",
     requiresArg: true
+  },
+  "json": {
+    describe: dedent`
+      Returns output as a machine-readable json object
+      (Only for 'ls' and 'updated' commands)
+    `,
+    type: "boolean",
+    default: false
   }
 };
 
@@ -69,6 +77,11 @@ export default class Command {
   constructor(input, flags, cwd) {
     log.pause();
     log.heading = "lerna";
+
+    // Default to silent log level for JSON output
+    if (flags.json) {
+      log.level = "silent";
+    }
 
     if (flags.loglevel) {
       log.level = flags.loglevel;
--- a/src/Command.js
+++ b/src/Command.js
@@ -62,14 +62,6 @@ export const builder = {
     describe: "Set max-buffer(bytes) for Command execution",
     type: "number",
     requiresArg: true
-  },
-  "json": {
-    describe: dedent`
-      Returns output as a machine-readable json object
-      (Only for 'ls' and 'updated' commands)
-    `,
-    type: "boolean",
-    default: false
   }
 };
 
@@ -77,11 +69,6 @@ export default class Command {
   constructor(input, flags, cwd) {
     log.pause();
     log.heading = "lerna";
-
-    // Default to silent log level for JSON output
-    if (flags.json) {
-      log.level = "silent";
-    }
 
     if (flags.loglevel) {
       log.level = flags.loglevel;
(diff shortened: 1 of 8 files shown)


ricky:
Node 4 test failure(s) _appear_ to be unrelated... 👀

test/UpdatedCommand.js
```diff
@@ -353,6 +353,45 @@ describe("UpdatedCommand", () => {
...
+      }, testDir);
+
+      updatedCommand.runValidations();
+      updatedCommand.runPreparations();
+
+      updatedCommand.runCommand(exitWithCode(0, (err) => {
+        if (err) return done.fail(err);
+        try {
+          const outputStr = consoleOutput();
+          expect(outputStr).toMatchSnapshot();
```
evocateur: I think we can just skip straight to the `JSON.parse(consoleOutput())` snapshot, as we really shouldn't be testing whether or not `JSON.stringify()` is working correctly.

src/Command.js
```diff
@@ -70,6 +78,11 @@ export default class Command {
     log.pause();
     log.heading = "lerna";
 
+    // Default to silent log level for JSON output
+    if (flags.json) {
```
evocateur: This isn't necessary, actually, as all logging is emitted to `stderr` and only the JSON string is emitted to `stdout`. When you pipe the output, it only includes `stdout` (unless you do fancy stdio redirects, of course).

It's too bad I didn't write a `lerna ls` integration test that could demonstrate this :/

src/Command.js
```diff
@@ -62,6 +62,14 @@ export const builder = {
...
     type: "number",
     requiresArg: true
+  },
+  "json": {
+    describe: dedent`
+      Returns output as a machine-readable json object
+      (Only for 'ls' and 'updated' commands)
+    `,
+    type: "boolean",
+    default: false
```
evocateur: This default is implicit for boolean options.

src/Command.js
```diff
@@ -62,6 +62,14 @@ export const builder = {
     describe: "Set max-buffer(bytes) for Command execution",
     type: "number",
     requiresArg: true
+  },
+  "json": {
+    describe: dedent`
```
evocateur: Let's move the `json` option descriptor into `LsCommand` and `UpdatedCommand` directly. It's an acceptable amount of duplication to maintain the benefit of targeted, relevant `--help` output. The description could be shorter, like "Show information in JSON format". (when in doubt, i copy npm's [docs](https://docs.npmjs.com/cli/ls#json))

evocateur:
And yes, the node v4 failure was due to a timeout. That's unfortunate, and hopefully rare?

src/Command.js
ricky: Fixed in ceb1fe0

src/Command.js
ricky: ceb1fe0

src/Command.js
ricky: Removed in ceb1fe0

test/UpdatedCommand.js
ricky: Fixed in 73f492a

Issue #501, opened 2017-01-16
[question] Using yarn commands with non-published lerna dependencies


timoxley:
Is it possible to use `npm`/`yarn` with `lerna` if you've got packages which are not published to any registry?

e.g.

* `./packages/A` depends on `B`, found in `./packages/B`.
* `B` exists solely for code sharing, is just an internal package. 
* edit: No intention to publish either `A` nor `B` to any registry. 

Currently, if you run any `yarn` command in `./packages/A` e.g. `yarn remove express` the command will fail because `yarn` looks for `B` and `B` isn't published in any registry, thus `error Couldn't find package "B" on the "npm" registry.`

Is there a way around this? ~~A way to specify `lerna` dependencies outside of "dependencies" perhaps? Maybe I missed something in the instructions. Perhaps this use-case just isn't supported.~~

edit: this seems like a `yarn` shortcoming really.

At the moment my workaround is terrible; I'm manually removing all `lerna` managed dependencies from `package.json` just prior to running commands like `yarn add`, then adding them back afterwards. This is truly not great.

timoxley:
Ideally `lerna bootstrap` would know how to symlink file dependencies:

```json
{
  "name": "A"
  "dependencies": {
    "B": "file:../B"
  }
}
```

kschzt:
FYI this seems to work correctly with npm.

evocateur:
If `B` is never published (a result of adding `"private": true` to its `package.json`), it certainly should never be specified in `A`'s `dependencies`, as any consumer of `A` will never install correctly.

Adding `B` to `A`'s `devDependencies` should be fine, and it is not necessary to specify the `file:` url to get the symlinking for free (the version just has to satisfy `version` in the root `lerna.json`).

timoxley:
> FYI this seems to work correctly with npm.

@kschzt sorry can you elaborate?

> any consumer of A will never install correctly.

@evocateur Note there are no consumers of any of these packages other than themselves, they're not shared libraries, they're all components of a single app and internal to the monorepo.

timoxley:
Also relevant to this issue is this npm/yarn RFC for changing `file:` semantics and adding `link:`  dependency specifiers: 

* https://github.com/npm/npm/pull/15900
* https://github.com/yarnpkg/rfcs/pull/34

kylecordes:
I've found the answer to the initial question to be 100% yes - I'm using it that way in the following.

https://github.com/OasisDigital/many-to-many-angular

[...]

PR #832, opened 2017-05-22
Make `lerna bootstrap` work when using local dependencies

trotzig:
## Description

I use local dependencies in one of my lerna projects (e.g. `"foo": "file:../foo"`). When `lerna bootstrap` is run in this project, I get an error:
```
    lerna info Installing external dependencies
    lerna ERR! execute Error: Command failed: npm install
    lerna ERR! execute npm ERR! addLocal Could not install /<redacted>/lerna-no-name-repro
/packages/a
    lerna ERR! execute npm ERR! Darwin 15.6.0
    lerna ERR! execute npm ERR! argv "/<redacted>/.nodenv/versions/6.1.0/bin/node" "/<reda
cted>/.nodenv/versions/6.1.0/bin/npm" "install"
    lerna ERR! execute npm ERR! node v6.1.0
    lerna ERR! execute npm ERR! npm  v3.8.6
    lerna ERR! execute
    lerna ERR! execute npm ERR! No name provided in package.json
```

If I switch `npmClient` to `yarn`, I get a different error:

```
lerna info Installing external dependencies
lerna ERR! execute Error: Command failed: yarn install --mutex network:42424
lerna ERR! execute warning No license field
lerna ERR! execute error Package "a@" doesn't have a "version".
```

The reason I use local dependencies is that I have packages that are never meant to be published. They're merely there as a simple way of sharing code.

## How Has This Been Tested?
Apart from updating existing unit tests, I created a minimal repro project to test things out:
https://github.com/trotzig/lerna-no-name-repro
This repo contains (I believe) the bare minimum setup to trigger the issue. One thing I found when creating this repo was that there is a race condition somewhere. The bug will only surface if a package 'A that some other package `B depends on hasn't finished installing its node_modules before B tries to resolve its dependencies. This is why the repro repository has a seemingly unrelated dependency ("react") specified for one of the packages.

## Types of changes

## Checklist:

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -33,6 +33,10 @@ export default class NpmUtilities {
     const packageJson = path.join(directory, "package.json");
     const packageJsonBkp = packageJson + ".lerna_backup";
 
+    // Grab the original package.json so that we can inject a few properties to
+    // the temporary fake replacement.
+    const originalPackage = JSON.parse(FileSystemUtilities.readFileSync(packageJson));
+
     log.silly("installInDir", "backup", packageJson);
     FileSystemUtilities.rename(packageJson, packageJsonBkp, (err) => {
       if (err) {
@@ -58,6 +62,8 @@ export default class NpmUtilities {
 
       // Construct a basic fake package.json with just the deps we need to install.
       const tempJson = {
+        name: originalPackage.name,
+        version: originalPackage.version,
         dependencies: dependencies.reduce((deps, dep) => {
           const [pkg, version] = splitVersion(dep);
           deps[pkg] = version || "*";
--- a/test/NpmUtilities.js
+++ b/test/NpmUtilities.js
@@ -239,6 +239,10 @@ describe("NpmUtilities", () => {
       stubExecOpts();
       ChildProcessUtilities.exec.mockImplementation(callbackSuccess);
       FileSystemUtilities.rename.mockImplementation(callbackSuccess);
+      FileSystemUtilities.readFileSync.mockImplementation(() => (`{
+        "name": "foo",
+        "version": "2.1.2"
+      }`));
       writePkg.mockImplementation(() => Promise.resolve());
     });
 
@@ -258,6 +262,8 @@ describe("NpmUtilities", () => {
         if (err) return done.fail(err);
 
         try {
+          expect(FileSystemUtilities.readFileSync).lastCalledWith(
+            path.join(directory, "package.json"));
           expect(FileSystemUtilities.rename).lastCalledWith(
             path.join(directory, "package.json"),
             path.join(directory, "package.json.lerna_backup"),
@@ -276,6 +282,8 @@ describe("NpmUtilities", () => {
                 caret: "^1.0.0",
                 exact: "1.0.0",
               },
+              name: "foo",
+              version: "2.1.2",
             },
           );
           expect(ChildProcessUtilities.exec).lastCalledWith("npm", ["install"], {
@@ -311,6 +319,8 @@ describe("NpmUtilities", () => {
                 "@scoped/tagged": "next",
                 tagged: "next",
               },
+              name: "foo",
+              version: "2.1.2",
             },
           );
           expect(ChildProcessUtilities.exec.mock.calls[0][2]).toMatchObject({
@@ -343,6 +353,8 @@ describe("NpmUtilities", () => {
                 "@scoped/foo": "latest",
                 foo: "latest",
               },
+              name: "foo",
+              version: "2.1.2",
             },
           );
           expect(ChildProcessUtilities.exec).lastCalledWith(
@@ -383,6 +395,8 @@ describe("NpmUtilities", () => {
                 "@scoped/something": "github:foo/bar",
                 something: "github:foo/foo",
               },
+              name: "foo",
+              version: "2.1.2",
             },
           );
           expect(ChildProcessUtilities.exec).lastCalledWith(
@@ -424,6 +438,8 @@ describe("NpmUtilities", () => {
                 "@scoped/something": "github:foo/bar",
                 something: "github:foo/foo",
               },
+              name: "foo",
+              version: "2.1.2",
             },
           );
           expect(ChildProcessUtilities.exec).lastCalledWith(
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
   ],
   "scripts": {
     "build": "babel src -d lib",
-    "dev": "babel -w src -d lib",
+    "dev": "npm run build -- --watch",
```
trotzig: This change is unrelated to the main change in this PR. I decided it was too small to merit being pushed as a separate PR, let me know if you disagree.

src/NpmUtilities.js
```diff
@@ -33,6 +33,10 @@ export default class NpmUtilities {
     const packageJson = path.join(directory, "package.json");
     const packageJsonBkp = packageJson + ".lerna_backup";
 
+    // Grab the original package.json so that we can inject a few properties to
+    // the temporary fake replacement.
+    const originalPackage = JSON.parse(FileSystemUtilities.readFileSync(packageJson));
```
trotzig: I initially had a regular `require(packageJson)` instead of reading the file through FileSystemUtilites. But that was hard to test, plus I wasn't fully sure what the side-effects would be (in terms of caching etc).

package.json
evocateur: I appreciate this desire, and I use the `--` pattern all the time with npm scripts, but I don't think it's appropriate here. The original command isn't that long or complicated, and despite a counterexample just two lines below this ("fix"), I'd rather not create another nested npm subprocess for a persistent, long-running command.

src/NpmUtilities.js
evocateur: Definitely prefer a library method instead of `require`.

src/NpmUtilities.js
```diff
@@ -55,9 +59,10 @@ export default class NpmUtilities {
         unregister();
         callback(err);
       };
-
```
evocateur: newlines are our friends

package.json
trotzig: Fair point, I hadn't thought about the nesting much. I'll remove this commit.

src/NpmUtilities.js
trotzig: Whoops, this was a mistake. I'll put it back.

trotzig:
Thanks for the quick review! I agree that this project isn't the best fit for lerna. However, we find the bootstrapping process is helping us out, especially after exploring the `--hoist` option. 

We could move to private packages, although it feels a little overkill since all the dependencies on the packages are in the same repo. So there's no real need to publish them, we just want a good internal structure.

evocateur:
I think you'll find the hoisted bootstrap process goes _even quicker_ when you employ lerna's default symlinking strategy (the deps just need to have a compatible version, like caret same-major).

evocateur:
(also, boy am I getting tired of the integration tests timing out on AppVeyor 😠 )

trotzig:
> I think you'll find the hoisted bootstrap process goes even quicker when you employ lerna's default symlinking strategy

Yeah, that's probably true. 

I should back up a few steps as well and explain why I even started exploring using `file:../` dependencies. We were already using the default symlinking strategy that lerna gave us. The packages were never published to any registry, and things worked fine. `lerna bootstrap` was doing its thing, hoisting seemed to work (although we never explored it fully). Outside of lerna, we use yarn, and it was because of how `yarn add/remove` would choke on any package not being a public one that I went down the path of local dependencies (unlike `npm`, `yarn` will check to make sure all dependencies are available on install). 

I'll leave it up to you whether you want to merge this or not. We're an outlier in terms of project setups, and we can definitely find a different setup that avoids the issue fixed in this PR. But I can't see much harm of making the temporary package.json slightly more realistic. 

Thanks for all the great work done in this project. Much appreciated. ❤️

trotzig:
After further exploring using local package dependencies (e.g. `"foo": "file:../foo") I've realized they're not a good fit for us, mainly because local dependencies are copied into the local `node_modules` folder on `npm install`, making local development hard because changes in depended-on packages aren't automatically visible to other packages. So we're back to using the default symlinking behavior from Lerna. 

I've worked around the issue with using `yarn add`/`yarn remove` by simply not using it, instead either manually editing the package.json file or using `npm install --save`. I've also removed all the individual `yarn.lock` files (expect the root one) because they were very hard to keep stable when using the hoisting option for Lerna. 

So right now this patch isn't needed by us. I'll leave it to you to close it however. For the (likely very) few projects out there using local dependencies, this is still a useful fix.

evocateur:
I appreciate the follow-up. I think in the long-term, clients like npm 5 and a future version of yarn will make this symlink vs relative path dance less frustrating overall.

In the mean time, it probably wouldn't hurt to leave more of the original manifest intact, especially if we want to support some of the install filtering flags like `--no-optional` or `--production`. We can address that elsewhere, however.

trompx:
Hello @evocateur,

After reading a lot of threads and browsing tenth of boilerplate repos, I have hard time to figure what is the best way to import local packages with the current version of lerna.
I have a monorepo with multiple websites, all packages are private (will probably never published them) so basically I could just import with relative path, but just like trotzig, I like the way lerna mutualize the common dependencies in the root node_modules and speed up my setup.

I followed the solved issues at the end of https://github.com/lerna/lerna/issues/501 but could not find a correct setup.

When you say 
> employ lerna's default symlinking strategy

do you mean, set package in dependencies with a fixed version like:
```
"dependencies": {
    "package-a": "^0.0.0"
}
```

Or the way to go is:
```
"dependencies": {
    "package-a": "file:../package-a"
}
```

A note on private packages and relative path would be great on the main readme.
Thank you for this great tool anyway!

evocateur:
The relative file: specifiers would be dramatically faster with npm 5+, no bootstrap required.

trompx:
Sorry but I don't get it. If I don't bootstrap, I will not get any symlink if multiple packages use the same external library (like react, node-sass, immutable etc.) thus installing them in the local node_modules instead of the root one.
What would be the command to get symlink for both external and local packages while using the file path for the local one (`"package-a": "file:../package-a"`), is `lerna bootstrap --hoist` all I need?

evocateur:
Local file: specifiers, then `npm i packages/*` (or if you have a single “entry” module, just that one is sufficient). It’s as easy as that. Lerna itself does it.

Issue #813, opened 2017-05-10
[Feature request] pass npmClient arguments


DxCx:
Hey,

I've checked the code and it's seems to be not supported yet, but i want to be able to pass arguments to npmClient, so for example, with npmClient `yarn` i can use `--prefer-offline` and `--pure-lockfile`

PR #834, opened 2017-05-22
Pass extra arguments to npmClient during bootstrap

xaka:
## Description

Allow to pass npmClient arguments.

## Motivation and Context
https://github.com/lerna/lerna/issues/813

## How Has This Been Tested?
```sh
# npm version
{ lerna: '2.0.0-rc.4',
  npm: '3.10.10',
  ares: '1.10.1-DEV',
  http_parser: '2.7.0',
  icu: '57.1',
  modules: '48',
  node: '6.9.4',
  openssl: '1.0.2j',
  uv: '1.9.1',
  v8: '5.1.281.89',
  zlib: '1.2.8' }

# uname -a
Darwin pstrashkin-mbp 15.6.0 Darwin Kernel Version 15.6.0: Tue Apr 11 16:00:51 PDT 2017; root:xnu-3248.60.11.5.3~1/RELEASE_X86_64 x86_64
```

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,23 @@ When run, this command will:
 2. Symlink together all Lerna `packages` that are dependencies of each other.
 3. `npm prepublish` all bootstrapped packages.
 
-`lerna bootstrap` respects the `--ignore`, `--scope` and `--include-filtered-dependencies` flags (see [Flags](#flags)).
+`lerna bootstrap` respects the `--ignore`, `--scope` and `--include-filtered-dependencies` flags (see [Flags](#flags)). 
+
+Pass extra arguments to npm client by placing them after `--`:
+
+```sh
+$ lerna bootstrap -- --production --no-optional
+```
+
+May also be configured in `lerna.json`:
+
+```js
+{
+  ...
+  "npmClient": "yarn",
+  "npmClientArgs": ["--production", "--no-optional"]
+}
+```
 
 #### How `bootstrap` works
 
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -10,10 +10,10 @@ import NpmUtilities from "../NpmUtilities";
 import PackageUtilities from "../PackageUtilities";
 
 export function handler(argv) {
-  return new BootstrapCommand(argv._, argv).run();
+  return new BootstrapCommand([...argv.args], argv).run();
 }
 
-export const command = "bootstrap";
+export const command = "bootstrap [args..]";
 
 export const describe = "Link local packages together and install remaining package dependencies";
 
@@ -46,12 +46,18 @@ export default class BootstrapCommand extends Command {
   }
 
   initialize(callback) {
-    const { registry, npmClient } = this.options;
+    const { registry, npmClient, npmClientArgs } = this.options;
 
     this.npmConfig = {
       registry,
       npmClient,
+      npmClientArgs,
     };
+
+    // lerna bootstrap ... -- <input>
+    if (this.input.length) {
+      this.npmConfig.npmClientArgs = [...(npmClientArgs || []), ...this.input];
+    }
 
     this.batchedPackages = this.toposort
       ? PackageUtilities.topologicallyBatchPackages(this.filteredPackages)
--- a/test/integration/__snapshots__/lerna-bootstrap.test.js.snap
+++ b/test/integration/__snapshots__/lerna-bootstrap.test.js.snap
@@ -56,6 +56,10 @@ package-3 cli1 OK
 package-3 cli2 package-2 OK"
 `;
 
+exports[`passes remaining arguments + npmClientArgs to npm client 1`] = `"install,--production,--no-optional"`;
+
+exports[`passes remaining arguments to npm client 1`] = `"install,--no-optional"`;
+
 exports[`simple: stderr 1`] = `
 "lerna info version __TEST_VERSION__
 lerna info Bootstrapping 4 packages
--- a/test/integration/lerna-bootstrap.test.js
+++ b/test/integration/lerna-bootstrap.test.js
@@ -1,8 +1,10 @@
 import execa from "execa";
+import fs from "fs-extra";
 import getPort from "get-port";
 import globby from "globby";
-import tempy from "tempy";
 import normalizePath from "normalize-path";
+import path from "path";
+import tempy from "tempy";
 
 import { LERNA_BIN } from "../helpers/constants";
 import initFixture from "../helpers/initFixture";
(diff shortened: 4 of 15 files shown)


README.md
```diff
@@ -790,6 +790,24 @@ May also be configured in `lerna.json`:
 }
 ```
 
+#### --npm-client-args [client]
+
+Pass extra arguments to npm client.
+
+```sh
+$ lerna bootstrap --npm-client=yarn --npm-client-args=--ignore-engine
```
evocateur: How do you pass multiple flags to `--npm-client-args`?

README.md
```diff
@@ -790,6 +790,24 @@ May also be configured in `lerna.json`:
 }
 ```
 
+#### --npm-client-args [client]
```
evocateur: This needs to be defined for yargs in the `builder` export of BootstrapCommand.

Even better, why not just accept all args after a `--` to `lerna bootstrap` to be passed to the client execution?

```
lerna bootstrap -- --ignore-engine --silent
```

To pass it via lerna.json, you'd need to pass an array of strings instead of a string:
```json
"npmClientArgs": ["--ignore-engine", "--silent"]
```

README.md
xaka: As `--npm-client-args` is just a string and simply proxies everything to underlying npm client, it'll be:

```
lerna bootstrap --npm-client=yarn --npm-client-args='--ignore-engine --flag1 --flag2 ...'
```

README.md
xaka: Do we want to isolate it to `bootstrap` command only? I thought it should be for all of them regardless of where npm client is used.

README.md
evocateur: Unfortunately that's not how `execa` works, you need to pass an array of arguments (or push multiple, in this case) if there are multiple arguments.

README.md
evocateur: `bootstrap` is currently the only place we alternate on the value of `npmClient`, and it seems cleaner that way (supporting multiple arguments properly without re-implementing yargs, for example)

xaka: It seems accepting all args after `--` is the lottery winner. My only concern is it's not very explicit and bootstrap isn't just a proxy of `[npm | yarn] install` so it might be confusing whether args get appended to entire flow of bootstrap or particular parts of it. 

I can live with that :)

xaka: @evocateur what if both npmClientArgs and `--` are provided, should we append one to another or replace? I vote for append as it'll work like "common arguments you always want to pass".

xaka:
@evocateur please take another look; i wanted to add integration tests, but couldn't find proper way to capture arguments and do assertion - feels like there is need for mock npm client for that

evocateur:
It would be sufficient to test whether it respects `--production --no-optional`, for example, by asserting that a modules under `devDependencies` and `optionalDependencies` didn't get installed.

xaka:
Good call. Will do.

src/commands/BootstrapCommand.js
```diff
@@ -46,13 +46,19 @@ export default class BootstrapCommand extends Command {
   }
 
   initialize(callback) {
-    const { registry, npmClient } = this.options;
+    const { registry, npmClient, npmClientArgs } = this.options;
```
evocateur: This needs a `npm-client-args` config in the `builder` object above

evocateur:
Also, don't worry about making several commits, I'll be squashing the PR anyway.

xaka:
D'oh! Unfortunately we cannot move forward with yargs until there is fix for https://github.com/yargs/yargs/issues/682. Next version (hasn't been released yet) adds `populate--` option to change the behavior, but it's unclear when it'll become available. Until then it's difficult to distinguish between positional arguments and arguments that simply haven't been parsed.

src/commands/BootstrapCommand.js
evocateur: Specifically, using this block starting at line 12:

```js
export function handler(argv) {
  return new BootstrapCommand([...argv.npmClientArgs], argv).run();
}

export const command = "bootstrap [npm-client-args..]";
```

This pattern is used in `lerna exec` (and `run`): https://github.com/lerna/lerna/blob/master/src/commands/ExecCommand.js#L8

Thus you can alias a collection of all unparsed positional arguments, as well as unparsed remainders behind `--`.

```
lerna bootstrap --scope foo -- --bar
lerna bootstrap --hoist -- --no-optional
```

It should still be retrievable from `this.options.npmClientArgs`, but it's also always present as an array (which might be empty).

xaka:
@evocateur ready for another round with integration tests included! thanks for patience 🍻

test/integration/__snapshots__/lerna-bootstrap.test.js.snap
```diff
@@ -56,6 +56,18 @@ package-3 cli1 OK
 package-3 cli2 package-2 OK"
 `;
 
+exports[`passes remaining arguments to npm client: node_modules 1`] = `
+Array [
+  "packages/package-1/node_modules/ansi-regex",
+  "packages/package-1/node_modules/clone",
+  "packages/package-1/node_modules/columnify",
+  "packages/package-1/node_modules/defaults",
+  "packages/package-1/node_modules/pify",
```
xaka: pify shouldn't be here, hmm

xaka:
One test is failing and it says `packages/package-1/node_modules/pify` is expected to be installed, but when i manually run npm with same arguments, it doesn't happen. I'm confused...

xaka:
Turned out it was `lerna` itself extracting and merging `dependencies` and `devDependencies` together, and passing it `npm` afterwards, which affected test at the end.

evocateur:
Yeah, it's because `pify` is a transitive dependency of `tempy`, and thus will always be installed no matter the flags. I'd recommend choosing dependencies that don't have any dependencies themselves, so that the scope of the assertion is tightly focused.

`normalize-newline`, `remove-trailing-separator`, and `balanced-match` all fit the bill.

That being said, you're correct! We might want to focus on different flags, perhaps? Perhaps it would be easier to use `--ignore-engines` instead?

xaka:
I ended up using fake npm client, which is less error prone, doesn't have side effects and comes with predictable behavior. Everything should stable now.

test/integration/lerna-bootstrap.test.js
```diff
@@ -1,8 +1,10 @@
 import execa from "execa";
+import fs from "fs-extra";
 import getPort from "get-port";
 import globby from "globby";
-import tempy from "tempy";
 import normalizePath from "normalize-path";
+import path from "path";
+import tempy from "tempy";
```
xaka: I sorted everything as I didn't know where to put new imports

xaka:
@evocateur knock knock!

DxCx:
@xaka thanks alot!
@evocateur any expectation when this will be released?

test/integration/lerna-bootstrap.test.js
```diff
@@ -65,6 +67,22 @@ describe("lerna bootstrap", () => {
       const stdout = await execa.stdout(LERNA_BIN, ["run", "test", "--", "--silent"], { cwd });
       expect(stdout).toMatchSnapshot("--npm-client yarn: stdout");
     });
+
+    test.concurrent("passes remaining arguments to npm client", async () => {
+      const cwd = await initFixture("BootstrapCommand/npm-client-args");
```
evocateur: We also need an integration test that _doesn't_ employ the `npmClientArgs` key in lerna.json (using the `BootstrapCommand/basic` fixture, for example). I'm pretty sure that fails right now, as is.

test/fixtures/BootstrapCommand/npm-client-args/npm
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+
+require("fs").writeFileSync(require("path").resolve(__dirname, "npm-debug.log"), process.argv.slice(2));
```
evocateur: I don't think we need to write a file, just `console.log` the argv and assert on the command's `stdout`.

xaka: unfortunately bootstrap doesn't proxy npm client output

evocateur:
Of course, you're right! I apologize for my scattered attention.

xaka:
@evocateur added another integration test for remaining arguments without `npmClientArgs` in `lerna.json`

Timer:
Hi! We'd love to see this merged for https://github.com/facebookincubator/create-react-app/pull/2673; is there anything I can do to help?
If not, take this as a friendly bump! 😄

evocateur:
I didn't realize the durable `npmClientArgs` from lerna.json and any additional CLI flags would be merged, not overridden, so my previous comment about the config naming is invalid. Still wish the integration test was a little less "fake out npm", but it works, and that's the ultimate goal.

Thanks @xaka for your patience. :)

xaka:
@evocateur just couple months, no prob man 🍻 😄 next time i'll start with FB to bribe you first, hehe

evocateur:
Augh, AppVeyor is exasperating. 2 out of 3 pass, so I'm calling it good.

Issue #323, opened 2016-08-26
[Bug] lerna exec exits early


doug-wade:
I'm trying to use [lerna exec](https://github.com/lerna/lerna#exec) on beta 26 in [React Server](https://github.com/redfin/react-server) to find all of my packages that have a dependency on [superagent](https://www.npmjs.com/package/superagent) with [npm ls](https://docs.npmjs.com/cli/ls).

Here's my console output:

```
dougwade code/react-server ‹master*› » lerna exec -- npm ls superagent                                                                                                        1 ↵
Lerna v2.0.0-beta.26
react-server@0.4.6 /Users/doug.wade/code/react-server/packages/react-server
├── UNMET PEER DEPENDENCY express@>=3.x
└── superagent@1.2.0  extraneous

npm ERR! peer dep missing: express@>=3.x, required by express-state@1.2.0
npm ERR! extraneous: superagent@1.2.0 /Users/doug.wade/code/react-server/packages/react-server/node_modules/superagent
Errored while running command 'npm' with arguments 'ls superagent' in 'react-server'
Errored while running ExecCommand.execute
Command failed: npm ls superagent
dougwade code/react-server ‹master*› » babel-plugin-react-server@0.4.4 /Users/doug.wade/code/react-server/packages/babel-plugin-react-server                                  1 ↵
└── (empty)

npm ERR! code 1
babel-preset-react-server@0.4.4 /Users/doug.wade/code/react-server/packages/babel-preset-react-server
└── (empty)

npm ERR! code 1
generator-react-server@0.4.6 /Users/doug.wade/code/react-server/packages/generator-react-server
└── (empty)

npm ERR! code 1

dougwade code/react-server ‹master*› » lerna ls                                                                                                                               1 ↵
Lerna v2.0.0-beta.26
- babel-plugin-react-server
- babel-preset-react-server
- generator-react-server
- react-server
- react-server-cli
- react-server-core-middleware
- react-server-data-bundle-cache
- react-server-gulp-module-tagger
- react-server-module-tagger
```

There's a couple of things to note.  The big thing is that only four of the nine packages are visited and have output -- for instance, the "core" module, `react-server` is never visited.  

There are, however, a couple of other minor surprises: private packages aren't listed with `lerna ls`, and the command seems to exit early (or, at least, I get a command prompt halfway through execution).

doug-wade:
[...]

PR #835, opened 2017-05-22
Flag for exec to bail upon child execution error

rsolomon:
## Description
Currently, if a subprocess created using `lerna exec` throws an error, it will kill the entire operation. This change adds a `bail` flag, which will ignore subprocess errors for the `exec` command. The default value for `bail` is true, thereby preserving the existing functionality.

This works by passing a `reject` flag to `execa` (documented [here](https://www.npmjs.com/package/execa#reject)). When this flag is set to `false`, it will resolve its Promise instead of the default rejection behavior.

## Motivation and Context
Motivated by #323 , which outlined that there are many cases where you want to run a command that may fail on all packages, and that the failure outcome should not be cause for halting the operation for the remaining packages.

## How Has This Been Tested?
* Added a new unit test, which confirms the passing of the `reject` flag to `execa`.
* Tested the following command on a local application repository with 23 packages, to confirm they were ran on all packages:

```
lerna exec --bail=false -- npm outdated
```

* Tested the following command on the same local application repository, confirming that the execution will halt upon a single instance of `npm outdated` failing within a package:

```
lerna exec npm outdated
```

## Types of changes

## Checklist:

### DIFF
--- a/test/integration/lerna-exec.test.js
+++ b/test/integration/lerna-exec.test.js
@@ -104,4 +104,48 @@ describe("lerna exec", () => {
     expect(stdout).toMatch("package-2: file-2.js");
     expect(stdout).toMatch("package-2: package.json");
   });
+
+  test.concurrent("--bail=false <cmd>", async () => {
+    const cwd = await initFixture("ExecCommand/basic");
+    const args = [
+      "exec",
+      "--bail=false",
+      "--concurrency=1",
+      "--",
+      "npm run fail-or-succeed",
+    ];
+
+    const { stdout, stderr } = await execa(LERNA_BIN, args, { cwd });
+    expect(stderr).toMatch(
+      "Failed at the package-1@1.0.0 fail-or-succeed script 'echo \"failure!\" && exit 1'."
+    );
+    expect(stdout).toMatch("failure!");
+    expect(stdout).toMatch("success!");
+  });
+
+  test.concurrent("--bail=true <cmd>", async () => {
+    const cwd = await initFixture("ExecCommand/basic");
+    const args = [
+      "exec",
+      "--bail=true",
+      "--concurrency=1",
+      "--",
+      "npm run fail-or-succeed",
+    ];
+
+    let hasError = false;
+    try {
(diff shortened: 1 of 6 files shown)


test/integration/lerna-exec.test.js
```diff
@@ -104,4 +104,48 @@ describe("lerna exec", () => {
...
+
+    const { stdout, stderr } = await execa(LERNA_BIN, args, { cwd });
+    expect(stderr).toMatch(
+      "Failed at the package-1@1.0.0 fail-or-succeed script 'echo \"failure!\" && exit 1'."
+    );
+    expect(stdout).toMatch("failure!");
+    expect(stdout).toMatch("success!");
+  });
+
+  test.concurrent("--bail=true <cmd>", async () => {
```
evocateur: Is `--no-bail` also supported? It might be more useful to test that variant here, instead of the default which we should trust `yargs` has already dealt with as we configured it (and all of our other "intentional failure" tests confirm).

rsolomon: @evocateur apologies, but what do you mean by `--no-bail`? The `bail=false` test case is above on line 108, but I assume you mean something else. This is my first time diving into the Lerna codebase, so I'm probably just missing something :)

evocateur:
Yes, sorry, I wasn't clear!

We already have tests for the --bail=true, and nobody would do that anyway.

Instead, yargs should support prefixing an option with "no-" that also yields false. Let's test that case here instead of the already-tested default, just for completion's sake.

(I personally find --no-bail to be the more "fluent" construction of this negation, if that makes sense)

rsolomon:
@evocateur Ah, that makes sense. Test case has been updated.

rsolomon:
Any advice on the Appveyor build? It doesn't look like the failure is related to my commit, but I don't see any obvious way to re-trigger.

evocateur:
It looks completely unrelated, like somebody kicked a server over. I'll try kicking the build manually.

evocateur:
Awesome, it worked. Gotta love Windows 🙄

PR #857, opened 2017-05-30
Print n/a when a package has no version key.

ben-eb:
## Description

When a package's `version` key is missing, Lerna would print `vundefined` when listing it via `lerna ls`. That's slightly jarring to see, so I replaced the string with `n/a` instead. I also right-aligned the version numbers because it looked slightly neater:

screen shot 2017-05-30 at 21 26 17

## Motivation and Context

The output looked ugly before. 😄 

## How Has This Been Tested?

I've included a snapshot test for this behaviour.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/LsCommand.js
+++ b/src/commands/LsCommand.js
@@ -44,7 +44,7 @@ export default class LsCommand extends Command {
       output(JSON.stringify(formattedPackages, null, 2));
     } else {
       formattedPackages.forEach((pkg) => {
-        pkg.version = chalk.grey(pkg.version ? `v${pkg.version}` : "n/a");
+        pkg.version = pkg.version ? chalk.grey(`v${pkg.version}`) : chalk.yellow("MISSING");
         pkg.private = pkg.private ? `(${chalk.red("private")})` : "";
       });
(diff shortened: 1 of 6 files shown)


ben-eb:
I'm not sure that the test failure is related to this patch but it was passing locally - also it's passing on AppVeyor...

evocateur:
Shouldn't the lack of a version be an error? It's benign when it's a private package, I guess, but I'd rather avoid letting users think it's going to work _without_ a proper version field.

ben-eb:
We could change "n/a" to be something more actionable if you like. Maybe a yellow "missing" instead?

evocateur:
A yellow "MISSING" would be perfect.

evocateur:
(and the test failure is indeed unrelated, a weird intermittent timeout)

ben-eb:
@evocateur Updated. We can probably squash the history for this branch. 😄

evocateur:
for sure!

/me kicks timed-out travis build

ben-eb:
Heh, no worries. 😄

PR #861, opened 2017-06-01
chore(travis): test against node 8 and npm 5

douglasduteil:


rgbkrk:
Thanks for doing this, it helps highlight some of what's going on with the npm5 / node 8 change.

.travis.yml
```diff
@@ -3,6 +3,7 @@ git:
 sudo: false
 language: node_js
 node_js:
+ - '8'
  - '7'
```
evocateur: I'd say we should replace 7 with 8, no reason to test 7 anymore.

evocateur:
Appveyor needs node 8 too. Unfortunately the test failures aren't especially descriptive (differences in npm output, more than anything).

douglasduteil:
:up: 
@evocateur I have difficulties finding out why Lerna is not working with npm@5. 
Could you enlighten me about the conflict between Lerna and npm@5 ?

- Is it the lock that is conflicting with the way Lerna is re-arrange the node_modules ?
- Do you have some kind of test that show us the difference between installation with npm@4 and npm@5 ?
- Are you planing on keeping lerna@2.x npm@4 only and working on lerna@3.x for npm@5 (as this is looking like a breaking change) ?

evocateur:
Judging by the latest Travis failures, it looks like some of the messages from npm 5 have changed. I'll have to run things locally to narrow it down.

Once npm 5 & yarn gain workspace support, a 2.x "legacy" branch seems reasonable while 3.x becomes master. In the immediate near-term, I'm pretty sure npm 5 won't necessitate a major bump.

douglasduteil:
A little update.
The current issue between `npm@5` and `lerna@2` is the **hoisting**.
As Lerna is adding modules in the root node_modules and Npm is cleaning the root node_modules following the created `packge-lock.json`.
I guess that a fix will be making Lerna alter the lock files too when hoisting.

jskrzypek:
Couldn't lerna just backup & replace the lock file like what is already done with the package.json?

Issue #859, opened 2017-05-31
lerna run: Don't error if no scoped packages are matched


treshugart:
I'm implementing #822 and have been attempting to integrate it into our workflow. However, I'm running into some issues where I want to scope some commands using that flag (`--since`) but don't want it to error if no packages are run.

In thinking about this, I'm not sure it's valuable to ever exit with non-zero if no scoped packages are matched; it doesn't seem like an exceptional state. Warning might be fine, but even then, it's fairly clear when no packages are matched.

## Expected Behavior

I would expect to be able to run `lerna run something --scope some-unmatched-package` without errors.

This is also inconsistent with how scoping is applied to built-in Lerna commands as well as `exec`. For example, `bootstrap` and `ls` don't error if a filtering option doesn't match any packages, it just logs that it didn't run in any packages.

## Current Behavior

The command errors. This means you can't do something like `lerna run eslint --since this-branch` to only lint against changed files, it will fail linting if no packages have been changed.

To work around this you can do something like `lerna exec --since ref -- eslint`.

## Possible Solution

The built-in commands log how many packages a command was run in, so maybe we can follow suit here for `run`.

## Steps to Reproduce (for bugs)

n/a

## Context

This seems only an issue for `lerna run`, so to work around it you could use `exec`, but it feels odd to have to do something like `lerna exec --scope package -- npm run test`.

## Your Environment

n/a

evocateur:
The inconsistency between scoped `ls`/`bootstrap` and `run` is certainly itchy. I am open to making it more consistent.

evocateur:
This was fixed waaaay back in v2.0.0 🤸‍♀️

PR #868, opened 2017-06-07
lerna run: Don't error if no scoped packages are matched

ricky:
## Description
Display a warning when `lerna run <script>` does not find any npm scripts to execute, rather than throw an error.

## Motivation and Context
Similar to @treshugart, we have a desire to run certain npm scripts against only changed packages. Sometimes no packages have changed, and sometimes only packages without the referenced script has changed. In most situations, this is not a failure state, and this behavior is more in line with other scoped commands (e.g. `ls`, `bootstrap`).

https://github.com/lerna/lerna/issues/859

## How Has This Been Tested?
Manual invocations of the command with scopes that do and don't contain the target npm script.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -65,7 +65,6 @@ export default class RunCommand extends Command {
 
     if (!this.packagesWithScript.length) {
       this.logger.warn(`No packages found with the npm script '${this.script}'`);
-      return callback(null, false);
     }
 
     if (this.options.parallel || this.options.stream) {
@@ -85,8 +84,10 @@ export default class RunCommand extends Command {
       if (err) {
         callback(err);
       } else {
-        this.logger.success("run", `Ran npm script '${this.script}' in packages:`);
-        this.logger.success("", this.packagesWithScript.map((pkg) => `- ${pkg.name}`).join("\n"));
+        if (this.packagesWithScript.length) {
+          this.logger.success("run", `Ran npm script '${this.script}' in packages:`);
+          this.logger.success("", this.packagesWithScript.map((pkg) => `- ${pkg.name}`).join("\n"));
+        }
         callback(null, true);
       }
     };
--- a/src/commands/RunCommand.js
+++ b/src/commands/RunCommand.js
@@ -65,6 +65,7 @@ export default class RunCommand extends Command {
 
     if (!this.packagesWithScript.length) {
       this.logger.warn(`No packages found with the npm script '${this.script}'`);
+      return callback(null, false);
     }
 
     if (this.options.parallel || this.options.stream) {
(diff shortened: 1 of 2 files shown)


src/commands/RunCommand.js
```diff
@@ -64,8 +64,7 @@ export default class RunCommand extends Command {
     }
 
     if (!this.packagesWithScript.length) {
-      callback(new Error(`No packages found with the npm script '${this.script}'`));
-      return;
+      this.logger.warn(`No packages found with the npm script '${this.script}'`);
     }
```
evocateur: How about we just exit early here, instead of proceeding to the `execute()` phase after batching an empty array?
```
return callback(null, false);
```

ricky: Good call. Fixed in bd83ae6.

ricky:
I didn't add a test because it didn't seem to be explicitly tested in the other commands, but I'm not opposed to doing so.

evocateur:
In current `master`'s `test/RunCommand.js`, this test fails:
```js
    it("does not error when no packages match", (done) => {
      const runCommand = new RunCommand(["missing-script"], {}, testDir);

      runCommand.runValidations();
      runCommand.runPreparations();

      runCommand.runCommand(exitWithCode(0, (err) => {
        if (err) return done.fail(err);

        try {
          expect(ranInPackages(testDir)).toMatchSnapshot("run <missing-script>");
          expect(output).not.toBeCalled();

          done();
        } catch (ex) {
          done.fail(ex);
        }
      }));
    });
```
```
  ● RunCommand › in a basic repo › does not error when no packages match

    No packages found with the npm script 'missing-script'
      
      at RunCommand.initialize (src/commands/RunCommand.js:67:18)
      at RunCommand._attempt (src/Command.js:377:2199)
      at RunCommand.runCommand (src/Command.js:377:1912)
      at Object.<anonymous> (test/RunCommand.js:84:18)
```

evocateur:
and actually, you don't need to assert on the snapshot, as `ranInPackages` should never be called.

ricky:
Yeah, I just created a very similar test in this branch, which caught that fact that the early exit in bd83ae6 actually [returns a non-zero exit code](https://github.com/lerna/lerna/blob/master/src/Command.js#L300-L302). So, I think I should revert back to the original behavior. Do y'all have a preference for doing so with a new commit or dropping the commit from the branch?

evocateur:
Yeah, I was afraid of that. I really need to fix that (non-errors exiting early shouldn't be non-zero exits).

Just push up a commit that reverts it, including "[skip ci]" in the commit message, and I'll squash the PR so it'll look clean :)

ricky:
Revert in 07df0d5, test in 61ef203

evocateur:
Thanks!

treshugart:
So stoked to see this go in so quick! Thank you @ricky?

PR #898, opened 2017-06-26
fix: Fix issue where Yargs default would override durable options.

### LINKED ISSUE #895: Boolean flags override even when not specified
When Yargs defaults are specified, they provide the value as flags to the command constructor. This means that when flags, command overrides, lerna json options and legacy options are merged, the flags will take precedence over any thing else because they have a value.

This can easily be reproduced by simply trying to use `stream: true` in the `lerna.json`. If you remove the option from, for example, `RunCommand` it works. Similarly, if you rename it from `boolean` to `string`, it continues to work. So it seems that while `_.defaults()` is doing the right thing (because `false` is a value, where `undefined` is not), providing a default that is considered a "value", will break anything following the durable option convention.

## Expected Behavior

Durable options should take precedence over Yargs defaults.

## Current Behavior

Defaults override user-defined values not specified as flags.

## Possible Solution

After digging a bit deeper, it seems that all we really have to do to get around this is to specify `default: undefined` to any option that is a `boolean` and doesn't already have a default (because they default to `false`. When `_.defaults()` is called, they will then be overridden if they're not specified by the user.

## Steps to Reproduce

1. Put `stream: true` in the `lerna.json`
2. Run a command that should stream.
3. It won't stream.

## Context

- References #893.

treshugart:
Fixes #895.

The solution to this requires that we stop using yargs defaults because we have several different ways of specifying options. If we specify yargs defaults, then they get set as flags and override any specified in the lerna.json. To get around this, we have to put defaults last in the chain by specifying them per command, as opposed to yargs which comes first in the chain as they merge user-defined options passed as flags with the corresponding defaults.

Unless I'm missing something (definitely could be), there's no way for us to insert our `lerna.json` options in between the options yargs parses and the defaults that we specify for them.

## Description

I've updated each command to:

1. Properly declare boolean options, if they weren't already specified (some weren't).
2. Ensure boolean options are specifying `undefined` as the default.
3. Move the default value that was specified before to be returned from `get defaultOptions` on the command class.

## Motivation and Context

As described in #895, options specified in the `lerna.json` wouldn't be taken if a `default` was specified to the Yargs option definition.

## How Has This Been Tested?

I've written a test to ensure that `defaultOptions` get inserted at the end of the merge chain for flags and lerna.json options.

Other than that, we'd have to test to make sure we're defining all boolean options with `undefined` defaults, or to test this behaviour with every single option that could be subject to the issue, which I'm not sure if this is a good cost-benefit. Open to thoughts on this.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -18,23 +18,33 @@ export const builder = {
     group: "Command Options:",
     describe: "Bail on exec execution when the command fails within a package",
     type: "boolean",
-    default: true,
+    default: undefined,
   },
   "only-updated": {
     group: "Command Options:",
     describe: "Run command in packages that have been updated since the last release only",
     type: "boolean",
+    default: undefined,
   },
   "parallel": {
     group: "Command Options:",
     describe: "Run command in all packages with unlimited concurrency, streaming prefixed output",
     type: "boolean",
+    default: undefined,
   },
 };
 
 export default class ExecCommand extends Command {
   get requiresGit() {
     return false;
+  }
+
+  get defaultOptions() {
+    return Object.assign({}, super.defaultOptions, {
+      bail: true,
+      onlyUpdated: false,
+      parallel: false,
+    });
   }
 
   initialize(callback) {
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -28,7 +28,7 @@ export const builder = {
   "canary": {
     group: "Command Options:",
     describe: "Publish packages after every successful merge using the sha as part of the tag.",
-    alias: "c"
+    alias: "c",
   },
   "cd-version": {
     group: "Command Options:",
@@ -59,7 +59,7 @@ export const builder = {
     defaultDescription: "origin",
     describe: "Push git changes to the specified remote instead of 'origin'.",
     type: "string",
-    requiresArg: true
+    requiresArg: true,
   },
   "yes": {
     group: "Command Options:",
@@ -72,19 +72,19 @@ export const builder = {
     describe: "Use a custom commit message when creating the publish commit.",
     alias: "m",
     type: "string",
-    requiresArg: true
+    requiresArg: true,
   },
   "npm-tag": {
     group: "Command Options:",
     describe: "Publish packages with the specified npm dist-tag",
     type: "string",
-    requiresArg: true
+    requiresArg: true,
   },
   "repo-version": {
     group: "Command Options:",
     describe: "Specify repo version to publish.",
     type: "string",
-    requiresArg: true
+    requiresArg: true,
   },
   "skip-git": {
     group: "Command Options:",
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -44,11 +44,15 @@ export const builder = {
   },
   "conventional-commits": {
     group: "Command Options:",
-    describe: "Use angular conventional-commit format to determine version bump and generate CHANGELOG."
+    describe: "Use angular conventional-commit format to determine version bump and generate CHANGELOG.",
+    type: "boolean",
+    default: undefined,
   },
   "exact": {
     group: "Command Options:",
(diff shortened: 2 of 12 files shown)


test/ExecCommand.js
```diff
@@ -125,6 +125,7 @@ describe("ExecCommand", () => {
             env: expect.objectContaining({
               LERNA_PACKAGE_NAME: "package-2",
             }),
+            reject: true,
```
treshugart: Implementing defaults actually caught this issue where this *should* have been passed (`bail` defaults to `true`) but was never actually defaulting to true due to the way we're testing it by bypassing yargs (we pass flags directly to the constructor).

evocateur: We test yargs usage in the integration tests, not unit tests. I've often thought of writing an arg bootstrapper to streamline this.

src/commands/ExecCommand.js
```diff
@@ -18,17 +18,19 @@ export const builder = {
     group: "Command Options:",
     describe: "Bail on exec execution when the command fails within a package",
     type: "boolean",
-    default: true,
+    default: undefined,
```
evocateur: ugh, this noise really bothers me, should be fixed upstream.

Glad you figured it out, all the same. :)

src/commands/PublishCommand.js
```diff
@@ -44,57 +44,80 @@ export const builder = {
...
+    describe: "Create a temporary tag while publishing.",
+    type: "boolean",
+    default: undefined,
   }
 };
 
 export default class PublishCommand extends Command {
+  get defaultOptions() {
+    return Object.assign({}, super.defaultOptions, {
+      conventionalCommits: false,
```
evocateur: My biggest frustration with this pattern is the disconnect between the `dash-case` options config and this `camelCase` defaults object. Again, glad it can be worked around, but squirmy about the solution.

src/commands/ExecCommand.js
treshugart: Yeah, I didn't like having to do it that way. Maybe something Yargs could do here (as a breaking change), would be to default to `undefined` instead of `false`. Other than that, this doesn't feel like an issue with Yargs, but more of how Lerna is using their defaults in conjunction with other forms of obtaining options that are merged in with them. The only alternative would be for Lerna to insert its durable options in between user options that yargs parses and their defaults they automatically merge in.

Do you think it's worth raising an issue for one of these for Yargs and referencing this PR?

src/commands/PublishCommand.js
treshugart: Same. I went for the cheapest option, in terms of amount of changes to Lerna and waiting for Yargs to address upstream (if they even would). See comment above about possibly raising an issue.

treshugart:
Re tests, I'm happy to write one. I may need a little guidance in the preferred way to do this, though. It wasn't immediately apparent if there was an existing method in how you would approach it.

src/commands/ExecCommand.js
evocateur: Yes, I think this would be a great example to reference when raising this upstream, which I also think would be a good idea.

treshugart: @evocateur I was just raising this issue, and in describing the rationale, I thought of a possible solution. What if instead of merging the durable options *after* yargs parses the command, we read them *before* and pass them in as the yargs options defaults?

evocateur: I was noodling about using yargs' [`config()`](https://github.com/yargs/yargs/blob/master/docs/api.md#config) method (as described [here](https://github.com/yargs/yargs/blob/master/docs/advanced.md#rc-files)) with the parsed lerna.json stuff, and it's certainly more elegant (and in practice the same idea as you propose).

The chicken & egg problem strikes us here, because the arg parsing (as currently constructed) has already finished by the time we read lerna.json and filter the command-specific durable config, etc. We should probably extract that chunk (config resolution and filtering) anyway, but it's a bit complex to contemplate.

Issue #91, opened 2016-03-25
[Feature] Allow custom publish subdirectory


jacobwgillespie:
**EDIT:** I overlooked `package/[name]/scripts/prepublish.js`, so that's awesome.  It would however be a bit difficult to customize the directory in which `npm publish` is run just with a prepublish script.  I'll adjust this issue to recommend adding a way to customize that publish directory, since it's hard-coded in `publishTaggedInDir`.

---

It would be awesome to be able to customize the publish behavior (like the publish directory) - for my use case, I build the packages like `packages/abc/src` into a directory called `packages/abc/dist`, then copy the package.json into `dist` and publish from there (to have a flat module structure).  I've gotten it working for now by overriding the exports of `npmUtils`, but in the future it would be awesome to be able to customize this step:

``` javascript
#!/usr/bin/env node

// Wrapper script that patches lerna with our custom publish path

import child from 'child_process';
import logger from 'lerna/lib/utils/logger';
import path from 'path';

const newPublishTaggedInDir = logger.logifyAsync('npmUtils.publishTaggedInDir', (tag, dir, callback) => {
  const cmd = `cd ${dir}/dist && cp ../package.json . && npm publish --tag ${tag}`;
  child.exec(cmd, (err, stdout, stderr) => {
    callback(stderr || err, stdout);
  });
});

// Manually require npmUtils to load it into the cache
require('lerna/lib/utils/npmUtils');

// Patch the exports
const modulePath = path.resolve(__dirname, '../node_modules/lerna/lib/utils/npmUtils.js');
require('module')._cache[modulePath].exports.publishTaggedInDir = newPublishTaggedInDir;

// Run the original binary
require('lerna/bin/lerna');
```

thejameskyle:
You can put package-specific scripts inside `packages/:package_name/scripts/prepublish.js` or `packages/:package_name/scripts/postpublish.js`. If you want to do something before lerna entirely, I recommend just running it before you run lerna with like an npm script or whatever.

jacobwgillespie:
Ah, nice, I overlooked the pre/post publish scripts.  I'm already doing stuff outside of lerna before running publish (with a Makefile).

I edited the title and issue description now to be specific about customizing the publish directory, as it would still be pretty difficult to change the directory from a prepublish script alone (it's kinda hard-coded in `publishTaggedInDir`).

thejameskyle:
[...]

PR #900, opened 2017-06-27
[Feature] Support <folder> option on npm publish

nathf:
## Description

Previously listed as wont fix in #91, but the npm command allows publishing from a specific directory which allows easier organisation for 'flat' module structures.
So this PR exposes the folder option to align with [npm publish](https://docs.npmjs.com/cli/publish) options.

This change is non-breaking.

Unsure if this aligns with how the maintainers want to proceed with this tool, thought this would be a good starting point.

## Motivation and Context

Align with npm options, simplify package structures.

## How Has This Been Tested?

Tested locally through npm link in another lerna repo. Appropriate errors are thrown if `publishDir` doesnt contain a package.json, won't fail silently.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,7 +594,8 @@ Running `lerna` without arguments will show all commands/options.
       "ignore": "component-*"
     }
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "publishDir": "lib/"
 }
 ```
 
@@ -604,6 +605,7 @@ Running `lerna` without arguments will show all commands/options.
 - `commands.bootstrap.ignore`: an array of globs that won't be bootstrapped when running the `lerna bootstrap` command.
 - `commands.bootstrap.scope`: an array of globs that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
+- `publishDir`: Relative path in each package that [`npm publish`](https://docs.npmjs.com/cli/publish) will run in. This folder must contain a `package.json`.
 
 
 ### Common `devDependencies`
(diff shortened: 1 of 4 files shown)


gauntface:
I would love this feature.

@hzoo is there any chance you could take a look at this?

MattCain:
This would be awesome, at the moment I can't publish the built versions of my packages without a redundant `dist` directory at the root of the distribution.

README.md
```diff
@@ -594,7 +594,8 @@ Running `lerna` without arguments will show all commands/options.
       "ignore": "component-*"
     }
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "publishDir": "lib/"
```
gauntface: Would it be worth changing this to `dist/` or `build/` it feels like a more common use case for developers

gauntface:
One wrinkle I just realised.

How does this work with bootstrap?

If you use `prepublish` to build your project before everything is npm linked by `lerna bootstrap` you'll get in a messy / hard to reason with state as the ``package.json will need to be in the root of each package to find the `prepublish` npm script that would then generate the build directory.

My understanding is that this is how you should be using lerna, which means this folder option may actually cause confusion.

nathf:
@gauntface Good point. During bootstrap symlinking happens before prepublish. So the build directory wouldn't exist at symlink time. 

If the order were reversed you could npm link from the build directory as it would need to contain a `package.json` for publishing.  Alternatively the build directory could be created during `preinstall` phase but those both seem a bit flakey/cause more confusion.

Also taking into account the addition of yarn workspaces this feature may be a stretch. 

Thoughts?

gauntface:
Yeah, the more I thought about it, the more I realised this wouldn't fit my needs.

alan-agius4:
Can we have this back please?

PR #904, opened 2017-07-02
feat(canary): Improved canary option functionality

Craga89:
## Description

This change changes the `canary` semantics and bumps the minor version, as well as overloading it allowing the user to pass a value that will be used for the meta suffix instead i.e. `--canary=beta`, which results in a `x.x.x-beta.<sha>` version tag.

This is especially useful in CI scenarios where you want to publish an `alpha`, `beta`, and/or `rc` version depending on the branch you're building i.e. `1.1.0-beta.e5fg52` for the `develop` branch.

BREAKING CHANGE: Now bumps to the next `minor` version instead of using current version, as per `GitVersion` and other such tools. This prevents issues the alphabetical ordering outlined in #277 

## Motivation and Context
Currently the `canary` option has a few limitations:
* Hard codes an `alpha` meta suffix on the resulting version
* Uses the current version, not the next minor version which causes [issues (#277)](https://github.com/lerna/lerna/issues/277)

## How Has This Been Tested?
Locally tested, functions as it did if no value is passed to the `canary` operator, except for the minor version bump as explained above

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ $ lerna publish --canary
 $ lerna publish --canary=beta
 ```
 
-When run with this flag, `publish` publishes packages in a more granular way (per commit). Before publishing to npm, it creates the new `version` tag by taking the current `version`, bumping it to the next /minor/ version, adding the provided meta suffix (defaults to `alpha`) and appending the current git sha (ex: `1.0.0-alpha.81e3b443`, `1.0.0-beta.81e3b443`).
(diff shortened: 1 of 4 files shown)


evocateur:
I was hoping the incrementing could be done with `git describe --long` semantics for the tagged part (`-alpha.{commits since nearest tag}.{sha}`), but this is better than our current `--canary` and I would like to get this breaking change in before v2.0.0 official.

Thanks @Craga89!

Craga89:
I really like the idea of `{commits since nearest tag}` as it matches well to the current `GitVersion` implementation we have at work. I'll look into it in a different commit ASAP so we could potentially get it in before 2.0.0? When are you cutting for it?

evocateur:
Already cut 2.0.0, no worries! I see no breaking change in slightly modifying the post-tag metadata.

Hypnosphi:
What do I do if I want alpha canary for the next major release?

evocateur:
@Hypnosphi As of v2.1.0, `lerna publish --canary --cd-version major`

PR #908, opened 2017-07-06
Add `--semver-ignore` option to Bootstrap Command

ndelangen:
I added an option to the BootstrapCommand: `--semver-ignore` (open to other names suggestions).

## Description
The new option tweaks the functionality of `symlinkPackages`.
The current behavior is packages that are known by name, but mismatch according to semver are installed from registry, and **not** linked.
This behavior is the default.
I added an optional boolean option to link known packages regardless of semver compatibility.

## Motivation and Context
I've tried to show my case well in the unittest provided in the PR, so have a look at that when this makes little sense to you:
It's common to make alpha / beta / next releases (prereleases). I ran into the problem that unknowingly packages were **not** linked when I expected them to. I traced the reason to version numbers like these: `3.2.0-alpha.7`.
The solution is to change the versionnumbers in the consuming packages to match exactly, but this is very cumbersome.

Furthermore I have never wanted once to **not** link known packages. I'd rather have them linked anyway and get a warning. (this is not implemented in this PR).

## How Has This Been Tested?
1. I changed lerna source directly in a project that already used lerna.
2. I added unit tests and they pass

## Types of changes

## Checklist:

I will update my PR with documentation changes *after* I get feedback on whether the feature is likely to get merged.

2 tests seem to be skipped, but this is not in the testfile I worked on? I'm guessing this is known and not related to my PR.

### DIFF
--- a/test/BootstrapCommand.js
+++ b/test/BootstrapCommand.js
@@ -538,6 +538,55 @@ describe("BootstrapCommand", () => {
     });
   });
 
+  describe("mismatching version numbers", () => {
+    let testDir;
+
+    beforeEach(() => initFixture("BootstrapCommand/version-mismatch").then((dir) => {
+      testDir = dir;
+    }));
+    beforeEach(stubSymlink);
+    afterEach(resetSymlink);
+
+    it("should link when option.semverIgnore is true", (done) => {
+      const bootstrapCommand = new BootstrapCommand([], { semverIgnore: true }, testDir);
+
+      bootstrapCommand.runValidations();
+      bootstrapCommand.runPreparations();
+
+      bootstrapCommand.runCommand(exitWithCode(0, (err) => {
+        if (err) return done.fail(err);
+
+        try {
+          expect(installedPackagesInDirectories(testDir)).toMatchSnapshot();
+          expect(symlinkedDirectories(testDir)).toMatchSnapshot();
+
+          done();
+        } catch (ex) {
+          done.fail(ex);
+        }
+      }));
+    });
+    it("should link when option.semverIgnore is false", (done) => {
+      const bootstrapCommand = new BootstrapCommand([], { semverIgnore: false }, testDir);
+
+      bootstrapCommand.runValidations();
+      bootstrapCommand.runPreparations();
+
+      bootstrapCommand.runCommand(exitWithCode(0, (err) => {
+        if (err) return done.fail(err);
+
+        try {
+          expect(installedPackagesInDirectories(testDir)).toMatchSnapshot();
+          expect(symlinkedDirectories(testDir)).toMatchSnapshot();
+
+          done();
+        } catch (ex) {
+          done.fail(ex);
+        }
+      }));
+    });
+  });
+
   describe("zero packages", () => {
     let testDir;
 
--- a/test/fixtures/BootstrapCommand/version-mismatch/package.json
+++ b/test/fixtures/BootstrapCommand/version-mismatch/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "independent"
+}
--- a/test/fixtures/BootstrapCommand/version-mismatch/packages/package-2/package.json
+++ b/test/fixtures/BootstrapCommand/version-mismatch/packages/package-2/package.json
@@ -1,0 +1,8 @@
+{
+  "name": "package-2",
+  "version": "1.0.0",
+  "dependencies": {
+    "package-1": "^1.0.0",
+    "external": "^2.0.0"
+  }
+}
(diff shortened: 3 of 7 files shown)


ndelangen:
Is there someone able to review this?

test/BootstrapCommand.js
```diff
@@ -538,6 +538,55 @@ describe("BootstrapCommand", () => {
...
+          expect(installedPackagesInDirectories(testDir)).toMatchSnapshot();
+          expect(symlinkedDirectories(testDir)).toMatchSnapshot();
+
+          done();
+        } catch (ex) {
+          done.fail(ex);
+        }
+      }));
+    });
+    it("should link when option.semverIgnore is false", (done) => {
```
evocateur: "should **not** link when options.semverIgnore is false" ? I think this case is handled, anyway, with the other linking tests that have non-matching semver.

test/fixtures/BootstrapCommand/version-mismatch/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "independent"
```
evocateur: I know this is copypasta, but let's at least customize the name to match the directory.

test/fixtures/BootstrapCommand/version-mismatch/packages/package-2/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "package-2",
+  "version": "1.0.0",
+  "dependencies": {
+    "package-1": "^1.0.0",
```
evocateur: So how does this get mismatched in the first place? If you did a `lerna publish` choosing `prerelease`, that should have also bumped `package-2`'s version and its dependency on `package-1`. Also, it would have bumped the patch, since `1.0.0-alpha.x` will _always_ be "less than" `1.0.0`.

mbecke:
I ran into a similar issue recently and can confirm that this change solves my issue.   As with the original author I always want a link to be created, regardless of whether or not the semver matches.

test/BootstrapCommand.js
ndelangen: ok, can remove this test, you mean?

evocateur: I would lean toward removing it, as it's not really necessary to prove a negative.

test/fixtures/BootstrapCommand/version-mismatch/packages/package-2/package.json
ndelangen: The mismatches happen after a few prereleases in a row, with some packages getting changes and some not. Eventually the versions stop matching and without warning get packages from the registry.

We're sometimes doing prereleases for minor or even patches. Could that be a reason too?

Irregardless, I'd like a flag to always link, never pull internal packages from registry. Sounds like @mbecke wants such a flag as well..

mbecke: Yep, we're working with version independent projects which ideally should be correctly updated across the board, but it doesn't always happen.  Rather than have to enforce that with some other mechanism I was looking to just force linking with a flag like this.

mbecke:
Any update on this pull request?

evocateur:
So, the main problem I have with this patch is that once enabled, you've set yourself up for an inevitable footgun. Start assuming your local packages are always linked, regardless of semver expression, and it's a question of "when", not "if", you ship code with untested/incompatible APIs.

I would much rather figure out why the versions drift in the first place. Do we need to improve our handling of pre-release versioning? Quite likely! There are a lot of subtleties about pre-* version comparison that are not well-tested in the current codebase.

We could almost certainly improve our matching algorithm without abandoning one of the core idioms of npm packages (that is, semver ranges that express semantic compatibility). Improving our messaging when these error cases are encountered would go a long way to helping understand why they occur in the first place.

evocateur:
As I suggested in #1082, I'd be more comfortable with this if it was renamed `--force-local`, avoiding the implicit condoning of ignoring semver. Yes, it still _does_ ignore the semver range under the covers, but the public API doesn't have to admit that. 😉

ndelangen:
Ow hey, I kinda had given up on this PR, I wasn't aware it was still open.

Reading the other PR, it seems to be adding the same functionality but with the flag you wanted?
Does that mean we can just close this? If you feel there's still value in this PR just let me know, I can set aside some time to fix it up next monday or so.

Though actually you kinda convinced me with https://github.com/lerna/lerna/pull/908#issuecomment-327383816

evocateur:
#1082 is only adding it for `lerna link`, not `lerna bootstrap`. It's a lot simpler there, and I kinda prefer it there vs bootstrap, which is already a _terribly_ complex command. If you're okay with `lerna link --force-local`, feel free to close this PR.

ndelangen:
Sorry for anyone hoping this to get merged, Feel free to argue why we should reconsider.

PR #940, opened 2017-07-26
Fixes #939 Bootstrap performance issue with --include-filtered-dependencies in large, cyclic repos

### LINKED ISSUE #939: Performance bug when using --include-filtered-dependencies
## Expected Behavior
When running bootstrap (scoped to a package) with the `--include-filtered-dependencies` flag the scoped packages and each of of deps should be bootstrapped once.

## Current Behavior
Some deps will be rebootstrapped multiple times making the bootstrapping take significantly longer than it needs to.

## Possible Solution
Im pretty sure I was the one who introduced this with the original `--include-filtered-dependencies` PR. In the `addDependencies` function in `PackageUtilities.js`

```javascript
while (fringe.length !== 0) {
  const pkg = fringe.shift();
  const pkgDeps = Object.assign({}, pkg.dependencies, pkg.devDependencies);

  Object.keys(pkgDeps).forEach((dep) => {
    // We correctly check that the new dep is not one we have found already or one we are looking at
    if (packageExistsInRepository(dep) && !packageAlreadyFound(dep) && !packageInFringe(dep)) {
      fringe.push(packageGraph.get(dep).package);
    }
  });

  // We fail to do the same check when re-adding the package we are looking at
  dependentPackages.push(pkg);
}
```

So the fix should be super simple. Writing a test might be a little more annoying.

This is hard to see in smaller repos, but in our repo (60+ packages with lots of dependency loops) it becomes very apparent when running bootstrap with deps of a single package

```
lerna bootrap --scope="package-a" --include-filtered-dependencies
```

```json
{
  "lerna": "2.0.0-rc.4",
  "npmClient": "yarn",
  "version": "independent",
  "stream": true
}

```

I will try to make a reproducible case in the tests for this.

## Context
Context above, very large repo, lots of cyclical deps. 

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.0.0-rc.5 |
| `npm --version`  | 4.2.0 |
| `yarn --version` | 0.23.4 |
| `node --version` | 7.10.0 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| OSX El Capitan | 10.11.4 |

lukebatchelor:
This is a very small change to fix a performance issue when using the include-filtered-dependencies flag in a large, cyclic repository.

The old behaviour would cause packages to be returned multiple times, hence bootstrapped multiple times.

## Description
The change was just to add the extra check that you havent seen a package before you add it to the list.

## Motivation and Context
Major performance concerns in bootstrapping

## How Has This Been Tested?
I have run this in our repo (>60 packages) and have found a 33-50% decrease in bootstapping times

For one small package it went from 24s to 14s (10 packages bootstrapped -> 6 packages bootstrapped), for another it went from 214s to 145s (58 packages boostrapped -> 36 packages bootstrapped)

I've also added another test fixture with a very specifically crafted set of deps that tests three things
* A package depending on itself isn't added twice
* A package being added twice in the same stage of the expansion isn't added twice
* A package that has already been processed wont get added twice

## Types of changes
- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -118,7 +118,9 @@ export default class PackageUtilities {
         }
       });
 
-      dependentPackages.push(pkg);
+      if (!packageAlreadyFound(pkg.name)) {
+        dependentPackages.push(pkg);
+      }
     }
 
     return dependentPackages;
--- a/test/PackageUtilities.js
+++ b/test/PackageUtilities.js
@@ -297,4 +297,34 @@ describe("PackageUtilities", () => {
       });
     });
   });
+
+  describe(".addDependencies()", () => {
+    let packages;
+    let packageGraph;
+
+    // for reference: 1->2, 1->3, 1->4, 2->4, 2->5, 3->4, 3->6, 4->1, 4->4,  5->4, 6->4, 7->4
+    // We design the package tree in a very specific way. We want to test several different things
+    // * A package depending on itself isn't added twice (package 4)
+    // * A package being added twice in the same stage of the expansion isn't added twice (package 4)
+    // * A package that has already been processed wont get added twice (package 1)
+    beforeAll(() => initFixture("PackageUtilities/cycles-and-repeated-deps").then((testDir) => {
+      packages = PackageUtilities.getPackages(new Repository(testDir));
+      packageGraph = PackageUtilities.getPackageGraph(packages);
+    }));
+
+    it("should add all transitive dependencies of passed in packages with no repeats", () => {
+      // we need to start with one package and have it add the deps required
+      const packagesToExpand = packages.filter((pkg) => pkg.name === "package-1");
+      const packagesWithDeps = PackageUtilities.addDependencies(packagesToExpand, packageGraph);
+
+      // should follow all transitive deps and pass all packages except 7 with no repeats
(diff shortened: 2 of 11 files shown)


lukebatchelor:
Have rerun this and am running into issues in the integration tests in appveyor. Are there any known issues there?

src/PackageUtilities.js
```diff
@@ -118,7 +118,9 @@ export default class PackageUtilities {
         }
       });
 
-      dependentPackages.push(pkg);
+      if (!packageAlreadyFound(pkg.name)) {
+        dependentPackages.push(pkg);
```
hzoo: looks like a good change 😄 , should probably just turn this array into a set?

evocateur: A `Set` seems reasonable here, although out of scope for this PR.

evocateur:
@lukebatchelor Sorry for the test breaks, those are actually not related to this change. (my own dumb snapshot "normalization" is causing this problem because lerna's version _happens_ to match that of one of the fixtures)

lukebatchelor:
Wow, thank you so much for the super quick responses!

PR #943, opened 2017-07-26
Return error and exit on invalid command

### LINKED ISSUE #809: Lerna should error when provided an invalid command
```
$ lerna boostrap
```

## Expected Behavior

```
lerna: invalid command "boostrap"
```
Exit code non-zero.

## Current Behavior

No output. Exit code zero.

## Context
We had a build script with a lerna command misspelled that took us way too long to diagnose due to lerna quietly exiting.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.0.0-rc.2 |
| `npm --version`  | 4.4.4 |
| `yarn --version` | 0.23.4 |
| `node --version` | v6.2.2 |

| OS | Version |
| --- | --- |
| Windows 10 Enterprise | 14393 |

Siilwyn:
Return error message and exit with non-zero code. Disable help on fail
so that the error is clearly visible.

Resolves #809

## Motivation and Context

See issue #809

## How Has This Been Tested?

It's not, only seeing command specific tests. If I should add a test please point to where it should be added.

## Types of changes

## Checklist:

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -35,10 +35,12 @@ export default function CLI(argv, cwd) {
     .usage("Usage: $0 <command> [options]")
     .options(globalOptions).group(globalKeys, "Global Options:")
     .commandDir("../lib/commands")
-    .demandCommand()
+    .demandCommand(1, "Pass --help to see all available commands and options.")
     .help("h").alias("h", "help")
     .version().alias("v", "version")
     .wrap(cli.terminalWidth())
+    .strict()
+    .showHelpOnFail(false, "A command is required.")
     .epilogue(dedent`
       When a command fails, all logs are written to lerna-debug.log in the current working directory.
 


src/cli.js
```diff
@@ -39,6 +39,8 @@ export default function CLI(argv, cwd) {
     .help("h").alias("h", "help")
     .version().alias("v", "version")
     .wrap(cli.terminalWidth())
+    .strict()
+    .showHelpOnFail(false)
```
evocateur: We should provide a message to `showHelpOnFail()`:
```js
    .showHelpOnFail(false, "Pass --help to see all available commands and options")
```

The existing `demandCommand()` usage should also include a message, to account for cases when `lerna` is called without arguments.
```js
    .demandCommand(1, "A command is required")
```

Siilwyn: Thanks for the feedback!
It even returns `A command is required.` when given an incorrect argument like `lerna nope`. But it seems `yargs` has no way of preventing that.

evocateur: Messing around locally, without the extra `"A command is required"` message to `demandCommand()`, I was getting less-than-useful feedback like `Unknown argument: nope`.

Siilwyn: So we're good to go right?

The CI is doing some funky stuff though.

evocateur: Yes, thank you!

PR #946, opened 2017-07-29
Swallow promise rejection in registerChild as it is handled via callback

viliusl:
## Description
`ChildProcessUtilities.registerChild()` attaches itself to an `execa` thenable, but then there is no catch, so if command executed fails, there is an `unhandledRejection`. Given `execa` promise is returned/handled via callback, it's ok to swallow error.

## Motivation and Context
Not clean - using `lerna` api for an extension and get unhandled rejections in tests:)

## How Has This Been Tested?
Cleaned-up tests, I could add regression test though.

## Types of changes

## Checklist:

### DIFF
--- a/src/ChildProcessUtilities.js
+++ b/src/ChildProcessUtilities.js
@@ -81,7 +81,7 @@ function registerChild(child) {
     if (children === 0) {
       emitter.emit("empty");
     }
-  });
+  }).catch(() => {});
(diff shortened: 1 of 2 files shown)


evocateur:
Thanks @viliusl!

viliusl:
No, thank you:)

PR #958, opened 2017-08-05
Support pre* options in --cd-version with optional --preid meta suffix

### LINKED ISSUE #567: force publish for prereleased packages

Hypnosphi:
## Description
`prerelease | prepatch | preminor | premajor` options were added to `--cd-version flag`. `--preid` flag was added (used as `semver.inc`'s [third argument](https://github.com/npm/node-semver#prerelease-identifiers)).

## Motivation and Context
Fixes #567

## How Has This Been Tested?
See changes in tests

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,13 +380,23 @@ Useful in [Continuous integration (CI)](https://en.wikipedia.org/wiki/Continuous
 #### --cd-version
 
 ```sh
-$ lerna publish --cd-version (patch | major | minor)
+$ lerna publish --cd-version (patch | minor | major | prepatch | preminor | premajor | prerelease)
 # uses the next semantic version(s) value and this skips `Select a new version for...` prompt
 ```
 
 When run with this flag, `publish` will skip the version selection prompt (in independent mode) and use the next specified semantic version.
 You must still use the `--yes` flag to avoid all prompts. This is useful when build systems need
 to publish without command prompts. Works in both normal and independent modes.
+
+#### --pre-id
+
+```sh
+$ lerna publish --cd-version (prepatch | preminor | premajor | prerelease) --pre-id beta
+# prefixes the prerelease tag with given identifier
+```
+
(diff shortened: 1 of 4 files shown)


Hypnosphi:
That's good, and it covers greater scope, so I'll close this in favor of commenting there

evocateur:
Thanks!

Issue #962, opened 2017-08-07
publish --canary should have some way of limiting publish to changed packages only


darrylhodgins:
## Expected Behavior
In much the same way that `lerna publish` behaves (by only publishing packages with changes, plus packages with dependencies on those), `lerna publish --canary` should have some way to restrict publishing to only those packages with changes.

## Current Behavior
Currently, publish --canary forces inclusion of all packages ([here, in UpdatedPackagesCollector.js](https://github.com/lerna/lerna/blob/950e26f503a955982e564f95dd99b1b89ed5abcd/src/UpdatedPackagesCollector.js#L160))

## Possible Solution
Use the `--since` flag to filter out unchanged packages.

## Context
In a project with a large number of packages, being able to create `--canary` builds on only packages that have actually changed since the last release would save significant time.  In my project, this would reduce the build time by at least 80%.

bradleat:
Is this going to be merged in soon?

evocateur:
#963 needs tests, and updating due to changes in #1020

bladedancer:
I'm hitting the exact same issue and started looking at the code to create an up-to-date PR but the more I dig the more something feels wrong here. So before spending too much time on it I wanted to check what's the acceptable behavior for canary.

Our CI system is publishing canary builds on feature branches to essentially create a testable but disposable artifact. We can then independently QA the features prior to merging down to master. The merge to master is publishing the normal build.

Anyway as noted here, the canary publish is publishing all packages. This is noisy, slower and also consumes a lot of space in our private registry (our CI system is doing this on every push). So we're looking into limiting this to just publishing deltas.

The --since option seems to be a workable solution but just in general it feels like I'm abusing the --canary option. The docs describe it's usage as being for per commit publishes and not as being for disposable builds.
Specifically code like this in [UpdatedPackagesCollector](https://github.com/lerna/lerna/blob/master/src/UpdatedPackagesCollector.js#L65)
seems to require it's per commit, i.e. that diff will just return the latest commit, if I've multiple commits in a single push then the canary only sees the latest. Which I suppose is in line with what the docs say. 

So long story short, if --canary got a --since flag would it be in line with the expectations of the canary functionality?

thanks.

evocateur:
Does the solution described in https://github.com/lerna/lerna/issues/1161#issuecomment-359626768 get you closer to this sort of thing? It should get you "partial" publishing, and you could even add `--since staging` in the feature branches to get exactly what has changed in the branch.

[...]

PR #963, opened 2017-08-07
Allow --since flag on publish --canary

darrylhodgins:
## Description
Add check for `--since` flag on `publish --canary`

## Motivation and Context
An effort to reduce the build time for canary publish on large projects.  See #962

## How Has This Been Tested?
Tested locally using a variety of options for `--since`.

## Types of changes

## Checklist:

### DIFF
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -57,7 +57,7 @@ export default class UpdatedPackagesCollector {
     let { since } = options;
 
     if (GitUtilities.hasTags(execOpts)) {
-      if (canary) {
+      if (canary && typeof since !== "string") {
         let currentSHA;
 
         if (canary !== true) {
@@ -157,7 +157,7 @@ export default class UpdatedPackagesCollector {
       return (
         this.updatedPackages[pkg.name] ||
         (this.options[SECRET_FLAG] ? false : this.dependents[pkg.name]) ||
-        this.options.canary
+        (this.options.canary && typeof this.options.since !== "string")
       );
     }).map((pkg) => {
       this.logger.verbose("has filtered update", pkg.name);
--- a/src/Command.js
+++ b/src/Command.js
@@ -39,7 +39,7 @@ export const builder = {
     describe: dedent`
       Restricts the scope to the packages that have been updated since
       the specified [ref], or if not specified, the latest tag.
-      (Only for 'run', 'exec', 'clean', 'ls', and 'bootstrap' commands)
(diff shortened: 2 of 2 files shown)


darrylhodgins:
Hmm… [the same same CI errors are happening on latest `master`](https://travis-ci.org/lerna/lerna/jobs/258257507)

darrylhodgins:
My local, totally non-scientific, tests were pretty reasonable, but absolutely: need something in the PR.  Will add some!  Thanks :)

evocateur:
I totally trust you, I just don't trust _myself_ ;)

darrylhodgins:
I keep telling the team here: tests are to help ensure that the functionality you are building today will continue to work tomorrow.  If there's no test, there's no guarantee that someone else won't break it next week…

Obligatory movie quote: "I trust everyone, I just don't trust the devil inside them."

src/UpdatedPackagesCollector.js
```diff
@@ -157,7 +157,7 @@ export default class UpdatedPackagesCollector {
       return (
         this.updatedPackages[pkg.name] ||
         (this.options[SECRET_FLAG] ? false : this.dependents[pkg.name]) ||
-        this.options.canary
+        (this.options.canary && typeof this.options.since !== "string")
```
GeorgeTaveras1231: (**Just my opinion**): This condition does not depend on the `pkg` argument (the subject of the filter) so it would be much clearer to understand how this condition affects `collectUpdates` if it was taken out of the `filter` callback. If I'm understanding this correctly, when `canary` is specified and `since` isn't, `collectUpdates` will collect **all** packages. This would be much clearer if written something like:

```js
collectUpdates() {
  const shouldCollectAllPackages = this.options.canary && typeof this.options.since !== 'string';
  let collectedPackages = this.packages;
  
  if (!shouldCollectAllPackages) {
    collectedPackages = this.packages.filter((pkg) => {
      return this.updatedPackages[pkg.name] || this.options[SECRET_FLAG] ? false : this.dependents[pkg.name];
    });
  }

  /* Omitted */
}
```

src/UpdatedPackagesCollector.js
```diff
@@ -57,7 +57,7 @@ export default class UpdatedPackagesCollector {
     let { since } = options;
 
     if (GitUtilities.hasTags(execOpts)) {
-      if (canary) {
+      if (canary && typeof since !== "string") {
```
GeorgeTaveras1231: This line conflicts with #1020. Just a heads up.

PR #965, opened 2017-08-09
Avoid false positives during integration test filtering

darrylhodgins:
…that won't just match any old instance of "2.0.0"

**Caution: I also increased the timeout for the integration tests from 70s to 300s.**  I know it's probably a bad idea, but some of those tests seem to take forever on node v.8

A slightly-too-complicated regular expression.  The `.` characters present in `constants.LERNA_VERSION` will still match any character, but…

[Here's more info on the regexp I added.](https://regex101.com/r/E8GYA8/1)

I had considered bumping all the version numbers in the snapshots up into the `1000000.0.0` range, but when I thought about it some more I saw that the existing regexp here would match `1000002.0.0`.  Since there are versions from `0.0.0` through `6.0.0` in the snapshots, I figured this was the safest approach (at least until lerna gets to `6.0.1`, at which point even the original naïve regexp would do just fine.

Also removed `strict()` from `cli.js` based on conversation with @evocateur.

## Motivation and Context
Builds have been failing in Travis since 2.0.0 was tagged.

## How Has This Been Tested?
`npm run integration` && `npm run test` locally.

## Types of changes

## Checklist:

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -39,7 +39,6 @@ export default function CLI(argv, cwd) {
     .help("h").alias("h", "help")
     .version().alias("v", "version")
     .wrap(cli.terminalWidth())
-    .strict()
     .showHelpOnFail(false, "A command is required.")
     .epilogue(dedent`
       When a command fails, all logs are written to lerna-debug.log in the current working directory.
--- a/test/helpers/serializePlaceholders.js
+++ b/test/helpers/serializePlaceholders.js
@@ -6,8 +6,9 @@ const _ = require("lodash");
 const normalizeNewline = require("normalize-newline");
 const constants = require("./constants");
 
-const VERSION_REGEX = new RegExp(`v?${constants.LERNA_VERSION}`, "gm");
-// TODO: maybe less naïve regex?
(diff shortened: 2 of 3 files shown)


darrylhodgins:
*grumble grumble grumble*

These two seem to take _forever_ on my end, but they do eventually finish… Maybe ~70~ 180 seconds wasn't quite enough for the timeout ;)

```
 PASS  test/integration/lerna-clean.test.js (53.974s)
  lerna clean
    ✓ global (13935ms)
    ✓ local npm (39441ms)
    ○ skipped 1 test
```

and

```
 PASS  test/integration/lerna-bootstrap.test.js (72.522s)
  lerna bootstrap
    from CLI
      ✓ bootstraps all packages (64923ms)
      ✓ respects ignore flag
      ✓ git repo check is ignored by default (1ms)
      ✓ --npm-client yarn
      ✓ passes remaining arguments to npm client
      ✓ passes remaining arguments + npmClientArgs to npm client
    from npm script
      ✓ bootstraps all packages (7056ms)
      ○ skipped 1 test
```

evocateur:
Thanks @darrylhodgins!

darrylhodgins:
Glad I could help & throw some regexp wizardry at it :)

Issue #887, opened 2017-06-19
Stack trace on no changes when running publish


SimenB:
## Expected Behavior

I expect to just get a nice message and exit 1 on no changes when running publish

## Current Behavior

```sh-session
$ yarn lerna publish
yarn lerna v0.24.6
$ "/Users/simbekkh/repos/podium/node_modules/.bin/lerna" publish
lerna info version 2.0.0-rc.5
lerna info versioning independent
lerna info Checking for updated packages...
lerna info Comparing with tag @finn-no/podium-client@1.0.0-alpha.11
lerna info No updated packages to publish.
lerna ERR! execute Error: On branch master
lerna ERR! execute Your branch is up-to-date with 'origin/master'.
lerna ERR! execute nothing to commit, working tree clean
lerna ERR! execute
lerna ERR! execute     at Function.module.exports.sync (/Users/simbekkh/repos/podium/node_modules/execa/index.js:277:26)
lerna ERR! execute     at Function.execSync (/Users/simbekkh/repos/podium/node_modules/lerna/lib/ChildProcessUtilities.js:59:30)
lerna ERR! execute     at Function.commit (/Users/simbekkh/repos/podium/node_modules/lerna/lib/GitUtilities.js:84:39)
lerna ERR! execute     at PublishCommand.gitCommitAndTagVersionForUpdates (/Users/simbekkh/repos/podium/node_modules/lerna/lib/commands/PublishCommand.js:578:30)
lerna ERR! execute     at PublishCommand.commitAndTagUpdates (/Users/simbekkh/repos/podium/node_modules/lerna/lib/commands/PublishCommand.js:560:26)
lerna ERR! execute     at PublishCommand.execute (/Users/simbekkh/repos/podium/node_modules/lerna/lib/commands/PublishCommand.js:246:16)
lerna ERR! execute     at PublishCommand._attempt (/Users/simbekkh/repos/podium/node_modules/lerna/lib/Command.js:265:21)
lerna ERR! execute     at /Users/simbekkh/repos/podium/node_modules/lerna/lib/Command.js:252:15
lerna ERR! execute     at /Users/simbekkh/repos/podium/node_modules/lerna/lib/Command.js:274:13
lerna ERR! execute     at PublishCommand.initialize (/Users/simbekkh/repos/podium/node_modules/lerna/lib/commands/PublishCommand.js:205:9)
lerna ERR! execute  callback with error
lerna ERR! execute  Error: On branch master
[...]

PR #807, opened 2017-05-05
Change exit codes for `updated` and `publish`

koddsson:
## Description
This patch:

- Changes the exit code for the `updated` command from 0 to 1 when there are no changes to publish
- Changes the exit code from `publish` from 1 to 0 and instead of throwing a error with a message that there is nothing to publish, logs that message out.

## Motivation and Context

This change enables users that want to publish packages from a CI to more easily create a flow that relies on the exit codes of lerna commands.

Before this change CI builds would fail when `lerna publish` would return a non-zero exit code and it was not possible to tack `lerna updated` in front of it (`!lerna updated & lerna publish`) since it would always return a zero exit code.

Fixes #802 

## How Has This Been Tested?

I added a unit test for `lerna updated` and a integration test for `lerna publish`.

The `lerna updated` explicitly checks the exit code of that the command returns is 1 but the integration relies on the fact that if `lerna publish` would return a non zero exit code the test would fail.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

(This might be a breaking feature if we consider changing the exit code a breaking change, I'm not sure what you think)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

The docs don't mention exit codes as far as I can see but I can certainly add something to that effect to the docs if you'd like :)

PR #966, opened 2017-08-09
Prevent crash on publish when no updated packages

alisd23:
## Description
Currently running `lerna publish` when no packages need updating causes the command to crash unnecessarily.

This small fix causes the command to complete if no updates are found, instead of trying to continue to the next section.

This change was missing from the PR: #807 
Relates to the issue: #887 

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -149,7 +149,7 @@ export default class PublishCommand extends Command {
 
     if (!this.updates.length) {
       this.logger.info("No updated packages to publish.");
-      callback(null, true);
+      callback(null, false);
       return;
     }
 


alisd23:
More complex solution than I thought. Also some tests are broken on master it seems, plus publish command prints to `stderr` even when there is no error, which I can't figure out.

evocateur:
`stderr` is where all logging goes, and does not mean an error is occurring.

alisd23:
Oh. Out of interest: why not use stdout? Maybe this fix is valid then.

evocateur:
stdout is for output that might be piped. stderr is for logging (it's a very common unix idiom).

PR #979, opened 2017-08-23
Run "prepare" lifecycle script during bootstrap

### LINKED ISSUE #421: [bootstrap] Support for "prepare" npm event-script

Andarist:
Followup to #499

## Description
Handle more lifecycle events during bootstrap - pairing in the behaviour with latest npms

## Motivation and Context
Handle correctly specified lifecycle events for npm4+. At the moment behaviours for regular npm install and lerna bootstrap are out of sync, which cause problems with preparation scripts i.e. not being run at all.
Fixes #421

## How Has This Been Tested?
Ran lerna's test suite + used this on my own project to check if added scripts are getting executed during bootstrap.

## Types of changes

## Checklist:

cc @wzrdzl and @evocateur

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -107,7 +107,9 @@ export default class BootstrapCommand extends Command {
         // postinstall bootstrapped packages
         (cb) => this.postinstallPackages(cb),
         // prepublish bootstrapped packages
-        (cb) => this.prepublishPackages(cb)
+        (cb) => this.prepublishPackages(cb),
+        // prepare bootstrapped packages
+        (cb) => this.preparePackages(cb)
       ], callback);
     }
 
@@ -173,6 +175,15 @@ export default class BootstrapCommand extends Command {
   prepublishPackages(callback) {
     this.logger.info("lifecycle", "prepublish");
     this.runScriptInPackages("prepublish", callback);
+  }
+
+  /**
+   * Run the "prepublish" NPM script in all bootstrapped packages
+   * @param callback
+   */
+  preparePackages(callback) {
+    this.logger.info("lifecycle", "prepare");
+    this.runScriptInPackages("prepare", callback);
   }
 
   /**
--- a/README.md
+++ b/README.md
@@ -197,7 +197,8 @@ When run, this command will:
 
 1. `npm install` all external dependencies of each package.
 2. Symlink together all Lerna `packages` that are dependencies of each other.
-3. `npm prepublish` all bootstrapped packages.
+3. `npm run prepublish` in all bootstrapped packages.
+4. `npm run prepare` in all bootstrapped packages.
 
 `lerna bootstrap` respects the `--ignore`, `--scope` and `--include-filtered-dependencies` flags (see [Flags](#flags)).
 
--- a/test/integration/__snapshots__/lerna-bootstrap.test.js.snap
+++ b/test/integration/__snapshots__/lerna-bootstrap.test.js.snap
@@ -16,6 +16,7 @@ lerna info Installing external dependencies
 lerna info Symlinking packages and binaries
 lerna info lifecycle postinstall
 lerna info lifecycle prepublish
+lerna info lifecycle prepare
 lerna success Bootstrapped 4 packages"
 `;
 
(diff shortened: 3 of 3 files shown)


Andarist:
Obviously I still need to add tests for the changes and probably adjust some docs, but would like to have this completed code-wise before I get to the docs and as for the tests I need to dig up more how do u write those for the project (dont worry though, they are comming! 😄 )

@evocateur You have mentioned [here](https://github.com/lerna/lerna/pull/499#issuecomment-324100681) using `npm install --ignore-scripts` during those lifecycles, but truth to be told I wasnt sure how that works and if I should add this now or was it already covered in the codebase.

src/commands/BootstrapCommand.js
```diff
@@ -107,7 +107,9 @@ export default class BootstrapCommand extends Command {
         // postinstall bootstrapped packages
         (cb) => this.postinstallPackages(cb),
         // prepublish bootstrapped packages
-        (cb) => this.prepublishPackages(cb)
+        (cb) => this.prepublishPackages(cb),
+        // prepare bootstrapped packages
+        (cb) => this.preparePackages(cb)
```
Andarist: dont think `bootstrap` needs any more work besides the tests, please correct me if im wrong

src/commands/PublishCommand.js
```diff
@@ -542,6 +542,8 @@ export default class PublishCommand extends Command {
 
     this.updates.forEach((update) => {
       this.execScript(update.package, "prepublish");
+      this.execScript(update.package, "prepare");
```
Andarist: this on the other hand could be parallelized to speed up the process, Im not sure what to use here though, I guess it would need to be a combo of `PackageUtilities.runParallelBatches` and `async.series`. I see the desired flow as:
- parallel execution of `async.series` per each package
- on completion one more parallel batch of publish tasks

I think its the safest one

evocateur: As I understand it, we want to ensure that a given set of lifecycle scripts finishes in a package before any dependent packages start their prepublish+ lifecycles. So `PackageUtilities.runParallelBatches` is definitely appropriate here. I'm not the sharpest with the async loops, but what you describe sounds like the right approach.

In any case, lots of testing!

evocateur: Actually, `this.execScript` is used for executing custom, non-npm lifecycle scripts that are just `.js` files living in a package-local `scripts` directory. They're a lerna-specific extension that allows you to do things _outside_ the normal npm lifecycles. Admittedly, not something I use, but largely inherited from the babel project.

I don't think we should extend this artificial pattern to the extra names, they're just arbitrary and never called during the normal `npm publish` lifecycles. `npm publish` is being called (in topological order) in `NpmUtilities.publishTaggedInDir` and it should automatically handle running the various lifecycles where appropriate, I think.

```
package-foo/
 - package.json
 - scripts/
    - prepublish.js
    - postpublish.js
 - src/
    - index.js
```

Andarist: huh, havent realized that, seemed like it was executing npm commands ;)

It seems that I should remove those lines and just let npm handle its lifecycle events on its own, right?

evocateur: Yep!

Andarist:
@evocateur ive updated my PR, please take a look. As far as I know its merge-ready. Last CI checks were passing (latest checks got triggered by a new commit containing only docs changes - so they should pass soon too ;) )

evocateur:
Thanks @Andarist!

Andarist:
@evocateur 
No problem. I have needed this :P so thanks for merging. Any ETA for a release?

evocateur:
How about right now? :shipit:

https://github.com/lerna/lerna/releases/tag/v2.1.2

Andarist:
Great! :) thank you

PR #970, opened 2017-08-13
Create configured "packages" directories during lerna init

Siilwyn:
## Description

Create the packages directory when `lerna init` is ran in an empty directory.

## Motivation and Context

https://github.com/lerna/lerna/issues/704

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #985, opened 2017-08-25
Update installation instructions to match lerna init behavior

sapegin:
## Description

`lerna init` will initialize a Git repo. It doesn’t create a `packages` folder but it should be fixed in https://github.com/lerna/lerna/pull/970


sapegin:
Looks like CI failures aren’t related to my changes — I’ve changed only documentation anyway.

evocateur:
Thanks @sapegin!

PR #1000, opened 2017-08-30
Add `lerna link` command

### LINKED ISSUE #341: [Feature] support method to /only/ link packages

Inkdpixels:
[Sponsored by Immowelt](https://stackshare.io/immowelt-group/)

@evocateur

## Description
This commit adds a new 'link' command to the CLI which will create symlinks across the lerna package dependencies. This command was requested in #341 and solves issues with using 'npm prune' with lerna based repositories, e.g. 'npm prune' removes extranous packages, which include the symlinks of lerna.

## Motivation and Context
Since the bootstrap command is a bit heavy, especially while hoisting, I opted for a seperate command since the bootsrap command is pretty complex as of now anyway and I don't think it will be particularly usefull to add option after option.

I also refactored most of the code, e.g. made it more purely functional, regarding the symlinkPackages method and extracted it into the PackageUtilities.

## How Has This Been Tested?
I've used the Bootstrap test as a base and included a test case with fixtures which checks if all of the symlinks are created accordingly with Jests snapshots feature.

All of the pre-existing tests including the code style checks are green. (Except for the Node 4 integration test, but this is also failing on master itself)

## Types of changes

## Checklist:

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -224,6 +224,9 @@ export default class PackageUtilities {
   /**
    * Symlink all packages to the packages/node_modules directory
    * Symlink package binaries to dependent packages' node_modules/.bin directory
+   * @param {Array.<Package>} packages
+   * @param {Object} packageGraph
+   * @param {Object} logger
    * @param {Function} callback
    */
   static symlinkPackages(packages, packageGraph, logger, callback) {
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -6,6 +6,7 @@ import path from "path";
 import readPkg from "read-pkg";
 
 import PackageGraph from "./PackageGraph";
+import FileSystemUtilities from "./FileSystemUtilities";
 import Package from "./Package";
 
 /**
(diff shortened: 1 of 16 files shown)


src/PackageUtilities.js
```diff
@@ -218,4 +219,147 @@ export default class PackageUtilities {
...
+
+  /**
+   * Symlink all packages to the packages/node_modules directory
+   * Symlink package binaries to dependent packages' node_modules/.bin directory
+   * @param {Array.<Package>} packages
+   * @param {Object} packageGraph
+   * @param {Object} logger
+   * @param {Function} callback
+   */
+  static symlinkPackages(packages, packageGraph, logger, callback) {
```
evocateur: I love this refactoring!

src/commands/LinkCommand.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const builder = {};
+
+export default class LinkCommand extends Command {
+  get defaultOptions() {
+    return {};
+  }
+
+  initialize(callback) {
+    this.exact = this.options.exact;
```
evocateur: I don't think this property is necessary?

src/commands/LinkCommand.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export const command = "link";
+
+export const describe = "Symlink together all packages which are dependencies of each other";
+
+export const builder = {};
+
+export default class LinkCommand extends Command {
+  get defaultOptions() {
```
evocateur: I believe this getter is not necessary, since you're not mixing in any defaults from the parent class.

src/commands/LinkCommand.js
```diff
@@ -0,0 +1,30 @@
+import Command from "../Command";
+import PackageUtilities from "../PackageUtilities";
+
+export function handler(argv) {
+  return new LinkCommand(argv._, argv).run();
```
evocateur: Would you updating this to the new pattern [here](https://github.com/lerna/lerna/blob/4d64a2813ff0c35fd46ed745bbca1833e5cb86f0/src/commands/BootstrapCommand.js#L13-L14)?

test/LinkCommand.js
```diff
@@ -0,0 +1,72 @@
...
+import FileSystemUtilities from "../src/FileSystemUtilities";
+
+// helpers
+import callsBack from "./helpers/callsBack";
+import exitWithCode from "./helpers/exitWithCode";
+import initFixture from "./helpers/initFixture";
+import normalizeRelativeDir from "./helpers/normalizeRelativeDir";
+
+// file under test
+import LinkCommand from "../src/commands/LinkCommand";
```
evocateur: The test needs to be updated to support the [yargsRunner pattern](https://github.com/lerna/lerna/commit/ab3c5c5cb42df9705e51eef970b1db263db9c813):

```js
import yargsRunner from "./helpers/yargsRunner";

// file under test
import * as commandModule from "../src/commands/LinkCommand";

const run = yargsRunner(commandModule);
```

The test case itself becomes much more compact:
```js
  describe("with local package dependencies", () => {
    beforeEach(stubSymlink);
    afterEach(resetSymlink);

    it("should symlink all packages", async () => {
      const testDir = await initFixture("LinkCommand/basic");
      const lernaLink = run(testDir);
      await lernaLink();

      expect(symlinkedDirectories(testDir)).toMatchSnapshot();
    });
  });
```

Thanks!

src/PackageUtilities.js
Inkdpixels: Awesome! Will add the feedback tomorrow :) 👍

kevroadrunner:
@evocateur Thanks again, your feedback has been added.

Issue #730, opened 2017-04-05
[Bug] Import failing (2.0.0-beta.38)


jmcpeak:
Trying to use lerna import. I saw in a few other issues that a rebase should be tried if import failed. I tried that first - no luck.

Here is the error message

`lerna import /Users/mcpeakj/projects/document-ui`
`Lerna v2.0.0-beta.38`
`About to import 248 commits from /Users/mcpeakj/projects/document-ui into packages/document-ui`
`? Are you sure you want to import these commits onto the current branch? Yes`
`Errored while running ImportCommand.execute`
`Failed to apply commit 5745ce9.`
`Error: Command failed: git am -3`
`error: Failed to merge in the changes.`

`Rolling back to previous HEAD (commit 0515496).`

This repo is hosted by JFrog/Artifactory, at my work - it is not a github repo.
I tried importing one of my public github repos and that worked fine.

[lerna-debug.log.zip](https://github.com/lerna/lerna/files/900592/lerna-debug.log.zip)

Robinfr:
I ran into the same problem when I tried to import a Git repository from a local path..

jmcpeak:
So if the -B flag is added (https://git-scm.com/docs/git-format-patch#git-format-patch) a lot more information is outputted.

am80l:
+1 same issue

evocateur:
@am80l Please don't add "+1" comments. If you have a reproducible example or other context, that would be very helpful.

kanzelm3:
+1

dmaksimovic:
I added a pull request https://github.com/lerna/lerna/pull/1019 that fixes this issue.

dmaksimovic:
I think this issue can be closed as fixed since pull request #1019 was merged

PR #1019, opened 2017-09-19
Add --flatten option to use when merge conflicts cannot be imported

dmaksimovic:
Add option "--flatten" to import each merge as a single change the merge introduced

## Description
Instead of doing "git format-patch", patches can be created by "git log" with a certain set of options that mimic format-patch, but with git log some additional options are available, like "--first-parent", which only follows main branch, so that side branch commits are not taken in separately, but instead only the merge commit is imported containing all changes from all branches commits that were merged there.

## Motivation and Context
This is required for situations when there were merge commits containing merge conflict resolutions in imported project. In the current implementation, "lerna import" fails because "git format-patch" does not include merge commits and its resolution, so such projects can't be imported. If "--flatten" is provided with the proposed solution, these problematic projects are successfully imported.

The issue was first reported in #730, and I personally had this issue so I had to fix it in order to use Lerna. Now I'm sharing the solution.

## How Has This Been Tested?
Tested with all sorts of importable repos, with several branches merged, with binary content, with conflicts and without conflicts, with my option and with the original option. Also all auto tests pass and one automatic test added.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.
- [x ] I have added tests to cover my changes.
- [x ] All new and existing tests passed.

### DIFF
--- a/src/commands/ImportCommand.js
+++ b/src/commands/ImportCommand.js
@@ -120,12 +120,27 @@ export default class ImportCommand extends Command {
   createPatchForCommit(sha) {
     let patch = null;
     if (this.options.flatten) {
-      const diff = this.externalExecSync("git", ["log", "--reverse", "--first-parent", "-p",
-        "-m", "--pretty=email", "--stat", "--binary", "-1", sha]);
+      const diff = this.externalExecSync("git", [
+        "log",
+        "--reverse",
+        "--first-parent",
+        "-p",
+        "-m",
+        "--pretty=email",
+        "--stat",
+        "--binary",
+        "-1",
+        sha
+      ]);
       const version = this.externalExecSync("git", ["--version"]).replace(/git version /g, '');
       patch = `${diff}\n--\n${version}`;
     } else {
-      patch = this.externalExecSync("git", ["format-patch", "-1", sha, "--stdout"]);
+      patch = this.externalExecSync("git", [
+        "format-patch",
+        "-1",
+        sha,
+        "--stdout"
+      ]);
     }
 
     const replacement = "$1/" + this.targetDir;
--- a/src/commands/ImportCommand.js
+++ b/src/commands/ImportCommand.js
@@ -23,10 +23,22 @@ export const builder = {
   "yes": {
     group: "Command Options:",
     describe: "Skip all confirmation prompts",
-  }
+  },
+  "flatten": {
+    group: "Command Options:",
+    describe: "Import each merge commit as a single change the merge introduced",
+  },
 };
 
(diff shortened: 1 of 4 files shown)


dmaksimovic:
I think the AppVeyor build is failing for its own reasons. I don't see any relationship of its output to my changes. Can somebody have a look at this change please? It's been 7 days now, I'd appreciate some feedback

evocateur:
Sorry, yeah, the appveyor build is flapping a lot. I've been extremely busy, I'll take a look when I can.

src/commands/ImportCommand.js
```diff
@@ -105,24 +117,39 @@ export default class ImportCommand extends Command {
     return ChildProcessUtilities.execSync(cmd, args, this.externalExecOpts);
   }
 
+  createPatchForCommit(sha) {
+    let patch = null;
+    if (this.options.flatten) {
+      const diff = this.externalExecSync("git", ["log", "--reverse", "--first-parent", "-p",
```
evocateur: I'd appreciate if the array of args was spread one-per-line for readability, thanks.

src/commands/ImportCommand.js
```diff
@@ -105,24 +117,39 @@ export default class ImportCommand extends Command {
     return ChildProcessUtilities.execSync(cmd, args, this.externalExecOpts);
   }
 
+  createPatchForCommit(sha) {
```
evocateur: Thanks for extracting this method :)

test/ImportCommand.js
```diff
@@ -57,6 +57,35 @@ describe("ImportCommand", () => {
       expect(await pathExists(packageJson)).toBe(true);
     });
 
+    it("imports a repo with conflicted merge commits when run with --flatten", async () => {
+      const cwdExternalDir = { cwd: externalDir };
+      const branchName = "conflict_branch";
+      const conflictedFileName = "conflicted-file.txt";
+      const conflictedFile = path.join(externalDir, conflictedFileName);
+      await fs.writeFile(conflictedFile, "initial content");
```
evocateur: Please put newlines before the `fs.writeFile()` bits so the logical steps are a bit easier to follow, thanks.

src/commands/ImportCommand.js
dmaksimovic: No worries, it was my pleasure. Changes applied as requested

evocateur:
Thanks!

dmaksimovic:
Cheers!

PR #1020, opened 2017-09-20
Use --canary=<value> as prerelease tag, not commit-ish

### LINKED ISSUE #1008: lerna publish --canary=${my_prerelease_tag} doesn't work
When I use the publish command like so:

```
lerna publish --canary=staging
```

I would expect changed modules to be published with versions similar to:

```
Changes:
  - my-module: 1.0.0 => 1.0.0-staging.dd3c916f
```

Instead it tries to use the value passed to the `--canary` option to work out a range of git commits:

```
$ lerna publish --yes --canary=staging
lerna info version 2.1.2
lerna info versioning independent
lerna info canary enabled
lerna info Checking for updated packages...
lerna info Comparing with staging^..staging.
lerna ERR! initialize Error: fatal: bad revision 'staging^..staging'
```

## Expected Behavior

The value passed to the canary flag should appear in the published version string.

## Current Behavior

The value passed to the canary flag is used as a git commit id which explodes.

1. Pass a value to the canary flag
2.
3.
4.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.1.2 |
| `npm --version`  | 5.4.0 |
| `yarn --version` | n/a |
| `node --version` | 6.9.1 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Sierra | 10.12.5 |

achingbrain:
## Description

The docs say [we should be able to override the meta suffix by passing the canary flag as a string value](https://github.com/lerna/lerna#--canary--c).  lerna takes this value and tries to use it to work out what range of commits it should consider when publishing a package so you can't actually override the meta suffix unless you use a commit id to do this.  Considering the commit id is added to the version number for canary releases anyway this seems redundant.

eg:

This works:
```
lerna publish --canary
...
Changes:
 - package-1: 1.0.0 => 1.1.0-alpha.0a2f1a97
 - package-1: 1.0.0 => 1.1.0-alpha.0a2f1a97
```

According to the docs this should work but doesn't:
```
lerna publish --canary=beta
...
lerna ERR! initialize Error: fatal: bad revision 'beta^..beta'
*kaboom!*
```

## Motivation and Context

It would be useful to be able to override the meta suffix as described in the docs.

Fixes #1008.

## How Has This Been Tested?

Added some unit tests for the `UpdatedPackagesCollector` module and integration tests for the canary flag.

## Types of changes

## Checklist:

### DIFF
--- a/test/UpdatedPackagesCollector.js
+++ b/test/UpdatedPackagesCollector.js
@@ -1,5 +1,28 @@
+// mocked modules
+import GitUtilities from "../src/GitUtilities";
+
 // file under test
 import UpdatedPackagesCollector from "../src/UpdatedPackagesCollector";
+
+jest.mock("../src/GitUtilities");
+
+const filteredPackages = [{
+  name: 'package-1',
+  location: 'location-1'
+}, {
+  name: 'package-2',
+  location: 'location-2'
+}];
+
+const logger = {
+  silly: () => {},
+  info: () => {},
+  verbose: () => {}
+};
+
+const repository = {
+  rootPath: 'root-path'
+};
 
 describe("UpdatedPackagesCollector", () => {
   it("should exist", () => {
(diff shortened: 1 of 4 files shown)


achingbrain:
Hmm, the appveyor builds seem to fail in different ways each time, none of which seem to be related to my changes.

GeorgeTaveras1231:
@achingbrain was going to work on this myself but glad to see you got this.

Additionally, I wanted to raise a concern with the lerna maintainers about the role that the `canary` option value is playing and whether it also needs to be used as the `pre-id`, even though there is already a flag to define the `pre-id`. It's not a huge deal but it just feels inconsistent and my response was 🤔 when I realized I could not use `--pre-id` for this.

achingbrain:
I see what you mean.  That said canary works well for my current project which requires per-commit automated CI deployment for internal testing with a manual public release gate.

@evocateur any thoughts on this PR?

evocateur:
I appreciate the PR, @achingbrain. I am not generally a fan of "polymorphic" options that we inherited from the more lenient `meow` cli parser. I would much prefer `--preid` to be used for this purpose, as @GeorgeTaveras1231 suggested (`--canary` would always be a boolean, never a string).

`--preid`'s implicit default is "alpha", which is exactly what the (implicit :P) default of `--canary`-as-string is now. I think that's the long-term solution. In the mean-time, this PR seems reasonable. It's not a breaking change, imo, because the functionality never worked as described during the non-beta 2.x cycle to date.

test/UpdatedPackagesCollector.js
```diff
@@ -1,10 +1,84 @@
...
+const repository = {
+  rootPath: 'root-path'
+};
+
 describe("UpdatedPackagesCollector", () => {
   it("should exist", () => {
     expect(UpdatedPackagesCollector).toBeDefined();
   });
 
   it("needs better tests");
```
evocateur: It now has them, thank you! (you can delete this empty test case, if you want)

test/integration/lerna-publish.test.js
```diff
@@ -13,6 +13,9 @@ import loadPkgManifests from "../helpers/loadPkgManifests";
 const lastCommitMessage = (cwd) =>
   execa.stdout("git", ["log", "-1", "--format=%B"], { cwd }).then(normalizeNewline);
 
+const lastCommitId = (cwd) =>
+  execa.stdout("git", ["rev-parse", "HEAD"], { cwd }).then((line) => line.trim().substring(0, 8));
```
evocateur: I thought `execa` automatically trimmed `stdout`?

test/integration/lerna-publish.test.js
```diff
@@ -66,6 +69,36 @@ describe("lerna publish", () => {
...
+      "publish",
+      "--canary",
+      "--skip-npm",
+      "--yes",
+    ];
+
+    const { stdout, stderr } = await execa(LERNA_BIN, args, { cwd });
+    const hash = await lastCommitId(cwd);
+    expect(stdout.replace(new RegExp(hash, 'g'), 'hash')).toMatchSnapshot("stdout: canary default version");
+    expect(stderr).toMatchSnapshot("stderr: canary beta version");
```
evocateur: This snapshot label should be `stderr: canary default version`, looks like copy+paste oversight.

test/UpdatedPackagesCollector.js
achingbrain: Heh, I left that in because I agree with the sentiment 😉 

Will remove.

test/integration/lerna-publish.test.js
achingbrain: Removed

test/integration/lerna-publish.test.js
achingbrain: Fixed

achingbrain:
@evocateur I've made the changes as requested, is there anything else that needs doing before this can be merged?  Thanks..

evocateur:
@achingbrain Thanks!

Issue #1024, opened 2017-09-25
How to detect which package has a circular dependency?


FezVrasta:
Sorry if I'm not following the template but I'm not reporting a bug nor requesting a feature.

I'm getting this error now that I enabled the Yarn workspaces but I don't know how to find out which packages are genereting the circular dependency.

```
lerna WARN ECYCLE Encountered a cycle in the dependency graph. This may cause instability!
```

Is there any way to find it out?

evocateur:
If you're using `lerna bootstrap`, add `--loglevel=silly` and it might become apparent? It's hard to say, since the warning itself could use improvement (although how much, I'm not certain, as it is a complex part of the codebase).

Grepping packages for references to other packages should be sufficient to identify circular dependencies, however. A -> B -> A is the sort of thing this warning is (attempting to) highlight.

FezVrasta:
Thanks for the tip! Unfortunately it doesn't provide any useful info :-(

```
❯ lerna bootstrap --loglevel=sillylerna sill input []
lerna sill flags { _: [ 'bootstrap' ], loglevel: 'silly', args: [] }
lerna verb rootPath /Users/federicozivolo/Acme/repos/stacks/ui-framework
lerna info version 2.2.0
lerna sill existsSync /Users/federicozivolo/Acme/repos/stacks/ui-framework/VERSION
lerna sill initialize attempt
lerna WARN ECYCLE Encountered a cycle in the dependency graph. This may cause instability!
lerna sill initialize successlerna sill execute attempt
lerna info Bootstrapping 16 packages
lerna sill installInDir ui-framework
lerna sill installInDir /Users/federicozivolo/Acme/repos/stacks/ui-framework/package.json
lerna sill getExecOpts { cwd: '/Users/federicozivolo/Acme/repos/stacks/ui-framework' }
lerna sill installInDir [ 'yarn', [ 'install', '--mutex', 'network:42424' ] ]
```

(it gets stuck there for several minutes)

I checked my dependencies manually but I couldn't find any circular dependency 😞

evocateur:
[...]

PR #1025, opened 2017-09-25
Log  which packages are throwing ECYCLE error

FezVrasta:
## Motivation and Context

Help for future cases of #1024

## How Has This Been Tested?

No testing, will do something tomorrow.

## Types of changes

Just added a log, is it a feature?

## Checklist:

### DIFF
--- a/src/PackageUtilities.js
+++ b/src/PackageUtilities.js
@@ -201,6 +201,8 @@ export default class PackageUtilities {
         batch.push(packages.reduce((a, b) => {
           return (refCounts[a.name] || 0) > (refCounts[b.name] || 0) ? a : b;
         }));
+        
+        log.silly("packages", batch.reduce((packages, package) => packages.push(package.name), []).join(', '));
       }
 
       batches.push(batch);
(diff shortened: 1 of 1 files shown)


src/PackageUtilities.js
```diff
@@ -201,6 +201,8 @@ export default class PackageUtilities {
         batch.push(packages.reduce((a, b) => {
           return (refCounts[a.name] || 0) > (refCounts[b.name] || 0) ? a : b;
         }));
+        
+        log.silly("packages", batch.reduce((pkgs, pkg) => pkgs.push(pkg.name), []).join(', '));
```
evocateur: [Array.prototype.push()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) actually returns the new length value, not the array pushed into. The `reduce` callback will need to be slightly longer (returning `pkgs` after pushing into it), and I don't think you'll need the `.join(', ')`, as the logger handles displaying arrays across multiple lines fairly well.

zxqfox: `log.silly("packages", batch.reduce((pkgs, pkg) => (pkgs.push(pkg.name), pkgs), []))` ^.^

prob something like that could be easier to read:
`log.silly("packages", [].concat.apply([], batch.map(pkg => pkg.name)))`

evocateur:
Or just `batch.map((pkg) => pkg.name)`, even.

Issue #774, opened 2017-04-21
Execute scripts preversion, version and postversion during lerna publish.


ncochard:
Hello

First of all, thank you for `lerna`, awesome tool.

The command `npm version ...` will execute the npm scripts `preversion`, `version` and `postversion`. See this link for more details: [https://docs.npmjs.com/misc/scripts](https://docs.npmjs.com/misc/scripts).

> * preversion: Run BEFORE bumping the package version.
> * version: Run AFTER bumping the package version, but BEFORE commit.
> * postversion: Run AFTER bumping the package version, and AFTER commit.

It would be great if the command `lerna publish` could execute these scripts when increasing the version number of a package.

noherczeg:
Lerna runs the `npm publish` command internally. I believe it should work currently as-is.

On the other hand `npm publish` does not run `version` internally if I'm not mistaken.

ncochard:
That is correct: `npm publish` does not run `npm version` internally.

Let me try to clarify my suggestion...

The typical workflow in a pure *npm* project would be the following:

1. We fist execute `npm version` to increase the version number and create a git tag.
   (This also executes our custom npm scripts `npm run preversion`, `npm run version` and `npm run postversion`.)
2. Then we can execute `npm publish`.
   (This also execute `npm run prepublish`, `npm run prepublishOnly`, etc.)

Our workflow is simplified with lerna because the command `lerna publish` figuratively speaking *encapsulates* the `npm version` command and the `npm publish` command.

Since lerna already honors the script `prepublish`; for consistency, it would be useful if it also honored the scripts `version`. This would allow us to migrate npm projects to lerna and benefit from the same features.

I can understand however that it is probably not a high priority feature: the scripts `preversion`, `version` and `postversion` are not often used.

evocateur:
I'm in favor of supporting lifecycle scripts whenever we can. I don't envision actually calling `npm version` because of all the side effects, but running the scripts at the appropriate times seems very doable.

cwmoo740:
@evocateur 

These lifecycle scripts are currently non-blocking, so things might not work 100% as expected. For example, if a preversion script fails, it doesn't halt the commit, and if a preversion script adds files to be committed, they might not make it in time.

I can open another PR to make them wait on the scripts if that's worthwhile.

PR #1029, opened 2017-09-27
Call version lifecycle scripts during publish

cwmoo740:
## Description

Following the guidelines here: https://docs.npmjs.com/cli/version

Lerna will now call preversion, version, and postversion scripts at the appropriate times in each package when running `lerna publish`.

## Motivation and Context

This allows more hooks into the lerna publish command. For example, I want to run a custom changelog generation script after package.json version is updated, but before the git commit.

Closes: https://github.com/lerna/lerna/issues/774

## How Has This Been Tested?

Added some tests to verify that the three lifecycle scripts are called. 

I wanted to mock Package.js but found no easy way to do that - the changes I was working on ended up affecting many of the tests in Publish.js, since it doesn't import or have access to the Package class directly.

## Types of changes

This may be a breaking change because it now calls version scripts on publish, whereas before they were ignored.

## Checklist:

- [ ] ?? My change requires a change to the documentation.

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -536,11 +536,25 @@ export default class PublishCommand extends Command {
       this.updatePackageDepsObject(pkg, "devDependencies", exact);
       this.updatePackageDepsObject(pkg, "peerDependencies", exact);
 
+      // exec preversion script
+      pkg.runScript("preversion", (err) => {
+        if (err) {
+          this.logger.error("publish", "error running preversion", pkg.name, err);
+        }
+      });
+
       // write new package
       writePkg.sync(packageJsonLocation, pkg.toJSON());
       // NOTE: Object.prototype.toJSON() is normally called when passed to
       // JSON.stringify(), but write-pkg iterates Object.keys() before serializing
       // so it has to be explicit here (otherwise it mangles the instance properties)
+
+      // exec version script
+      pkg.runScript("version", (err) => {
+        if (err) {
+          this.logger.error("publish", "error running version", pkg.name, err);
+        }
+      });
 
       // we can now generate the Changelog, based on the
       // the updated version that we're about to release.
@@ -595,7 +609,7 @@ export default class PublishCommand extends Command {
   }
 
   gitCommitAndTagVersionForUpdates() {
-    const tags = this.updates.map(({ package: { name } }) =>
+    const tags = this.updates.map(({ "package": { name } }) =>
       `${name}@${this.updatesVersions[name]}`
     );
     const subject = this.options.message || "Publish";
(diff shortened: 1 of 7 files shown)


src/commands/PublishCommand.js
```diff
@@ -595,7 +609,7 @@ export default class PublishCommand extends Command {
   }
 
   gitCommitAndTagVersionForUpdates() {
-    const tags = this.updates.map(({ package: { name } }) =>
+    const tags = this.updates.map(({ "package": { name } }) =>
```
cwmoo740: I can back out this change, but package is a reserved keyword and shouldn't be referenced without quotes.

evocateur: Indeed, good call!

test/PublishCommand.js
```diff
@@ -2,7 +2,6 @@ import chalk from "chalk";
 import log from "npmlog";
 import normalizeNewline from "normalize-newline";
 import path from "path";
-
```
evocateur: These newlines are significant (and mostly consistent across the other test files), please keep them.

test/PublishCommand.js
```diff
@@ -995,4 +992,48 @@ describe("PublishCommand", () => {
...
+    });
+
+    it("should not call version lifecycle scripts for a package missing them", async () => {
+      await run(testDir)();
+      scripts.forEach(script => {
+        expect(NpmUtilities.runScriptInDir).not.toHaveBeenCalledWith(
+          script,
+          [],
+          path.resolve(testDir, "packages", "package-2"),
+          jasmine.any(Function)
```
evocateur: Pretty sure `expect.any(Function)` should work here? Or is this Jest 19 (not upgraded yet for reasons) biting us again?

cwmoo740:
I corrected the tests to work with windows path separators - it should rerun the build.

test/PublishCommand.js
cwmoo740: Will do

cwmoo740:
The appveyor test failed on node 4, but it seems to be unrelated - can those tests be re-run?

evocateur:
Nah, those appveyor tests have been flapping due to some unrelated reason. It's quite likely a re-run would just fail randomly in the node 8 matrix. We're fine here.

evocateur:
Looks great, thanks @cwmoo740!

evocateur:
Oh, and I don't see this as breaking, since they were never run previously, and moreover _shouldn't_ have been run by a manual `npm version` inside the leaf package's directory.

src/commands/PublishCommand.js
```diff
@@ -536,12 +536,26 @@ export default class PublishCommand extends Command {
       this.updatePackageDepsObject(pkg, "devDependencies", exact);
       this.updatePackageDepsObject(pkg, "peerDependencies", exact);
 
+      // exec preversion script
+      pkg.runScript("preversion", (err) => {
```
gustaff-weldon: This will run script asynchronously, wouldn't it? Same goes for other hooks that were added.
Are we fine with that?

evocateur: That’s true. I forgot that almost everything in publish is synchronous, unlike bootstrap where parallelism is important.

alex-pex:
If I read correctly, hooks are called in handled package, not on the root package.

Can we also call hooks on the rootpackage? I'd like to call a hook on postversion, when everything is done (I'd like to create a merge commit between two branches).

If we can't, I'm stuck with a wrapper command
```
  "scripts": {
    "publish-all": "lerna publish && ./hooks/postversion.js"
  }
```

PR #1030, opened 2017-09-27
Add runScriptSync for use in version lifecycle methods

cwmoo740:
follow up for https://github.com/lerna/lerna/pull/1029 and https://github.com/lerna/lerna/issues/774

Makes npm preversion/version/postversion scripts synchronous

## Description

Add Package.runScriptSync, NpmUtilities.runScriptInDirSync
Converted Publish.js to use these to run lifecycle scripts at appropriate times

## Motivation and Context

The prior pr ran the scripts async, which is slightly confusing when a preversion/version script performs side effects like `git add abc`.

## How Has This Been Tested?

Added tests for all three edited files.

## Types of changes

## Checklist:

It would be nice to run all the preversion/version/postversion scripts in parallel instead of sequentially, but still block on them. I thought about doing this but there were significant code changes needed that would add promises everywhere. If that's preferred I can do that too.

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -156,6 +156,13 @@ export default class NpmUtilities {
     ChildProcessUtilities.exec("npm", ["run", script, ...args], opts, callback);
   }
 
+  static runScriptInDirSync(script, args, directory, callback) {
+    log.silly("runScriptInDirSync", script, args, path.basename(directory));
+
+    const opts = NpmUtilities.getExecOpts(directory);
+    ChildProcessUtilities.execSync("npm", ["run", script, ...args], opts, callback);
+  }
+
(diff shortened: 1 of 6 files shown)


evocateur:
I think we would want to use the concurrent batched async loop for this, rather than sync executions? It's a sticky part of the codebase, to be sure, but matching what the other lifecycle executions do should be sufficient?

gustaff-weldon:
https://github.com/lerna/lerna/pull/1029#pullrequestreview-65783844

@evocateur I believe in this case the whole point of being able to add eg. custom changelog is to wait for `preversion/version/postversion` scripts to be done before proceeding with anything else.

evocateur:
Indeed, I totally spaced that the publish command does not generally worry about parallelism, unlike bootstrap, for example.

PR #1033, opened 2017-09-29
Add support for git-hosted urls as sibling package dependencies

gustaff-weldon:
## Description

This PR introduces optional support for package dependency version to be stored as url to [git hosted](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) repositories in `package.json` eg. given `package-1` that depends on `package-2`, a `package.json` of the former might look like:

```
// packages/package-1/package.json
{
  version: "0.0.1",
  ...
  dependencies: {
    "package-2": "git://github.com/user/package-2.git#v0.0.1"
  }
}
```

- implementation is based around the idea that `package.json` file contains a `serialized` version of package dependency versions, as full urls, that are read internally as semver ones ie. eg. `0.0.1`.
- uses `useGitVersion` and `gitVersionPrefix` options in `lerna.json` to enable the feature and optionally configure it
  ```
  {
    "lerna": "2.2.0",
    "packages": [
      "packages/*"
    ],
    "version": "0.0.1",
    "useGitVersion": true,
    "gitVersionPrefix": "v"
  } 
  ```
- introduces `VersionSerializer` that handles deserializing/serializing dependency version from/to `package.json` and `GitVersionParser` to handle git-specific urls.

## Motivation and Context
Lerna assumes packages to be distributed on npm upon publish. However, when private repo is not an option, packages can be distributed on private GH repositories.

We run a private Github Lerna monorepo with Ember addons and [we publish our packages to separate read-only repositories](https://www.dotconferences.com/2016/05/fabien-potencier-monolithic-repositories-vs-many-repositories), from which addons are installable separately (and can bring in other subpackages as dependencies).

With versions written down as `0.0.1` installation of git-hosted downstream repositories would fail when resolving other package dependencies. Storing urls in GH forms solves the issues, also keeps history and tags to point to an identical version of `package.json` files in monorepo and read-only repositories.

This PR shares the work we've done with @synaptiko to streamline this process. We've done our best to make it least invasive. We are open to improvement suggestions to make it even more suitable for inclusion in Lerna. 

## Known limitations:
- feature requires dependency url to contain [`committish`](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) part 
- `publish` command should be used with `--exact` since using `^` or `~` in commitish part is not supported

## How Has This Been Tested?
We run code from this PR internally to bootstrap and publish our monorepo. We have also updated Lerna tests to cover added code.

## Types of changes

## Checklist:
- [x] All new and existing tests passed. (we checked with `run ci`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,8 +663,6 @@ Running `lerna` without arguments will show all commands/options.
     }
   },
   "packages": ["packages/*"]
-  "useGitVersion": true,
-  "gitVersionPrefix": "v"
 }
 ```
 
@@ -674,24 +672,6 @@ Running `lerna` without arguments will show all commands/options.
 - `commands.bootstrap.ignore`: an array of globs that won't be bootstrapped when running the `lerna bootstrap` command.
 - `commands.bootstrap.scope`: an array of globs that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
-- `useGitVersion`: a boolean (defaults to `false`) indicating if [git hosted urls](https://github.com/npm/hosted-git-info) should be allowed instead of plain version number. If enabled, Lerna will attempt to extract and save the interpackage dependency versions using git url-aware parser.
-    This allows packages to be distributed via git repos if eg. packages are private and [private npm repo is not an option](https://www.dotconferences.com/2016/05/fabien-potencier-monolithic-repositories-vs-many-repositories).
-    Please note that using `gitVersion` requires `publish` command to be used with `--exact` and is limited to urls with [`committish`](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) part present.
-
-    Example assuming 2 packages where `my-package-1` depends on `my-package-2`, for which `package.json` of `my-package-1` could be:
-  ```
-  {
-    name: "my-package-1",
-    version: "1.0.0",
-    bin: "bin.js",
-    dependencies: { "my-package-2": "github:example-user/my-package-2#1.0.0" },
-    devDependencies: { "my-dev-dependency": "^1.0.0" },
-    peerDependencies: { "my-peer-dependency": "^1.0.0" }
-  }
-  ```
-- `gitVersionPrefix`: version prefix string (defaults to 'v') ignored when extracting version number from commitish part of git url. eg. given `github:example-user/my-package-2#v1.0.0`
- and `gitVersionPrefix: 'v'` version will be read as `1.0.0`. Ignored if `useGitVersion` is set to `false`.
-
 
 ### Common `devDependencies`
 
@@ -942,6 +922,63 @@ The root-level package.json must also include a `workspaces` array:
 ```
 This list is broadly similar to lerna's `packages` config (a list of globs matching directories with a package.json),
 except it does not support recursive globs (`"**"`, a.k.a. "globstars").
+
+#### --use-git-version
+
+Allow target versions of dependent packages to be written as [git hosted urls](https://github.com/npm/hosted-git-info) instead of a plain version number.
+If enabled, Lerna will attempt to extract and save the interpackage dependency versions from `package.json` files using git url-aware parser.
+
+Eg. assuming monorepo with 2 packages where `my-package-1` depends on `my-package-2`, `package.json` of `my-package-1` could be:
+```
+// packages/my-package-1/package.json
+{
+  name: "my-package-1",
+  version: "1.0.0",
+  bin: "bin.js",
+  dependencies: {
+    "my-package-2": "github:example-user/my-package-2#v1.0.0"
+  },
+  devDependencies: {
+    "my-dev-dependency": "^1.0.0"
+  },
+  peerDependencies: {
+    "my-peer-dependency": "^1.0.0"
+  }
+}
+```
+For the case above Lerna will read the version of `my-package-2` dependency as `1.0.0`.
+
+This allows packages to be distributed via git repos if eg. packages are private and [private npm repo is not an option](https://www.dotconferences.com/2016/05/fabien-potencier-monolithic-repositories-vs-many-repositories).
+
+Please note that using `--use-git-version`
+- is limited to urls with [`committish`](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) part present (ie. `github:example-user/my-package-2` is invalid)
+- requires `publish` command to be used with `--exact`
+
+May also be configured in `lerna.json`:
+```js
+{
+  ...
+  "useGitVersion": true
+}
+```
+
+#### --git-version-prefix
+
+Defines version prefix string (defaults to 'v') ignored when extracting version number from a commitish part of git url.
+Everything after the prefix will be considered a version.
+
+
+Eg. given `github:example-user/my-package-2#v1.0.0` and `gitVersionPrefix: 'v'` version will be read as `1.0.0`.
+
+Only used if `--use-git-version` is set to `true`.
+
+May also be configured in `lerna.json`:
+```js
+{
+  ...
+  "gitVersionPrefix": "v"
+}
+```
 
 #### --stream
 
--- a/README.md
+++ b/README.md
@@ -663,6 +663,8 @@ Running `lerna` without arguments will show all commands/options.
     }
   },
   "packages": ["packages/*"]
+  "useGitVersion": true,
+  "gitVersionPrefix": "v"
 }
 ```
 
@@ -672,6 +674,23 @@ Running `lerna` without arguments will show all commands/options.
 - `commands.bootstrap.ignore`: an array of globs that won't be bootstrapped when running the `lerna bootstrap` command.
 - `commands.bootstrap.scope`: an array of globs that restricts which packages will be bootstrapped when running the `lerna bootstrap` command.
 - `packages`: Array of globs to use as package locations.
+- `useGitVersion`: a boolean (defaults to `false`) indicating if [git hosted urls](https://github.com/npm/hosted-git-info) should be allowed instead of plain version number. If enabled, Lerna will attempt to extract and save the interpackage dependency versions using git url-aware parser.
+    This allows packages to be distributed via git repos if eg. packages are private and [private npm repo is not an option](https://www.dotconferences.com/2016/05/fabien-potencier-monolithic-repositories-vs-many-repositories).
+    Please note that using `gitVersion` requires `publish` command to be used with `--exact` and is limited to urls with [`committish`](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) part present.
+
+    Example assuming 2 packages where `my-package-1` depends on `my-package-2`, for which `package.json` of `my-package-1` could be:
+  ```
+  {
+    name: "my-package-1",
+    version: "1.0.0",
+    bin: "bin.js",
+    dependencies: { "my-package-2": "github:example-user/my-package-2#1.0.0" },
+    devDependencies: { "my-dev-dependency": "^1.0.0" },
+    peerDependencies: { "my-peer-dependency": "^1.0.0" }
+  }
+  ```
+- `gitVersionPrefix`: version prefix string (defaults to 'v') ignored when extracting version number from commitish part of git url. eg. given `github:example-user/my-package-2#v1.0.0`
+ and `gitVersionPrefix: 'v'` version will be read as `1.0.0`. Ignored if `useGitVersion` is set to `false`.
 
 
 ### Common `devDependencies`
@@ -896,9 +915,9 @@ May also be configured in `lerna.json`:
 
 #### --use-workspaces
 
-Enables integration with [Yarn Workspaces](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-workspaces-install-phase-1.md) (available since yarn@0.27+).  
-The values in the array are the commands in which Lerna will delegate operation to Yarn (currently only bootstrapping).    
-If `--use-workspaces` is true then `packages` will be overridden by the value from `package.json/workspaces.`  
+Enables integration with [Yarn Workspaces](https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-workspaces-install-phase-1.md) (available since yarn@0.27+).
+The values in the array are the commands in which Lerna will delegate operation to Yarn (currently only bootstrapping).
+If `--use-workspaces` is true then `packages` will be overridden by the value from `package.json/workspaces.`
 May also be configured in `lerna.json`:
 
 ```js
--- a/src/Package.js
+++ b/src/Package.js
@@ -62,13 +62,7 @@ export default class Package {
   }
 
   set versionSerializer(versionSerializer) {
-    const previousSerializer = this._versionSerializer;
-
     this._versionSerializer = versionSerializer;
-
-    if (previousSerializer) {
-      this._package = previousSerializer.serialize(this._package);
-    }
 
     if (versionSerializer) {
       this._package = versionSerializer.deserialize(this._package);
--- a/src/Package.js
+++ b/src/Package.js
@@ -2,6 +2,7 @@ import dedent from "dedent";
 import log from "npmlog";
 import path from "path";
 import semver from "semver";
+import _ from "lodash";
 
 import dependencyIsSatisfied from "./utils/dependencyIsSatisfied";
 import NpmUtilities from "./NpmUtilities";
@@ -60,12 +61,27 @@ export default class Package {
     return this._package.scripts || {};
   }
 
+  set versionSerializer(versionSerializer) {
+    const previousSerializer = this._versionSerializer;
+
+    this._versionSerializer = versionSerializer;
+
+    if (previousSerializer) {
+      this._package = previousSerializer.serialize(this._package);
+    }
+
+    if (versionSerializer) {
+      this._package = versionSerializer.deserialize(this._package);
+    }
+  }
+
   isPrivate() {
     return !!this._package.private;
   }
 
   toJSON() {
-    return this._package;
+    const pkg = _.cloneDeep(this._package);
+    return this._versionSerializer ? this._versionSerializer.serialize(pkg) : pkg;
   }
 
   /**
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -8,9 +8,6 @@ import semver from "semver";
 
 import dependencyIsSatisfied from "./utils/dependencyIsSatisfied";
 import Package from "./Package";
-import PackageUtilities from "./PackageUtilities";
-import GitVersionParser from "./GitVersionParser";
-import VersionSerializer from "./VersionSerializer";
 
 const DEFAULT_PACKAGE_GLOB = "packages/*";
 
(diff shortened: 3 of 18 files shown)


test/GitVersionParser.js
```diff
@@ -0,0 +1,79 @@
...
+      });
+
+      expect(parser.parseVersion("git@github.com:user-foo/project-foo#0.0.5")).toEqual({
+        prefix: "git@github.com:user-foo/project-foo#",
+        version: "0.0.5"
+      });
+    });
+  });
+
+  describe("parseVersion - with prefix", () => {
```
Calanthe: [minor]: I would rename `with prefix` into `with versionPrefix`. The title confused me with the `prefix` property returned by the parser.

test/Package.js
```diff
@@ -105,15 +105,127 @@ describe("Package", () => {
...
+      };
+
+      pkg.versionSerializer = mockSerializer;
+
+      expect(mockSerializer.deserialize.mock.calls.length).toBe(1);
+      expect(mockSerializer.deserialize.mock.calls[0][0]).toBe(pkg._package);
+      expect(mockSerializer.serialize.mock.calls.length).toBe(0);
+    });
+
+    it("should call 'deserialize' on old and and 'serialize' on new serializer'", () => {
```
Calanthe: [minor] double `and`

test/Package.js
```diff
@@ -105,15 +105,127 @@ describe("Package", () => {
...
+          return pkg;
+        }),
+        deserialize: jest.fn((pkg) => {
+          return pkg;
+        })
+      };
+
+      pkg.versionSerializer = mockSerializer;
+
+      expect(mockSerializer.deserialize.mock.calls.length).toBe(1);
```
Calanthe: Either test or just it's description is wrong. You call `deserialize` method, not `serialize` like the title says.

test/Package.js
```diff
@@ -105,15 +105,127 @@ describe("Package", () => {
...
+      expect(mockSerializer.deserialize.mock.calls[0][0]).toBe(pkg._package);
+      expect(mockSerializer.serialize.mock.calls.length).toBe(0);
```
Calanthe: If I understand it correctly, `serialize` is not called at all.

test/PackageGraph.js
```diff
@@ -0,0 +1,76 @@
...
+  describe("constructor", () => {
+    it(".get should return dependencies", () => {
+      const [ pkg1, pkg2 ] = createPackages("0.0.1");
+      const graph = new PackageGraph([ pkg1, pkg2 ]);
+
+      expect(graph.get(pkg1.name).dependencies).toEqual([]);
+      expect(graph.get(pkg2.name).dependencies).toEqual([pkg1.name]);
+    });
+
+    it(".get should not return the dependencies for urecognized versions", () => {
```
Calanthe: [minor] `unrecognised`

test/PackageGraph.js
```diff
@@ -0,0 +1,76 @@
...
+      const [pkg1, pkg2] = createPackages("0.0.1", "github:user-foo/project-foo#v0.0.1");
```
Calanthe: To keep consistency, add spaces before and after `[]` like in the test before.

test/PackageGraph.js
```diff
@@ -0,0 +1,76 @@
...
+      const graph = new PackageGraph([pkg1, pkg2]);
+
+      expect(graph.get(pkg1.name).dependencies).toEqual([]);
+      expect(graph.get(pkg2.name).dependencies).toEqual([]);
+    });
+
+    it(".get should not return the dependencies for urecognized versions", () => {
```
Calanthe: [minor] `unrecognized` and spaces before and after `[]`

test/PackageGraph.js
Calanthe: Also, I am not sure if you need this test if you have the second `.get should not return the dependencies for urecognized versions` more detailed test.

test/VersionSerializer.js
```diff
@@ -0,0 +1,151 @@
...
+        dependencies: { "my-dependency": "^1.0.0" },
+        devDependencies: { "my-package-2": "^1.0.0" },
+        peerDependencies: { "my-package-3": "^1.0.0" }
+      };
+
+      expect(serializer.serialize(pkg)).toEqual(pkg);
+    });
+
+    it("should write back version strings transformed by deserialize", () => {
+        expect(serializer.deserialize({
```
evocateur: This looks like a copy+paste error? Not sure why `deserialize` is being tested inside `describe("serialize")` block.

src/Package.js
```diff
@@ -60,12 +61,27 @@ export default class Package {
     return this._package.scripts || {};
   }
 
+  set versionSerializer(versionSerializer) {
+    const previousSerializer = this._versionSerializer;
+
+    this._versionSerializer = versionSerializer;
+
+    if (previousSerializer) {
```
evocateur: When does this ever happen? Just being defensive?

(It seems strange to assign the versionSerializer multiple times)

src/Repository.js
```diff
@@ -118,8 +120,22 @@ export default class Repository {
   }
 
   buildPackageGraph() {
-    this._packages = PackageUtilities.getPackages(this);
-    this._packageGraph = PackageUtilities.getPackageGraph(this._packages);
+    const packages = PackageUtilities.getPackages(this);
+    const packageGraph = PackageUtilities.getPackageGraph(packages, false, this.lernaJson.useGitVersion
+      ? new GitVersionParser(this.lernaJson.gitVersionPrefix)
```
evocateur: I'd prefer the version parser be instantiated outside of the `getPackageGraph()` arguments, this ternary initially made me think you had forgotten the closing paren. Also, the version parser could be reused in the subsequent version serializer loop.

```js
const { useGitVersion, gitVersionPrefix } = this.lernaJson;
// FIXME: should be destructured from command.options to support CLI flag overrides

const versionParser = useGitVersion && new GitVersionParser(gitVersionPrefix);
const packages = PackageUtilities.getPackages(this);
const packageGraph = PackageUtilities.getPackageGraph(packages, false, versionParser);

if (useGitVersion) {
  packages.forEach((pkg) => {
    pkg.versionSerializer = new VersionSerializer({
      monorepoDependencies: packageGraph.get(pkg.name).dependencies,
      versionParser,
    });
  });
}
```

src/Repository.js
```diff
@@ -118,8 +120,22 @@ export default class Repository {
...
+      : null);
+
+    if (this.lernaJson.useGitVersion) {
+      packages.forEach((pkg) => {
+        pkg.versionSerializer = new VersionSerializer({
+          monorepoDependencies: packageGraph.get(pkg.name).dependencies,
```
evocateur: `graphDependencies` sounds clearer to me, here. It's not the dependencies of the entire monorepo, after all.

src/Package.js
```diff
@@ -60,12 +61,27 @@ export default class Package {
...
+    }
+  }
+
   isPrivate() {
     return !!this._package.private;
   }
 
   toJSON() {
-    return this._package;
+    const pkg = _.cloneDeep(this._package);
```
evocateur: Can we move this `cloneDeep` into the version serializer methods? It seems overly defensive for the non-serialized case.

src/Repository.js
```diff
@@ -118,8 +120,22 @@ export default class Repository {
   }
 
   buildPackageGraph() {
```
evocateur: We should pass an options object derived from `this.options` when it is called from [`Command.js`](https://github.com/lerna/lerna/blob/5ea7e2d9a216bb2274e217401eb47db8661f6068/src/Command.js#L306) and use those properties instead of `this.lernaJson`. This allows us to support CLI flags (`--use-git-version`, `--git-version-prefix`) and eventually argument validation with yargs.

```diff
diff --git a/src/Command.js b/src/Command.js
index e5bb0c4..032cd01 100644
--- a/src/Command.js
+++ b/src/Command.js
@@ -288,7 +288,7 @@ export default class Command {
   }
 
   runPreparations() {
-    const { scope, ignore, registry, since } = this.options;
+    const { scope, ignore, registry, since, useGitVersion, gitVersionPrefix } = this.options;
 
     if (scope) {
       log.info("scope", scope);
@@ -303,7 +303,7 @@ export default class Command {
     }
 
     try {
-      this.repository.buildPackageGraph();
+      this.repository.buildPackageGraph({ useGitVersion, gitVersionPrefix });
       this.packages = this.repository.packages;
       this.packageGraph = this.repository.packageGraph;
       this.filteredPackages = PackageUtilities.filterPackages(this.packages, { scope, ignore });
```
(Sorry for the scattered review)

README.md
```diff
@@ -672,6 +674,23 @@ Running `lerna` without arguments will show all commands/options.
...
+    This allows packages to be distributed via git repos if eg. packages are private and [private npm repo is not an option](https://www.dotconferences.com/2016/05/fabien-potencier-monolithic-repositories-vs-many-repositories).
+    Please note that using `gitVersion` requires `publish` command to be used with `--exact` and is limited to urls with [`committish`](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies) part present.
+
+    Example assuming 2 packages where `my-package-1` depends on `my-package-2`, for which `package.json` of `my-package-1` could be:
+  ```
+  {
+    name: "my-package-1",
+    version: "1.0.0",
+    bin: "bin.js",
+    dependencies: { "my-package-2": "github:example-user/my-package-2#1.0.0" },
```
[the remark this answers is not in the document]
evocateur: (the same stuff in various tests is fine on one line)

gustaff-weldon:
@evocateur thanks for a detailed review, much appreciated! I will get back to you once the changes are done.

src/Package.js
gustaff-weldon: Yes, it's just being defensive. Currently, it is not likely that second serializer would be used in one run. YAGNI and remove?

src/Package.js
gustaff-weldon: We can, there's a couple of reasons why this is done like this:
- `toJSON` was returning its internal `pkg` object, any outside changes on it will affect Package
- cloning only for serializer case would make `toJSON` behave differently - in one case it will return its internal object the other time it would return a deep copy

src/Repository.js
gustaff-weldon: Any suggestion what would be the best way to deal with `buildPackageGraph` being called from
[`packages` and `packageGraph` getters](https://github.com/lerna/lerna/blob/master/src/Repository.js#L70-L81)?

For now I have removed those implicit calls, since otherwise I think we would need to set current command options on the repository and then pass them:
```
  get packageGraph() {
    if (!this._packageGraph) {
      this.buildPackageGraph(this.options); // this would have to be set eg. from command which looks a bit messy...
    }

    return this._packageGraph;
  }
```

Unless there's an acces to currently running command from Repository?

src/Package.js
evocateur: Yes, please.

src/Package.js
evocateur: Despite `.toJSON()` _currently_ being used only in one place, I think you've raised some good points that will ultimately help future use cases (such as when I actually get around to refactoring lerna into a lerna-managed monorepo and the Package class is exposed in a package of its own). I was being overly-cautious about performance in a place where it really, really doesn't matter (package.json files aren't particularly large, and not deeply nested at all).

Thanks for the clear explanation. :)

src/Repository.js
evocateur: Yeah, I totally understand the mixed feelings. I have them myself, about the Repository class in general. It's doing too many disparate things for my taste, but it's been low on the priority list to refactor...

So: It occurs to me that Repository itself _shouldn't_ be storing `packages` and `packageGraph` in getters since there are literally two places in the entire codebase that they are referenced:
1. in Command, right after the `buildPackageGraph()` is called
2. in [UpdatedPackagesCollector](https://github.com/lerna/lerna/blob/b350871597d46da602eb54733bd398ff0c9b8d07/src/UpdatedPackagesCollector.js#L40), where it violates separation of concern by reaching into (sub-)properties of the command instance just to retrieve mostly-static (i.e., unchanging) values.

Here's what we can do: Move the creation of both the `packages` list and the `packageGraph` into Command, where it belongs, and completely remove the `get packages` and `get packageGraph` from Repository.

This means all the additional logic added here will be contained inside the place where `this.options` already exists. Something like this:
```diff
diff --git a/src/Command.js b/src/Command.js
index 58659bb..aa4bc7f 100644
--- a/src/Command.js
+++ b/src/Command.js
@@ -294,7 +294,8 @@ export default class Command {
   }
 
   runPreparations() {
-    const { scope, ignore, registry, since } = this.options;
+    const { rootPath, packageConfigs } = this.repository;
+    const { scope, ignore, registry, since, useGitVersion, gitVersionPrefix } = this.options;
 
     if (scope) {
       log.info("scope", scope);
@@ -309,10 +310,22 @@ export default class Command {
     }
 
     try {
-      this.repository.buildPackageGraph();
-      this.packages = this.repository.packages;
-      this.packageGraph = this.repository.packageGraph;
-      this.filteredPackages = PackageUtilities.filterPackages(this.packages, { scope, ignore });
+      const versionParser = useGitVersion && new GitVersionParser(gitVersionPrefix);
+      const packages = PackageUtilities.getPackages({ rootPath, packageConfigs });
+      const packageGraph = PackageUtilities.getPackageGraph(packages, false, versionParser);
+
+      if (useGitVersion) {
+        packages.forEach((pkg) => {
+          pkg.versionSerializer = new VersionSerializer({
+            graphDependencies: packageGraph.get(pkg.name).dependencies,
+            versionParser,
+          });
+        });
+      }
+
+      this.packages = packages;
+      this.packageGraph = packageGraph;
+      this.filteredPackages = PackageUtilities.filterPackages(packages, { scope, ignore });
 
       // The UpdatedPackagesCollector requires that filteredPackages be present prior to checking for
       // updates. That's okay because it further filters based on what's already been filtered.
diff --git a/src/Repository.js b/src/Repository.js
index 8316083..7f1e138 100644
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -8,7 +8,6 @@ import semver from "semver";
 
 import dependencyIsSatisfied from "./utils/dependencyIsSatisfied";
 import Package from "./Package";
-import PackageUtilities from "./PackageUtilities";
 
 const DEFAULT_PACKAGE_GLOB = "packages/*";
 
@@ -67,22 +66,6 @@ export default class Repository {
       .map(parentDir => path.resolve(this.rootPath, parentDir));
   }
 
-  get packages() {
-    if (!this._packages) {
-      this.buildPackageGraph();
-    }
-
-    return this._packages;
-  }
-
-  get packageGraph() {
-    if (!this._packageGraph) {
-      this.buildPackageGraph();
-    }
-
-    return this._packageGraph;
-  }
-
   get packageJson() {
     if (!this._packageJson) {
       try {
@@ -117,11 +100,6 @@ export default class Repository {
     return this.version === "independent";
   }
 
-  buildPackageGraph() {
-    this._packages = PackageUtilities.getPackages(this);
-    this._packageGraph = PackageUtilities.getPackageGraph(this._packages);
-  }
-
   hasDependencyInstalled(depName, version) {
     log.silly("hasDependencyInstalled", "ROOT", depName, version);
 
diff --git a/src/UpdatedPackagesCollector.js b/src/UpdatedPackagesCollector.js
index 67d8ba5..5e51b60 100644
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -31,7 +31,7 @@ export default class UpdatedPackagesCollector {
     this.logger = command.logger;
     this.repository = command.repository;
     this.packages = command.filteredPackages;
-    this.packageGraph = command.repository.packageGraph;
+    this.packageGraph = command.packageGraph;
     this.options = command.options;
   }
 
diff --git a/test/Repository.js b/test/Repository.js
index 1c21bfc..8b01475 100644
--- a/test/Repository.js
+++ b/test/Repository.js
@@ -142,32 +142,6 @@ describe("Repository", () => {
     });
   });
 
-  describe("get .packages", () => {
-    it("returns the list of packages", () => {
-      const repo = new Repository(testDir);
-      expect(repo.packages).toEqual([]);
-    });
-
-    it("caches the initial value", () => {
-      const repo = new Repository(testDir);
-      expect(repo.packages).toBe(repo.packages);
-    });
-  });
-
-  describe("get .packageGraph", () => {
-    it("returns the graph of packages", () => {
-      const repo = new Repository(testDir);
-      expect(repo.packageGraph).toBeDefined();
-      expect(repo.packageGraph).toHaveProperty("nodes", []);
-      expect(repo.packageGraph).toHaveProperty("nodesByName", {});
-    });
-
-    it("caches the initial value", () => {
-      const repo = new Repository(testDir);
-      expect(repo.packageGraph).toBe(repo.packageGraph);
-    });
-  });
-
   describe("get .packageJson", () => {
     afterEach(() => {
       readPkg.sync = readPkgSync;
 ```

evocateur:
Golly, I really need to fix that AppVeyor node v4 build. :/

Looking great so far!

gustaff-weldon:
I have updated the code. Please let me know if something can be further improved. If accepted, I will rebase this PR into one commit.

evocateur:
@gustaff-weldon Superb, thank you! Don't worry about rebasing, I will squash merge.

mattbrunetti:
Awesome @gustaff-weldon! I can't wait to try this out!

gustaff-weldon:
@mattbrunetti kudos also to @synaptiko, we worked on that together. 
I'm glad you find this useful!

ramasilveyra:
This is awesome! Thanks @gustaff-weldon and @synaptiko!

jamiehill:
Thanks for the heads up @ramasilveyra 

This is fantastic news indeed!

mattbrunetti:
Did not end up using this :( Could not find resources/support on the workflow of this "git-only" way of distributing packages.. Decided to try out npm private packages instead..

synaptiko:
For the future reference, there are multiple ways how to achieve "git-only" distribution from monorepo.

We've tried both of those:

1. [splitsh](https://github.com/splitsh/lite)
2. regular `git filter-branch --subdirectory-filter packages/<package>`

Both are quite fast, splitsh is faster (we have less then 200 commits though) and both require to have prepared "downstream" repositories in advance. There is one bummer with `filter-branch` related to tags. You have to either restore original tags after the push of the "split". Or you have to create a fresh clone of the repository before split.

@mattbrunetti It was easier for to setup this than to manage private npm repository (which requires every developer to setup his npm/yarn accordingly).

chinchang:
@synaptiko I have one major confusion regarding all this. Please help me clear this:

"This PR only allows putting versions as git urls. That's all. Pushing individual packages to downstream git repos has to be still handled on our own (using splitish etc)."

Am I correct on this?

synaptiko:
@chinchang yes, you are correct… I understand why it's confusing but unfortunately it would make lerna even more complex to support it directly… it also depends on what CI you use and so on

JacopKane:
As a side tip, I'm experimenting with this tool with git hooks/husky:
https://github.com/ingydotnet/git-subrepo

Seems like doing the job well without the need to keep repositories read-only. The only gotcha was I needed to filter my git hooks for the only root origin in order to get rid of endless push/pulls

PR #1044, opened 2017-10-03
Throw publish validation errors before version selection

evocateur:
## Description
When attempting to publish in a branch restricted by `--allow-branch`, the error should be thrown (and the publish process halted) _before_ the version selection prompt is rendered.

Also, `--canary` releases should _always_ be allowed, even if the branch isn't whitelisted by `--allow-branch`. Effectively, we should consider `--allow-branch` to describe "the branch (or branches) allowed to publish non-canary versions on".

## Motivation and Context
It makes the error cases much less confusing, since you don't spend a lot of effort choosing versions and whatnot before the validation error occurs.

## How Has This Been Tested?
added unit tests, local testing

## Types of changes

## Checklist:

### DIFF
--- a/src/Command.js
+++ b/src/Command.js
@@ -74,7 +74,7 @@ export const builder = {
   }
 };
 
-class ValidationError extends Error {
+export class ValidationError extends Error {
   constructor(prefix, message) {
     super(message);
     this.name = "ValidationError";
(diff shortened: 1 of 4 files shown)


evocateur:
This also modifies the error message when `--allow-branch` is violated to be less "here's how you override this config that was set for a reason".

Issue #1058, opened 2017-10-08
"import" throws with various projects


swernerx:
I am trying to use lerna for the [Edge Platform](https://github.com/sebastian-software/edge). Right now I am in the phase of importing some old projects. Unfortunately, I have reproducible import issues which always seem to stop at specific commits. I am unaware of any specifics of these commits. The error message from Lerna is also not very helpful.

## Expected Behavior

Importing should work on all packages.

## Current Behavior

Currently Lerna throws during import with hard errors:

```
lerna info 1ebef2c 
lerna info b1075d6 
lerna info 3c0a15e 
lerna ERR! execute callback with error
(node:58112) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): TypeError: Cannot read property 'split' of undefined
(node:58112) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

## Possible Solution

Check where the "split" call happens and why there is a possible `null` value inside the code.

## Steps to Reproduce (for bugs)

All the code is open source so it should be possible for others to reproduce:

- New Multi Package Repository: https://github.com/sebastian-software/edge
- Packages which fail:
  - https://github.com/sebastian-software/prepublish (at 3c0a15e)
  - https://github.com/sebastian-software/readable-code (at 3998aa0)

This happens also when using a freshly checked-out repository. So it is probably nothing about local changes.

Vers

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.4.0",
  "packages": [
    "packages/*"
  ],
  "version": "independent",
  "npmClient": "yarn",
  "useWorkspaces": true
}
```
</p></details>

<details><summary>lerna-debug.log</summary><p>

```txt
<!-- If you have a `lerna-debug.log` available, please paste it here -->
<!-- Otherwise, feel free to delete this <details> block -->
```
</p></details>

## Context

Starting to use lerna.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.4.0 |
| `npm --version`  | 5.5.1 |
| `yarn --version` | 1.1.0 |
| `node --version` | 8.6.0 |
| `git --version` | 2.14.2 |

| OS | Version |
| --- | --- |
| macOS Sierra | 10.12.6 |

swernerx:
[...]

PR #1059, opened 2017-10-08
Improve "import" edgecases + (i18n fixes + git tweaks)

swernerx:
## Description

Originally this just contains the fixes for handling issues detecting in https://github.com/lerna/lerna/issues/1058

- Outputting added for errors when something unintended happens during importing of projects.
- Added new handling to automatically skip empty commits. These are an anti-pattern I figure, but I saw them accidentally in some of my own repositories. As `git` does not seem to have a useful parameter for its `am` command we have to `skip` the command when detecting the empty commits error (my approach analyzes the error message from `git`).
- Disabling GPG signing of all "git commit" runs as this seems to break the system. Lerna should not rely on the user to disabling this behavior manually.
- We have some tests which compare the exact output of some commands. As these commands might be executed under the local env variables of the user they might differ from our tested ones. I added `cross-env` and added a `LC_ALL` setting in front of the `npm` script to run `jest` to overcome this.

## Motivation and Context

This fixes to import my projects from Edge Platform into a single repository. 

Without the change on skipping empty commits it just failed. Without the logging change, it failed without any useful user-facing error.

## How Has This Been Tested?

The test suite and linting pass without any errors.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "preintegration": "npm pack",
     "integration": "jest --config test/config/integration.json",
     "pretest": "npm run lint -- --cache",
-    "test": "jest",
+    "test": "cross-env LC_ALL=en-US jest",
     "ci": "npm test -- --coverage --verbose && npm run integration",
     "prepublish": "npm run build"
   },
@@ -76,6 +76,7 @@
     "babel-eslint": "^7.2.3",
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-preset-env": "^1.6.0",
+    "cross-env": "^5.0.5",
     "eslint": "^4.5.0",
     "eslint-config-babel": "^7.0.2",
     "eslint-plugin-babel": "^4.1.2",
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -42,7 +42,7 @@ export default class GitUtilities {
 
   static commit(message, opts) {
     log.silly("commit", message);
-    const args = ["commit"];
+    const args = ["commit", "--no-gpg-sign"];
 
     if (message.indexOf(EOL) > -1) {
       // Use tempfile to allow multi\nline strings.
--- a/src/commands/ImportCommand.js
+++ b/src/commands/ImportCommand.js
@@ -172,14 +172,23 @@ export default class ImportCommand extends Command {
       // due to merge history.
       ChildProcessUtilities.exec("git", ["am", "-3"], this.execOpts, (err) => {
         if (err) {
-          // Give some context for the error message.
-          err = `Failed to apply commit ${sha}.\n${err}\n` +
-                `Rolling back to previous HEAD (commit ${this.preImportHead}).\n` +
-                `You may try with --flatten to import flat history.`;
+          const isEmptyCommit = err.stdout.indexOf("Patch is empty.") === 0;
+          if (isEmptyCommit) {
+            // Automatically skip empty commits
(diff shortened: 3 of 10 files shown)


natesilva:
Thanks @swernerx. I’m testing Lerna, importing a big legacy project, and your branch was a lifesaver. I hope this gets merged ASAP.

package.json
```diff
@@ -15,7 +15,7 @@
     "preintegration": "npm pack",
     "integration": "jest --config test/config/integration.json",
     "pretest": "npm run lint -- --cache",
-    "test": "jest",
+    "test": "cross-env LC_ALL=en-US jest",
```
evocateur: Thanks for this! I was hoping to get away from the brittle localized snapshots eventually, but this is a good addition regardless.

src/commands/ImportCommand.js
```diff
@@ -191,7 +200,10 @@ export default class ImportCommand extends Command {
 
       if (!err) {
         this.logger.success("import", "finished");
+      } else {
+        this.logger.error("import", err);
```
evocateur: I can't believe we forgot to log the error 😅

src/GitUtilities.js
```diff
@@ -42,7 +42,7 @@ export default class GitUtilities {
 
   static commit(message, opts) {
     log.silly("commit", message);
-    const args = ["commit"];
+    const args = ["commit", "--no-gpg-sign"];
```
evocateur: I'm a little leery of adding this flag to _all_ executions of `git commit`. I have no idea how common GPG signing of commits is, although I definitely appreciate your reasoning.

I'm okay with taking this now, but we'll have to note this very well in the changelog and be open to the possibility that we may need to revert it (or put it behind an option) in the future.

rgbkrk:
Thanks for doing this, it was super helpful to see the error I'm running into:

```
lerna info 8520709
lerna ERR! import Failed to apply commit 8520709.
lerna ERR! import Error: Command failed: git am -3
lerna ERR! import error: Failed to merge in the changes.
lerna ERR! import Applying: chore(RefactorContents): Cleanup ajax settings.
lerna ERR! import Using index info to reconstruct a base tree...
lerna ERR! import M	packages/rx-jupyter/test/contents-spec.js
lerna ERR! import Falling back to patching base and 3-way merge...
lerna ERR! import Auto-merging packages/rx-jupyter/test/contents-spec.js
lerna ERR! import CONFLICT (content): Merge conflict in packages/rx-jupyter/test/contents-spec.js
lerna ERR! import Patch failed at 0001 chore(RefactorContents): Cleanup ajax settings.
lerna ERR! import The copy of the patch that failed is found in: .git/rebase-apply/patch
lerna ERR! import When you have resolved this problem, run "git am --continue".
lerna ERR! import If you prefer to skip this patch, run "git am --skip" instead.
lerna ERR! import To restore the original branch and stop patching, run "git am --abort".
lerna ERR! import
lerna ERR! import Rolling back to previous HEAD (commit 78c3e73fc680daca40dfb47cb9fd407066413338).
lerna ERR! import You may try with --flatten to import flat history.
lerna ERR! execute callback with error
```

I'm probably just going to rebase the originating repo on a branch and then importing from a version lerna can work with.

rgbkrk:
Here's what I did to "fix" these commits:

```
git rebase -i --root
# Find commit 8520709
# Switch from `pick` to `reword`
# Also do this with the following commit
```

annnnd somehow I was able to import it after doing this 
¯\\\_(ツ)\_/¯

Issue #521, opened 2017-01-23
Lerna push to github repo


MatthD:
I am sorry but I can't reopen my previous ticket on this ...

I read docs but it appears that it does not push anything to my github monorepo when I tried it.

It clearly increment packages versions & create a commit of it but it DOES NOT push anything to my github repository (I added it in package.json too ... ) 

I tried on this repo https://github.com/MatthD/sisyphe 

Command use & result (Id does not say anything after "Yes"): 

![lerna](https://cloud.githubusercontent.com/assets/7901366/22200446/60a41dfa-e15f-11e6-9de7-c2ab9763bfdf.png)

May i forgot something ?

MatthD:
Ok, I understand what append,

https://github.com/lerna/lerna/blob/aeb775d1fec9988f9165cbfad82db2d4fb796c7b/src/commands/PublishCommand.js#L130

i was using --skip-npm (because I did not wanted to publish on npm & with this option it does not git push in fact)

gigabo:
Thanks for reporting @MatthD.

Hmm... is this a bug?  Should `--skip-npm` push the commit and release tags?

noherczeg:
I think that the `npm version` way should be used. [link](https://docs.npmjs.com/cli/version)

So if you want a release like you said you could do something like this:
```
"scripts": {
  "mypublish": "npm run lerna publish --yes --skip-npm",
  "postmypublish": "git push && git push --tags"
}
```

This way you could run `npm run mypublish`, and I assume you'd get what you wanted. If this is sufficient then the docs should be updated letting the users know about git as well.

PR #1060, opened 2017-10-09
Add --skip-git-push flag

jmherbst:
Adding a `--skip-git-push` flag 

## Description
The `--skip-git-push` flag could be used as an alternative to the `--skip-git` flag in order to have the commits + tags still created locally, just not pushed up to the `gitRemote`.

## Motivation and Context
While attempting to get `lerna publish` working in our internal CI environment we needed a way for lerna publish the build packages, along with adding the git commits+tags, but not `git push` so that it could be done at a later time in CI.

We still want to leverage lerna's ability to generate the commits and tags.

## How Has This Been Tested?
I added a test case in order to ensure the files and tags were added and a commit is generated, but the `pushWithTags` is not performed.

I also tested locally using `npm run build && npm link` and a local repository we have lerna configured on.   With the local testing I was able to verify that `--skip-git-push` works as intended and no behavior changes were made to `--skip-git`

## Types of changes

## Checklist:
  - I Believe it does: `npm run lint` does not return any errors
  - I was not entirely sure what type of test to add to the `test/integration/lerna-publish.test.js`
  - Also, I was not sure if committing the `test/__snapshots__/PublishCommand.js.snap` was correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,16 @@ When run with this flag, `publish` will publish to npm without running any of th
 
 > Only publish to npm; skip committing, tagging, and pushing git changes (this only affects publish).
 
+#### --skip-git-push
+
+```sh
+$ lerna publish --skip-git-push
+```
+
+When run with this flag, `publish` will publish to npm without pushing the commit and tags up to git remote.
+
+> Publish to npm without pushing commit and tags up to git remote (this only affects publish).
+
 #### --skip-npm
 
 ```sh
--- a/test/PublishCommand.js
+++ b/test/PublishCommand.js
@@ -387,6 +387,32 @@ describe("PublishCommand", () => {
   });
 
   /** =========================================================================
+   * NORMAL - SKIP GIT PUSH
+   * ======================================================================= */
+
+  describe("normal mode with --skip-git-push", () => {
+    let testDir;
+
+    beforeEach(() => initFixture("PublishCommand/normal").then((dir) => {
+      testDir = dir;
+    }));
+
+    it("should publish the changed packages, commit, and tag, but not push", () => {
+      return run(testDir)(
+        "--skip-git-push"
+      ).then(() => {
+        expect(GitUtilities.addFile).toBeCalled();
+        expect(GitUtilities.commit).toBeCalled();
+        expect(GitUtilities.addTag).toBeCalled();
+        expect(GitUtilities.pushWithTags).not.toBeCalled();
+
+        expect(publishedTagInDirectories(testDir))
+          .toMatchSnapshot("[normal --skip-git-push] npm publish --tag");
+      });
+    });
+  });
+
+  /** =========================================================================
    * NORMAL - SKIP NPM
    * ======================================================================= */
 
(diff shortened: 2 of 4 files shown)


jmherbst:
I'm not sure why the AppVeyor build is failing, but I'm not able to get those integration tests to run cleanly locally using `master` branch either. 🤔 

Could use some help troubleshooting that test as well as any other feedback on this PR.

Thanks!

jmherbst:
Oh, AppVeyor build is passing now, so I must have done somet

jmherbst:
Didn't intend to close this 😆
Let me know what you think and any feedback you might have! 
Thanks!

jmherbst:
Possibly tangentially related: https://github.com/lerna/lerna/issues/521 ?

test/PublishCommand.js
```diff
@@ -460,6 +486,52 @@ describe("PublishCommand", () => {
   });
 
   /** =========================================================================
+   * NORMAL - SKIP GIT PUSH AND SKIP NPM
+   * ======================================================================= */
+
+  describe("normal mode with --skip-git-push and --skip-npm", () => {
```
evocateur: Unless there is some sort of interaction between `--skip-git-push` and `--skip-npm`, I'm not sure how important this test case is. We've already established what `--skip-npm` does, after all, and `--skip-git-push` doesn't change anything about that?

test/__snapshots__/PublishCommand.js.snap
```diff
@@ -355,6 +355,37 @@ Array [
...
+Object {
+  "packages/package-1": "1.0.1",
+  "packages/package-2": "1.0.1",
+  "packages/package-3": "1.0.1",
+  "packages/package-4": "1.0.1",
+  "packages/package-5": "1.0.1",
+}
+`;
+
+exports[`[normal --skip-git-push] npm publish --tag 1`] = `
```
evocateur: When adding a `toMatchSnapshot()` in a test, it is appropriate to commit the update to the snapshot file.

However, in this case, as I mentioned in the test above, I don't think this snapshot is necessary (it's almost verbatim from another snapshot).

README.md
```diff
@@ -334,6 +334,16 @@ When run with this flag, `publish` will publish to npm without running any of th
 
 > Only publish to npm; skip committing, tagging, and pushing git changes (this only affects publish).
 
+#### --skip-git-push
+
+```sh
+$ lerna publish --skip-git-push
+```
+
+When run with this flag, `publish` will publish to npm without pushing the commit and tags up to git remote.
```
evocateur: I must admit my gut-reaction to this PR was "this is going to be abused and it will cause far more damage than it is worth." Let me explain.

Lerna's core value proposition, from my perspective, is that it handles the boring, repetitive, but _necessary_ steps of managing multiple npm packages according to community best-practices. One of those best-practices is that (if you're managing the version field in source control) you should always push the version bump commit and git tag to origin after calling `npm publish`. `npm` does not _force_ you to do this, but it is nonetheless important to keep public git history consistent with public npm registry state.

Until we get to a place where lerna's internals are properly modularized and published as separate packages, I'm very reticent to add a footgun like this to address a perplexing CI requirement. If you've already accepted that a CI process will be injecting commits into your public history, what can't you accomplish with the current all-in-one approach? Adding generated changelogs to the commit?

KingScooty: @evocateur an issue i've come across with Lerna pushing a commit along with a tag when using CI, is it results in perpetual running of builds. CI runs lerna publish, lerna pushes to git, CI detects a git change, and triggers a new build, and the cycle continues.

Am i missing something? I'm finding it very difficult with the current flags to get Lerna working nicely with CI. You must have an ideal workflow in your head when integrating Lerna into CI, what is it?

PR #1069, opened 2017-10-11
feat: implement lerna add <pkg>[@version] [--dev]

marionebl:
* adresses #998
* implement AddCommand
* add AddCommand unit tests
* add lerna-add integration tests
* add pkgMatchers for better manifest assertions

## Description

**Given**
```sh
.
├── lerna.json
├── package.json
└── packages
    ├── package-1
    │   └── package.json # 1.0
    ├── package-2
    │   └── package.json # 2.0
    ├── package-3
    │   └── package.json # 1.0
    └── package-4
        └── package.json # 1.0
```

```sh
# adds @test/package-1@^1.0.0 to package-2,3,4
lerna add @test/package-1 

 # adds @test/package-1@~1 to package-2,3,4
lerna add @test/package-1@~1

# fails with ENOTFOUND
lerna add @test/package-5

# fails with ENOTSATISFIED
lerna add @test/package-1@2 

# noop
lerna add @test/package-1 --scope=@test/package-1 

# adds @test/package-1@^1.0.0 to package-2
lerna add @test/package-1 --scope=@test/package-2 

 # adds @test/package-1@^1.0.0 to package-3,4
lerna add @test/package-1 --ignore=@test/package-2

# @test/package-1 => 2,3,4
# @test/package-2 => 1,3,4  
lerna add @test/package-{1,2} --ignore=@test/package-2 
```

## Motivation and Context

* adresses #998
* solution designed according to https://github.com/lerna/lerna/issues/998#issuecomment-326413910
* focused on `add` as first step, `remove` and `upgrade` in separate PRs?

### Before
* To depend on local, unpublished dependency you have to
  * add `<dependency>` to `<depending>/package.json` by hand
  * execute `lerna bootstrap --scope=<depending>`
### After
* `lerna add <dependency> --scope=<depending>`

## How Has This Been Tested?
* Added unit tests
* Added integration tests
* Some smoke tests in private projects

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,29 @@ Let's use `babel` as an example.
 - `babel-generator` and `source-map` (among others) are dependencies of `babel-core`.
 -  `babel-core`'s [`package.json`](https://github.com/babel/babel/blob/13c961d29d76ccd38b1fc61333a874072e9a8d6a/packages/babel-core/package.json#L28-L47) lists both these packages as keys in `dependencies`, as shown below.
 
+### add
+
+```sh
+$ lerna add <package>[@version] [--dev]
+```
+
+Add `package` as dependency to packages in in the current Lerna repo.
+
+When run, this command will:
+
+1. Add `package` to each applicable package. Applicable are packages that are not `package` and are in scope
+2. Bootstrap packages with changes to their manifest file (`package.json`)
+
+`lerna add` respects the `--ignore`, `--scope` and `--include-filtered-dependencies` flags (see [Flags](#flags)).
+
+#### Examples
+
+```
+lerna add module-1 --scope=module-2 # Install module-1 to module-2
+lerna add module-1 --scope=module-2 --dev # Install module-1 to module-2 in devDependencies
+lerna add module-1 # Install module-1 in all modules
+```
+
 ```js
 // babel-core package.json
 {
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ Let's use `babel` as an example.
 $ lerna add <package>[@version] [--dev]
 ```
 
-Add `package` as dependency to packages in in the current Lerna repo.
+Add local or remote `package` as dependency to packages in in the current Lerna repo.
 
 When run, this command will:
 
@@ -243,7 +243,8 @@ When run, this command will:
 ```
 lerna add module-1 --scope=module-2 # Install module-1 to module-2
 lerna add module-1 --scope=module-2 --dev # Install module-1 to module-2 in devDependencies
-lerna add module-1 # Install module-1 in all modules
+lerna add module-1 # Install module-1 in all modules except module-1
+lerna add babel-core # Install babel-core in all modules
 ```
 
 ```js
--- a/src/Command.js
+++ b/src/Command.js
@@ -443,7 +443,6 @@ export default class Command {
 
   _logError(method, description, err) {
     log.error(method, description);
-
     // npmlog does some funny stuff to the stack by default,
     // so pass it directly to avoid duplication.
     log.error("", cleanStack(err, this.className));
--- a/src/Command.js
+++ b/src/Command.js
@@ -443,6 +443,7 @@ export default class Command {
 
   _logError(method, description, err) {
     log.error(method, description);
+
     // npmlog does some funny stuff to the stack by default,
     // so pass it directly to avoid duplication.
     log.error("", cleanStack(err, this.className));
--- a/src/commands/AddCommand.js
+++ b/src/commands/AddCommand.js
@@ -27,12 +27,6 @@ export default class AddCommand extends Command {
     return false;
   }
 
-  constructor(input, flags, cwd) {
-    super(input, flags, cwd);
-    this.flags = flags;
-    this.cwd = cwd;
-  }
-
   initialize(callback) {
     const pkgs = this.input
       .filter(input => typeof input === 'string' && input.trim() !== '')
--- a/src/commands/AddCommand.js
+++ b/src/commands/AddCommand.js
@@ -152,15 +152,14 @@ export default class AddCommand extends Command {
       .then(() => {
         this.logger.info(`Changes require bootstrap in ${this.packagesToChange.length} packages`);
 
-        // Bootstrap all changed packages
         const options = Object.assign({}, this.options, {
           scope: this.packagesToChange.map(pkgToChange => pkgToChange.name)
         });
 
         return new BootstrapCommand([], options, this.repository.rootPath).run()
-          .then(() => callback())
-          .catch(callback);
-      });
+      })
+      .then(() => callback())
+      .catch(callback);
   }
 
   getPackage(name) {
--- a/src/commands/AddCommand.js
+++ b/src/commands/AddCommand.js
@@ -1,0 +1,222 @@
+import path from "path";
+import dedent from "dedent";
+import {JSONFile} from "json-file-plus";
+import semver from "semver";
+import BootstrapCommand from "./BootstrapCommand";
+import Command, { ValidationError } from "../Command";
+import FileSystemUtilities from "../FileSystemUtilities";
+
+export const command = "add [args..]";
+
+export const describe = "Add packages as dependency to matched packages";
+
+export const builder = {
+  dev: {
+    describe: `Save as devDependency`
+  },
+};
+
+export function handler(argv) {
+  new AddCommand([argv.script, ...argv.args], argv, argv._cwd)
+    .run()
+    .then(argv._onFinish, argv._onFinish);
+}
+
+export default class AddCommand extends Command {
+  get requireGit() {
+    return false;
+  }
+
+  constructor(input, flags, cwd) {
+    super(input, flags, cwd);
+    this.flags = flags;
+    this.cwd = cwd;
+  }
+
+  initialize(callback) {
+    const pkgs = this.input
+      .filter(input => typeof input === 'string' && input.trim() !== '')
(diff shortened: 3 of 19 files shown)


gausie:
Is there anything stopping this being merged?

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+            const id = `${pkgToInstall.name}@${range}`;
+            const message = `Add ${id} as ${this.dependencyType} in ${pkgToChange.name}`;
+            this.logger.verbose(message);
+            results[pkgToInstall.name] = range;
+            return results;
+          }, {});
+
+        const source = FileSystemUtilities.readFileSync(manifestPath);
+
+        const file = new JSONFile(manifestPath, source);
```
evocateur: Is it really necessary to add `JSONFile`? You should be using `write-pkg`, an existing dependency, to write the package.json back to disk, as it already handles alphabetizing the updated (dev-)dependencies.

If you use `read-pkg` instead of reading the string, you can just merge `applicable` into the targeted `dependencyType` and write it back without testing, since the serialization is always alphabetized anyway.

```js
const manifestJson = readPkg.sync(manifestPath, { normalize: false });
const manifestDeps = manifestJson[dependencyType] || {};

// overwrite updated dependencies
manifestJson[dependencyType] = {
  ...manifestDeps,
  ...applicable,
};

writePkg.sync(manifestPath, manifestJson);
```

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+      });
+
+      this.logger.info(`Changes require bootstrap in ${this.packagesToChange.length} packages`);
+
+      // Bootstrap all changed packages
+      const flags = Object.assign({}, this.flags, {
+        scope: this.packagesToChange.map(pkgToChange => pkgToChange.name)
+      });
+
+      return new BootstrapCommand([], flags, this.cwd).run()
```
evocateur: Ah, I see why the `constructor` is there. Actually, you don't need `this.flags`, you can use `this.options` to merge the updated scope onto. `this.cwd` should be `this.repository.rootPath`, which is equivalent for the purposes of calling another command.

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+  getPackage(name) {
+    return this.packages.find(pkg => pkg.name === name);
+  }
+
+  packageExists(name) {
+    return this.packages.some(pkg => pkg.name === name);
+  }
+
+  packageSatisfied(name, versionRange) {
+    const pkg = this.packages.find(pkg => pkg.name === name);
```
evocateur: Wondering why not `const pkg = this.getPackage(name);` here?

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+        scope: this.packagesToChange.map(pkgToChange => pkgToChange.name)
+      });
+
+      return new BootstrapCommand([], flags, this.cwd).run()
+        .then(() => callback())
+        .catch(callback);
+  }
+
+  getPackage(name) {
+    return this.packages.find(pkg => pkg.name === name);
```
evocateur: This and the body of `packageExists()` seems to wish `this.packages` was a`Map()<String(name), Package>`...

...or maybe `Set()<Package>`? Noodling for higher-level refactoring, no worries in this PR.

test/AddCommand.js
```diff
@@ -0,0 +1,185 @@
...
+const commandFlags = (mock) => {
+  return mock.mock.calls[0][1];
+};
+
+jest.mock("../src/NpmUtilities");
+jest.mock("../src/commands/BootstrapCommand");
+
+expect.extend(pkgMatchers);
+
+describe("AddCommand", () => {
```
evocateur: These are brilliant tests, thanks!

test/helpers/pkgMatchers.js
```diff
@@ -0,0 +1,50 @@
+import semver from "semver";
+
+const matchDependency = dependecyType => {
```
evocateur: should be `dependencyType`. I go cross-eyed typing "dependency" a lot 😭

test/integration/lerna-add.test.js
```diff
@@ -0,0 +1,67 @@
...
+    const cwd = await initFixture("AddCommand/basic");
+    await cli(["@test/package-1"], {cwd});
+    const pkgs = await getPkgs(cwd);
+    expect(pkgs['@test/package-1']).not.toDependOn('@test/package-1');
+    expect(pkgs['@test/package-2']).toDependOn('@test/package-1');
+    expect(pkgs['package-3']).toDependOn('@test/package-1');
+    expect(pkgs['package-4']).toDependOn('@test/package-1');
+  });
+
+  test("add to scoped package", async () => {
```
evocateur: Given the current functionality, I'm not sure we need any more integration tests beyond the first. (variations based on args and whatnot are already covered in unit tests)

README.md
```diff
@@ -223,6 +223,29 @@ Let's use `babel` as an example.
...
+
+```sh
+$ lerna add <package>[@version] [--dev]
+```
+
+Add `package` as dependency to packages in in the current Lerna repo.
+
+When run, this command will:
+
+1. Add `package` to each applicable package. Applicable are packages that are not `package` and are in scope
```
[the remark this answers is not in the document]
evocateur: (it occurs to me that the `--hoist` behaviour is already dealt with by calling bootstrap at the end of execute, _as long as_ we are also passing along `this.options` instead of the flags)

src/Command.js
```diff
@@ -437,7 +437,6 @@ export default class Command {
 
   _logError(method, description, err) {
     log.error(method, description);
-
```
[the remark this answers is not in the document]
marionebl: Addressed via https://github.com/lerna/lerna/pull/1069/commits/b34a031401f1d7aeb4c0aed1faae081868e73c50

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+export const describe = "Add packages as dependency to matched packages";
+
+export const builder = {
+  dev: {
+    describe: `Save as devDependency`
+  },
+};
+
+export function handler(argv) {
+  new AddCommand([argv.script, ...argv.args], argv, argv._cwd)
```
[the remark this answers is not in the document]
marionebl: Adressed via https://github.com/lerna/lerna/pull/1069/commits/e1d5fe55a0c1201420a49ee59496d4143a9071cb

src/commands/AddCommand.js
```diff
@@ -0,0 +1,222 @@
...
+    .run()
+    .then(argv._onFinish, argv._onFinish);
+}
+
+export default class AddCommand extends Command {
+  get requireGit() {
+    return false;
+  }
+
+  constructor(input, flags, cwd) {
```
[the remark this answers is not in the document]
marionebl: Adressed via https://github.com/lerna/lerna/pull/1069/commits/e6bfdff89128453513861697db06da551cb7ec45

src/commands/AddCommand.js
marionebl: I used `json-file-plus` to preserve `package.json` formatting. As this differs in behaviour from `npm` and other `lerna` commands and does not handle key alphabetization, I'll refactor as outlined.

src/commands/AddCommand.js
marionebl: Addressed via https://github.com/lerna/lerna/pull/1069/commits/a3a09a253e20aa90405901c8f438cf0311f28e71

test/helpers/pkgMatchers.js
marionebl: Addressed via https://github.com/lerna/lerna/pull/1069/commits/e4c9f0c6affcab5d07e1e3f78fb27c9a99b52268

test/integration/lerna-add.test.js
marionebl: Adressed via https://github.com/lerna/lerna/pull/1069/commits/efa5cc3e964978e50cd8cd62d34a8f8d4dc6a94d

README.md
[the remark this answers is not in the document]
marionebl: Addressed installing external dependencies via https://github.com/lerna/lerna/pull/1069/commits/4be0d91870596bf91f6e3ad58b178e66213806be. 

The change got bigger than I'd like due to the fact I switched to an async control flow. 

This is needed due to the network requests we send to determine the correct remote package version ranges to persist.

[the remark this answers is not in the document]
marionebl: This type of thing could be easier to build with `async - await`. Is there a reason is not used in the codebase so far?

src/commands/AddCommand.js
marionebl: I had the impression that iterating over all (filtered) packages was more common in the code base, and `<Package>[]` makes that very easy.

evocateur:
Sorry for the force push, I should really stop trying to use the dumb Github UI for resolving merge conflicts (it never, ever works)...

kachkaev:
Thank you for for adding this feature @marionebl!

I noticed that placing `--dev` before package names fails the script and it took a while to investigate. Just curious: is this intentional?

```bash
## not in a mono-repo
yarn add something --dev ## works
yarn add --dev something ## works

## monorepo
lerna add something --dev ## works
lerna add --dev something ## ERROR: Missing list of packages to add to your project.
```

Should not `--dev` (as well as its alias `-D`) be used as flags so that they don't swallow an argument?

kachkaev:
Created a new issue https://github.com/lerna/lerna/issues/1386

PR #1070, opened 2017-10-11
Fixing diffing on Windows

the0neWhoKnocks:
## Description
We started using Lerna at work, and folks on Macs don't have any issues when using the `--since` flag during testing. Noticed on Windows boxes the tests would never get run when there were code changes.

- `diffSinceIn` wasn't accounting for Windows paths, so when the diff would run it'd never detect changes and tests wouldn't run.
- Had to patch up some files so that
  - `cleanStack` would output errors
  - Promises wouldn't get stuck, not resolve, and end with a timeout error in the `ImportCommand` tests

## Motivation and Context
Allows devs on Windows machines to utilize a key feature of Lerna.

## How Has This Been Tested?
- Ran the `test` command multiple times
- Had a coworker run my code on his Mac
- Added in a `watch` command to make working with tests a little easier

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "integration": "jest --config test/config/integration.json",
     "pretest": "npm run lint -- --cache",
     "test": "jest",
+    "test:watch": "jest --watchAll --no-watchman",
     "ci": "npm test -- --coverage --verbose && npm run integration",
     "prepublish": "npm run build"
   },
--- a/src/Command.js
+++ b/src/Command.js
@@ -214,6 +214,7 @@ export default class Command {
     return new Promise((resolve, reject) => {
       const onComplete = (err, exitCode) => {
         if (err) {
+          if(typeof err === 'string') err = { stack:err };
           err.exitCode = exitCode;
           reject(err);
         } else {
@@ -472,7 +473,7 @@ export function commandNameFromClassName(className) {
 }
 
 function cleanStack(err, className) {
-  const lines = err.stack.split('\n');
+  const lines = (err.stack) ? err.stack.split('\n') : err.split('\n');
   const cutoff = new RegExp(`^    at ${className}._attempt .*$`);
   const relevantIndex = lines.findIndex((line) => cutoff.test(line));
   return lines.slice(0, relevantIndex).join('\n');
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -129,9 +129,19 @@ export default class GitUtilities {
   }
 
   static diffSinceIn(since, location, opts) {
-    log.silly("diffSinceIn", since, location);
+    let normalizedPath = location;
 
-    const diff = ChildProcessUtilities.execSync("git", ["diff", "--name-only", since, "--", location], opts);
+    /**
+     * There are cases on Windows where `location` comes in as the absolute path
+     * so when the diff runs it always reports there is no diff.
+     */
+    if(normalizedPath.indexOf('packages') > -1){
+      normalizedPath = `packages${ normalizedPath.split('packages')[1] }`.replace('\\', '/');
+    }
+
+    log.silly("diffSinceIn", since, normalizedPath);
+
+    const diff = ChildProcessUtilities.execSync("git", ["diff", "--name-only", since, "--", normalizedPath], opts); // eslint-disable-line
     log.silly("diff", diff);
 
     return diff;
--- a/test/ImportCommand.js
+++ b/test/ImportCommand.js
@@ -84,10 +84,10 @@ describe("ImportCommand", () => {
       expect(await lastCommitInDir(externalDir)).toBe("Branch merged");
 
       await lernaImport(externalDir, "--flatten");
-      expect(await lastCommitInDir(testDir)).toBe("Branch merged");
+      expect(await lastCommitInDir(externalDir)).toBe("Branch merged");
 
-      const newFilePath = path.join(testDir, "packages", path.basename(externalDir), conflictedFileName);
-      expect(await pathExists(newFilePath)).toBe(true);
+      const newFilePath = path.join(externalDir, conflictedFileName);
+      expect(pathExists.sync(newFilePath)).toBe(true);
     });
 
     it.skip("works with --max-buffer", async () => {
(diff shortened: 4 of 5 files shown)


the0neWhoKnocks:
@gigabo I'm curious about how long it usually takes to get a PR merged and published for this repo. The fix is needed for work, so if there's anything else you need from me, please let me know.

src/GitUtilities.js
```diff
@@ -129,9 +129,19 @@ export default class GitUtilities {
...
 
   static diffSinceIn(since, location, opts) {
-    log.silly("diffSinceIn", since, location);
+    let normalizedPath = location;
+
+    /**
+     * There are cases on Windows where `location` comes in as the absolute path
+     * so when the diff runs it always reports there is no diff.
+     */
+    if(normalizedPath.indexOf('packages') > -1){
```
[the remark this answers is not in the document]
evocateur: Also, hard-coding `packages` is not appropriate here, as there could be multiple package locations that have any arbitrary directory name. We need to determine _why_ Windows sometimes gives us absolute paths for the `location` property of Package objects, as passed in [UpdatedPackagesCollector](https://github.com/lerna/lerna/blob/34fd08aaac474693f8d9e0cc46bf22db862880ba/src/UpdatedPackagesCollector.js#L170).

src/GitUtilities.js
```diff
@@ -129,9 +129,19 @@ export default class GitUtilities {
...
+      normalizedPath = `packages${ normalizedPath.split('packages')[1] }`.replace('\\', '/');
+    }
+
+    log.silly("diffSinceIn", since, normalizedPath);
 
-    const diff = ChildProcessUtilities.execSync("git", ["diff", "--name-only", since, "--", location], opts);
+    const diff = ChildProcessUtilities.execSync("git", ["diff", "--name-only", since, "--", normalizedPath], opts); // eslint-disable-line
```
evocateur: If the line is too long, the argument array should be spread across multiple lines.

test/ImportCommand.js
```diff
@@ -84,10 +84,10 @@ describe("ImportCommand", () => {
       expect(await lastCommitInDir(externalDir)).toBe("Branch merged");
 
       await lernaImport(externalDir, "--flatten");
-      expect(await lastCommitInDir(testDir)).toBe("Branch merged");
+      expect(await lastCommitInDir(externalDir)).toBe("Branch merged");
```
evocateur: The `"Branch merged"` commit is no longer imported into the target repo? That doesn't seem right.

test/ImportCommand.js
```diff
@@ -84,10 +84,10 @@ describe("ImportCommand", () => {
...
 
-      const newFilePath = path.join(testDir, "packages", path.basename(externalDir), conflictedFileName);
-      expect(await pathExists(newFilePath)).toBe(true);
+      const newFilePath = path.join(externalDir, conflictedFileName);
```
evocateur: This is definitely not correct. We're testing that the conflicted file was imported into the target repo, here.

test/ImportCommand.js
```diff
@@ -98,14 +98,14 @@ describe("ImportCommand", () => {
...
-      const newFilePath = path.join(testDir, "packages", path.basename(externalDir), "new-file");
+      const newFilePath = path.join(externalDir, "new-file");
 
       await execa("git", ["mv", "old-file", "new-file"], { cwd: externalDir });
       await execa("git", ["commit", "-m", "Moved old-file to new-file"], { cwd: externalDir });
 
       await lernaImport(externalDir);
 
-      expect(await lastCommitInDir(testDir)).toBe("Moved old-file to new-file");
+      expect(await lastCommitInDir(externalDir)).toBe("Moved old-file to new-file");
```
evocateur: literally not the test, please revert this entire file.

evocateur:
@the0neWhoKnocks I appreciate the effort, but please understand that we are an entirely volunteer-run project. We get to PRs when we get to them.

src/Command.js
```diff
@@ -214,6 +214,7 @@ export default class Command {
     return new Promise((resolve, reject) => {
       const onComplete = (err, exitCode) => {
         if (err) {
+          if(typeof err === 'string') err = { stack:err };
```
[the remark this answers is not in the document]
the0neWhoKnocks: Agreed, but when I was running tests, that's what I was seeing. The Promises were stalling out, and after digging through the code I found that the errors weren't being reported correctly because this bit of the code was expecting an Object to set `exitCode` on. This check doesn't hurt anything, and makes it work in such cases.

test/ImportCommand.js
the0neWhoKnocks: Dug through code, and I didn't see in the tests where you were actually adding those files to `testDir`, just saw them being created in `externalDir`. Is there a function that's supposed to copy those over in this test context?

the0neWhoKnocks: When I ran tests on `master` (with *no changes*), I was getting Promise errors, and errors for the tests dealing with `testDir`.

test/ImportCommand.js
the0neWhoKnocks: What method is supposed to import that conflicted file?

src/GitUtilities.js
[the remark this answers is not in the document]
the0neWhoKnocks: You have multiple linting configs, which will fail tests if something doesn't follow the styles specified. That's not one of them, should I add it so someone else doesn't come across this issue?

I'll keep digging into the pathing issue. I'd previously traced it back to https://github.com/lerna/lerna/blob/34fd08aaac474693f8d9e0cc46bf22db862880ba/src/PackageUtilities.js#L80 and just figured by that point I'd tackle the issue where the error was surfacing. Didn't know you could pass in a custom `packages` path, thanks for that info.

package.json
```diff
@@ -16,6 +16,7 @@
     "integration": "jest --config test/config/integration.json",
     "pretest": "npm run lint -- --cache",
     "test": "jest",
+    "test:watch": "jest --watchAll --no-watchman",
```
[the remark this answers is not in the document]
the0neWhoKnocks: It was added for easier development. `yarn jest` will only work if `jest` is installed globally. Also, one of the pitfalls that Windows users fall into with `jest` is they'll use `--watch` which has issues where as `watchAll` does not. Just trying to make the repo friendly to work in.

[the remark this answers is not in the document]
the0neWhoKnocks: Also, the CONTRIBUTING doc calls out `npm` for most procedures, should it be updated to `yarn`?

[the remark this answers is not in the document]
evocateur: `yarn jest` calls the `jest` cli installed in local `node_modules`, actually. `npx jest` does the same thing in npm5.

Alternatively, `npm test -- --watchAll --no-watchman` will do the same thing, using any version of npm.

We have a `yarn.lock` file, which implies we use yarn to manage our dependencies. As such, that means yarn is used to run scripts and such (most npm commands still work).

We should update the contributing guide, sorry for the confusion.

the0neWhoKnocks:
@evocateur  by chance do you know the specs of the system that last ran these tests in Windows successfully? At work I'm running Windows 10, node 6.9.5, yarn 1.1.0, and Cygwin for terminal. I ask because I see specific code calling out fixes for Windows.

evocateur:
We run our test suite on AppVeyor, which is a Windows-based CI service. I don’t know the specs offhand.

Issue #1075, opened 2017-10-13
Allow `lerna publish` to skip files that are in the `.gitignore`


acarl005:
We have a slightly different workflow with Lerna. Within `packages/*`, most of the packages are on GitHub and NPM. However, there is one package in there that is not part of the monorepo. It was placed inside the monorepo in order to get cross-linked, but it is `.gitignore`'d.

The following change will accommodate a more customizable workflows. 

## Expected Behavior
We would like `lerna publish` to work on all the **checked-in** packages, as it currently does. But additionally, we would like it to update the `package.json` in the **git-ignored** package, but not try to `git add` it.

## Current Behavior
`lerna publish` command fails due to the git-ignored package.

## Possible Solution
Invoke `git check-ignore` before attempting `git add`, skipping the file to avoid an exit.
PR #1076

## Steps to Reproduce (for bugs)

```sh
📂 our-monorepo
├──📂.git
├──.gitignore  # /packages/api-service
├──📂 packages
│  ├──📂 a
│  │  └──package.json
│  ├──📂 b
│  │  └──package.json
│  └──📂 api-service
│     ├──📂 .git
│     └──package.json
├──package.json
└──lerna.json
```

```sh
$ lerna publish
# THIS FAILS
```

We put `📂 api-service` inside the monorepo because the cross-linking greatly helps local development and keeping its dependencies up-to date. But it maintains its own, separate git history for ops reasons.

matryoshcow:
Wouldn't tagging your `api-service` package as private solve your issue? `lerna publish` ignores packages that are marked as private AFAIK. This behavior is documented here: https://github.com/lerna/lerna/blob/master/README.md#publish

acarl005:
@matryoshcow Marking it as private only causes it to skip `npm publish`, not the git stuff.

PR #1076, opened 2017-10-13
publish command now respects the .gitignore

acarl005:
This change makes `lerna publish` skip files in the `.gitignore`.

## Description
Before attempting to `git add` files, `git check-ignore` is run first. This allows `lerna publish` to proceed if a file is ignored, rather than exiting.

## Motivation and Context
 Presently, Lerna assumes that all the sub-packages are part of the mono-repo. This change allows more flexible workflows, like if a package inside the monorepo still wants to benefit from the cross-linking and version-updating features of Lerna but isn't actually checked into the monorepo.

More details in this issue: #1075

## How Has This Been Tested?
We use our fork for our internal workflow and it works. Test suite is still passing, although no new unit tests were added.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -591,7 +591,14 @@ export default class PublishCommand extends Command {
     });
 
     if (this.gitEnabled) {
-      changedFiles.forEach((file) => GitUtilities.addFile(file, this.execOpts));
+      changedFiles.forEach((file) => {
+        const ignored = GitUtilities.checkIgnore(file, this.execOpts)
+        if (ignored) {
+          this.logger.info('git', `skipping "${file}" because its git-ignored`)
+        } else {
+          return GitUtilities.addFile(file, this.execOpts)
+        }
+      });
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/commands/PublishCommand.js
```diff
@@ -591,7 +591,14 @@ export default class PublishCommand extends Command {
     });
 
     if (this.gitEnabled) {
-      changedFiles.forEach((file) => GitUtilities.addFile(file, this.execOpts));
+      changedFiles.forEach((file) => {
+        const ignored = GitUtilities.checkIgnore(file, this.execOpts);
```
evocateur: How about instead we just add [`--update`](https://git-scm.com/docs/git-add#git-add---update) to the `git add` command? It's been around for almost ten years, and disallows adding new files (silently ignoring ignored files), which is okay since we're never adding new files anyway when publishing.

I'd rather not do a bunch of extra calls to git when all we need to do is add an extra flag to an existing call.

acarl005: Ah yes I hadn't considered `--update`. I will do it that way instead.

StephanBijzitter: I was just browsing through pull requests to see what's coming up and this is the first time I hear of the this amazing git feature. Completely off-topic, I know, but reading this made me a happy(ier) man.

adwigren:
This is fixed in [v3.0.0-alpha.0](https://github.com/lerna/lerna/releases/tag/v3.0.0-alpha.0) 👍

PR #1077, opened 2017-10-16
Republish prereleased packages during subsequent non-prerelease publish

### LINKED ISSUE #567: force publish for prereleased packages

spudly:
## Description
When you publish with --cd-version pre-*, and then publish with --cd-version major/minor/patch, some of your packages will still have prerelease version numbers, which is not desirable.

When you're finished with all your changes, all your packages should have non-prerelease version numbers.

This change updates the code to make it release both changed and prereleased versions if the version increment is a non-prerelease increment.

## Motivation and Context
Fixes #567

I manage a lerna monorepo with ~150 packages. Our current strategy for getting rid of prerelease version numbers is to run `lerna publish --force-publish=*`. Because of this, we often get stuck in merge hell when merging feature branches, because that updates every package's package.json file. Making this change means having conflicts in only a few package.json files instead of hundreds.

## How Has This Been Tested?
Added a few unit tests and verified that they passed without breaking other unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,8 +384,9 @@ $ lerna publish --cd-version (major | minor | patch | premajor | preminor | prep
 ```
 
 When run with this flag, `publish` will skip the version selection prompt (in independent mode) and use the next specified semantic version.
-You must still use the `--yes` flag to avoid all prompts. This is useful when build systems need
-to publish without command prompts. Works in both normal and independent modes.
+You must still use the `--yes` flag to avoid all prompts. This is useful when build systems need to publish without command prompts. Works in both normal and independent modes.
+
+If you have any packages with a prerelease version number (e.g. `2.0.0-beta.3`) and you run `lerna publish` with `--cd-version` and a non-prerelease version increment (major / minor / patch), it will publish those packages in addition to the packages that have changed since the last release.
 
 #### --preid
 
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -1,6 +1,7 @@
 import _ from "lodash";
 import minimatch from "minimatch";
 import path from "path";
+import semver from "semver";
 
 import GitUtilities from "./GitUtilities";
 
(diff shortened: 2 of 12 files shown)


spudly:
I'll add another commit to update the docs in a few...

spudly:
Ready for review.

Also, I should note that I'm not 100% sure whether this should be a breaking change or not. For me, it feels like a new feature (sorely needed by me), but technically it is changing the default behavior, so I checked the "major" box.

evocateur:
Thanks @spudly, this looks great!

I would even argue that this is a bugfix, because this really is what we should have been doing all along. I mean, who _wouldn't_ want this behaviour? If one uses prereleases at all, this is definitely an improvement.

spudly:
🎉

spudly:
Any plans to release a new version soon? I'd really love to start using this!

Issue #1015, opened 2017-09-13
root changelog for fixed-mode --conventional-commits


rsouthgate:
Our monorepo is typically distributed as a whole, though each package can live separately. We operate lerna in fixed mode and the recent changes in 2.1.0 that fix #922 are great. One thing that would help us though is the ability to combine the changelog messaging into a root changelog in addition to the package changelogs currently generated in that mode.

ZauberNerd:
I dug into this and there are a couple of complications:

* The main problem is that [conventional-changelog-core](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core) needs to know the `version` that is being / has been published.
* It does so by reading the `package.json` file in the current working directory (and if not found, searching upwards for a `package.json` file - see [conventional-changelog-core/merge-config](https://github.com/conventional-changelog/conventional-changelog/blob/431598a3b12a404aee63f8ab3e51263121cde485/packages/conventional-changelog-core/lib/merge-config.js#L97) and [conventional-changelog-cli/cli](https://github.com/conventional-changelog/conventional-changelog/blob/1fc561217d79978b9d0248612b75e87c4b4a2d0b/packages/conventional-changelog-cli/cli.js#L90)).
* The [conventional-changelog-cli](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli) can be configured to the location of the `package.json` with the `--pkg` options (which is [done by lerna](https://github.com/lerna/lerna/blob/d721b2a7579b702477e00c551eb684f25dd06305/src/ConventionalCommitUtilities.js#L66)).

Right now I can see two ways to make this work:
1. Let lerna update the `version` field in the root `package.json` file if it is configured to run in `fixed-mode`.
2. [Overwrite the `version`](https://github.com/conventional-changelog/conventional-changelog/blob/431598a3b12a404aee63f8ab3e51263121cde485/packages/conventional-changelog-core/lib/merge-config.js#L158) through a [context](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#context) that can be passed to the [conventional-changelog-cli as `--context`](https://github.com/conventional-changelog/conventional-changelog/blob/1fc561217d79978b9d0248612b75e87c4b4a2d0b/packages/conventional-changelog-cli/cli.js#L110)

If choosing option 2 there are two ways to make this work:
[...]

PR #1078, opened 2017-10-16
Generate root changelog in fixed mode

ZauberNerd:
## Description

The integration of [conventional-changelog](https://github.com/conventional-changelog/conventional-changelog) into lerna (via the `--conventional-commits` flag) currently generates one `CHANGELOG.md` file per package.

This is fine as long as lerna is configured in "independent mode", but when using lerna in "fixed mode" it might be desirable to additionally generate a consolidated `CHANGELOG.md` in the root directory of the repository.  
This "root changelog" contains changes for all packages, while the per package changelogs only list changes affecting their package.

## Motivation and Context

Based on the discussion in #1015 

## How Has This Been Tested?

* Added unit tests for changelog generation
* Updated unit tests for publish command
* locally tested this in a repository
* updated integration tests, activated them, compared snapshots and deactivated them
  * they were deactivated before, because the changelog contains dates and commit sha ids that are not stable across multiple runs. I therefore deactivated these tests again until a viable solution can be found (maybe a custom snapshot serializer to modify the files beforehand and remove all unstable values from it).

## Types of changes

## Checklist:

### DIFF
--- a/src/ConventionalCommitUtilities.js
+++ b/src/ConventionalCommitUtilities.js
@@ -69,6 +69,15 @@ export default class ConventionalCommitUtilities {
     ConventionalCommitUtilities.updateChangelog(pkg, opts, "updateFixedChangelog", args);
   }
 
+  static updateFixedRootChangelog(pkg, opts) {
+    const args = [
+      CHANGELOG_CLI,
+      "-p", "angular",
+      "--context", path.resolve(__dirname, "..", "lib", "ConventionalChangelogContext.js"),
+    ];
+    ConventionalCommitUtilities.updateChangelog(pkg, opts, "updateFixedRootChangelog", args);
+  }
+
   static updateChangelog(pkg, opts, type, args) {
     log.silly(type, "for %s at %s", pkg.name, pkg.location);
 
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -590,6 +590,21 @@ export default class PublishCommand extends Command {
       changedFiles.push(packageJsonLocation);
     });
 
+    if (this.options.conventionalCommits) {
+      if (!this.repository.isIndependent()) {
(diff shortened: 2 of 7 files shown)


src/ConventionalCommitUtilities.js
```diff
@@ -69,6 +69,15 @@ export default class ConventionalCommitUtilities {
     ConventionalCommitUtilities.updateChangelog(pkg, opts, "updateFixedChangelog", args);
   }
 
+  static updateFixedRootChangelog(pkg, opts) {
+    const args = [
+      CHANGELOG_CLI,
+      "-p", "angular",
+      "--context", path.resolve(__dirname, "..", "lib", "ConventionalChangelogContext.js"),
```
evocateur: Hah, awesome!

src/commands/PublishCommand.js
```diff
@@ -590,6 +590,21 @@ export default class PublishCommand extends Command {
       changedFiles.push(packageJsonLocation);
     });
 
+    if (this.options.conventionalCommits) {
+      if (!this.repository.isIndependent()) {
```
evocateur: I probably would have combined this into one compound conditional (`this.options.conventionalCommits && !this.repository.isIndependent()`), but I am okay with this.

test/integration/lerna-publish.test.js
```diff
@@ -243,10 +243,61 @@ describe("lerna publish", () => {
 
   });
 
-  // TODO: stabilize timestamp of changelog output
-  // TODO: make interesting git history for meaningful snapshots
-  test.skip("--conventional-commits", async () => {
-    const cwd = await initFixture("PublishCommand/independent");
+  // TODO: stabilize timestamp and commit sha of changelog output
```
evocateur: Thanks for the additions here, I hope we can stop skipping these tests soon...

PR #1080, opened 2017-10-18
Add test for skipping empty patches during import --flatten

gyoshev:
Currently, `lerna import --flatten` fails to import repositories with empty commits. This PR fixes this by detecting the case and skipping any empty commits.

## Description

This patch skips empty commits when `lerna import --flatten` is used. This is done by detecting empty patches and using `git am --skip`.

## Motivation and Context

Using `lerna import --flatten` on repos with empty commits yields a cryptic error, because `git am` does not support importing of empty commits:

```
lerna ERR! execute callback with error
(node:24090) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): TypeError: Cannot read property 'split' of undefined 
```

## How Has This Been Tested?

I added a unit test, and used the new import on a repository that previously failed with the above error.

## Types of changes

## Checklist:

### DIFF
--- a/test/ImportCommand.js
+++ b/test/ImportCommand.js
@@ -110,6 +110,22 @@ describe("ImportCommand", () => {
       expect(await pathExists(newFilePath)).toBe(true);
     });
 
+    it("skips empty patches with --flatten", async () => {
+      const cwdExternalDir = { cwd: externalDir };
+      const filePath = path.join(externalDir, "file.txt");
+
+      await fs.writeFile(filePath, "non-empty content");
+      await execa("git", ["add", filePath], cwdExternalDir);
+      await execa("git", ["commit", "-m", "Non-empty commit"], cwdExternalDir);
+
+      await execa("git", ["commit", "--allow-empty", "-m", "Empty commit"], cwdExternalDir);
+
+      const { exitCode } = await lernaImport(externalDir, "--flatten");
+
+      expect(await lastCommitInDir(testDir)).toBe("Non-empty commit");
+      expect(exitCode).toBe(0);
+    });
+
     it("exits early when confirmation is rejected", async () => {
       PromptUtilities.confirm = jest.fn(callsBack(false));
 


src/commands/ImportCommand.js
```diff
@@ -171,6 +171,11 @@ export default class ImportCommand extends Command {
       // Fall back to three-way merge, which can help with duplicate commits
       // due to merge history.
       ChildProcessUtilities.exec("git", ["am", "-3"], this.execOpts, (err) => {
+        if (err && /^patch is empty/i.test(err.stdout)) {
```
evocateur: This was already fixed in https://github.com/lerna/lerna/pull/1059/files#diff-84077e6291e8b6b2a596aff7f258b0dd, but the PR view doesn't seem to incorporate those changes.

I like your unit test, but you need to rebase and remove this block before I can merge it.

test/ImportCommand.js
```diff
@@ -109,6 +109,22 @@ describe("ImportCommand", () => {
       expect(await pathExists(newFilePath)).toBe(true);
     });
 
+    it("skips empty patches with --flatten", async () => {
```
evocateur: Does it _require_ `--flatten` to skip the empty commits? The code (here or in #1059) doesn't seem to imply this.

gyoshev: If I understand this correctly, the problem occurs only with `--flatten`, because `createPatchForCommit` uses `git log`, instead of `git format-patch`. These have different behavior for empty commits:

```
(~) git format-patch -1 f3121c --stdout                                             
(~) git log --reverse --first-parent -p -m --pretty=email --stat --binary -1 f3121c 
From f3121c... Mon Sep 17 00:00:00 2001
From: Alex Gyoshev <___@___.__>
Date: Thu, 12 Oct 2017 14:23:10 +0300
Subject: ...
(~) 
```

gyoshev:
I've rebased the branch and left only the test case. The test probably should have been along with the other PR, but at least now there is some protection against regressions.

test/ImportCommand.js
evocateur: Oh, interesting! We should probably address this disparity in the future.

evocateur:
Thanks @gyoshev!

PR #1082, opened 2017-10-19
Add --force-local to link command

jiverson:
Add ignore sem versioning. Addresses https://github.com/lerna/lerna/pull/425, https://github.com/lerna/lerna/pull/908, and https://github.com/lerna/lerna/pull/928.

## Description
Add in `--force-local` option for linking only. 

## Motivation and Context
Being able to switch branches during dev mode as well as using local packages. As well as using git subtree instead of mono repos.

## How Has This Been Tested?
Yes added in integration tests.

## Types of changes

## Checklist:


README.md
```diff
@@ -637,6 +637,14 @@ $ lerna link
 
 Symlink together all Lerna `packages` that are dependencies of each other in the current Lerna repo.
 
+#### --ignore-semver
+
+```sh
+$ lerna link --ignore-semver=<boolean>
```
evocateur: Can we rename this `--force-local`, please? Even though it _does_ ignore semver, in practice, I don't want to implicitly condone ignoring semver. It's pure semantics, yes, but I think `--force-local` better describes what is actually being accomplished, rather that how it is implemented.

Also, for boolean CLI options, it's not necessary to pass a value:
```sh
$ lerna link --force-local
```
(omitting the flag entirely is the equivalent of `false`, the default)

src/commands/BootstrapCommand.js
```diff
@@ -521,6 +521,7 @@ export default class BootstrapCommand extends Command {
    * @param {Function} callback
    */
   symlinkPackages(callback) {
-    PackageUtilities.symlinkPackages(this.filteredPackages, this.packageGraph, this.logger, callback);
+    const { filteredPackages, packageGraph, logger } = this;
+    PackageUtilities.symlinkPackages(filteredPackages, packageGraph, logger, false, callback);
```
evocateur: Can you label the `false` please? I dislike long lists of arguments with sudden literals in the middle.
```diff
+const forceLocal = false;
const { filteredPackages, packageGraph, logger } = this;
-PackageUtilities.symlinkPackages(filteredPackages, packageGraph, logger, false callback);
+PackageUtilities.symlinkPackages(filteredPackages, packageGraph, logger, forceLocal, callback);
```

README.md
```diff
@@ -637,6 +637,14 @@ $ lerna link
...
+```
+
+This flag signifies whether or not the `link` command should ignore semantic version validation. Its default value is `false`.
```
evocateur: Renaming to `--force-local`, a better description would be
```txt
When passed, this flag causes the `link` command to always symlink local dependencies regardless of matching version range.
```

jiverson:
Force local makes more sense appreciate the feed back.

evocateur:
Just waiting for the gummed-up CI to finish, then I'll merge. Thanks for responding quickly!

Kikobeats:
one thing, because `boostrap` command implicitly make a `link`, have sense if  `boostrap` supports `--force-local` flag as well?

PR #1089, opened 2017-10-28
ignore package-lock.json

rgbkrk:
When installing the dev install as documented in `CONTRIBUTING.md`, a `package-lock.json` is created with `npm`. This sets the `.npmrc` to not create `package-lock`s as well as gitignore them in the repository.

The motivation for this is to ensure that following the dev install keeps a clean git tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@ coverage/
 .nyc_output/
 .eslintcache
 *.tgz
+
+package-lock.json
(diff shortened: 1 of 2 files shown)


evocateur:
I updated the guide to use only `yarn` commands for now. Thanks for raising the issue!

rgbkrk:
Ok cool, thanks!

PR #1095, opened 2017-10-31
Symlink peerDependencies that are in the same monorepo.

loganfsmyth:
## Description
Make `lerna bootstrap` symlink `peerDependencies` in the same monorepo. 

## Motivation and Context
Babel is adding a peerDependency on `@babel/core` to all of our plugins, and currently Lerna just ignores that instead of linking it.

## How Has This Been Tested?
`npm link lerna`ed into Babel and bootstrap added the symlinks.

## Types of changes

I don't see where this would be a breaking change, but I guess technically it is possible that someone was relying on the symlink _not_ being added?

## Checklist:

### DIFF
--- a/src/Package.js
+++ b/src/Package.js
@@ -57,6 +57,15 @@ export default class Package {
     );
   }
 
+  get fullDependencies() {
+    return Object.assign(
+      {},
+      this.peerDependencies,
+      this.devDependencies,
+      this.dependencies
+    );
+  }
+
   get scripts() {
     return this._package.scripts || {};
   }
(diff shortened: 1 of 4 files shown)


evocateur:
Shouldn't `@babel/core` be added to the `devDependencies` of the plugins as well, then? That's my understanding of the "best practice" when it comes to `peerDependencies`.

I'm concerned about the disparity between local peers and external peers: this patch would enable the omission of a local peer from any other dependency type and yet _require_ an external peer to also be included in `devDependencies` (or optional or "regular", but that's weird).

loganfsmyth:
Ah interesting, I guess including it in devDependencies would work. I'd be fine with that for our usecase, just hadn't considered it.

> I'm concerned about the disparity between local peers and external peers: this patch would enable the omission of a local peer from any other dependency type and yet require an external peer to also be included in devDependencies (or optional or "regular", but that's weird).

I don't totally follow the situation that you're describing. I think for me I was struggling to think of a case where you'd have a peerDependency on a package that lives in the monorepo where would expect it not to be symlinked. Like I said though, the devDependency approach is fine with me.

evocateur:
I'm mostly wary of potential footguns where lerna enables/promotes a behavior that is not possible during "normal" package development (1-1 package to repository). Adding peers to dev dependencies is an established pattern when using `peerDependencies`.

PR #1110, opened 2017-11-10
Add `--stream` option to `lerna exec`

alan-agius4:
## Description
At the moment when you run `lerna exec --parallel` You will have detailed information on where the command is running by the means of `Stout`, example each log message will be prefixed by the package name.

However, this is not the case when --parallel flag is not used. I want to make these consistant. 

## How Has This Been Tested?
I tested the changes by linking

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -20,6 +20,12 @@ export const builder = {
     type: "boolean",
     default: undefined,
   },
+  "stream": {
+    group: "Command Options:",
+    describe: "Stream output with lines prefixed by package.",
+    type: "boolean",
+    default: undefined,
+  },
   "parallel": {
     group: "Command Options:",
     describe: "Run command in all packages with unlimited concurrency, streaming prefixed output",
@@ -50,11 +56,11 @@ export default class ExecCommand extends Command {
     const { filteredPackages } = this;
 
     try {
-    this.batchedPackages = this.toposort
-      ? PackageUtilities.topologicallyBatchPackages(filteredPackages, {
-        rejectCycles: this.options.rejectCycles
-      })
-      : [filteredPackages];
+      this.batchedPackages = this.toposort
+        ? PackageUtilities.topologicallyBatchPackages(filteredPackages, {
+          rejectCycles: this.options.rejectCycles
+        })
+        : [filteredPackages];
     } catch (e) {
       return callback(e);
     }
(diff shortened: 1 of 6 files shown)


src/commands/ExecCommand.js
```diff
@@ -100,7 +100,7 @@ export default class ExecCommand extends Command {
   }
 
   runCommandInPackage(pkg, callback) {
-    ChildProcessUtilities.spawn(this.command, this.args, this.getOpts(pkg), (err) => {
+    ChildProcessUtilities.spawnStreaming(this.command, this.args, this.getOpts(pkg), pkg.name, (err) => {
```
evocateur: This needs to be behind an option, `--stream`, just like [RunCommand does](https://github.com/lerna/lerna/blob/master/src/commands/RunCommand.js#L18-L23). Otherwise, it's a breaking change, and I'd like to avoid that.

alan-agius4:
Updated

test/NpmUtilities.js
```diff
@@ -199,7 +199,7 @@ describe("NpmUtilities", () => {
     });
 
     it("trims trailing whitespace in tag parameter", () => {
-      NpmUtilities.publishTaggedInDir("trailing-tag  ", directory, callback);
+      NpmUtilities.publishTaggedInDir("trailing-tag", directory, callback);
```
evocateur: Please revert this, it is actually testing that it trims trailing whitespace.

test/ExecCommand.js
```diff
@@ -164,5 +164,16 @@ describe("ExecCommand", () => {
...
+      await lernaExec("--stream", "ls");
+
+      expect(execInPackagesStreaming(testDir)).toEqual([
+        "packages/package-1 ls",
+        "packages/package-2 ls",
+      ]);
+    });
   });
-});
+});
```
evocateur: There should be a trailing newline in this file

src/commands/ExecCommand.js
```diff
@@ -100,11 +106,19 @@ export default class ExecCommand extends Command {
   }
 
   runCommandInPackage(pkg, callback) {
-    ChildProcessUtilities.spawn(this.command, this.args, this.getOpts(pkg), (err) => {
+    const callbackMethod = (err) => {
```
evocateur: Let's name this `done`, it's less likely to be confused with `callback`

alan-agius4:
updated

evocateur:
Thanks @alan-agius4!

Issue #850, opened 2017-05-26
[Question] How to use conventional changelog


danielo515:
Hello everyone.

I have a lerna monorepo using independent mode with several packages. Lerna has solved several problems we had at our company (code splitting, dependency management, logic independence, etc), but it has introduced several challenges that I was unable to solve. The most importants are:

- changelog generation
- releasing
- making commits

The first one depends on the third. We were using github-changelog-generator on each repository, but it does not plays well with lerna.

What I would LOVE to do is generate a changelog for each package and maybe a global changelog with each package as a section.

Now that you have recently added support for conventional-changelog I feel that it should be possible. Several closed issues (like #665 ) makes me think that this should be possible, and people commenting about this is very excited. However, I have no clue about how to do any of the following: 

1. How to setup lerna to use conventional-changelog
1. How can I invoke conventional-changelog through lerna
1. How can I generate an automatic CHANGELOG file under each package directory
1. How can I enable conventional changelog by default on lerna.json
1. How can I scope/relate one specific commit message to the package it belongs to
1. How can I enforce conventional-messages with some kind of commit wizard

I think that the fifth task is done automatically by lerna by using git log scoped to certain directory (not sure where I read that)

The reason I'm asking this here is because I know that lerna guys know about lerna stuff and there are plenty of people here that seems to know how to use this new feature.

Many thanks and regards

gastonelhordoy:
Hey man, hope this help. 

I'm using [cz-lerna-changelog](https://github.com/atlassian/cz-lerna-changelog) for commit wizard (6) which also detects which package was/were affected by the changes (5). Regarding commit message enforcement, I'm using [husky](https://www.npmjs.com/package/husky) and [validate-commit-msg](https://github.com/kentcdodds/validate-commit-msg).

And then for generating changelog, I started doing some tests with [lerna-semantic-release](https://github.com/atlassian/lerna-semantic-release) but didn't complete them because of lack of time... So if you make some progress on that side share your notes!

danielo515:
Thank you @gastonelhordoy 

Could you provide me some guides about how to configure cz-lerna-changelog ? I tried installing it and trying to execute, but it does not work.

danielo515:
@gastonelhordoy  Ok, I finally figured it out. Now I'm testing it

I would prefer to provide commit messages per package. I noticed that cz-lerna-changelog detects what packages have changed. Is it possible to configure it to promt per package?

gastonelhordoy:
[...]

PR #1111, opened 2017-11-10
Add --changelog-preset option to customize --conventional-commits output

alan-agius4:
…gular`

## Description
Lerna uses `angular` as changelog preset, in my case my lerna repo is hosted on a private `bitbucket` server. For the `angular` preset is built for usage with `github` in my case I want to use some other preset such as `angular-bitbucket`

## Motivation and Context
this issue has already been highlighted in  #850

## How Has This Been Tested?
Unit Tests

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -69,6 +69,12 @@ export const builder = {
     type: "boolean",
     default: undefined,
   },
+  "changelog-preset": {
+    group: "Command Options:",
+    describe: "Use another conventional-changelog preset rather than angular.",
+    type: "string",
+    default: undefined,
+  },
   "exact": {
     group: "Command Options:",
     describe: "Specify cross-dependency version numbers exactly rather than with a caret (^).",
@@ -152,6 +158,18 @@ export default class PublishCommand extends Command {
     });
   }
 
+  get changelogOpts() {
+    if (!this._changelogOpts) {
+      const { changelogPreset, } = this.options;
+
+      this._changelogOpts = Object.assign({}, this.execOpts, {
+        changelogPreset,
+      });
+    }
+
+    return this._changelogOpts;
+  }
+
   initialize(callback) {
     this.gitRemote = this.options.gitRemote || "origin";
     this.gitEnabled = !(this.options.canary || this.options.skipGit);
@@ -201,14 +219,14 @@ export default class PublishCommand extends Command {
 
     this.packagesToPublishCount = this.packagesToPublish.length;
     try {
-    this.batchedPackagesToPublish = this.toposort
-      ? PackageUtilities.topologicallyBatchPackages(this.packagesToPublish, {
-        // Don't sort based on devDependencies because that would increase the chance of dependency cycles
-        // causing less-than-ideal a publishing order.
-        depsOnly: true,
-        rejectCycles: this.options.rejectCycles
-      })
-      : [this.packagesToPublish];
+      this.batchedPackagesToPublish = this.toposort
+        ? PackageUtilities.topologicallyBatchPackages(this.packagesToPublish, {
+          // Don't sort based on devDependencies because that would increase the chance of dependency cycles
+          // causing less-than-ideal a publishing order.
+          depsOnly: true,
+          rejectCycles: this.options.rejectCycles
+        })
+        : [this.packagesToPublish];
     } catch (e) {
       return callback(e);
     }
@@ -419,7 +437,8 @@ export default class PublishCommand extends Command {
         version: update.package.version,
         location: update.package.location
       };
-      const recommendedVersion = recommendVersionFn(pkg, this.execOpts);
+
+      const recommendedVersion = recommendVersionFn(pkg, this.changelogOpts);
       callback({ pkg, recommendedVersion });
     });
   }
@@ -544,7 +563,7 @@ export default class PublishCommand extends Command {
   }
 
   updateUpdatedPackages() {
-    const { exact } = this.options;
+    const { exact, conventionalCommits } = this.options;
     const changedFiles = [];
 
     this.updates.forEach((update) => {
@@ -574,18 +593,17 @@ export default class PublishCommand extends Command {
 
       // we can now generate the Changelog, based on the
       // the updated version that we're about to release.
-      if (this.options.conventionalCommits) {
+      if (conventionalCommits) {
         if (this.repository.isIndependent()) {
           ConventionalCommitUtilities.updateIndependentChangelog({
             name: pkg.name,
             location: pkg.location
-          }, this.execOpts);
+          }, this.changelogOpts);
         } else {
-
           ConventionalCommitUtilities.updateFixedChangelog({
             name: pkg.name,
             location: pkg.location
-          }, this.execOpts);
+          }, this.changelogOpts);
         }
 
         changedFiles.push(ConventionalCommitUtilities.changelogLocation(pkg));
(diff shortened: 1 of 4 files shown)


src/commands/PublishCommand.js
```diff
@@ -419,7 +425,12 @@ export default class PublishCommand extends Command {
         version: update.package.version,
         location: update.package.location
       };
-      const recommendedVersion = recommendVersionFn(pkg, this.execOpts);
+
+      const changelogOptions = Object.assign({}, this.execOpts, {
```
evocateur: Let's make this a lazy getter, so we only have to merge the options once:

```patch
diff --git a/src/commands/PublishCommand.js b/src/commands/PublishCommand.js
index 9edf8ad5..65222478 100644
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -152,6 +152,20 @@ export default class PublishCommand extends Command {
     });
   }
 
+  get changelogOpts() {
+    if (!this._changelogOpts) {
+      const {
+        changelogPreset,
+      } = this.options;
+
+      this._changelogOpts = Object.assign({}, this.execOpts, {
+        changelogPreset,
+      });
+    }
+
+    return this._changelogOpts;
+  }
+
   initialize(callback) {
     this.gitRemote = this.options.gitRemote || "origin";
     this.gitEnabled = !(this.options.canary || this.options.skipGit);
```

Then just pass `this.changelogOpts` to the various functions.

src/commands/PublishCommand.js
```diff
@@ -574,18 +589,17 @@ export default class PublishCommand extends Command {
 
       // we can now generate the Changelog, based on the
       // the updated version that we're about to release.
-      if (this.options.conventionalCommits) {
+      if (conventionalCommits) {
```
evocateur: Thanks, this was bugging me. :)

alan-agius4:
@evocateur, thanks for the feedback. I updated it :)

evocateur:
Thanks!

engfragui:
@alan-agius4  This `--changelog-preset` option to customize `--conventional-commits` is exactly what I need right now, since my organization hosts code on bitbucket as well (I know...).

However this new and exciting option doesn't seem to work for me. I'm using lerna v2.9.0 (but I also tried with v2.6.0), yarn 1.3.2, node 8.

I'm trying to run `lerna publish` this way: `lerna publish --skip-npm --skip-git --conventional-commits --changelog-preset angular-bitbucket` but I get the same result of `lerna publish --skip-npm --skip-git --conventional-commits` (same command, but without `--changelog-preset` option).

It makes me think that `--changelog-preset` gets completely ignored, also because I get the same result even if I do `--changelog-preset foo-bar` or passing any other invalid changelog preset.

Should I add anything to my `package.json` and/or install additional yarn dependencies?

Thank you

evocateur:
The changelog preset in lerna 2.x is unfortunately not passed into the recommended-bump cli (choosing the version), only to the conventional-changelog cli (that generates the actual changelog output).

Lerna 3.0’s first alpha (available under the “next” dist-tag) passes the preset to both APIs programmatically, perhaps that might solve your issue?

alan-agius4:
Yes, you need to install this
https://www.npmjs.com/package/conventional-changelog-angular-bitbucket

The reason why the preset is only passed to the conventional-changelog is
because of issue https://github.com/lerna/lerna/pull/1185 and if you try to
pass preset to recommanded bump, you’ll get an error.

alan-agius4:
Yes, you need to install this
https://www.npmjs.com/package/conventional-changelog-angular-bitbucket

The reason why the preset is only passed to the conventional-changelog is
because of issue #1185 and if you try to pass preset to recommanded bump, you’ll get an error

engfragui:
Thank you both. I just updated my previous comment, clarifying what it's the result that I expect (CHANGELOG.md with bitbucket-style links), and what is the one I'm actually getting (CHANGELOG.md with github-style links). 

From your answers it looks like the bitbucket preset should correctly be passed to conventional-changelog and I should get a bitbucket-style changelog, which is not what's happening at the moment.

I tried installing https://www.npmjs.com/package/conventional-changelog-angular-bitbucket and then `lerna publish` again, but it made no difference.

alan-agius4:
When you say git-style links are you referring to the commit links or issue
link? Are you setting the proper urls for bitbucket in package.json?

Can you create an example small sample?, as i am certain that this feature
work, because at work we are using this and it seems to be working.


Are you using bitbucket enterprise & which version of bitbucket sre you
using?

Also can you try to use --changelogPreset instead of  --changelog-preset?

alan-agius4:
Hey @engfragui,

Sorry that I didn't provide a lot of info earlier but I was getting on a plane.

I tried things out a bit on one of my bitbuckets repos, and you need to do a couple of things.

> Install the `conventional-changelog-angular-bitbucket`
> In each of the packages/libs `package.json` you need to specify the below
```json
  "repository": {
    "type": "git",
    "url": "https://bitbucketsson.betsson.local/projects/WF/repos/obg.ngx.packages" // note should not end with `.git`
  },
  "bugs": {
    "url": "https://jirasson.betsson.local/browse" // this can be a link to your Jira. If you don't have you can omit this section
  },
```

Also, if you provide an invalid preset, no error will be thrown. I am not sure if this is something not handled by `lerna` or by `convential-changelog`, but properly we should also look into this,

engfragui:
Thank you @alan-agius4, your comment was extremely helpful.

In the end it was sufficient to add:
```
"repository": {
    "type": "git",
    "url": "https://bitbucketsson.betsson.local/projects/WF/repos/obg.ngx.packages" // note, this should not end with `.git` and this is typically the url you use to browse the repo.
  },
```

to the root level `package.json` (in my case there was no need to add it into each of the `packages/package.json`).

My mistake was that I didn't have them at all. For some reason I assumed that they were not needed since the CHANGELOG.md that was generated via `lerna publish` had links to my bitbucket repo (even though they were github-style and not bitbucket-style like I wanted).

Just for the sake of future readers, the few lines that need to be add to `lerna.json` are:
```
"command": {
    "publish": {
      "conventionalCommits": true,
      "changelogPreset": "angular-bitbucket"
    }
  }
```

(`"publish"` needs to be inside `"command"`).

Thank you again for your help, I would buy you lunch for a week if I could.

evocateur:
Thanks @alan-agius4!

alan-agius4:
Glad to hear the issue is solved 😁

bertho-zero:
How to use a custom preset? the `-p` option is used for an official preset but for a local config file you have to use the `-n` option of conventione-changelog.

bertho-zero:
@evocateur When I use the `changelogPreset` option with the value `@openagenda/oa`, it's works with:
```
./node_modules/.bin/lerna publish
```

But the following command ignore my preset:
```
lerna publish
```

The two lerna have the same version (2.9.0), `@openagenda/oa` is replaced by `@openagenda/conventional-changelog-oa` by `conventional-changelog` but not found/used by my global lerna.

evocateur:
Your preset is also installed globally?

bertho-zero:
No my preset is in my lerna monorepo, but I am in it (it's my current working directory).

evocateur:
If I had to guess, it's a limitation of how `conventional-changelog-cli` resolves its preset argument. A similar limitation for `eslint` plugins/config that will fail if a global `eslint` is used (instead of the locally-installed copy).

Lerna 3 always uses the locally-installed version, even when called from a global installation. Also, `conventional-changelog-cli` has been replaced by direct `conventional-changelog` API calls, with a much better custom preset experience overall.

mocheng:
I have a customized preset installed at root `node_modules`. For lerna v2.6.0, it is interesting that publish command can find the preset with argument `--canary`, but fail without `--canary`. 

Upgrade to v3 beta version solves the problem.

PR #1113, opened 2017-11-13
fix: pass process.env to sub commands

### LINKED ISSUE #721: `NODE_ENV=production` is not honored when installing packages

nikku:
Closes #721

## Additional Rationale

I'd expect passing down process.env to sub commands to be the default behavior.

For instance in my monorepo project I'd do something like 

```
TEST_BROWSERS=Chrome lerna run test
```

to execute the test cases in all sub-projects against the `Chrome` browser.

## Types of changes

## Checklist:

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -157,11 +157,10 @@ export default class NpmUtilities {
       cwd: directory,
     };
 
-    if (registry) {
-      opts.env = Object.assign({}, process.env, {
-        npm_config_registry: registry,
-      });
-    }
+    // pass environment variables to sub commands
+    opts.env = Object.assign({}, process.env, registry ? {
+      npm_config_registry: registry,
+    } : {});
 
     log.silly("getExecOpts", opts);
     return opts;
--- a/test/NpmUtilities.js
+++ b/test/NpmUtilities.js
@@ -134,6 +134,7 @@ describe("NpmUtilities", () => {
       const scriptArgs = ["run", "foo", "--bar", "baz"];
       const opts = {
         cwd: directory,
+        env: process.env,
       };
       expect(ChildProcessUtilities.exec).lastCalledWith(cmd, scriptArgs, opts, expect.any(Function));
     });
@@ -153,6 +154,7 @@ describe("NpmUtilities", () => {
       const scriptArgs = ["run", "foo", "--bar", "baz"];
       const opts = {
         cwd: directory,
+        env: process.env,
       };
       expect(ChildProcessUtilities.execSync).lastCalledWith(cmd, scriptArgs, opts, expect.any(Function));
     });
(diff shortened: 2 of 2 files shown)


nikku:
Test timeout locally but pass [on CI](https://travis-ci.org/lerna/lerna/builds/301569930).

src/NpmUtilities.js
```diff
@@ -157,11 +157,10 @@ export default class NpmUtilities {
       cwd: directory,
     };
 
-    if (registry) {
-      opts.env = Object.assign({}, process.env, {
-        npm_config_registry: registry,
-      });
-    }
+    // pass environment variables to sub commands
```
evocateur: How is this different from what [execa does by default](https://github.com/sindresorhus/execa/blob/dc7e21b0754985b75a5e89917c9035beac1f1f2e/index.js#L19-L26) in v0.8.0? I'd rather not incur the expensive iterating of `process.env` twice per call, if we can help it. We could even get rid of it in the existing conditional, given what execa already does.

A recent investigation of Appveyor failures due to undefined env vars led to these PRs (the root cause mysteriously cleaned up with an updated version of a console emulator):
https://github.com/IndigoUnited/node-cross-spawn/issues/80
https://github.com/kentcdodds/kcd-scripts/pull/5

evocateur:
Like I [posted](https://github.com/lerna/lerna/issues/721#issuecomment-343967103), this isn't actually the problem.

nikku:
Please re-read my PR description. What I'd personally like to accomplish is passing down environment variables to npm scripts invoked via `lerna run <scriptname>`:

```
TEST_BROWSERS=Chrome lerna run test
```

nikku:
I've reinstalled the current version of lerna. As you mentioned it fixes the issue.

PR #1115, opened 2017-11-14
Keep non-patch subject parts when importing repository

koenpunt:
## Description

Currently when you import a repo that follows a commit message convention that prefixes commits with something between brackets; e.g. `[TICKET-123] something changed`, the prefix is stripped due to how `git am` (or actually `git mailinfo`) works.

## Motivation and Context

Since these prefixes are relevant, since they often refer to a project management system, or define the category of the commit, they should not be stripped. 

Especially since the README states the following:

> Original commit authors, dates and **messages** are preserved.

## How Has This Been Tested?

I ran the `git am` command in the `ImportCommand.js`, and tested with the command arguments, where the `--keep-non-patch` argument (forwarded as `git mailinfo -b ...`) resulted in the desired effect of only the `[PATCH]` prefix being stripped, and other prefixes remained.

~I still have to try and see if I can add an assertion to the tests for this behavior.~
_Added a test._

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/ImportCommand.js
+++ b/src/commands/ImportCommand.js
@@ -171,7 +171,7 @@ export default class ImportCommand extends Command {
       //
       // Fall back to three-way merge, which can help with duplicate commits
       // due to merge history.
-      ChildProcessUtilities.exec("git", ["am", "-3"], this.execOpts, (err) => {
+      ChildProcessUtilities.exec("git", ["am", "-3", "--keep-non-patch"], this.execOpts, (err) => {
         if (err) {
           const isEmptyCommit = err.stdout.indexOf("Patch is empty.") === 0;
           if (isEmptyCommit) {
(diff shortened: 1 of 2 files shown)


koenpunt:
There's 1 test failing but I don't believe that one's related to my changes..

evocateur:
Yeah, that's an environmental problem with npm on Windows, as far as I can tell. I kicked it again to see if we win the dice roll.

src/commands/ImportCommand.js
```diff
@@ -171,7 +171,7 @@ export default class ImportCommand extends Command {
       //
       // Fall back to three-way merge, which can help with duplicate commits
       // due to merge history.
-      ChildProcessUtilities.exec("git", ["am", "-3"], this.execOpts, (err) => {
+      ChildProcessUtilities.exec("git", ["am", "-3", "--keep-non-patch"], this.execOpts, (err) => {
```
evocateur: More detail: https://git-scm.com/docs/git-mailinfo#git-mailinfo--b

PR #1122, opened 2017-11-17
Use local lerna if available

marionebl:
## Description
This will change the CLI behaviour to use `lerna` from the nearest `node_modules` relative to `process.cwd()` if available.

## Motivation and Context
* Better dev ergonomics
* Enables local usage without aliasing in `package.json`
* Solution discussed back in #138 

## How Has This Been Tested?
* Additional unit tests for cli

## Types of changes

## Checklist:

## Breaking Change:
lerna now prefers installs resolvable relative to the current working directory.

For situations where a global install was used and an outdated lerna version
is present in the node_modules folder, this might cause breakage.

The issue can be resolved by installing a lerna version that fulfills the version
requirement in lerna.json


bin/lerna.js
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
 "use strict";
+const resolveCwd = require("resolve-cwd");
```
evocateur: I like the (more recent) [import-local](https://github.com/sindresorhus/import-local#usage) pattern, which encapsulates the tricky stuff for us:

```js
const importLocal = require("import-local");

if (importLocal(__filename)) {
    require("npmlog").verbose("cli", "using local version of lerna");
} else {
    require("../lib/cli")().parse(process.argv.slice(2));
}
```

test/cli.js
```diff
@@ -0,0 +1,31 @@
...
+import initFixture from "./helpers/initFixture";
+
+const BIN = path.join(__dirname, "../bin/lerna.js");
+const bin = (args, options) => execa(BIN, args, options);
+
+describe("cli", () => {
+  it("should throw without command", async () => {
+    try {
+      await bin([]);
+      fail(`Expected an exception, but it didn"t throw anything`); // eslint-disable-line
```
evocateur: Where does `fail` come from? Seems a little odd to templatize a string that isn't interpolated.

marionebl: `fail` is exposed by jasmine. I changed those to match the jest recommendations: https://facebook.github.io/jest/docs/en/tutorial-async.html#async-await

evocateur: Ah, that's why I didn't recognize it. I really wish we could upgrade to Jest 21, but something in Jest 20 broke how we run the integration tests...

evocateur:
So, about the "breaking change"-ness...

I don't see this change as breaking for existing calls from npm scripts, which is (especially in CI) by far the most common usage? I agree that it _does_ constitute a very different flow for those that call `lerna` directly in their terminal. Semver is hard. 😛 

I suppose I should suck it up and sketch out a 3.0 milestone, bundling a few more breaking changes (#174, ...?).

marionebl:
Yeah, scratched my head quite a bit on categorizing this, too. 

I tend to play it safe with semver and mark changes as breaking as soon as there is potential breakage, especially in uncommon usage scenarios (e.g. global installation, even on CI).

My personal approach to major releases is to handle them in no special way, so I'd release this without waiting for other changes. Def. your call though.

zxqfox:
If people using lerna as a CLI with some API and you breaking this API then it's breaking. If you don't breaking this API — then it's not.
Just decide what supported API is for this moment.

PR #1123, opened 2017-11-18
WIP: dogfeed lerna

marionebl:
Begin breaking lerna into submodules that can be managed with lerna for lernaception.

## Description
This is an outline how I'd approach the split into distinct packages, I think of this as a discussion starter.

* packages scoped  via`@lerna` 
* use `yarn` with workspaces
* distinct `@lerna/cli` package
* use `@lerna/core` as catch-all to break out pieces from over time
* use a test util package `@lerna/test` (should be private)
* remove most lifecycle scripts from packages
* call wanted scripts via `lerna run explicitly`

All of them are up for debate!

## ToDo
* [ ] come up with an idea where to put integration tests (cli?)
* [ ] package readmes
* [ ] centralize babel config in package
* [ ] centralize jest config`
* [ ] add prettier?
* [ ] add a "develop all the things" top-level run-script
* [ ] update contribution guide
* [ ] break out `@lerna/repository`
* [ ] break out `@lerna/command`
* [ ] break out `@lerna/package`

## Motivation and Context
* #174
* enables #167

## How Has This Been Tested?
- [ ] not tested yet

## Types of changes

## Checklist:

## Breaking change
The install names of lerna will change when releasing this.


CHANGELOG.md
```diff
@@ -1,3 +1,40 @@
+# Change Log
```
evocateur: Why did this file get all the whitespace formatting removed? This looks like bad publishes?

.travis.yml
```diff
@@ -25,7 +25,9 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
   - export PATH=$HOME/.yarn/bin:$PATH
 
-install: yarn --frozen-lockfile
+install:
+  - yarn --frozen-lockfile
+  - yarn run setup
```
evocateur: This should be a `postinstall` lifecycle, not a separate call?

packages/cli/cli.js
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+"use strict";
+const importLocal = require("import-local");
+
+if (importLocal(__filename)) {
```
evocateur: This shouldn't be built on top of #1122, it's polluting the diff.

packages/cli/package.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "homepage": "https://lernajs.io/",
+  "dependencies": {
+    "@lerna/core": "^2.5.1",
+    "import-local": "^0.1.1",
+    "npmlog": "^4.1.2"
+  },
+  "devDependencies": {
+    "@lerna/test": "^2.5.1",
+    "jest": "^19.0.2"
```
evocateur: All of the jest stuff (config, deps,  and running) should be done from the root.

packages/core/CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+<a name="2.5.1"></a>
+## [2.5.1](https://github.com/lerna/lerna/compare/1.0.0...2.5.1) (2017-11-18)
```
evocateur: This doesn't seem like correct output?

packages/test/src/copy-fixture.js
```diff
@@ -1,12 +1,14 @@
 import fs from "fs-extra";
 import path from "path";
+import pkgDir from "pkg-dir";
 import replaceStream from "replacestream";
 import * as constants from "./constants";
 
 export default copyFixture;
```
evocateur: This file is named `copyFixture.js` because it provides the `copyFixture` function. Not sure why the dash-cased rename?

packages/test/src/init-fixture.js
```diff
@@ -1,11 +1,11 @@
 import tempy from "tempy";
-import copyFixture from "./copyFixture";
-import gitInit from "./gitInit";
+import copyFixture from "./copy-fixture";
+import gitInit from "./git-init";
 
-export default initFixture;
+module.exports = initFixture;
```
evocateur: Why `module.exports` instead of `export default`?

yarn.lock
```diff
@@ -2674,6 +2695,49 @@ lcid@^1.0.0:
   dependencies:
     invert-kv "^1.0.0"
 
+lerna@^2.5.1:
+  version "2.5.1"
+  resolved "https://registry.yarnpkg.com/lerna/-/lerna-2.5.1.tgz#d07099bd3051ee799f98c753328bd69e96c6fab8"
```
evocateur: I don't think this kind of dependency will work, we need to be calling the locally built lerna.

yarn.lock
```diff
@@ -4205,3 +4292,7 @@ yargs@~3.10.0:
     cliui "^2.1.0"
     decamelize "^1.0.0"
     window-size "0.1.0"
+
+yarn@^1.3.2:
```
evocateur: Not sure why there's a dev dependency on `yarn`?

evocateur:
I think I'm going to take a stab at this over the holiday weekend. I have some Rollup plugins and package generator scripts that I need to open-source that would facilitate this much better.

.travis.yml
marionebl: Using a postinstallhere currently is blocked by #1125

packages/cli/package.json
marionebl: Understood. I tried to keep the modules completely self contained regarding any executable code, including tests. Is that no design goal for you?

.travis.yml
evocateur: Hah, of course, you're right. I've been trying (and largely failing) to make tests reasonable for that one.

packages/cli/package.json
evocateur: No, I'm a big believer in hoisting all testing and building to the root of a lerna repo, per-package stuff is a recipe for increased maintenance costs for little (if any) gain.

packages/test/src/init-fixture.js
marionebl: Moving this to a separate package disables jest's automatic compilation for the sources, thus ES2015 modules are unavailable.

I aimed at making this work for Node v8 for this draft, we will have to come up with a babel build for this package eventually

yarn.lock
marionebl: I thought we could track the second to latest version here. The issue with using the local version would be that there is nothing there to do the bootstraping then (if we e.g. built via babel)

yarn.lock
marionebl: Ther lerna setup uses `yarn` as npmClient. I thought it would be wise to specify that dependency explicitly instead of assuming there was some globally installed version present

packages/cli/package.json
marionebl: Ok, I'll change this accordingly

packages/core/CHANGELOG.md
marionebl: I'll have a look at the broken changelog, don't know what happened there

packages/test/src/copy-fixture.js
marionebl: I'll revert this

yarn.lock
evocateur: If we're going for node v8.x compatibility, we might as well convert stuff to CommonJS (require) and avoid the transpilation entirely. Nice and fast, but a lot more work.

yarn.lock
evocateur: We have a yarn.lock, this means developers should be using yarn, not npm, to install dependencies. I think it's safe to assume this, just as we didn't have a local dependency on npm when we used that to install packages.

marionebl: Ok, will revert

marionebl:
> I was envisioning moving the subcommands to their own packages, too. I don't like the subpath import style (import foo from "@scope/package/lib/foo";), it promotes tight-coupling.

Yes, I am def. on the same page. This PR is intended to bring and up and (partly) clarify all those detailed questions popping up in the review so far. 

Defining a split of commands into separate commands as I could imagine there would be enough value in intermediary steps where stuff that is not broken down into a package yet remains in core until we come around moving it.

This approach might make changes like this more likely to actually land while avoiding to disrupt all other feature development that might be going on.

What do you think?

> Also, we need to keep lerna as the public CLI package name.

I thought we might introduce an "alias" package for this kind of thing? I'll commit an example how this might look like.

marionebl:
> I think I'm going to take a stab at this over the holiday weekend. I have some Rollup plugins and package generator scripts that I need to open-source that would facilitate this much better.

Is this PR valuable for the changes you have in mind / does it inform your decisions a bit? 
Don't want to waste your time if it does not.

yarn.lock
marionebl: That still has the issue we have no lerna to do the root + leaf installs before said install, right?

evocateur:
> This approach might make changes like this more likely to actually land while avoiding to disrupt all other feature development that might be going on.

Yes, I agree, thanks for clarifying my own scattered thoughts. :D

> Is this PR valuable for the changes you have in mind / does it inform your decisions a bit?

Yes! I didn't mean to imply I was going to ignore this. I think your explanation of the transition steps helped sell me on it. I started sketching out a directory structure, but I didn't really get anywhere meaningful yet:
```txt
cli/
└── package.json (name: lerna)
 
commands/
├── add
├── bootstrap
├── clean
├── diff
├── exec
├── import
├── init
├── link
├── list
├── publish
├── run
└── updated

internals/
├── command
├── config
├── conventional-changelog
├── git-utils
├── package
├── pkg-utils
├── process-utils
└── prompt-utils
```
```json
  "workspaces": [
    "cli",
    "commands/*",
    "internals/*"
  ]
```
The `internals` in this sketch are incomplete, as it is missing quite a few existing equivalents. I thought about naming `internals` "core", but that felt weird. I also dislike the `-utils` suffix, but what can you do?

yarn.lock
evocateur: hrm, I'm not sure, with yarn workspaces in play. I feel like a local path as a dependency in the root package should be possible?
```json
"dependencies": {
  "lerna": "./packages/cli"
}
```

lerna.json
```diff
@@ -1,8 +1,7 @@
 {
-  "lerna": "2.0.0-beta.32",
-  "packages": [
-    "."
-  ],
+  "lerna": "2.5.1",
+  "npmClient": "yarn",
+  "useWorkspaces": true,
   "changelog": {
```
evocateur: I'm not sure, but I think we might need to still duplicate the `packages` config in lerna.json for the benefit of `lerna-changelog`, the consumer of this config....

yarn.lock
marionebl: Worth a try :)

marionebl: The current early CI failures are what I tried to avoid: https://travis-ci.org/lerna/lerna/jobs/306655406#L542

wtgtybhertgeghgtwtg:
Would it be possible to move the `conventional` stuff into another package, as well?

evocateur:
Yeah, there’s a lot of room for better modularity.

evocateur:
Implemented in #1307. Thanks for the exploration, @marionebl!

Issue #1133, opened 2017-11-21
`lerna publish` fails with some versions of Git on Windows


zenflow:
## Expected Behavior

As a Windows user with Git installed via Cygwin, I expect to be able to run `lerna publish` successfully.

## Current Behavior

After confirming the publish (i.e. typing "yes"), I get the following error in my console: 

<details><summary><strong>console error</strong></summary><p>

```txt
lerna ERR! execute caught error
lerna ERR! Error: On branch master
lerna ERR! Your branch is up-to-date with 'origin/master'.
lerna ERR!
lerna ERR! Changes not staged for commit:
lerna ERR!      modified:   apps/mtdirect-api/package.json
lerna ERR!      modified:   apps/mtdirect-proxy/package.json
lerna ERR!      modified:   apps/mtdirect-ui-nextjs/package.json
lerna ERR!
lerna ERR! no changes added to commit
lerna ERR!
lerna ERR!     at Function.module.exports.sync (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\execa\index.js:303:26)
lerna ERR!     at Function.execSync (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\lerna\lib\ChildProcessUtilities.js:60:30)
lerna ERR!     at Function.commit (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\lerna\lib\GitUtilities.js:88:39)
lerna ERR!     at PublishCommand.gitCommitAndTagVersionForUpdates (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\lerna\lib\commands\PublishCommand.js:726:30)
lerna ERR!     at PublishCommand.commitAndTagUpdates (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\lerna\lib\commands\PublishCommand.js:702:26)
lerna ERR!     at PublishCommand.execute (C:\Users\mbrunetti\Documents\git\mtdirect-monorepo\node_modules\lerna\lib\commands\PublishCommand.js:307:14)
(node:4536) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: On branch master
Your branch is up-to-date with 'origin/master'.

Changes not staged for commit:
        modified:   apps/mtdirect-api/package.json
        modified:   apps/mtdirect-proxy/package.json
        modified:   apps/mtdirect-ui-nextjs/package.json

[...]

PR #1134, opened 2017-11-21
Normalize path used for `git add` in respect to OS/environment

zenflow:
## Description

Added code to transform file paths used in `git add` to OS- (POSIX vs Windows) neutral form, for improved portability. 

## Motivation and Context

See issue #1133

## How Has This Been Tested?

Added two tests to the `GitUtilities` test suite.

Also `npm link`ed the package to my monorepo project to verify that my changes resolve the problem in #1133

## Types of changes

## Checklist:

> All new and existing tests passed

Actually some tests were failing on my Windows 7 Pro system after a fresh clone and install. Ignoring those, yes, all new and existing tests passed. I expect tests to pass on CI.

### DIFF
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -38,7 +38,8 @@ export default class GitUtilities {
 
   static addFile(file, opts) {
     log.silly("addFile", file);
-    ChildProcessUtilities.execSync("git", ["add", file], opts);
+    const portablePath = path.relative(opts.cwd, path.resolve(opts.cwd, file)).replace(/\\/g, '/');
+    ChildProcessUtilities.execSync("git", ["add", portablePath], opts);
   }
 
   static commit(message, opts) {
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -2,6 +2,7 @@ import { EOL } from "os";
 import log from "npmlog";
 import path from "path";
 import tempWrite from "temp-write";
+import slash from "slash"
 
 import ChildProcessUtilities from "./ChildProcessUtilities";
 
@@ -38,7 +39,8 @@ export default class GitUtilities {
 
   static addFile(file, opts) {
     log.silly("addFile", file);
-    const portablePath = path.relative(opts.cwd, path.resolve(opts.cwd, file)).replace(/\\/g, '/');
+    const relativePath = path.relative(opts.cwd, path.resolve(opts.cwd, file));
+    const portablePath = slash(relativePath);
     ChildProcessUtilities.execSync("git", ["add", portablePath], opts);
   }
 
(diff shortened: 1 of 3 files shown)


test/GitUtilities.js
```diff
@@ -65,6 +65,23 @@ describe("GitUtilities", () => {
...
+      const cwd = path.resolve("test");
+      const opts = { cwd };
+      GitUtilities.addFile("foo", opts);
+      expect(ChildProcessUtilities.execSync).lastCalledWith(
+        "git", ["add", "foo"], opts
+      );
+    });
+    it("works with absolute paths for file and cwd", () => {
+      const cwd = path.resolve("test");
+      const file = path.resolve("test", "foo");
```
evocateur: Shouldn't this be `const file = path.resolve(cwd, "foo");`? Otherwise I don't see how this test is any different than the previous one.

src/GitUtilities.js
```diff
@@ -38,7 +38,8 @@ export default class GitUtilities {
 
   static addFile(file, opts) {
     log.silly("addFile", file);
-    ChildProcessUtilities.execSync("git", ["add", file], opts);
+    const portablePath = path.relative(opts.cwd, path.resolve(opts.cwd, file)).replace(/\\/g, '/');
```
evocateur: It's a pity there's no built-in way to coerce paths to posix... (`path.posix.*` methods don't work with Windows paths, near as I can tell)

test/GitUtilities.js
zenflow: @evocateur 

In the previous test only `cwd` is absolute; the file path is still relative.

In this one both are absolute.

Your line of code is equivalent to mine, i.e. `path.resolve(path.resolve("test"), "foo")` (your code but with `cwd` inlined) returns the same value as `path.resolve("test", "foo")`.

Should I change it?

src/GitUtilities.js
zenflow: @evocateur 

Would it be worthwhile to pull in one of these dependencies?

https://www.npmjs.com/package/path-posix
https://www.npmjs.com/package/ensure-posix-path (no README but tests look good)

zenflow:
@evocateur Added some commits.

I think my changes to the two new tests helps to clarify the difference (there's only 1 line different between them now, `const file = "foo";` vs `const file = path.resolve(cwd, "foo");`) though the 1st test might be a bit gratuitous. We could probably do without it, as long as we still have the 2nd test. Should I just remove the 1st test or is it ok to remain?

I pulled in [`'slash'`](https://github.com/sindresorhus/slash) for coercing the paths to POSIX format. If you look at [the source](https://github.com/sindresorhus/slash/blob/master/index.js), it's a very light weight package, made just for what we need it for, and no more. And it helps to more explicitly declare the intent of what we're doing. Think this is acceptable?

Also added one more test to make sure Windows paths are always transformed to POSIX paths for `git add`.

PR #1135, opened 2017-11-22
Support publish command using npmClient config parmas

JackPu:
## Description
I change some code to support users define custom publish npm client (CNPM Yarn eg.). 

## Motivation and Context
We have our own npm registry and npm clint. But we meet some erros when lerna publish the packages So I made some changes in the NpmUtilities.js to support custom npm client.

## How Has This Been Tested?
All tests passed!

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "preintegration": "npm pack",
     "integration": "cross-env LC_ALL=en-US jest --config test/config/integration.json",
     "lint": "eslint . --ignore-path .gitignore",
-    "prepublish": "npm run build",
     "pretest": "npm run lint -- --cache",
     "test": "cross-env LC_ALL=en-US jest",
     "test:watch": "cross-env LC_ALL=en-US jest --watchAll --no-watchman",
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -17,8 +17,10 @@ function execInstall(directory, {
   // build command, arguments, and options
   const opts = NpmUtilities.getExecOpts(directory, registry);
   const args = ["install"];
-  let cmd = npmClient || "npm";
-
+  let cmd = "npm";
+  if (npmClient) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -15,7 +15,6 @@
     "preintegration": "npm pack",
     "integration": "cross-env LC_ALL=en-US jest --config test/config/integration.json",
     "lint": "eslint . --ignore-path .gitignore",
-    "prepublish": "npm run build",
```
evocateur: Revert this, please. You're literally breaking the entire package with this deletion.

src/NpmUtilities.js
```diff
@@ -145,11 +146,17 @@ export default class NpmUtilities {
...
   }
 
-  static publishTaggedInDir(tag, directory, registry, callback) {
+  static publishTaggedInDir(tag, directory, registry, config, callback) {
     log.silly("publishTaggedInDir", tag, path.basename(directory));
-
     const opts = NpmUtilities.getExecOpts(directory, registry);
-    ChildProcessUtilities.exec("npm", ["publish", "--tag", tag.trim()], opts, callback);
+    let cmd = 'npm';
+    if (typeof config === 'function') {
```
evocateur: All these conditionals are too complex. We literally control all callers of this method, if the signature changes then we can assume we can change it everywhere it is called.

src/NpmUtilities.js
```diff
@@ -17,8 +17,10 @@ function execInstall(directory, {
   // build command, arguments, and options
   const opts = NpmUtilities.getExecOpts(directory, registry);
   const args = ["install"];
-  let cmd = npmClient || "npm";
```
evocateur: This is perfectly acceptable, I'd prefer you didn't change code unrelated to the goal of the PR.

(this includes all the random deleted newlines, they're there for a reason)

PR #1149, opened 2017-11-29
Add support for root-level version lifecycle

bigtimebuddy:
## Description
This change introduces an optional, root-level `postversion` script that can be run when publishing. This happens after all per-package `postversion` scripts have been called.

## Motivation and Context
Currently, `postversion` hooks can only be run per-package. It's useful to have a `postversion` hook immediately before publishing _and_ after version bumping to rebuild or do anything else which might need to use the updated global version number. In my particular circumstance, I'm using Rollup to build all the packages at the top-level and not relying on `lerna run` (for performance and memory reasons). While my circumstance is admittedly somewhat specific, the utility of adding a hook here seems pretty generalized. 

## How Has This Been Tested?
It has been run locally within a project by running `lerna publish --skip-npm` within a project.  This triggers `postversion` if it's set in the root-level **package.json** scripts. It has been run with and without the hook and performs as expected.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -662,6 +662,9 @@ export default class PublishCommand extends Command {
     this.updates.forEach((update) => {
       this.runSyncScriptInPackage(update.package, "postversion");
     });
+
+    // run postversion, if set in the root directory
+    this.runSyncScriptInPackage(this.repository.package, "postversion");
   }
 
   gitCommitAndTagVersionForUpdates() {


evocateur:
What I'm thinking, additionally:
```diff
diff --git a/src/commands/PublishCommand.js b/src/commands/PublishCommand.js
index f1d3d994..f76c703b 100644
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -570,6 +570,9 @@ export default class PublishCommand extends Command {
     const { exact, conventionalCommits } = this.options;
     const changedFiles = [];
 
+    // exec preversion lifecycle in root (before all updates)
+    this.runSyncScriptInPackage(this.repository.package, "preversion");
+
     this.updates.forEach((update) => {
       const pkg = update.package;
       const packageLocation = pkg.location;
@@ -632,6 +635,9 @@ export default class PublishCommand extends Command {
       }
     }
 
+    // exec version lifecycle in root (after all updates)
+    this.runSyncScriptInPackage(this.repository.package, "version");
+
     if (this.gitEnabled) {
       changedFiles.forEach((file) => GitUtilities.addFile(file, this.execOpts));
     }
```

PR #1152, opened 2017-12-01
fix(NpmUtilities): include process.env in getExecOpts

Craga89:
## Description
Adds missing `env` option passed to `childProcess.exec`

## Motivation and Context
Currently, the `getExecOpts` command in `NpmUtilties` only includes the `process.env` variables if there is a registry option set. This ensures they are always passed down regardless.

## How Has This Been Tested?
Adjusted existing unit tests to over these cases and ran locally against a repo I know to be working with the `--registry` setting

## Types of changes

## Checklist:

### DIFF
--- a/src/NpmUtilities.js
+++ b/src/NpmUtilities.js
@@ -155,11 +155,12 @@ export default class NpmUtilities {
   static getExecOpts(directory, registry) {
     const opts = {
       cwd: directory,
+      env: Object.assign({}, process.env)
     };
 
     if (registry) {
-      opts.env = Object.assign({}, process.env, {
-        npm_config_registry: registry,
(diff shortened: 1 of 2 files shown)


evocateur:
`execa` (the library that wraps our `child_process` calls) is [already doing this](https://github.com/sindresorhus/execa/blob/8fb8b6d6cdad24598b665aa4e8e4ae501306bbb4/index.js#L26) for us.

Craga89:
It's been a while since I looked at this, so `execa` might have changed in the mean time to fix the issue I was experiencing. The tests included might give some insight into if this is the case; I'll re-run when I get a chance and see if that's the case.

PR #1185, opened 2018-01-09
fix: Error: Preset: "custom-preset" does not exist

alan-agius4:
## Description
 `conventional-recommended-bump` cannot accept custom preset and thus this should always be `angular` or `eslint` https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-recommended-bump/presets

When pass a custom preset is passed it this shouldn't be passed to `conventional-recommended-bump` because an error will be thrown `Error: Preset: "custom-preset" does not exist `

## Motivation and Context
Described above.

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/ConventionalCommitUtilities.js
+++ b/src/ConventionalCommitUtilities.js
@@ -19,20 +19,24 @@ const CHANGELOG_CLI = require.resolve("conventional-changelog-cli/cli");
 
 export default class ConventionalCommitUtilities {
   static recommendIndependentVersion(pkg, opts) {
+    // `-p` here is overridden because `conventional-recommended-bump`
+    // cannot accept custom preset.
     const args = [
       RECOMMEND_CLI,
       "-l", pkg.name,
       "--commit-path", pkg.location,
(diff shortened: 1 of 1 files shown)


alan-agius4:
@evocateur

alan-agius4:
One of the CI is failing due to `obsolete snapshots` however I tried to run `yarn run integration -- -u` but had no updates to commit

evocateur:
Huh, weird, I’ll take a look. I’m in the process of upgrading to Jest 22 in another PR, which dramatically refactored the integration tests. I’ll probably just take this PR and fix it later, once we have an updated Jest.

(Odd that Appveyor passed!)

evocateur:
(We should definitely figure out a way to pass the custom preset to bump later, maybe it needs upstream fixes?)

alan-agius4:
Well, the custom preset is this passed to convental-changelog-cli which is
really where it’s needed.

evocateur:
Ah, okay, that makes sense. Still weird that the recommended bump cli doesn’t match the same options as conventional commits cli.

Issue #1018, opened 2017-09-18
peerDependencies shouldn't be updated by `lerna publish`


ianstormtaylor:
Given the following setup and using `independent` versioning:

```
packages
| -- foo
| -- bar
```
```js
// foo/package.json
{
  "version" "0.2.1"
}
```
```js
// bar/package.json
{
  "peerDependencies": {
    "foo": ">=0.2.0"
  },
  "devDependencies": {
    "foo": "^0.2.1"
  }
}
```

If you make changes in `foo`, and run `lerna publish` (patch)...

1. `foo` will updated to `0.2.2`, as expected.
2. `bar`'s dev dependency on `foo` will be updated to `^0.2.2`, also as expected.
3. **But** `bar`'s peer depdenncy on `foo` will be overwritten to `^0.2.2`, unexpectedly!

I think the `peerDependencies` tracking of Lerna should be disabled, since it's impossible for Lerna to know whether a change would have forced a peer dependency upgrade or not. 

Peer dependencies should remain as loose as possible, with the absolute earliest accept version that works, reducing warning noise and increasing interop. With the current logic of forcing a bump with each publish, Lerna forces dependents to upgrade even if there's no need to.

Related to https://github.com/lerna/lerna/issues/955

ljharb:
peer deps should definitely be loose; any peer dep change that removes a valid version is semver-major.

ianstormtaylor:
Also, I forgot there's _another_ unexpected thing...

Because of the unexpected bump of the peer dependencies, when you make a small change to the "core" package that is peer-depended on, Lerna will think that any packages that have a peer dependency on it have also been "changed", so when you go to publish you'll be publishing changes for **every** package, instead of just the one you actually changed.

Can any maintainers comment on this? 

As far as I can tell, using `independent` versioning with `peerDependencies` is impossible without a broken UX for semver...

evocateur:
I agree that peer dependencies, when used at all, should be as flexible as possible. The current handling of peer dependencies in lerna could use some improvement, definitely.

That said, if you aren't _also_ expressing a dev dependency on a sibling package that you've expressed a peer dependency for, how are you even running your unit tests? Peer dependencies aren't installed by default, after all. (This pattern applies for _all_ peer dependencies, not just lerna siblings)

[...]

PR #1187, opened 2018-01-09
Completely ignore peerDependencies during publish

evocateur:
## Description
Changes to a peer version range are always semver major, and should be as broad as possible.

Until we can get fancier, we should never automatically modify them to match the new version being published (which is the current incorrect behavior).

## Motivation and Context
see #1018

## How Has This Been Tested?
new test assertions

## Types of changes

## Checklist:

### DIFF
--- a/test/Package.js
+++ b/test/Package.js
@@ -24,7 +24,7 @@ describe("Package", () => {
         scripts: { "my-script": "echo 'hello world'" },
         dependencies: { "my-dependency": "^1.0.0" },
         devDependencies: { "my-dev-dependency": "^1.0.0" },
-        peerDependencies: { "my-peer-dependency": "^1.0.0" }
+        peerDependencies: { "my-peer-dependency": ">=1.0.0" }
       },
       "/path/to/package"
     );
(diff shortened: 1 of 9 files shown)


test/Package.js
```diff
@@ -24,7 +24,7 @@ describe("Package", () => {
         scripts: { "my-script": "echo 'hello world'" },
         dependencies: { "my-dependency": "^1.0.0" },
         devDependencies: { "my-dev-dependency": "^1.0.0" },
-        peerDependencies: { "my-peer-dependency": "^1.0.0" }
+        peerDependencies: { "my-peer-dependency": ">=1.0.0" }
```
ljharb: um, this isn't a good change; "as broad as possible" requires `^`, because declaring compatibility with `^1` is not the same thing as declaring infinite future compatibility across future semver-major lines.

Can this be fixed in a followup?

evocateur: This is just to make very _very_ explicit that we are never using `peerDependencies` for version "pairing" (that is, identifying lerna-managed sibling dependencies by name + semver) in this test.

The relevant test is here: https://github.com/evocateur/lerna/blob/5f38efe0227578757fb64f02a4a891468047f408/test/PublishCommand.js#L145-L148

Lerna never ever ever ever touches `peerDependencies` now, not even to compute sibling relationships. The `Package` getter is just leftover because I was too squeamish about removing it completely (with the idea that we would eventually reuse it for the improved experienced described in #1018).

ljharb: ok cool, i was worried that this indicated that lerna actually *changed* `^` to `>=`. Never touching peer deps is better than doing the wrong thing, that's for sure (but not as good as doing the right thing, which is keeping the existing semver range but widening it as needed to include the new version)

PR #1192, opened 2018-01-11
Set `matchBase` to `false`

PepijnSenders:
## Description
At the moment ignored files are being matched with `matchBase: true`, but it's more flexible when you match the base of a file by prefixing it with `**/`.

See the following cases:

```
1. minimatch('src/index.js', 'index.js', { matchBase: false })
false
2. minimatch('src/index.js', 'index.js', { matchBase: true })
true
3. minimatch('index.js', 'index.js', { matchBase: false })
true
4. minimatch('src/index.js', '**/index.js', { matchBase: false })
true
5. minimatch('index.js', '**/index.js', { matchBase: false })
true
```

For my use case I need `1` to be `false` and `3` to be `true`. This is not possible at the moment, due to the `matchBase: true` option. If you would want to accomplish that both `1` and `3` are true, you could prefix the pattern with the `**/` wildcard like I showed in `4` and `5`.

## Motivation and Context
I want to match a file in the root folder of the package, but it at the moment it matches all files that are named like in all subfolders of that package.

## How Has This Been Tested?
I tested this in my own project and it worked.

## Types of changes

## Checklist:

### DIFF
--- a/src/UpdatedPackagesCollector.js
+++ b/src/UpdatedPackagesCollector.js
@@ -215,7 +215,7 @@ export default class UpdatedPackagesCollector {
 
     if (this.options.ignore) {
       changedFiles = changedFiles.filter(
-        file => !_.find(this.options.ignore, pattern => minimatch(file, pattern, { matchBase: true })),
+        file => !_.find(this.options.ignore, pattern => minimatch(file, pattern, { matchBase: false })),
       );
     }
 


evocateur:
What does your `lerna.json` look like? I don't understand the (very buried) request, files in the root are always ignored when determining if a package has changed.

PepijnSenders:
This is what my `lerna.json` looks like:

```json
{
  "lerna": "2.5.1",
  "npmClient": "yarn",
  "useWorkspaces": true,
  "version": "1.32.0",
  "commands": {
    "publish": {
      "ignore": [
        "__tests__",
        "entry.js",
        "*.md"
      ]
    }
  }
}
```

All I want is to ignore `entry.js` just in the root and not in any other subfolder, which is impossible at the moment.

evocateur:
> I also meant the root folder of the package and not the root folder of the project 😅, updated that in the description.

Oh good, thanks for clarifying!

Wouldn't a subsequent negation (to _enable_ the subdirectories) work? (`["entry.js", "!*/entry.js"]`)

I don't recall why I used `matchBase: true`, maybe I was smart and noted it in a commit message?

PepijnSenders:
That might work and I will test it later, but looking at the code it looks like it won't.

```
    if (this.options.ignore) {
      changedFiles = changedFiles.filter(
        file => !_.find(this.options.ignore, pattern => minimatch(file, pattern, { matchBase: true })),
      );
    }
```

The code basically tries to find the first `true` in the array of patterns. In my case it will always `find` one `true` since it's `entry.js` will be `true` for all files in all folders. If subsequently `!*/entry.js` is `false` it won't mean much to the outcome.

PR #1193, opened 2018-01-11
Throw error when lerna.json or package.json have invalid syntax

### LINKED ISSUE #1067: If lerna.json is not valid JSON, the file is reported as missing
If `lerna.json` is not valid JSON, lerna reports the file as missing, rather than saying there was an error parsing it, causing confusion to beginners as to whether they might have used the wrong name for the file, or put it in the wrong place.

## Expected Behavior
```
lerna ERR! EINVALIDLERNA Cannot parse `lerna.json`: <detail message from the loadJsonFile.sync() exception string>
```

## Current Behavior
```
lerna ERR! ENOLERNA `lerna.json` does not exist, have you run `lerna init`?
```

## Possible Solution
Adjust the try-catch here to distinguish the difference between `lerna.json` being missing or invalid:
https://github.com/lerna/lerna/blob/5b8795c952b452e9e2f450673dd2955ca9fe41e6/src/Repository.js#L35-L39
...since contrary to the comment there, the difference does matter IMO.

Or alternatively if that's not desired, the existing error message could be adjusted to say `lerna.json does not exist OR is invalid JSON`.

1. Run `yarn add lerna`
2. Create a `lerna.json` with the following content:
```js
{
  "lerna": "2.4.0",
  "version": "1.0.0",
  "packages": [
    "packages/*"
  ],
}
```
3. Run `yarn lerna bootstrap`

## Your Environment
| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.4.0 |
| `npm --version`  | 5.4.2 |
| `yarn --version` | 1.2.0 |
| `node --version` | 8.6.0 |

| OS | Version |
| --- | --- |
| Windows 10 | 1703 |

evocateur:
Dependency updates:
- load-json-file 4
- read-pkg 3

Fixes #1067

(this isn't a breaking change because it was already broken in these scenarios)


gaearon:
Thanks

PR #1202, opened 2018-01-15
Add package location env variable

houfio:
## Description
This PR adds a new environment variable to the `exec` command - `LERNA_PACKAGE_PATH`. This variable can be used to get the exact path of the package. This is useful for cases where the name of the package differs from the name of the parent folder.

## Motivation and Context
The reason I made this PR is because in order for my shared webpack config to work, I need access to the full path of the packages. The names of the packages are namespaced, and thus have a different folder name.

## How Has This Been Tested?
In order to test this change, I first modified the test cases which involve the `LERNA_PACKAGE_NAME` variable. I later reverted the change, because the test cases weren't used to test the variables. After that, I made a new build to test against my use case, and everything seemed to work fine.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -653,10 +653,10 @@ To spawn long-running processes, pass the `--parallel` flag:
 $ lerna exec --parallel -- babel src -d lib -w
 ```
 
-You may also get the name of the current package through the environment variable `LERNA_PACKAGE_NAME`:
+You may also get the name and path of the current package through the environment variables `LERNA_PACKAGE_NAME` and `LERNA_PACKAGE_PATH`:
 
 ```sh
-$ lerna exec -- npm view \$LERNA_PACKAGE_NAME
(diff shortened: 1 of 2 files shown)


evocateur:
`lerna exec` already runs the command in each targeted package (`cwd: pkg.location`), is `process.cwd()` not sufficient?

houfio:
That works just fine, thank you!

houfio:
Looks like I found another use case. I'm fairly sure there is no other way to do with without creating node scripts.
```json
    "clean": "lerna exec -- rimraf \\$LERNA_PACKAGE_PATH/lib",
    "typings": "lerna exec -- tsc --declaration --outDir \\$LERNA_PACKAGE_PATH/lib",
```

evocateur:
No, those should work fine as well, because the relative “lib” is resolved to the current working directory. I’ve used `lerna exec -- rm -rf lib` many times.

PR #1212, opened 2018-01-17
Throw friendly error when workspaces are not configured correctly

craigbilner:
## Description
If you add `useWorkspaces` to your lerna config but don't make the necessary changes in your `package.json`, `lerna bootstrap` will fail [here](https://github.com/lerna/lerna/blob/d0e8d9c5af654f52634119d2c2bd2a888b9488f0/src/PackageUtilities.js#L58).

## Motivation and Context
This PR adds a validation error to make it clearer where the problem is, rather than pointing at the call to `some` on an `undefined` value for example.

Potentially it should throw if not an `Array<string>` too?

## How Has This Been Tested?
There didn't appear to be any repo workspace tests so added an example for future use using the naming convention that seemed to exist. Added one test to confirm workspaces were correctly returned as a default and another for when `useWorkspaces` is configured only.

As an aside all tests run and pass for me if I use `yarn test` but `yarn test:watch` fails on several of them. I can look at this separately if needed?

## Types of changes

## Checklist:

### DIFF
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -61,6 +61,14 @@ class Repository {
 
   get packageConfigs() {
     if (this.lernaJson.useWorkspaces) {
+      if (!this.packageJson.workspaces) {
+        throw new ValidationError(
+          "Invalid workspaces",
+          "workspaces need to be defined in the root package.json." +
+            "See: https://github.com/lerna/lerna#--use-workspaces"
+        );
+      }
+
       return this.packageJson.workspaces;
     }
     return this.lernaJson.packages || [DEFAULT_PACKAGE_GLOB];
--- a/src/Repository.js
+++ b/src/Repository.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const dedent = require("dedent");
 const findUp = require("find-up");
 const globParent = require("glob-parent");
 const loadJsonFile = require("load-json-file");
@@ -63,9 +64,11 @@ class Repository {
     if (this.lernaJson.useWorkspaces) {
       if (!this.packageJson.workspaces) {
         throw new ValidationError(
-          "Invalid workspaces",
-          "workspaces need to be defined in the root package.json." +
-            "See: https://github.com/lerna/lerna#--use-workspaces"
+          "EWORKSPACES",
+          dedent`
+            Yarn workspaces need to be defined in the root package.json.
+            See: https://github.com/lerna/lerna#--use-workspaces
+          `
         );
       }
 
--- a/test/Repository.js
+++ b/test/Repository.js
@@ -19,11 +19,9 @@ log.level = "silent";
 
 describe("Repository", () => {
   let testDir;
-  let testDirWithWorkspaces;
 
   beforeAll(async () => {
     testDir = await initFixture("Repository/basic");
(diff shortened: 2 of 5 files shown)


test/Repository.js
```diff
@@ -19,9 +19,11 @@ log.level = "silent";
 
 describe("Repository", () => {
   let testDir;
+  let testDirWithWorkspaces;
 
   beforeAll(async () => {
     testDir = await initFixture("Repository/basic");
+    testDirWithWorkspaces = await initFixture("Repository/yarn-workspaces");
```
evocateur: I think we can move this into the one test case that actually needs it, instead of copying over the directory fixture every single test.

```patch
-    it("returns workspace packageConfigs", () => {
+    it("returns workspace packageConfigs", async () => {
+      const testDirWithWorkspaces = await initFixture("Repository/yarn-workspaces");
       const repo = new Repository(testDirWithWorkspaces);
```

test/Repository.js
```diff
@@ -137,6 +139,17 @@ describe("Repository", () => {
...
+
+    it("returns workspace packageConfigs", () => {
+      const repo = new Repository(testDirWithWorkspaces);
+      expect(repo.packageConfigs).toEqual(["packages/*"]);
+    });
+
+    it("throws with friendly error if workspaces are not configured", () => {
+      const repo = new Repository(testDir);
+      repo.lernaJson.useWorkspaces = true;
+      expect(() => repo.packageConfigs).toThrowErrorMatchingSnapshot();
```
evocateur: I think we can avoid a solitary snapshot with a concise `.toThrow(/workspaces need to be defined/)` assertion here.

src/Repository.js
```diff
@@ -61,6 +61,14 @@ class Repository {
 
   get packageConfigs() {
     if (this.lernaJson.useWorkspaces) {
+      if (!this.packageJson.workspaces) {
+        throw new ValidationError(
+          "Invalid workspaces",
```
evocateur: This should be `"EWORKSPACES"`, to be consistent with other `ValidationError`s

test/Repository.js
evocateur: And if you use the `"EWORKSPACES"` prefix, the regex can be even simpler: `.toThrow(/EWORKSPACES/)`

src/Repository.js
```diff
@@ -61,6 +61,14 @@ class Repository {
...
+          "workspaces need to be defined in the root package.json." +
```
evocateur: This should be a multiline tagged template string (using `dedent`, several examples of usage in src/Command.js).

Should probably call them "yarn workspaces", too, to clarify their origin.

craigbilner:
@evocateur I've tidied it up

evocateur:
I’m not sure what’s up with the test:watch script, I just use `yarn jest --watch` most of the time. If you’re interested in figuring it out, go for it, but it’s not high priority.

craigbilner:
@evocateur fyi - looks like it's to do with the way the tests borrow each others files so the tests can't run in parallel. Running in band works. Would be a larger rewrite of the tests.

evocateur:
Ah, yes, that is a problem right now. Thanks for looking into it.

PR #1219, opened 2018-01-19
Avoid triggering pre-commit and commitmsg hooks during publish

alan-agius4:
At the moment lerna is publish is trigger a precommit hook and a commitmsg hook

## Motivation and context.
At the moment lerna triggers precommit hooks which for publishing shouldn’t be triggered I think.

## Types of changes

## Checklist:

### DIFF
--- a/src/GitUtilities.js
+++ b/src/GitUtilities.js
@@ -51,7 +51,7 @@ function addFile(file, opts) {
 
 function commit(message, opts) {
   log.silly("commit", message);
-  const args = ["commit", "--no-gpg-sign"];
+  const args = ["commit", "--no-gpg-sign", "--no-verify"];
 
   if (message.indexOf(EOL) > -1) {
     // Use tempfile to allow multi\nline strings.
--- a/test/GitUtilities.js
+++ b/test/GitUtilities.js
@@ -92,7 +92,7 @@ describe("GitUtilities", () => {
 
       expect(ChildProcessUtilities.execSync).lastCalledWith(
         "git",
-        ["commit", "--no-gpg-sign", "-m", "foo"],
+        ["commit", "--no-gpg-sign", "--no-verify", "-m", "foo"],
         opts
       );
       expect(tempWrite.sync).not.toBeCalled();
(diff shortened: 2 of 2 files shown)


evocateur:
This seems reasonable. I've just opened #1220, though, which makes conflicting changes. The tests also need to be updated, so I'd wait until I merge #1220 first.

alan-agius4:
Okay i’ll wait for your change

evocateur:
oh bother, i mucked up the spacing after a comma during the online conflict resolution 🚯

evocateur:
Thanks!

jure:
Ah! Finally found out what's causing our issues :)

Context: we use `prettier` to format our JSONs, keep them consistent across many packages. When lerna bumps the version, it also does things like:

```
-  "keywords": ["pubsweet", "cli"],
+  "keywords": [
+    "pubsweet",
+    "cli"
+  ],
```

This is not how `prettier` wants JSONs to look. Because of --no-verify now being used in lerna, the release commit comes in "malformatted". And because we're using `lint-staged`, whenever you merge a branch in that has these changes, and you do conflict resolution (which is often), `prettier` then fixes all these JSONs back to what it thinks they should look like. And then we release again, and the cycle is replenished.

What's the reasoning behind making the release commit using `--no-verify`? What makes it special? I couldn't gather from the issue description what the usecase was. Would you be open to a PR that makes this optional (either way)?

evocateur:
> This is not how prettier wants JSONs to look.

Prettier is wrong, in this case.

More specifically, Prettier should not be mangling `package.json` files; those are the domain of package managers, which universally use `JSON.stringify()` to serialize their edits (`npm i <pkg>`, `npm up`, `npm rm`, etc).

> What's the reasoning behind making the release commit using `--no-verify`?

Lerna allows customization (to a point) of the commit message (using `--message` or the `command.publish.message` key in lerna.json), so presumably if a project _cares_ about such things, there is a way to ensure that it _automatically_ follows the commit message standards.

Additionally, as `lerna publish` creates an automated commit, we don't want to fail because of overly-strict/spurious validation. It is incumbent on the developer (or CI process) that calls `lerna publish` to have a pre-release validation process in place. You can even use the `preversion` or `prepublishOnly` lifecycle in the root package.json to do this.

> Would you be open to a PR that makes this optional (either way)?

No, I really don't think that needs to be an option, given the purpose and behaviour of `lerna publish`. Maybe documentation, sure, but there's lots of documentation missing right now.

jure:
Fair enough, package.jsons being excluded from prettier makes sense in that light. We do have complex/nested configuration in there that's not packaga manager configuration (e.g. jest), that should be consistently formatted. But actually that would happen on every lerna release anyway (through JSON.stringify), so it's not likely to become an issue.

Thanks for the reasoning!

PR #1222, opened 2018-01-20
Use realpath when creating relative symlinks

andrewdotn:
## Description

If any parent directories are symlinks, then lerna can create invalid relative symlinks. This pull request fixes that by using `realpath` to resolve symlinks before computing relative paths.

## Motivation and Context

If a monorepo contains two packages `foo` and `bar`, and `foo` depends on `bar`, then `lerna bootstrap` always creates the symlink `packages/foo/node_modules/bar` pointing at `../../bar`. This target value comes from using node’s `path.relative` function, which only looks at strings and not the filesystem.

But if `node_modules` is a symlink off to somewhere else, then `../../bar` might not even exist from inside `foo/node_modules`.

To address the context here:

First, I gather from some googling that there have been some arguments in the JS community around Node’s handling of symlinks. Some people argue that Node handles symlinks ‘wrong’, and that instead of populating large `node_modules` directories with copies of code, `node_modules` should contain only symlinks to centrally-installed packages.

I’m not trying to argue for that at all.

What I have found to be helpful is to make `node_modules` a symlink off to a different part of the disk, e.g., `/not-backed-up/$project`. `npm install` and `require` both work happily with `node_modules` being a symlink. Then large `node_modules` directories from forgotten throwaway projects do not clog backups, and go into a central location where it’s easy to see large, potentially unused and easily recreatable directories, and free up that disk space. Lerna currently chokes on that, and this change would help support that use case.

And I think that, even without that specific use case, this is a fix for a valid bug: when there are other symlinks around, lerna can create symlinks which do not resolve.

## How Has This Been Tested?

Existing test suite passes. Added tests for new code.

Ran locally to verify that valid symlinks are now created where invalid ones were created before.

## Types of changes

## Checklist:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,21 +50,10 @@ $ yarn lint
 If you want to test out Lerna on local repos:
 
 ```sh
-$ yarn build
 $ yarn link
 ```
 
 This will set your global `lerna` command to the local version.
-
-Note that Lerna needs to be built after changes are made. So you can either run
-`yarn build` to run it once, or you can run:
-
-```sh
-$ yarn dev
-```
-
-Which will start a watch task that will continuously re-build Lerna while you
-are working on it.
 
 If you would like to check test coverage, run the coverage script, then open
 `coverage/lcov-report/index.html` in your favorite browser.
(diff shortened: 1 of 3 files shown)


evocateur:
If you're really concerned about backup space, any backup program worth its salt can be configured to ignore `node_modules` directories. lerna doesn't support an externally-linked `packages` location, and never will.

I appreciate the effort, but this is far too much added complexity for the purported benefit. Thanks for understanding.

(The edits to the README are much appreciated! I cherry-picked them into master)

PR #1233, opened 2018-01-24
Issue #962: publish --canary should have some way of limiting publish

bladedancer:
## Description
The current behavior for publish --canary is that it publishes all packages whether they've been updated or not. This PR is to support the --since flag on publish --canary to limit publishing to changed
modules only.  

## Motivation and Context
Avoid publishing unchanged modules when doing canary builds. See #962

The referenced issue discusses this in terms of performance, for us the issue isn't performance but rather storage and a clearer view of what the modules the feature branch is affect. In our CI system we have feature branches that we want to QA prior to merging. These branches are built and published using the --canary flag to our internal repository. However publishing every package in the monorepo even if it has no deltas is a waste of resources. 

## How Has This Been Tested?
The unit tests are updated with additional test cases to ensure it behaves as expected.
It's also been sanity checked in our gitlab CI environment.

## Types of changes

## Checklist:

### DIFF
--- a/test/UpdatedPackagesCollector.js
+++ b/test/UpdatedPackagesCollector.js
@@ -72,6 +72,36 @@ describe("UpdatedPackagesCollector", () => {
       expect(GitUtilities.diffSinceIn).toBeCalledWith("deadbeef^..deadbeef", "location-2", undefined);
     });
 
+    it("should use the last tag for commit ranges when the canary flag is set and since is set", () => {
+      new UpdatedPackagesCollector({
+        options: {
+          canary: true,
+          since: "",
+        },
+        repository,
+        logger,
+        filteredPackages,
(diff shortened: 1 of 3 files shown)


test/UpdatedPackagesCollector.js
```diff
@@ -83,5 +113,19 @@ describe("UpdatedPackagesCollector", () => {
       expect(GitUtilities.diffSinceIn).toBeCalledWith("lastTag", "location-1", undefined);
       expect(GitUtilities.diffSinceIn).toBeCalledWith("lastTag", "location-2", undefined);
     });
+
+    it("should use the tag in non-canary mode for commit ranges when since is set to tag", () => {
```
evocateur: I appreciate the clean tests!

PR #1246, opened 2018-01-31
Resolve #902

### LINKED ISSUE #902: Commands Publish on lerna.json are ignored
Commands configs on lerna.json are being ignored

## Expected Behavior
I wanted every lerna publish to take in account both the conventionalCommits and message configs that exists on the lerna.json file

## Current Behavior
It is ignoring all the configs under the publish command

```json
{
  "lerna": "2.0.0-rc.5",
  "packages": [
    "src/*"
  ],
  "registry": "https://tools-int.caseonit.net/npm",
  "commands": {
    "publish": {
      "conventionalCommits": true,
      "message": "chore: Publish"
    }
  },
  "version": "independent"
}

```

As you can see, both message and conventional commits are included as part of lerna.json. However they are not being applied.

I'm running lerna this way:

`lerna publish --conventional-commits`

That is the only way to make conventional commits argument to work.

I was trying this without any luck

`lerna publish`

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` |  2.0.0-rc.5 |
| `npm --version`  | 4.1.2 |
| `node --version` | 7 |

M3kH:
#  Description
Solves #902 add replace `%v` in `commitAndTagUpdates`

# Motivation and Context
We want to upgrade the single packages in a meaningful manner by maintaining an individual version for each sub-package.
The commit message of the publish should be more meaningful then just `Publish` or a `static message`, so we wanna be capable of passing the `master version` to the custom `commit message`, also when individual mode is enabled, during the Publish.

## How Has This Been Tested?
Has not been tested, but is kind of straight forward.
✅ Test pass. **Does it need to be unit covered?**

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6591 @@
+{
+  "name": "lerna",
+  "version": "3.0.0-dev",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0-beta.39",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.39.tgz",
+      "integrity": "sha512-PConL+YIK9BgNUWWC2q4fbltj1g475TofpNVNivSypcAAKElfpSS1cv7MrpLYRG8TzZvwcVu9M30hLA/WAp1HQ==",
+      "dev": true,
+      "requires": {
+        "chalk": "2.3.0",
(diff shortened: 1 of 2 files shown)


evocateur:
Replacing a version placeholder in an independent commit message doesn't make sense. The override of the message itself is already validated here: https://github.com/lerna/lerna/blob/1d7eb9eeff65d5a7de64dea73613b1bf6bfa8d57/test/PublishCommand.js#L680-L688

PR #1249, opened 2018-02-01
Expansion of functionality --cd-version flag

jarskii:
Expansion of functionality --cd-version flag

## Description
If you use CLI command for publishing scope of packages, you can publish all components only with single of version type. There are cases, when need publishing components with different version types.
For example, we have 3 components in list after `lerna updated`:

```sh
lerna info result:
   - someComponent
   - anotherComponent
   - oneMoreComponent
```

We want publish someComponent with minor version, anotherComponent with major version, rest components with patch

```sh
$ lerna publish --cd-version='someComponent:minor,anotherComponent:major,patch'
```

## Motivation and Context
For cases, when you can't use prompt interface

## How Has This Been Tested?
run integration test
Make test publish on local monorepo

## Types of changes

## Checklist:


evocateur:
Sorry, this is way too complicated. Use `--conventional-commits` if you need this level of granular control.

PR #1258, opened 2018-02-11
Add ability to run top level prepublish script with flag

philquinn:
## Description
Adds a flag to run the the main packages prepublish script. This will run once the changed packages are updated but before git phase and publish phase.

## Motivation and Context

Motivation for this is to run some automated doc updates (with newest versions that will be published) and to allow it to be automatically committed with the publish.

For example, I would like to run `yarn run doc:update && git add README.md` and have it committed by lerna as part of the current lifecycle.

## How Has This Been Tested?

Added test to test if npm script is run.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,6 +451,14 @@ $ lerna publish --repo-version 1.0.1
 When run with this flag, `publish` will skip the version selection prompt and use the specified version.
 Useful for bypassing the user input prompt if you already know which version to publish.
 
+#### --run-prepublish
+
+```sh
+$ lerna publish --run-prepublish
+```
+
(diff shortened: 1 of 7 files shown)


evocateur:
The `preversion`/`version`/`postversion` lifecycle is [already supported in the root](https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L626), which seems perfectly appropriate for your use case, and I'd rather not add a flag just to support a [deprecated lifecycle script](https://docs.npmjs.com/misc/scripts#deprecation-note).

PR #1261, opened 2018-02-12
Refactor publish command to be 98% async

evocateur:
This builds on #1260, utilizing the new `Map`-and-`Set`-based `PackageGraph` idioms. In the process, some algorithms were dramatically simplified, and performance improved noticeably (especially in the integration tests).

### `PublishCommand`

* Literally every callback except the superclass lifecycle replaced with Promise-based patterns.
* Integration test now actually tests `--conventional-commits` completely, shaving off a few seconds of execution time on average.

### `ConventionalCommitUtilities`

* Use programmatic (async) `conventional-commits` APIs instead of synchronous CLI calls.
* Now supports presets for `conventional-recommended-bump`, as long as they support the [new preset API](https://github.com/conventional-changelog/conventional-changelog/pull/262).
* Only the default (`conventional-commits-angular`) is included as a dependency, anything custom will need to be installed by the consumer.


src/ConventionalCommitUtilities.js
```diff
@@ -1,108 +1,186 @@
...
 const FileSystemUtilities = require("./FileSystemUtilities");
+const ValidationError = require("./utils/ValidationError");
 
-const CHANGELOG_NAME = "CHANGELOG.md";
-const CHANGELOG_HEADER = dedent(`# Change Log
+const cfgCache = new Map();
 
-  All notable changes to this project will be documented in this file.
-  See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.`);
+function getChangelogConfig(changelogPreset = "conventional-changelog-angular") {
```
evocateur: We could just let [`conventional-changelog-preset-loader`](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-preset-loader#usage) do the dirty work, but then we wouldn't get caching and more flexibility with scopes and such.

PR #1272, opened 2018-02-16
doc: sample to install dependencies in production

clakech:
doc: sample to install dependencies in production 

## Description
When you use lerna to deploy a NodeJS project to production, you should not deploy dev dependencies. So we can't use lerna bootstrap because dev dependencies are installed with this command

## Types of changes

- [x] Documentation

## Checklist:

See https://github.com/lerna/lerna/issues/1243

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,12 +199,6 @@ When run, this command will:
 
 `lerna bootstrap` respects the `--ignore`, `--ignore-scripts`, `--scope` and `--include-filtered-dependencies` flags (see [Flags](#flags)).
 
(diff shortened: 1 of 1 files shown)


evocateur:
You should not use lerna to deploy, for the same reason you don't use npm to deploy code into production: it is not a deployment tool.

Lerna helps you publish packages. Those packages should be installed in a build (taking care of the devDependencies "problem" automatically), that build should generate a tarball artifact, and that artifact should be what is used to deploy the app into production.

PR #1287, opened 2018-02-26
Add "name" and "version" to temp "package.json"

compulim:
## Description

When we generate temp `package.json`, in addition to `dependencies`, also add `name` and `version`.

## Motivation and Context

When lerna run `npm install`, npm will update `package-lock.json` based on the content of `package.json`. If `name` and `version` is missing from `package.json`, npm will remove `name` and `version` from existing `package-lock.json`.

This means after running `lerna bootstrap`, the `package-lock.json` will have `name` and `version` removed, and it might generate unnecessary changes.

The changes is depicted as the screenshot below:

image

## How Has This Been Tested?

Run `lerna bootstrap`. It should no longer generate any changes for `package-lock.json`.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

Existing repos which have `name` and `version` stripped from their `package-lock.json`. Running `lerna bootstrap` will see their `package-lock.json` updated with `name` and `version` re-added.

This is expected behavior. On repos without lerna, running `npm install` will update `package-lock.json`, NPM will copy both `name` and `version` from `package.json` to `package-lock.json`.

Since lerna previously accidentally removed them in temp `package.json`, after the user run `lerna bootstrap` the first time after this PR, they will see changes in their `package-lock.json` with `name` and `version` added back again. This new behavior is expected and match what NPM does without lerna.

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1290, opened 2018-02-27
Preserve package.json structure during bootstrap mangling

### LINKED ISSUE #1136: Lerna is not installing optionalDependencies in workers
Hi Lerna!

I am facing a problem, my lerna programm is not installing optionalDependencies,
If I go to my worker & run "npm install" it install it correctly .

the module to install could be only installed on darwin/linux : 
https://github.com/istex/popplonode/blob/master/package.json#L6

The worker which not install it's optionalDependencies: 
https://github.com/istex/sisyphe/blob/pdfforall/src/worker/pdf/package.json

What could be the problem ?

evocateur:
## Description

Instead of nuking the entire package.json, just edit what we don't want installed (the sibling packages that are instead symlinked). This helps avoid incorrect `package-lock.json` output due to missing `package.json` properties (name and version, specifically).

* Adds an actual integration test for `lerna bootstrap --hoist`
* Replaces bloated `cash-cp` fixture dep with basic node.js copy

## Motivation and Context

Credit @compulim for the initial idea, #1287

Fixes #1136

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/BootstrapCommand.js
+++ b/src/commands/BootstrapCommand.js
@@ -425,7 +425,7 @@ class BootstrapCommand extends Command {
         }
 
         return npmInstall
-          .dependencies(rootPkg.location, depsToInstallInRoot, this.npmConfig)
+          .dependencies(rootPkg, depsToInstallInRoot, this.npmConfig)
           .then(() =>
             // Link binaries into dependent packages so npm scripts will
             // have access to them.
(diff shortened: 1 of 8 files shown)


evocateur:
Hah, whoops, I forgot `fs.copyFileSync` isn't in node 6.x

PR #1296, opened 2018-03-01
Update README.md

nchandran3:
Adds missing flag in example


README.md
```diff
@@ -242,7 +242,7 @@ When run, this command will:
 ```
 lerna add module-1 --scope=module-2 # Install module-1 to module-2
 lerna add module-1 --scope=module-2 --dev # Install module-1 to module-2 in devDependencies
-lerna add module-1 # Install module-1 in all modules except module-1
+lerna add module-1 --ignore=module-2 # Install module-1 in all modules except module-2
```
evocateur: Actually, the original example is correct. `lerna add module-1` won't add `module-1` to itself.

Issue #1293, opened 2018-02-28
lerna publish --skip-git doesn't work in non git repository


rogeriorc:
## Expected Behavior
The flag --skip-git will publish to npm without running any of the git commands.

## Current Behavior
Using the flag in a svn repository throw an error "git binary missing, or this is not a git repository."

## Possible Solution
Skip the validation and git commands, and only update the package.json of the packages, using the --cd-version flag.

## Steps to Reproduce (for bugs)
1. run "lerna publish --skip-git" on a non git repository

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.9.0",
  "packages": [
    "widgets",
    "webapp-frontend",
    "webapp-backend"
  ],
  "version": "independent"
}
```
</p></details>

<details><summary>lerna-debug.log</summary><p>

```txt
lerna info version 2.9.0
lerna ERR! ENOGIT git binary missing, or this is not a git repository. Did you already run `git init` or `lerna init`?
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.0.0-alpha.1 |
| `npm --version`  | 5.6.0 |
| `yarn --version` |   |
| `node --version` | 8.9.4 |

| OS | Version |
| --- | --- |
| Windows 10 | 1709 |

evocateur:
There's literally no way to determine _what_ version to publish without git available. Sorry.

evocateur:
(Not even `--cd-version` provides that information)

rogeriorc:
package.json:
```json
{
"version": "1.0.0"
}
```

```bash
lerna publish --cd-version=major
```

```json
{
"version": "2.0.0"
}
```

evocateur:
Without git to identify which packages have changed, this version bump would be applied to all packages, every single time, regardless of their internal changes.

Basically, “skip git” is a terrible option name, as it does nothing more than skip the git commit and push, just as “skip npm” doesn’t actually prevent all execution of npm/yarn, it just omits the publish command.

vrachels:
version bump being applied to all packages seems totally acceptable in this situation.
Maybe an additional warning to that effect would be acceptable?

cwardcode:
I completely agree that bumping the version should be an acceptable answer. Good luck trying to publish a package via Lerna if you're using AWS codebuild. 👎

PR #1297, opened 2018-03-02
lerna publish --skip-git supported in non-git repository

vrachels:
## Description
This change overrides Command::requiresGit in PublishCommand::requiresGit and adds a check of that property in the collectUpdates utility function.

## Motivation and Context
lerna publish --skip-git fails if the current directory is not a git repo, because validation and supporting functions are not being notified of the --skip-git state

## How Has This Been Tested?
Used publish --skip-git in our repository that was recently moved from git -> p4 that has many subpackages.
Used again with --skip-git --skip-npm to confirm consecutive updates of packages.
Ran npx jest: 
 running from 3.0.0-alpha.1 got 2 failed tests
 with patch fails 14 tests, 9 snapshots

looks like there are definitely issues, but seems like a deeper problem
first (new) failed test, for example, is
```
Command › .filteredPackages › --since should return packages updated since the last tag
Expected value to have length:
      2
    Received:
      ...
      5
```
and others relating to git history / diff features.

noted below that this could be a breaking change, but not that invested in this to determine what is the problem with the broken git features.  I don't mind following up if there are suggested additional changes to try.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -497,6 +497,10 @@ class PublishCommand extends Command {
     );
   }
 
+  get requiresGit() {
+    return !(this.options.canary || this.options.skipGit);
+  }
+
   runLifecycle(pkg, scriptName) {
     if (pkg.scripts[scriptName]) {
       return npmRunScript(scriptName, {
--- a/src/utils/collect-updates.js
+++ b/src/utils/collect-updates.js
@@ -9,7 +9,7 @@ const makeDiffPredicate = require("./make-diff-predicate");
 
 module.exports = collectUpdates;
 
-function collectUpdates({ filteredPackages, packageGraph, options, rootPath, execOpts, logger }) {
+function collectUpdates({ filteredPackages, packageGraph, options, rootPath, execOpts, logger, gitEnabled }) {
   const packages =
     filteredPackages.length === packageGraph.size
       ? packageGraph
@@ -19,7 +19,7 @@ function collectUpdates({ filteredPackages, packageGraph, options, rootPath, exe
 
   let { since: committish } = options;
 
-  if (GitUtilities.hasTags(execOpts)) {
+  if (gitEnabled && GitUtilities.hasTags(execOpts)) {
     if (options.canary) {
       const sha = GitUtilities.getShortSHA(execOpts);
 


evocateur:
As I already noted just today in #1293, `--skip-git` is a misleading option name, as it only skips the commit and push, not _all_ git operations.

vrachels:
And yet it's still a useful feature/tool without git, and didn't require major changes to get it working.

evocateur:
It cannot choose the correct packages to publish, thus it is worthless without git.

vrachels:
Er, It was able to update and publish all packages.  I don't see that as worthless.  Heavy-handed, perhaps, and not the greatest long term solution, but definitely still useful to our project.

Especially with --scope capability, we can manually do (or write our own tooling to do) the only-publish-what-is-updated task.

evocateur:
You can’t scope publishes.

testower:
The rejection of this PR seems a bit harsh. I can easily see use cases for this, in cases where in CI git is not available, just the cloned project files. npm is perfectly capable of publishing packages without the presence of git, why not lerna? If I know the version, why can't I simply provide it as an argument? And if the default behavior of lerna is to update all packages, whenever one of them changes (non-independent), this doesn't seem like a far-fetched addition.

evocateur:
1. That’s not how lerna publish works.

2. If you “know the version”, what’s stopping you from looping through the packages and running `npm publish` yourself?

testower:
1. Not really sure what you mean. I clearly can provide it with a version of my own as my workaround demonstrates:

```
git init . && lerna publish --skip-git --force-publish=* --repo-version $(node -e \"console.log(require('./package.json').version)\") --yes
```

2. Absolutely nothing is stopping me from doing everything myself, except it would not be very fruitful for me to reinvent the wheel.

evocateur:
> if the default behavior of lerna is to update all packages

This was the quote I was responding to. That's not how `lerna publish` works.

"[Updating] all packages" is a _side-effect_ of having no existing tags. The "normal" publish process depends on git for determining what has changed since the last release (signified by an annotated git tag), and _only_ publishes the packages that have changed since then.

testower:
I understand, thanks for clarifying. I have misunderstood the distinction betweeen fixed/locked and independent modes and assumed in fixed/locked, all packages would be updated regardless. Of course I see now that this is not the case, and explains why I needed `--force-publish=*` in my workaround.

That said, I read some recent discussions around version 3 of lerna, where you indicate that you intend to split version from publish. I see this as another indication that versioning is git business and publishing is npm business. You also mentioned that lerna version would likely be called by lerna publish. So an approach could be to allow a --skip-version.

My workaround is not pretty, and perhaps illustrates why I should roll my own in this particular case. It's just that I found lerna useful to manage my mono-repo, but I had odd requirements during publishing. I appreciate that you cannot accommodate every peculiar use case.

testower:
Although I'm finding this odd:

> 2. If you “know the version”, what’s stopping you from looping through the packages and running `npm publish` yourself?

Given this comment I'm curious about why there is a --repo-version argument:

```
When run with this flag, publish will skip the version selection prompt and use the specified version. Useful for bypassing the user input prompt if you already know which version to publish.
```

testower:
For anyone interested I ended up doing this:

```
// CI has already bumped version in root package.json,
// so first we replicate that in all the packages with lerna exec
lerna exec -- npm version $(node -e \"console.log(require('./package.json').version)\")

// Then run npm publish in all the packages
lerna exec -- npm publish
```

Issue #1292, opened 2018-02-28
Lerna postversion lifecycle hook executed twice


mikestead:
I have top level `publish` lifecycle scripts for `version` and `postversion`, however `postversion` is being executed immediately after `version` I assume by yarn/npm's [run-script](https://docs.npmjs.com/misc/scripts) naming convention, and then again by lerna.

## Expected Behavior
I'd expect `postversion` to only be run after lerna has tagged the version.

## Current Behavior
`postversion` is executed twice, the first time before lerna has tagged the version.

## Possible Solution
I'm not really sure. If it is the naming convention triggering it then it may only be possible with a different one.

## Steps to Reproduce (for bugs)
1. create a simple lerna project with a package
2. add a `version` script and `postversion` script to the top level `package.json`
3. in `postversion` exit with code `1`
4. publish a package and see how `postversion` is executed twice, causing `version` to fail too

[Example project](https://github.com/mikestead/lerna-lifecycle-issue)

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.8.0",
  "packages": [
    "packages/*"
  ],
  "version": "1.0.0",
  "npmClient": "yarn",
  "useWorkspaces": true,
  "registry": "https://registry.npmjs.org/"
}
```
</p></details>

<details><summary>package.json</summary><p>

```json
{
  "private": true,
  "workspaces": [
    "packages/*"
  ],
  "scripts": {
    "release": "lerna publish --yes --loglevel silly --skip-npm",
    "version": "echo \"version!\"",
    "postversion": "echo \"postversion!\" && exit 1"
  },
  "devDependencies": {
    "lerna": "^3.0.0-alpha.1"
  }
}
```
</p></details>

<details><summary>Log output</summary><p>

```txt
lerna info auto-confirmed
lerna sill addFiles [ 'lerna-lifecycle-issue/lerna.json' ]
lerna sill npmRunScript version [ '--silent' ] lerna-lifecycle-issue
lerna sill getExecOpts { cwd: 'lerna-lifecycle-issue' }
lerna ERR! publish error running version in undefined
lerna ERR! publish  Error: Command failed: yarn run version --silent
lerna ERR! publish error Command failed with exit code 1.
lerna ERR! publish $ echo "version!" --silent
lerna ERR! publish version! --silent
lerna ERR! publish $ echo "postversion!" && exit 1
lerna ERR! publish postversion!
[...]

PR #1302, opened 2018-03-03
Use npm-lifecycle to run solitary lifecycle phases

evocateur:
## Description

Instead of shelling out to `npm run`, use [npm-lifecycle](https://www.npmjs.com/package/npm-lifecycle) and [npm-conf](https://www.npmjs.com/package/npm-conf) to run a single lifecycle phase instead of `npm run`'s automatically expanded list.

## Motivation and Context

#1292 pointed out that the `postversion` hook was being executed twice because we were calling `npm run version` _and_ `npm run postversion` for a given package (and the root) during `lerna publish`. This is obviously sub-optimal.

## How Has This Been Tested?

A bunch of new tests, as well as improvements to the existing tests in this area.

## Types of changes

## Checklist:


wtgtybhertgeghgtwtg:
Is this worth the native dependency?

evocateur:
It does exactly what we want the same way npm itself does it, which is a huge win. What native dependency are you referring to?

wtgtybhertgeghgtwtg:
I worded that really poorly and misunderstood the use.
It [depends on `node-gyp`](https://github.com/npm/lifecycle/blob/latest/package.json#L34) and then spawns a process that may or may not use that.  But since it doesn't seem to use it itself and just provides it like NPM does, I don't think it'll be a concern outside of really rare edge cases.
Sorry.

evocateur:
No worries, thanks for clarifying!

PR #1310, opened 2018-03-07
Add gitHead property to package.json before publish

evocateur:
## Description

This mimics [read-package-json](https://github.com/npm/read-package-json/blob/b5a9f47c437d9b01d4e86372a3fc1bd04a89313f/read-json.js#L327)'s annotation of the current git `HEAD` during `npm publish`.

## Motivation and Context

This helps users and automated processes understand "when" a given version was published in terms of a raw SHA1. Sometimes the tags get nuked, mutated, or otherwise perverted. The SHA of a git revison (theoretically) is simpler to back-track.

## How Has This Been Tested?

Updated tests.

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/__tests__/__fixtures__/relative-file-specs/packages/package-1/package.json
+++ b/commands/publish/__tests__/__fixtures__/relative-file-specs/packages/package-1/package.json
@@ -1,6 +1,7 @@
 {
   "name": "package-1",
   "version": "1.0.0",
+  "changed": false,
   "dependencies": {
     "tiny-tarball": "^1.0.0"
   }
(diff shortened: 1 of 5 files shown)


KingScooty:
I've noticed when running `lerna publish --skip-npm` the gitHead entries get added to the package.json but don't get added to the version commit. So they're left unstaged, away from the rest of the publish. Is this intentional?

evocateur:
@KingScooty Whoops, no, the `gitHead` is only meant to be present _during_ `npm publish`, and not part of the git commit. `gitHead` is literally the revision just created, so it's kinda redundant, and only used for registry metadata. The working tree is always intended to be "clean" after `lerna publish`.

Thanks for the bug report! It looks like a pretty easy fix, just gotta move the `annotateGitHead()` call inside the right conditional.

Of course, the _deeper_ issue is that we're blocking certain git operations when we "skip npm", which isn't ideal...

evocateur:
Fixed in [v3.0.0-beta.9](https://github.com/lerna/lerna/releases/tag/v3.0.0-beta.9) 🎉

PR #1317, opened 2018-03-11
feat: Check for upstream changes before attempting to publish

### LINKED ISSUE #1177: Lerna Publish should first check if publishing is possible
Currently, Lerna publish starts publishing to npm and git without checking for potential git problems like git remote branch being ahead.

This can lead to consistency issues. 

In our scenario, we run a CI build on each push to master that calls: `lerna publish --exact --cd-version minor --yes`. If we merge 2 PRs quickly the CI build is stuck. What happens:
* After we merge 2 PRs quickly -> publish is called for the first build -> npm packages are published successfully with the new version but the git push fails.
* On the next build -> publish is called -> it uses the same version as the previous build because git push failed last time -> npm publishing fails because the same version already exists.

## Expected Behavior
npm packages should not be published if git remote branch is ahead.

## Current Behavior
npm packages are published even though git push is impossible.

## Possible Solution
Check if git push is possible before npm publish. Consider a flag if the build should fail.

```json
{
  "lerna": "2.5.1",
  "packages": [
    "packages/*"
  ],
  "version": "1.17.0",
  "npmClient": "yarn",
  "useWorkspaces": true
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.5.1 |
| `npm --version`  | 3.10.10 |
| `yarn --version` | 1.3.2 |
| `node --version` | v6.11.5 |

jezzay:
## Description

Check the git remote upstream for changes before attempting to publish. Exit with an error if upstream changes are detected. 

## Motivation and Context

Partly fixes #1177, in the case that there are remote changes at the time when `lerna publish` is run.

It is still possible that during the publish to npm, changes are added to the remote; which would make the git push fail. 

## How Has This Been Tested?

- unit tests
- manually publishing locally with `lerna publish --skip-npm`

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -84,6 +84,18 @@ class PublishCommand extends Command {
           `
         );
       }
+
+      if (GitUtilities.isBehindUpstream(this.gitRemote, this.execOpts)) {
+        const remote = `${this.gitRemote}/${currentBranch} `;
+        throw new ValidationError(
+          "EBEHIND",
+          dedent`
+            Local branch '${currentBranch}' is behind remote upstream ${remote}
+            Please merge in remote changes into '${currentBranch}'. 
+            Remote changes can be fetched and merged with git pull 
+          `
+        );
+      }
     }
 
     this.conf = npmConf(this.options);
--- a/core/git-utils/index.js
+++ b/core/git-utils/index.js
@@ -209,6 +209,42 @@ function hasCommit(opts) {
   return retVal;
 }
 
+function fetchGitRemote(opts) {
+  log.silly("fetchGitRemote");
+  ChildProcessUtilities.execSync("git", ["remote", "update"], opts);
+}
+
+function isBehindUpstream(gitRemote, opts) {
+  log.silly("isBehindUpstream");
+  exports.fetchGitRemote(opts);
+
+  const status = exports.aheadBehindCount(gitRemote, opts);
+  const behind = status.behind >= 1;
+
+  log.verbose("isBehindUpstream", behind);
+  return behind;
+}
+
+function aheadBehindCount(gitRemote, opts) {
+  const branchName = exports.getCurrentBranch(opts);
+  const branchComparator = `${gitRemote}/${branchName}...${branchName}`;
+  const rawAheadBehind = ChildProcessUtilities.execSync(
+    "git",
+    ["rev-list", "--left-right", "--count", branchComparator],
+    opts
+  );
+
+  const aheadBehind = rawAheadBehind.split("\t");
+  const behind = parseInt(aheadBehind[0], 10);
+  const ahead = parseInt(aheadBehind[1], 10);
+
+  log.silly(
+    "aheadBehindCount",
+    `behind ${gitRemote}/${branchName} by ${behind} commits; ${branchName} is ahead by ${ahead} commits`
+  );
+  return { ahead, behind };
+}
+
 exports.isDetachedHead = isDetachedHead;
 exports.isInitialized = isInitialized;
 exports.addFiles = addFiles;
@@ -229,3 +265,6 @@ exports.getShortSHA = getShortSHA;
 exports.checkoutChanges = checkoutChanges;
 exports.init = init;
 exports.hasCommit = hasCommit;
+exports.fetchGitRemote = fetchGitRemote;
+exports.isBehindUpstream = isBehindUpstream;
+exports.aheadBehindCount = aheadBehindCount;
--- a/integration/lerna-publish.test.js
+++ b/integration/lerna-publish.test.js
@@ -8,6 +8,7 @@ const writeJsonFile = require("write-json-file");
 const loadJsonFile = require("load-json-file");
 const path = require("path");
 const os = require("os");
+const tempy = require("tempy");
 
 const cliRunner = require("@lerna-test/cli-runner");
 const gitAdd = require("@lerna-test/git-add");
@@ -168,5 +169,19 @@ describe("lerna publish", () => {
 
     const { stdout } = await cliRunner(cwd)(...args);
     expect(normalizeTestRoot(cwd)(stdout)).toMatchSnapshot();
+  });
+
+  test("exits with EBEHIND when behind upstream remote", async () => {
+    const { cwd, repository } = await cloneFixture("normal");
+    const cloneDir = tempy.directory();
+
+    // simulate upstream change from another clone
+    await execa("git", ["clone", repository, cloneDir]);
+    await fs.outputFile(path.join(cloneDir, "README.md"), "upstream change");
+    await gitAdd(cloneDir, "-A");
+    await gitCommit(cloneDir, "upstream change");
+    await execa("git", ["push", "origin", "master"], { cwd: cloneDir });
+
+    await expect(cliRunner(cwd)("publish")).rejects.toThrowError(/EBEHIND/);
   });
 });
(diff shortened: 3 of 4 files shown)


commands/publish/index.js
```diff
@@ -84,6 +84,16 @@ class PublishCommand extends Command {
           `
         );
       }
+
+      GitUtilities.fetchGitRemote(this.execOpts);
+      if (GitUtilities.isBehindUpstream(this.gitRemote, this.execOpts)) {
```
jezzay: I'm wondering if there should be an option to either opt-in / opt-out of this check?

jezzay: On second thoughts, checking `gitEnabled && !skipNpm`  should be sufficient, as if you are publishing to npm, it is expected that there is a git remote to push tags to.

core/git-utils/index.js
```diff
@@ -209,6 +209,41 @@ function hasCommit(opts) {
   return retVal;
 }
 
+function fetchGitRemote(opts) {
+  log.silly("fetchGitRemote");
+  ChildProcessUtilities.execSync("git", ["remote", "update"], opts);
```
evocateur: Hah, 10+ years using git, and I've literally never used this `git remote` subcommand. Sweet!

core/git-utils/index.js
```diff
@@ -209,6 +209,41 @@ function hasCommit(opts) {
...
+function aheadBehindCount(gitRemote, opts) {
+  const branchName = exports.getCurrentBranch(opts);
+  const branchComparator = `${gitRemote}/${branchName}...${branchName}`;
+  const rawAheadBehind = ChildProcessUtilities.execSync(
+    "git",
+    ["rev-list", "--left-right", "--count", branchComparator],
+    opts
+  );
+
+  const aheadBehind = rawAheadBehind.split("");
```
evocateur: The `--count` flag always separates the values with a tab, we should split on `"\t"` here (`execa` already trims the `stdout` result). Unless you're accounting for other cases I'm not seeing?

core/git-utils/index.js
```diff
@@ -209,6 +209,41 @@ function hasCommit(opts) {
...
+  const behind = parseInt(aheadBehind[0], 0);
```
evocateur: `parseInt()` should be called with radix `10` here, not `0`.

core/git-utils/index.js
```diff
@@ -209,6 +209,41 @@ function hasCommit(opts) {
...
+  log.verbose("isBehindUpstream", behind);
+  return behind;
+}
+
+function aheadBehindCount(gitRemote, opts) {
+  const branchName = exports.getCurrentBranch(opts);
+  const branchComparator = `${gitRemote}/${branchName}...${branchName}`;
+  const rawAheadBehind = ChildProcessUtilities.execSync(
+    "git",
+    ["rev-list", "--left-right", "--count", branchComparator],
```
evocateur: I'm learning all sorts of `git` tricks today, brilliant!

commands/publish/index.js
```diff
@@ -84,6 +84,18 @@ class PublishCommand extends Command {
           `
         );
       }
+
+      if (!this.options.skipNpm) {
+        GitUtilities.fetchGitRemote(this.execOpts);
```
evocateur: I think we could simply call the fetch from `isBehindUpstream()`? I don't see much value in separating it, kinda strikes me as an implementation detail of `isBehindUpstream()`.

commands/publish/index.js
evocateur: At the same time, publishing to npm and pushing to git are (intended to be) separate concerns (even _if_ the implementation is currently kinda muddled). If the local checkout is behind upstream when making git commits and tags and pushing is enabled, it should always error, as that would inevitably error _anyway_ once the push to remote was attempted later in the process.

That is, _regardless_ of whether publishing to npm is enabled, if `this.gitEnabled === true` it is always an error if the upstream is ahead when publishing.

commands/publish/index.js
```diff
@@ -84,6 +84,18 @@ class PublishCommand extends Command {
...
+        if (GitUtilities.isBehindUpstream(this.gitRemote, this.execOpts)) {
+          const remote = `${this.gitRemote}/${currentBranch} `;
+          this.logger.warn(dedent`
```
evocateur: This should be throwing a `new ValidationError("EBEHIND", msg)` with this message so it sets the `process.exitCode` to `1`, indicating an error status. A `return false` keeps the `process.exitCode` as the default `0`, which in unix-land means "no errors, everything is okay".

We never want lerna attempting to push a non-fast-forward commit/tag, after all. Instead of "Skipping publishing", maybe indicate what command to run to fix the problem? (Usually as simple as `git pull`, in my experience)

integration/lerna-publish.test.js
```diff
@@ -56,7 +56,7 @@ describe("lerna publish", () => {
 
   test("exit 0 when no updates", async () => {
     const cwd = await initFixture("normal");
-    const args = ["publish"];
+    const args = ["publish", "--skip-npm"];
```
evocateur: Clearly we need to improve our git setup during integration tests. In fact, I think a bare repository initialized `beforeAll` and then cloned into each `initFixture` temp directory is the way to go. I'll work on implementing that today.

evocateur: To this end, I've added a new `cloneFixture()` helper and an example of usage in https://github.com/lerna/lerna/commit/efbd77beac113e8e6436ecf68494df61ba02b7fc:

```js
const { cwd, repository } = cloneFixture("normal");
const cloneDir = tempy.directory();

// simulate upstream change from another clone
await execa("git", ["clone", repository, cloneDir]);
await fs.outputFile(path.join(cloneDir, "README.md"), "upstream change");
await gitAdd(cloneDir, "-A");
await gitCommit(cloneDir, "upstream change");
await execa("git", ["push", "origin", "master"], { cwd: cloneDir });

try {
  await cliRunner(cwd)("publish");
} catch (err) {
  expect(err.message).toMatch("upstream ahead");
}
```

commands/publish/index.js
jezzay: Good point, I exited with `0` so that in a CI build scenario, the build would still be marked as successful if the publish step was aborted; but exiting with an error (and thus failing a build) is probably more accurate.

integration/lerna-publish.test.js
jezzay: Awesome! I'll add an integration test for this

jezzay:
@evocateur I've updated the PR with the changes, can you re-review? Thanks!

Vistur:
>It is still possible that during the publish to npm, changes are added to the remote; which would make the git push fail.

Maybe I'm saying something stupid but, would `git pull` right before [GitUtilities.pushWithTags](https://github.com/lerna/lerna/blob/master/commands/publish/index.js#L227) solve the issue? Maybe there is a reason that I can't think of not to do that.
I could open a PR with the changes if approved.

evocateur:
A pull at that moment has the potential to leave the tree in a severely conflicted state, and in situations (CI) where the resolution is not possible or straightforward.

Vistur:
Not being able to push also leaves the remote in a conflicted state with npm registry though 😢. Not sure what would be a good solution to the problem.

Thank you so much for the super fast reply btw!
Love your job ❤️

evocateur:
> Not being able to push also leaves the remote in a conflicted state with npm registry

This is a fundamental problem with making commits in CI (which I generally dislike for this and other reasons). This is why `semantic-release` only adds tags, for example. We can probably come up with a better solution for this sort of thing.

PR #1343, opened 2018-03-22
feat: set LERNA_ROOT_PATH when invoking npm scripts

bajtos:
## Description

Previously, the environment variable LERNA_ROOT_PATH was set only for `lerna exec` command (see #873). This patch modifies the way how `npm` scripts are executed to set LERNA_ROOT_PATH too.

## Motivation and Context

In LoopBack (see https://github.com/strongloop/loopback-next), we are using Lerna together with Typescript. One issue we are facing that TypeScript compilation errors report paths relative to current working directory, which is a per-package directory. This makes it difficult for tooling like VisualStudio Code to correctly parse errors and convert them to clickable/navigate-able entries in Problems window.

We would like to modify our build script to report paths relative to monorepo root (to fix VS Code integration), but only when the per-package build script is invoked via Lerna (see https://github.com/strongloop/loopback-next/issues/1010) - to keep individual per-package scripts still useful when run on their own via `npm`.

As I was reading Lerna documentation, I discovered LERNA_ROOT_PATH. This env variable is perfect for our needs, except the fact that it's not set by `lerna run` command. And so this patch was born.

## How Has This Been Tested?

I added a new test case to the integration tests to verify my new use case.

FWIW, `npm run lint` is failing in a node_modules folder for me, I think `.gitignore` file needs to be changed to ignore _all_ node_module directories, not only the top/monorepo-level one. I am leaving that change out of my pull request though, as it's not related.

## Types of changes

## Checklist:

### DIFF
--- a/integration/__snapshots__/lerna-bootstrap.test.js.snap
+++ b/integration/__snapshots__/lerna-bootstrap.test.js.snap
@@ -1,25 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`lerna bootstrap --npm-client yarn: stderr 1`] = `
-lerna info version __TEST_VERSION__
-lerna info Bootstrapping 4 packages
-lerna info lifecycle package-4@1.0.0~preinstall: package-4@1.0.0
-lerna info Installing external dependencies
-lerna info Symlinking packages and binaries
-lerna info lifecycle @integration/package-3@1.0.0~postinstall: @integration/package-3@1.0.0
-lerna info lifecycle @integration/package-2@1.0.0~prepublish: @integration/package-2@1.0.0
-lerna info lifecycle @integration/package-1@1.0.0~prepare: @integration/package-1@1.0.0
-lerna success Bootstrapped 4 packages
-`;
-
-exports[`lerna bootstrap --npm-client yarn: stdout 1`] = `
-package-1
-package-2
-cli package-2 OK
-package-3 cli1 OK
-package-3 cli2 package-2 OK
-`;
-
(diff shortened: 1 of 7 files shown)


integration/__snapshots__/lerna-bootstrap.test.js.snap
```diff
@@ -1,25 +1,5 @@
...
-`;
-
-exports[`lerna bootstrap --npm-client yarn: stdout 1`] = `
-package-1
-package-2
-cli package-2 OK
-package-3 cli1 OK
-package-3 cli2 package-2 OK
-`;
-
```
bajtos: I think this change is not related to my work, but it's needed to make `npm run ci` pass.

evocateur:
Why can't the scripts be run from the root?

I'm not a fan of this change, tbh. It's trivial to export the env var yourself when calling `lerna run`:

```
"scripts": {
  "lint": "LERNA_ROOT_PATH=`pwd` lerna run ts-lint"
}
```

bajtos:
We ended up writing a short wrapper to set LERNA_ROOT_PATH before executing `lerna` itself, see https://github.com/strongloop/loopback-next/blob/63c2f372ad5071a10828db0535dee38a23377694/bin/run-lerna.js#L25.

Since it looks like the project maintainers are opposed to my proposed changes, there is little value in keeping this pull request opened.

Having said that, I would still prefer to have `lerna` set `LERNA_ROOT_PATH`, so that we don't have to maintain our wrapper. If the maintainers ever change their opinion, then I am happy to update this pull request so that it can be landed.

PR #1348, opened 2018-03-26
feat: Execute atomic publish lifecycle during lerna publish

strax:
## Description
Changes Lerna to run the following [npm lifecycle scripts][1] when running `lerna publish`:
1. in root package:
    1. `prepublish`
    2. `prepare`
    3. `prepublishOnly`
2. sequentially in each subpackage:
    1. `prepublish`
    2. `prepare`
    3. `prepublishOnly`
3. parallel in each subpackage:
    1. (the npm package is published)
    2. `postpublish`
4. in root package:
    1. `postpublish`

[1]: https://docs.npmjs.com/misc/scripts#description

These scripts will only be run when the `--skip-npm` flag is **not** preset and after creating git commits, so `preversion`, `version` and `postversion` scripts will execute before these. This means that a failing `prepublish` will not prevent the creation of git commits or tags, but will prevent any of the packages from being published to npm.

## Motivation and Context
Lifecycle scripts are standard way in npm to execute functionality after certain events. It is convenient to be able to add a `prepublish` or `prepare` script to either the root package or a subpackage to for instance run the project's tests and compile source files for publishing. Lerna currently executes `npm publish` for each subpackage which in turns calls the scripts, but this can lead to a corrupted state where some of the packages have been successfully published and some have not. By running prepublish scripts before any of the packages is actually published, any error in the scripts will prevent any of the packages from being published.

## How Has This Been Tested?
Unit tested by augmenting the current lifecycle mock expectations in `publish-command.test.js`.

## Types of changes

## Checklist:

There are some tests in `master` which are already failing. I wanted to gather feedback before making changes to documentation.

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -583,6 +583,11 @@ class PublishCommand extends Command {
     // if we skip temp tags we should tag with the proper value immediately
     const distTag = this.options.tempTag ? "lerna-temp" : this.getDistTag();
 
+    const rootPkg = this.repository.package;
+    this.runPackageLifecycle(rootPkg, "prepublish");
+    this.runPackageLifecycle(rootPkg, "prepare");
+    this.runPackageLifecycle(rootPkg, "prepublishOnly");
+
     this.updates.forEach(({ pkg }) => {
       this.execScript(pkg, "prepublish");
       this.runPackageLifecycle(pkg, "prepublish");
@@ -604,9 +609,10 @@ class PublishCommand extends Command {
       });
     };
 
-    return pFinally(runParallelBatches(this.batchedPackages, this.concurrency, mapPackage), () =>
-      tracker.finish()
-    );
+    return pFinally(runParallelBatches(this.batchedPackages, this.concurrency, mapPackage), () => {
+      this.runPackageLifecycle(rootPkg, "postpublish");
+      return tracker.finish();
+    });
   }
 
   npmUpdateAsLatest() {
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -591,14 +591,14 @@ class PublishCommand extends Command {
 
     const rootPkg = this.project.package;
 
-    const promise = Promise.resolve()
-      .then(() => this.runPrepublishScripts(rootPkg))
-      .then(() =>
-        pMap(this.updates, ({ pkg }) => {
-          this.execScript(pkg, "prepublish");
-          return this.runPrepublishScripts(pkg);
-        })
-      );
+    let chain = this.runPrepublishScripts(rootPkg);
+
+    chain = chain.then(() =>
+      pMap(this.updates, ({ pkg }) => {
+        this.execScript(pkg, "prepublish");
+        return this.runPrepublishScripts(pkg);
+      })
+    );
 
     tracker.addWork(this.packagesToPublish.length);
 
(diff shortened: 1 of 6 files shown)


commands/publish/index.js
```diff
@@ -578,12 +578,29 @@ class PublishCommand extends Command {
     }
   }
 
+  runPrepublishScripts(pkg) {
+    return Promise.resolve(this.runPackageLifecycle(pkg, "prepublish"))
```
evocateur: I don't think we should support "legacy" `prepublish`, as it is deprecated.

commands/publish/index.js
```diff
@@ -578,12 +578,29 @@ class PublishCommand extends Command {
...
     const distTag = this.options.tempTag ? "lerna-temp" : this.getDistTag();
 
-    this.updates.forEach(({ pkg }) => this.execScript(pkg, "prepublish"));
+    const rootPkg = this.project.package;
+
+    const promise = Promise.resolve()
+      .then(() => this.runPrepublishScripts(rootPkg))
+      .then(() =>
+        Promise.all(
+          this.updates.map(({ pkg }) => {
```
evocateur: Instead of `Promise.all()`, this should be `pMap(this.updates, ({ pkg }) => {...})`. `this.updates` will likely be a `Set` soon, and we may need to tune the concurrency.

commands/publish/index.js
```diff
@@ -595,11 +612,16 @@ class PublishCommand extends Command {
         tracker.completeWork(1);
 
         this.execScript(pkg, "postpublish");
+        this.runPackageLifecycle(pkg, "postpublish");
```
evocateur: This should be returned (just like `this.runPrepublishScripts`) so we don't drop any rejections.

commands/publish/index.js
```diff
@@ -595,11 +612,16 @@ class PublishCommand extends Command {
...
       });
     };
 
-    return pFinally(runParallelBatches(this.batchedPackages, this.concurrency, mapPackage), () =>
-      tracker.finish()
+    return pFinally(
+      promise.then(() => runParallelBatches(this.batchedPackages, this.concurrency, mapPackage)),
+      () => {
+        this.runPackageLifecycle(rootPkg, "postpublish");
```
evocateur: Since this needs to be returned inside the Promise chain too, I'd prefer if the actions in this method were composed into a chain similar to the `execute()` method:

```js
let chain = this.runPrepublishScripts(rootPkg);

chain = chain.then(() =>
  pMap(this.updates, ({ pkg }) => {
    this.execScript(pkg, "prepublish");

    return this.runPrepublishScripts(pkg);
  })
);

tracker.addWork(this.packagesToPublish.length);

const mapPackage = pkg => {
  // ...
};

chain = chain.then(() => runParallelBatches(this.batchedPackages, this.concurrency, mapPackage));
chain = chain.then(() => this.runPackageLifecycle(rootPkg, "postpublish"));

return pFinally(chain, () => tracker.finish());
```

commands/publish/index.js
strax: Would that cause a failing `postpublish` to skip the remaining updates? My concern is that since `postpublish` executes after the package is published, it might lead to corrupted states where some of the packages have been published but some have not.

evocateur: Well, I mean, an uncaught rejection is going to halt then entire process in the next major version of Node, so I don't see how we avoid this, unless we wait to run all the `postpublish` lifecycles after _all_ the publishes have succeeded.

And even then, it's not guaranteed. I guess I'm willing to let a vanishingly-rare `postpublish` break the build if someone is foolish enough to write a `postpublish` script that exits non-zero? We've done the best we can, we can't account for everything.

evocateur:
Thanks!

commands/publish/index.js
evocateur: Oh, and also, we already catch and log lifecycle errors in [`runPackageLifecycle`](https://github.com/lerna/lerna/blob/45efa2404933e890e5561ac4c3e0dd0c4c08cb04/commands/publish/index.js#L426-L432), so it's basically just good hygiene to return it.

PR #1352, opened 2018-03-29
feat: Support --no-prefix for streaming output

raymondfeng:
## Description

By default, stream output is prefixed with the originating package name. This flag allows us to disable the prefix (`--no-prefix`) so that the output can be parsed by a tool like Visual Studio Code to highlight the result by type for all packages in the same project.

## Motivation and Context

We set up the monorepo with npm scripts so that we can run `tsc` and `mocha`. When the scripts are 
triggered inside Visual Studio Code, the stream prefix gets in the way and `tsc` problem matcher cannot
recognize the output.

## How Has This Been Tested?

A few test cases are added to the PR.

## Types of changes

## Checklist:

### DIFF
--- a/src/commands/ExecCommand.js
+++ b/src/commands/ExecCommand.js
@@ -27,6 +27,14 @@ export const builder = {
     type: "boolean",
     default: undefined,
   },
+  // This option controls prefix for stream output so that it can be disabled to be friendly
+  // to tools like Visual Studio Code to highlight the raw results
+  prefix: {
+    group: "Command Options:",
+    describe: "Enable prefix for stream output",
+    type: "boolean",
+    default: undefined,
+  },
   parallel: {
     group: "Command Options:",
     describe: "Run command in all packages with unlimited concurrency, streaming prefixed output",
@@ -44,6 +52,7 @@ export default class ExecCommand extends Command {
     return Object.assign({}, super.defaultOptions, {
       bail: true,
       parallel: false,
+      prefix: true,
     });
   }
 
@@ -107,7 +116,8 @@ export default class ExecCommand extends Command {
 
     async.parallel(
       this.filteredPackages.map(pkg => done => {
-        ChildProcessUtilities.spawnStreaming(this.command, this.args, this.getOpts(pkg), pkg.name, done);
+        const prefixStr = this.options.prefix ? pkg.name : "";
+        ChildProcessUtilities.spawnStreaming(this.command, this.args, this.getOpts(pkg), prefixStr, done);
       }),
       callback
     );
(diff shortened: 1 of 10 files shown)


src/commands/ExecCommand.js
```diff
@@ -27,6 +27,14 @@ export const builder = {
     type: "boolean",
     default: undefined,
   },
+  // This option disables prefix for stream output so that the log can be handled by
+  // tools like Visual Studio Code
+  "disable-stream-prefix": {
```
evocateur: Can we rename this `--prefix`, so the auto-negated `--no-prefix` works, as well as the lerna.json-based `"prefix": false` works? `--prefix` certainly has different meanings in other contexts, but it's pretty safe for `lerna exec` and `lerna run`.

You'll also need to add `prefix: true,` to the `get defaultOptions()` block in ExecCommand and RunCommand.

raymondfeng:
@evocateur Thank you for the prompt response. I have updated the PR to address your comments. PATL.

I also wonder how to port it to 3.x if it's accepted.

evocateur:
I think you need to run `yarn integration -u` so the tests pass?

I can port over to `master`, no worries.

src/commands/ExecCommand.js
```diff
@@ -27,6 +27,14 @@ export const builder = {
     type: "boolean",
     default: undefined,
   },
+  // This option controls prefix for stream output so that it can be disabled to be friendly
+  // to tools like Visual Studio Code to highlight the raw results
+  prefix: {
+    group: "Command Options:",
+    describe: "Enable prefix for stream output",
+    type: "boolean",
+    default: true,
```
evocateur: So the weird thing about our "durable" options and yargs is that we need to specify `default: undefined` here, otherwise yargs overwrites our lerna.json-based config. Sorry for the weirdness.

test/integration/__snapshots__/lerna-run.test.js.snap
```diff
@@ -41,13 +58,36 @@ lerna success - package-3
...
+`;
+
 exports[`stdout: test --stream 1`] = `
 "package-3: package-3
 package-4: package-4
 package-1: package-1
 package-2: package-2"
 `;
+
+exports[`stdout: test --stream 2`] = `
```
evocateur: I'm not sure where this snapshot is coming from? Or the `stderr: test --stream 2` addition.

raymondfeng:
@evocateur I fixed the issues. The CI should be green now.

evocateur:
Thanks!

evocateur:
Published in v2.9.1 and v3.0.0-beta.12

Issue #16, opened 2016-01-20
Error running `lerna updated`


doctyper:
Hi, I'm trying to run lerna on an [existing monorepo](https://github.com/nfl/react-wildcat). I've run `lerna bootstrap` without issues, but I'm seeing this error when running `lerna updated`:

```
Checking packages...
fatal: No annotated tags can describe 'af2c265fda2e7a93c4f6881b747e6f398d4e7c8d'.
However, there were unannotated tags: try --tags.
child_process.js:484
    throw err;
    ^

Error: Command failed: git describe af2c265fda2e7a93c4f6881b747e6f398d4e7c8d
fatal: No annotated tags can describe 'af2c265fda2e7a93c4f6881b747e6f398d4e7c8d'.
However, there were unannotated tags: try --tags.

    at checkExecSyncError (child_process.js:441:13)
    at Object.execSync (child_process.js:481:13)
    at execSync (/usr/local/lib/node_modules/lerna/lib/commands/updated.js:63:16)
    at Object.exports.checkUpdatedPackages (/usr/local/lib/node_modules/lerna/lib/commands/updated.js:30:23)
    at Object.exports.execute (/usr/local/lib/node_modules/lerna/lib/commands/updated.js:9:33)
    at Object.<anonymous> (/usr/local/lib/node_modules/lerna/bin/lerna.js:41:9)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
```

Any idea what could be wrong?

hzoo:
Not sure, I extracted out what `lerna publish` was doing. 

It looks like it is erroring on [line 30](https://github.com/kittens/lerna/blob/5a74fd3524f298fd1ac168a6813dca7801963a77/lib/commands/updated.js#L30). `git describe sha-number`

After reading [stack overflow](http://stackoverflow.com/questions/4154485/git-describe-ignores-a-tag), it looks like `git describe` only uses `annotated tags` by default and I guess none of the tags in the repo are annotated. 

It looks like we could potentially do `git describe --tags` and it would work. @kittens?

danielo515:
Is this considered to be fixed ? I'm facing the exact same issue on `2.0.0-rc.5`

nevir:
Ditto on 2.0.0

danielo515:
Well, here I am again, facing this problem at 2.0.0

[...]

PR #1355, opened 2018-03-30
fix(publish): Allow tag check to fail with strong warning

pablolmiranda:
## Description
When using the `lerna publish` command the `--skip-git` option is not taking in consideration.

In this case, even when using the `--skip-git` option, before the `npm publish`, lerna is trying to get and create the tag for the publishing.

Passing the `s=--skip-git` option should avoid the git interaction on this flow.

## Motivation and Context
Related to https://github.com/lerna/lerna/issues/16

## How Has This Been Tested?
 - Current state:
`lerna publish --skip-git` will check and create the tags before publishing the packages
- This PR:
`lerna publish --skip-git` will skip the tags check and creation and publish the modules to NPM

## Types of changes

## Checklist:

- [-] I have added tests to cover my changes.
- [-] All new and existing tests passed.

Some tests are not passing, but looks like it is not related to my changes.

I need some help writing a test to cover this flow.

### DIFF
--- a/core/git-utils/index.js
+++ b/core/git-utils/index.js
@@ -70,8 +70,13 @@ function addTag(tag, opts) {
 
 function hasTags(opts) {
   log.silly("hasTags");
+  let yes = false;
 
-  const yes = !!ChildProcessUtilities.execSync("git", ["tag"], opts);
+  try {
+    yes = !!ChildProcessUtilities.execSync("git", ["tag"], opts);
+  } catch (err) {
+    log.verbose("hasTags error", err);
+  }
   log.verbose("hasTags", yes);
 
   return yes;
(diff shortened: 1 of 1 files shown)


evocateur:
So, `--skip-git` is a terribly-named flag. It doesn't actually skip _all_ git operations, just making commits and pushing to the remote. git tags are still necessary to determine the version.

Because `lerna publish` is actually the conflation of `npm version` and `npm publish` (not my choice), we're kinda stuck with this. What are you trying to accomplish, specifically?

evocateur:
If you don't have any git tags reachable from that branch, that's a deeper problem. Presumably lerna has successfully published on `master` before?

(I don't recommend `lerna publish` on a branch, for this and many other reasons)

pablolmiranda:
I had a successful publish from master.

pablolmiranda:
@evocateur, you suggestion worked. Take a look and let me know if that was what you have in mind.

Issue #1324, opened 2018-03-14
Add support for `npm ci` to `lerna bootstrap`


casey-speer:
Npm 5.7 includes a new install command, `npm ci`, that installs dependencies directly by scanning the `package-lock.json` and foregoes any additional dependency resolution. This can be much much faster than running the usual `npm install`, saving time on build. Lerna should have an option to use this command instead of `npm install` when running `lerna bootstrap`.

## Expected Behavior
Able to use `npm ci` on request to install packages with `lerna bootstrap`.

## Current Behavior
Currently `bootstrap` uses only the `npm install` command to install packages.

## Possible Solution
Add a `--npm-ci` flag or similar to the `bootstrap` command that causes it to use `npm ci` instead of `npm install` when present. 

## Context
In a CI environment, we wish to have builds run as quickly as possible and `npm ci` is faster than `npm install`.

evocateur:
I would just skip `lerna bootstrap` completely. Use [relative `file:` specifiers](https://github.com/npm/npm/blob/latest/doc/spec/file-specifiers.md) between your dependencies, `npm i packages/*` in the root, and call `npm ci`* directly. Lerna itself does exactly this.

\* Caveat:
Until npm 5.8.0 becomes `latest`, there is a bug in `npm ci` that mangles local `file:` directory package.jsons. Installing either `npm@next` or `cipm` (and running it instead of `npm ci`) will workaround this problem for the time being.

casey-speer:
Hmmm maybe I'm missing something about `file:` specifiers, but wouldn't this mean that I have to always manually include the `file:` dependencies of a package X from my lerna repo in any host app that wants to use X?

evocateur:
Nope, the `file:` specifiers are overwritten with the current version range during `lerna publish`, so they're fully-functional npm packages that you treat like any other when consuming. (The versioned source code always preserves the `file:` specifiers)

Compare `@lerna/cli`'s dependencies:

[source](https://github.com/lerna/lerna/blob/1ee15326adf778c32e4b136bbfb27e1c899716a1/core/cli/package.json#L35-L53):
```
  "dependencies": {
    "@lerna/add": "file:../../commands/add",
    "@lerna/bootstrap": "file:../../commands/bootstrap",
    "@lerna/changed": "file:../../commands/changed",
    "@lerna/clean": "file:../../commands/clean",
    "@lerna/diff": "file:../../commands/diff",
    "@lerna/exec": "file:../../commands/exec",
[...]

PR #1360, opened 2018-04-04
feat(bootstrap): Use `npm ci` with `--ci` option

mndewitt:
This is an attempt at solving #1324 

## Description
This adds an argument to `lerna bootstrap`, `--npm-ci-mode`, which allows the user to install dependencies using `npm ci` instead of `npm install`.  Definitely open to commentary on what the exact naming should be.

## Motivation and Context
As a user I want to use npm's new install command, `npm ci`.  This provides better stability as to the exact packages developers are using locally and in our CI server.

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:


commands/bootstrap/index.js
```diff
@@ -34,7 +34,15 @@ class BootstrapCommand extends Command {
...
   initialize() {
-    const { registry, rejectCycles, npmClient = "npm", npmClientArgs, mutex, hoist } = this.options;
+    const {
+      registry,
+      rejectCycles,
+      npmClient = "npm",
+      npmClientArgs,
+      mutex,
+      hoist,
+      npmCiMode,
```
evocateur: I think we should take advantage of the existing (hidden) `--ci` flag, which is set automatically when `lerna` is called in a CI env (Travis, AppVeyor, etc). It would only take effect when `npmClient === "npm"`, of course, and would be passed to `npmInstall()` as `this.npmConfig.subCommand` (defaulting to `"install"`).

commands/bootstrap/index.js
```diff
@@ -66,6 +74,7 @@ class BootstrapCommand extends Command {
       npmClient,
       npmClientArgs,
       mutex,
+      npmCiMode,
```
evocateur: This should be removed, in favor of conditional assignment of `this.npmConfig.subCommand`:

```js
if (npmClient === "npm" && this.options.ci && hasNpmCI()) {
  this.npmConfig.subCommand = "ci";
}
```

I sketched out the implementation of `hasNpmCI()` [here](https://runkit.com/embed/auv7hpnsx7f4), where you can see test examples too.

`commands/bootstrap/lib/make-npm-feature-predicate.js`:
```js
"use strict";

const semver = require("semver");

module.exports = makeNpmFeaturePredicate;

function makeNpmFeaturePredicate(range) {
  return () => Boolean(semver.satisfies(getNpmVersion(), range));
}

function getNpmVersion() {
  const userAgent = process.env.npm_config_user_agent;

  return (/npm\/(\d+\.\d+\.\d+)/.exec(userAgent) || [""]).pop();
}
```

`commands/bootstrap/index.js`:
```js
const makeNpmFeaturePredicate = require("./lib/make-npm-feature-predicate");

// https://github.com/npm/npm/releases/tag/v5.7.0
const hasNpmCI = makeNpmFeaturePredicate(">=5.7.0");
```

utils/npm-install/npm-install.js
```diff
@@ -12,13 +12,20 @@ const getExecOpts = require("@lerna/get-npm-exec-opts");
 module.exports = npmInstall;
 module.exports.dependencies = npmInstallDependencies;
 
-function npmInstall(pkg, { registry, npmClient, npmClientArgs, npmGlobalStyle, mutex, stdio = "pipe" }) {
+function npmInstall(
+  pkg,
+  { registry, npmClient, npmClientArgs, npmGlobalStyle, mutex, stdio = "pipe", npmCiMode = false }
```
evocateur: This options block should include `subCommand = "install"` instead of `npmCiMode`

utils/npm-install/npm-install.js
```diff
@@ -12,13 +12,20 @@ const getExecOpts = require("@lerna/get-npm-exec-opts");
...
+) {
   // build command, arguments, and options
   const opts = getExecOpts(pkg, registry);
-  const args = ["install"];
+  const args = npmCiMode ? ["ci"] : ["install"];
```
evocateur:
```js
const args = [subCommand];
```

commands/bootstrap/__tests__/bootstrap-command.test.js
```diff
@@ -166,6 +167,16 @@ describe("BootstrapCommand", () => {
     });
   });
 
+  describe("with --npm-ci-mode", async () => {
+    const testDir = await initFixture("basic");
+
+    await lernaBootstrap(testDir)("--npm-ci-mode");
```
evocateur: You'll need to pass explicit `"--ci"` here, as the unit test command runner does not decorate the `ci` config like `core/cli` does.

It would also be nice to have a negative test against Yarn, as well as insufficient npm versions. Overwriting `process.env.npm_config_user_agent` with the values from the [example tests](https://runkit.com/embed/auv7hpnsx7f4) would be necessary.

mndewitt:
All makes sense, thanks for the clear feedback!  I will try to makes these updates in the next couple of days.

mndewitt:
@evocateur I think I captured all the requested changes.  One thing I was unsure on is the best practice for stubbing process.env in the tests - saving the original, mutating, and resetting to the original felt OK.

commands/bootstrap/__tests__/bootstrap-command.test.js
```diff
@@ -166,6 +166,44 @@ describe("BootstrapCommand", () => {
     });
   });
 
+  describe("with --ci", async () => {
+    const originalNpmUserAgent = process.env.npm_config_user_agent;
+
+    afterEach(() => {
+      process.env.npm_config_user_agent = originalNpmUserAgent;
```
evocateur: This is exactly the right approach (as far as my experience with jest goes)

commands/bootstrap/__tests__/bootstrap-command.test.js
```diff
@@ -166,6 +166,44 @@ describe("BootstrapCommand", () => {
...
+    });
+
+    it("should call npmInstall with ci subCommand if on npm 5.7.0 or later", async () => {
+      const testDir = await initFixture("basic");
```
evocateur: Due to the requirements of `npm ci`, this means we should indeed create another fixture with the correct `package-lock.json` files (the publish integration tests also use the fixtures used by the unit tests). I would just add one package, have it depend on `tiny-tarball` (simple, no transitive dependencies, very small package-lock), and that should be sufficient.

evocateur:
A "more correct" solution to fixing the integration tests is passing `"--no-ci"` to the existing bootstrap calls and add another test that explicitly passes `"--ci"`, with the appropriate fixture.

mndewitt:
Made a 'ci' fixture in the unit tests and cloned it into the integration tests - is that what you meant?

commands/bootstrap/__tests__/__fixtures__/ci/packages/package-1/package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "package-1",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "tiny-tarball": {
+      "version": "1.0.0",
+      "resolved": "http://artifactory.aws.athenahealth.com:80/api/npm/npm/tiny-tarball/-/tiny-tarball-1.0.0.tgz",
```
evocateur: This should be resolved to the public registry, `https://registry.npmjs.org/`

evocateur:
Yep, that's the pattern. I forgot bootstrap had its own integration fixtures, but no worries!

commands/bootstrap/__tests__/__fixtures__/ci/packages/package-1/package-lock.json
mndewitt: Oops :).  Nice catch, fixed...

PR #1362, opened 2018-04-06
Stop creating extra spaces in the changelog file

bertho-zero:
It's due to neested `indent` package usage.

cf: https://github.com/lerna/lerna/commit/6ff74aa7033ba3bac9606ceefc1dd9fbb3d75a59, https://github.com/lerna/lerna/commit/64916d650233f939d49d34ec52c691037f89106c

### DIFF
--- a/src/ConventionalCommitUtilities.js
+++ b/src/ConventionalCommitUtilities.js
@@ -1,3 +1,4 @@
+import os from 'os';
 import dedent from "dedent";
 import log from "npmlog";
 import path from "path";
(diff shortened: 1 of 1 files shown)


bertho-zero:
The tests that doesn't pass are due to the
```js
const BLANK_LINE = os.EOL + os.EOL;
```

Probably...

evocateur:
I really appreciate you picking this up, and I'm sorry the unit tests on the 2.x branch for this feature are so brittle (they're dramatically improved in `master`, but not easily ported due to many other changes).

Let's simplify the 2.x [tests](https://github.com/lerna/lerna/blob/2.x/test/ConventionalCommitUtilities.js) by replacing the brittle `dedent`-ed argument assertions (like [here](https://github.com/lerna/lerna/blob/2.x/test/ConventionalCommitUtilities.js#L88-L99)) with simple `expect.stringContaining("I should be placed in the CHANGELOG")`, and so on. We can just take as a given that the rest of the boilerplate is working (I could have sworn there was an integration test to that effect in the 2.x branch).

evocateur:
@bcoe Can you check why the tests are actually failing? Appveyor output is inconclusive.

PR #1363, opened 2018-04-06
Allow custom conventional-changelog config or preset

bertho-zero:
## Description

This PR allow you tu use a custom conventional-changelog config, like:
```json
{
  "lerna": "2.8.0",
  "packages": [
    "packages/*"
  ],
  "version": "independent",
  "command": {
    "publish": {
      "conventionalCommits": true,
      "changelogPreset": "@org/conventional-changelog-org"
    }
  }
}
```

## Motivation and Context

Before it was possible to choose only among the official configs:
```js
const allowedPresets = ["angular", "atom", "codemirror", "ember", "eslint", "express", "jquery", "jscs", "jshint"];
```
With this PR you can use any config or preset resolvable by `require.resolve` (from the monorepo root), for example:
- `@org/conventional-changelog-org`, that is a package in your monorepo (scoped or not).
- `conventional-changelog-custom`, that is a installed package at the root.
- `./config/conventional-changelog.js`, that is a relative path to the root.

## Types of changes

## Checklist:


src/ConventionalCommitUtilities.js
```diff
@@ -2,10 +2,24 @@ import dedent from "dedent";
 import log from "npmlog";
 import path from "path";
 import semver from "semver";
+import conventionalChangelogPresetLoader from "conventional-changelog-preset-loader";
+import resolveFrom from "resolve-from";
 
 import ChildProcessUtilities from "./ChildProcessUtilities";
 import FileSystemUtilities from "./FileSystemUtilities";
 
+const allowedPresets = [
```
evocateur: Not a huge fan of copying in these names, but we might as well make it a `Set` so the test in `changelogConfigPath` can avoid the array `.includes` with `allowedPresets.has(presetName)`

src/ConventionalCommitUtilities.js
```diff
@@ -134,7 +147,12 @@ export default class ConventionalCommitUtilities {
     return path.join(pkg.location, CHANGELOG_NAME);
   }
 
-  static changelogPreset(opts) {
-    return opts && opts.changelogPreset ? opts.changelogPreset : "angular";
+  static changelogConfigPath(pkg, opts) {
```
evocateur: Where does `pkg` get used in this method?

src/ConventionalCommitUtilities.js
```diff
@@ -134,7 +147,12 @@ export default class ConventionalCommitUtilities {
...
+    const presetName = opts && opts.changelogPreset ? opts.changelogPreset : "angular";
+    const changelogPreset = allowedPresets.includes(presetName)
+      ? conventionalChangelogPresetLoader.presetLoader(str => str)(presetName)
```
evocateur: The current integration tests seem to be choking on the result of this. Why can't you use the default export of `conventionalChangelogPresetLoader`?

evocateur:
Is there a particular reason you can't use `lerna@next`? I switched to using the updated `conventional-changelog` API directly, and it's much cleaner.

evocateur:
> I corrected the tests and replaced the Array with a Set, I wait to see if it goes (on Node 4 which could also be removed in 3 weeks)

lerna 2.x transpiles with Babel, so we shouldn't have to worry about `Set()`. I don't plan on removing Node v4 support from lerna 2.x.

> [...] modules not found if we launch the global lerna

#1122 (released in [v3.0.0-alpha.0](https://github.com/lerna/lerna/releases/tag/v3.0.0-alpha.0)) makes a global lerna always use the local install. It's expected that lerna is always locally-installed, since this is the general npm best-practice (and required by CI, etc). I'm not convinced we need to support a global changelog preset from a global invocation of lerna.

evocateur:
The [resolution](https://github.com/conventional-changelog/conventional-changelog/blob/c297dfb1c20f1faa11fa639db3117d7e96d157b4/packages/conventional-changelog-cli/cli.js#L133-L135) of the `conventional-changelog-cli` `-n` (`--config`) option is very interesting:

```js
config = require(path.resolve(process.cwd(), flags.config))
```

I had never investigated that option before. TIL...

evocateur:
Published as [`v2.10.0`](https://github.com/lerna/lerna/releases/tag/v2.10.0)

fenduru:
@evocateur this change is breaking me. I had previously submitted work upstream to conventional-changelog to support this exact situation, so it was already supported. However the API Lerna has chosen is different than the API that conventional-changelog chose.

Can we please revert this and add documentation that mentions that you can use scoped presets by passing `@scope/name` (which resolves to `@scope/conventional-changelog-name`)

evocateur:
At some point, this PR was actually using [conventional-changelog-preset-loader](https://www.npmjs.com/package/conventional-changelog-preset-loader), but I don't recall why it was replaced with `resolve-from`.

That being said, I don't have the bandwidth to maintain lerna 2.x, and `lerna@next` should work just fine, since it actually uses the new API directly.

fenduru:
@evocateur I understand the lack of bandwidth, however I do believe this is a breaking change and should have been released as a major or should be reverted. This is causing sudden breakage across many projects at my company.

evocateur:
@fenduru Fair enough, I'll revert it.

evocateur:
https://github.com/lerna/lerna/releases/tag/v2.10.1

PR #1376, opened 2018-04-14
fix(git-utils): Pass --no-verify to `git push`

### LINKED ISSUE #1366: lerna publish --no-verify option
## Expected Behavior
Running `lerna publish` publishes to npm and pushes changes to git without going through prepush validation scripts.

## Current Behavior
When you run `lerna publish` and it goes to push the changes using git, it will silently trigger any prepush scripts. In the case where those scripts a while, it looks like lerna is hanging.

## Possible Solution
Add a `--no-verify` flag to `lerna publish` that pushes to git using the `--no-verify` flag. That or show the logging of the prepush scripts.

hjylewis:
## Description
Adds the `--no-verify` flag to the git argument in `pushWithTags()`

Also, condenses the two `git push` calls into one call.

Closes #1366

## Motivation and Context
#1366

## How Has This Been Tested?
This function has tests to check that the tags were pushed.
@evocateur stated in #1366 that an integration test to check for the `--no-verify` change would be unnecessary.

## Types of changes

## Checklist:

### DIFF
--- a/core/git-utils/index.js
+++ b/core/git-utils/index.js
@@ -106,9 +106,7 @@ function pushWithTags(remote, tags, opts) {
   log.silly("pushWithTags", [remote, tags]);
 
   return Promise.resolve(exports.getCurrentBranch(opts)).then(branch =>
-    ChildProcessUtilities.exec("git", ["push", remote, branch], opts).then(() =>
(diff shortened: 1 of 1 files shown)


evocateur:
Thanks!

Issue #1385, opened 2018-04-20
lerna with CI/CD pipeline


bochen2014:
## introduction
We have a monorepo project comprised of 5 packages. We are using lerna to manage package versions, with independent strategy. Overall, apart from one particular big surprise, the experience with lerna is good. 

## our  pipeline setup
5 developers work on different features, once their work is finished, code is merged into master, which will trigger a build in Bamboo. The bamboo build server will build the code, bump the version of changed packages and publish them into a private npm registry.

## the problem
the problem happens when people merge their code into master while there is an active build in progress.
For example, lets say my project has `A@1.0.0`, `B@1.0.0` and `C@1.0.0` to begin with.  Here is what happened (in time order)
1. developer A merged his code into master which contains `chore(A)` and `feat(B)`. Bamboo picked it up and starts build#1;
2. while build#1 is under way, developer B merged his code which contains `feat(C)` and `chore(A)`. This triggers another build#2.  Now we have both build#1 and build#2 running at the same time;
3. build#1 bumped packages to `A@1.0.1` and `B@1.1.0`; and successfully published them into npm registry. But when it tries to `git push` all those  `package.json` files and `changelog.md`s back, it failed because it doesn't contains latest change(due to developer B 's merge). So build#1 failed with  `A@1.0.1` and `B@1.1.0`  published into npm registry. No commits or tags were pushed to origin.
4. build#2 bumped packages to `C@1.1.0` , `A@1.0.1` and `B@1.1.0` (note here that `B` will be picked up as well because build#1 didn't push the new tags. even if it did, build#2 may not be able to see it due to timing condition). It successfully published `C@1.1.0` but failed on publishing `A@1.0.1` because that package already exists. So build#2 also failed with `C@1.1.0` published. No commits or tags were pushed to origin. 

Now, imagine you sit in front a dev machine and do a `git pull`. All your local `package.json` files are still stating `A@1.0.0`, `B@1.0.0` and `C@1.0.0`, no new git tags introduced,  but all 3 npm packages have been published. What do you do?  😵 😵 😱  😱 

## possible solutions : disable concurrent builds in bamboo
you may think that by making bambo build sequential the problem may be solved. But it wont' help because build#1 still fails due to not having developer B 's merge. build#2 also will fail because build#1 didn't push its local commits&tags.  so disabling concurrent build literally has no effect at all. 

## your solution
I do have a solution but it's not ideal and I would like to hear from others before I share it here.

[...]

PR #1387, opened 2018-04-21
run npm publish after git push

bochen2014:
this PR is to address 'double publish' issue. i.e. while one developer commits a change and is running `lerna publish`, another developer commits a new change and starts running `lerna publish` as well. 
At the moment both `lerna publish` will fail but packages will be published, rendering the project into a chaotic state.  An realistic example is illustrated in https://github.com/lerna/lerna/issues/1385

To solve this issue, ideally, we should wrap `npm publish` and `git push with tags` into an transaction and make it atomic. By putting `git push` ahead of `npm publish`,  in the case where `lerna publish` is not running based on latest code,  `git push` will fail and `npm publish` won't be run,  and therefore atomicity is better reflected.

I believe this will have a huge benefit in production use of lerna

### DIFF
--- a/src/commands/PublishCommand.js
+++ b/src/commands/PublishCommand.js
@@ -273,6 +273,12 @@ export default class PublishCommand extends Command {
     if (this.options.skipNpm) {
       callback(null, true);
     } else {
+      if (this.gitEnabled) {
+        this.logger.info("git", "Pushing tags...");
+        this.logger.silly(`new tags: ${this.tags}`);
+        GitUtilities.pushWithTags(this.gitRemote, this.tags, this.execOpts);
+        this.logger.info("git", "git push with tags finished.");
+      }
       this.publishPackagesToNpm(callback);
     }
   }
@@ -298,11 +304,6 @@ export default class PublishCommand extends Command {
         if (updateError) {
           callback(updateError);
           return;
-        }
-
-        if (this.gitEnabled) {
-          this.logger.info("git", "Pushing tags...");
-          GitUtilities.pushWithTags(this.gitRemote, this.tags, this.execOpts);
         }
 
         const message = this.packagesToPublish.map(pkg => ` - ${pkg.name}@${pkg.version}`);


jezzay:
FYI @bochen2014 in lerna 3.x a check has been added to prevent publishing when behind the git remote:  https://github.com/lerna/lerna/blob/master/commands/publish/index.js#L101

This sounds like it solves your issue in CI?

evocateur:
Yeah, I don't want to fundamentally change the order of operations in a maintenance branch. Probably gonna switch lerna 3.x to be `latest` in npm soon, anyway.

bochen2014:
If you check before npm publish, there is still a chance that thing goes wrong. What if new commits come in after your check, but during you npm publish. I know npm publish is parallel,but it can still take a while and that give window for errors.

I still believe git push should happen before npm publish. Is there a reason why it’s not designed that way?

evocateur:
When an `npm publish` fails, now your public git history is at odds with what is present in the registry. Failed npm publish means you can go into the local directories and fix the problem. Once git is public, it's less tractable.

bochen2014:
You are talking about all npm publish fail. What if 3 passed but 2 failed. Once published, you can’t I publish them, git is not public,so next build would fail.

Also, what if new commits come in during nom publish? We have 45 packages in our project so I am a bit concerned

evocateur:
The answer is to only run one build at a time for branches that push commits back to origin. I don't know how else to contain the insanity, _even_ pushing before `npm publish` is still liable to hit race conditions in such a chaotic situation.

bochen2014:
1. fair enough.  I agree that the issue can be solved by having a git check latest plus one build at a time. The drawback is that we need extra configuration on build server, as by default it allow multi (4 by default I think).

2. I don't think put `git push` before `npm publish` has the same race condition. `git push` will make sure only the build that has latest code can proceed; all others get knocked back.  `git push` is atomic by itself, meaning even 2 build try to `git push` at the same time, only one will succeed. I can't see a reason why we shouldn't do `git push` first

bochen2014:
Git push makes sure only one

bochen2014:
sorry, accidently close this.. how can I re-open it?

bochen2014:
Sorry, actually one build a time won’t solve the issue. If new commits in during npm publish, both build will fail

bochen2014:
The first build will fail because it doesn’t have latest code(even it pass check before)
and therefor failed push

The second build would also fail because can’t publish the same npm package version again

bochen2014:
`semantic-release` push right after commit/tag; 
they recommend run `--post-success "npm publish" `

so, `git push` will happen before `npm publish`

```
>>> Your release branch is: master
>> No SemVer tag found. It seems like your first release? Initial release will be set to v1.0.0 as per npmjs specification.
>>> Bump type is minor
>>> update version on package.json...
v1.0.0
>>> About to write/append contents to CHANGELOG.md...
>>> About to add and commit package.json and CHANGELOG...
[master 1012a2d] chore(release): v1.0.0 [ci skip] ***NO_CI***
 2 files changed, 11 insertions(+), 1 deletion(-)
 create mode 100644 CHANGELOG.md
>> Time to create the Semantic Tag: v1.0.0
>>...and push to remote...
To github.com:bochen2014/semantic-release-test.git
   ac903c8..1012a2d  master -> master
To github.com:bochen2014/semantic-release-test.git
 * [new tag]         v1.0.0 -> v1.0.0
Done in 15.30s.
```

bochen2014:
think about this, what if there is a commit merged in ( just merge in, not build ), right after `lerna` doing this line
https://github.com/lerna/lerna/blob/master/commands/publish/index.js#L101

what will happen?
and we don't have control over when developer will merge in their code

bochen2014:
this problem is critical to us as it has happened twice so far (in 2 weeks time), the result is chaotic as we can't unpublished npm packages - I had to force publish twice, but the version info won't be liner. 

if this issue won't be fixed in lerna@3,  i'm afraid that we will have to host our own version of lerna.

evocateur:
We always have control over what commits are generated in CI. You can always do `lerna publish --canary` to avoid generating conflicting commits. Not ideal, I'll grant you, but it's what we have right now.

By the way, (semantic-release)[https://github.com/semantic-release/semantic-release/blob/218aa32f7933559cc577c43b95118cdad3354de1/lib/definitions/plugins.js#L46] seems to run its [npm plugin](https://github.com/semantic-release/npm/blob/18a4d64b6d7b9a264e106a46f262cd2806bbb613/index.js#L85) before its github plugin during the publish phase, so it's not clear to me that your assertion is correct?

bochen2014:
"We always have control over what commits are generated in CI."

I don't think we are talking about the same thing. 
I'm talking about in a team develop environment, where multiple developers can check in or merge into master at anytime. If  right after lerna runs this line
https://github.com/lerna/lerna/blob/2838260b7d0058249b2f0aee61e41d7de6bbe228/commands/publish/index.js#L101
another developer push some changes into `master`. what will happen?

I don't think this issue is related to  lerna's commit message

evocateur:
Lerna publish in CI _without_ `--canary` really blows right now, to be blunt. I’m saying spare yourself some pain and don’t use it right now, it’s clearly causing you way more pain than it is worth.

bochen2014:
Thanks for your advice, really appreciate it.

Are you talking about lerna 2.x or 3.x?

bochen2014:
She’ll I create a PR in 3.x swapping npm publish and git push? 

I think I have explained enough, and it’s worth putting git push before non publish

evocateur:
I’ve started a branch off master separating `lerna version` from `lerna publish`, which effectively runs `git push` before `npm publish`. It’s kind of a giant task, however, and I won’t likely be done until next week (at the earliest).

In the meantime, I can cut a 2.x prerelease under a special dist-tag that reverses it for you, if you like. I can’t make it a normal 2.x release due to semver, but at least it would stop the bleeding.

bochen2014:
thanks you so much,  really appreciate your help!

I have already bumped lerna to 3.x in our code base, as we recently noticed `lerna publish` in `2.x` returns success (exit code 0) even when it fails to push. This caused a CI build marked as passed, but really should be failed.  This is a defect that we can't afford in production. 

So, thanks for your offer but since we don't use lerna 2.x anymore, there is no need to create a new dist-tag.

evocateur:
Thanks again for your patience @bochen2014, but I don't think we're going to land this on 2.x, it's a breaking change that I can't afford to make on what is in reality a maintenance branch. Let me know if lerna 3 needs changes to make this happen (I'm still working on the split of publish into version and publish phases...).

Can you open a new issue about the incorrect exit code after a push failure, please? We should really figure that out, it's not ideal.

bochen2014:
thanks for your reply
1.  `git push` should happen before `npm publish` , i thought we have reached agreement on this? in `3.x` there is a `isBehindUpstream` check but that is insufficient because new commits could be pushed right after that check. So, yes , we do need this in 3.x as an important change
2.  in our repository setup, we disallow direct commit to master. Every commit into master much has a PR first, but we forgot to create an exception for CI. As a result, once CI bumps the versions and tries to push commits & tags back, it failed. *however* that build was marked as green. After this I immediately bumped to `3.x` and can confirm this has been fixed in `3.x`

evocateur:
Yes, we agreed that `git push` should happen before `npm publish`. That would be part of the new `lerna version` command I'm working on, extracted from the current `lerna publish`.

In the near future, calling `lerna publish` will automatically call `lerna version` when necessary, which effectively reorders `git push` before `npm publish`.

Issue #1386, opened 2018-04-20
lerna add --dev something  <> lerna add something --dev


kachkaev:
I noticed that placing `--dev` before package names fails the script and it took a while to investigate. 

```bash
## not in a mono-repo
yarn add something --dev ## works
yarn add --dev something ## works

## monorepo + lerna + yarn
lerna add something --dev ## works
lerna add --dev something ## ERROR: Missing list of packages to add to your project.
```

Should not `--dev` work as flags so that it does not swallow an argument? I'd also expect `-D` to work the same as `--dev` for consistency with npm and yarn.

## Error details

```
$ lerna add --dev something
lerna info version 2.10.2
lerna info versioning independent
lerna ERR! EINPUT Missing list of packages to add to your project.
lerna ERR! initialize callback with error
lerna ERR! ValidationError: Missing list of packages to add to your project.
lerna ERR!     at new ValidationError (/usr/local/lib/node_modules/lerna/lib/utils/ValidationError.js:25:120)
lerna ERR!     at AddCommand.initialize (/usr/local/lib/node_modules/lerna/lib/commands/AddCommand.js:110:19)
```

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.10.2 |
| `npm --version`  | 5.8.0 |
| `yarn --version` | 1.6.0 |
| `node --version` | 9.11.1 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Sierra | 10.13.4 |

bochen2014:
https://github.com/lerna/lerna/pull/1389
https://github.com/lerna/lerna/pull/1390

PR #1389, opened 2018-04-21
fix(add): Configure `--dev` as boolean option

bochen2014:
#1386 for lerna 2.x


evocateur:
Published in [v2.11.0](https://github.com/lerna/lerna/releases/tag/v2.11.0)

PR #1390, opened 2018-04-21
fix(add): Configure `--dev` as boolean option

### LINKED ISSUE #1386: lerna add --dev something  <> lerna add something --dev

bochen2014:
Fixes #1386

### DIFF
--- a/commands/add/__tests__/add-command.test.js
+++ b/commands/add/__tests__/add-command.test.js
@@ -87,7 +87,7 @@ describe("AddCommand", () => {
     expect(await readPkg(testDir, "packages/package-2")).toDependOn("@test/package-1", "~1");
   });
 
-  it("should reference to devDepdendencies", async () => {
+  it("should add target package to devDependencies", async () => {
     const testDir = await initFixture("basic");
 
     await lernaAdd(testDir)("@test/package-1", "--dev");
@@ -97,14 +97,12 @@ describe("AddCommand", () => {
     expect(await readPkg(testDir, "packages/package-4")).toDevDependOn("@test/package-1");
   });
 
-  it("should reference to devDepdendencies flexibly", async () => {
+  it("should add target package to devDependencies with alias", async () => {
     const testDir = await initFixture("basic");
 
-    await lernaAdd(testDir)("--dev", "@test/package-1");
+    await lernaAdd(testDir)("-D", "@test/package-1");
 
     expect(await readPkg(testDir, "packages/package-2")).toDevDependOn("@test/package-1");
-    expect(await readPkg(testDir, "packages/package-3")).toDevDependOn("@test/package-1");
-    expect(await readPkg(testDir, "packages/package-4")).toDevDependOn("@test/package-1");
   });
 
   it("should not reference packages to themeselves", async () => {
--- a/commands/add/__tests__/add-command.test.js
+++ b/commands/add/__tests__/add-command.test.js
@@ -97,6 +97,16 @@ describe("AddCommand", () => {
     expect(await readPkg(testDir, "packages/package-4")).toDevDependOn("@test/package-1");
   });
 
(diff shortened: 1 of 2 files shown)


bochen2014:
I don't think failed unit test is caused by my change, and it passed on node 6 env. 

can you please take a look? @evocateur 
```
FAIL  commands/publish/__tests__/publish-command.test.js (85.035s)
  normal mode
      ✕ should publish the changed packages (5008ms)
      ✕ throws an error when --independent is passed (1049ms)
```

commands/add/__tests__/add-command.test.js
```diff
@@ -53,20 +53,19 @@ describe("AddCommand", () => {
 
     await lernaAdd(testDir)("tiny-tarball");
 
-    expect(await readPkg(testDir, "packages/package-1")).toDependOn("tiny-tarball");
-    expect(await readPkg(testDir, "packages/package-2")).toDependOn("tiny-tarball");
-    expect(await readPkg(testDir, "packages/package-3")).toDependOn("tiny-tarball");
-    expect(await readPkg(testDir, "packages/package-4")).toDependOn("tiny-tarball");
+    [1, 2, 3, 4].forEach(async index => {
```
evocateur: Don’t rewrite these into a loop, please.

commands/add/__tests__/add-command.test.js
```diff
@@ -88,13 +87,29 @@ describe("AddCommand", () => {
   });
 
   it("should reference to devDepdendencies", async () => {
-    const testDir = await initFixture("basic");
```
evocateur: Separate test cases, please. One initFixture per test case, just like all the others.

After the existing initial dev test, you only need to make assertions on one package.

bochen2014:
PR updated per your instruction before. please take a look

doug-wade:
yo @bochen2014 I'm not seeing the test for `-D` -- did you forget to push?

bochen2014:
I used to have test covering `-D` flag and then Daniel mentioned something like 'one setup fixture in a test'  in his code review; After that I removed it.
The reason being to test `-D` we will have to have a new fixture setup, which means we will have a new test case.  I feel like this is an overkill. It isn't that important for us,  and I feel like we are testing `yargs`

doug-wade:
huh.  I must've misunderstood [this comment](https://github.com/lerna/lerna/pull/1390#discussion_r183456193)

evocateur:
Thanks for your patience, @bochen2014

Issue #1383, opened 2018-04-18
Conventional Commit with Fixed versioning


codeNoobie:
Do we support publish with conventional-commits for fixed versioning?

codeNoobie:
Okay. Then help me identify if below is a bug or an expected behavior.

I have two packages currently managed by lerna. I am using default versioning with conventional commit. I am ensuring that commit messages are in conventional commit format. After every commit, I run the below command.

`lerna publish --conventional-commits --changelog-preset angular-bitbucket --skip-npm --yes`

Let's say the initial versions are:
```
package-1 -> 0.0.0
package-2 -> 0.0.0
```

For any commit, that requires minor and patch version update according to conventional commits guidelines, Lerna correctly the versions of all packages even those which did not have any change since the last release. This is good.

But when i create a Breaking Change commit which only modifies one package, say package-1,  the above Lerna publish command only updates the major version of package-1. But, package-2's version is not modified. I would expect the major version of package-2 to be modified as well.

```
package-1 -> 1.0.0
package-2 -> 0.0.0
```

Now I create another patch commit involving only package-1. This time, all the package versions are now modified. Moreover, the major version of package 2 also gets updated.
It does successfully update the version of

```
package-1 -> 1.0.1
package-2 -> 1.0.1
```

Is this an expected behavior? If not, am i missing something with config or this is indeed a bug?

Here is my lerna.json:

```
{
  "lerna": "2.10.2",
  "packages": [
    "libs/*",
    "studio"
  ],
  "commands": {
    "publish": {
      "conventional-commits" : "true"
    }
  },
  "npmClient": "yarn",
  "useWorkspaces": true,
  "loglevel": "silent"
}
```

```
yarn -version
1.5.1

node -v
v9.8.0

lerna -v
2.10.2

OsX High sierra
```

evocateur:
This would have been useful information to include in the original bug report.

evocateur:
Why the `--skip-npm`? Probably not the cause, but definitely odd, if you do it after _every_ commit.

codeNoobie:
@evocateur  Yes, I agree I should have added the details with the original description.

I am in the process of setting up a new Angular 5 mono repo. The two lerna packages are component libraries. We are using https://github.com/dherges/ng-packagr to transpile and build the distribution in Angular Package format. This is an important use case for us. 

[...]

PR #1391, opened 2018-04-21
fix(publish): Include all packages during global major bump

bochen2014:
https://github.com/lerna/lerna/issues/1383


bochen2014:
unit test seems to be well written and is sophisticated. I have definitely learned something during fix old&create new test case. The issue is that due to 'symbolic link' check, we have to modify `process.platform` , and that feature has been disabled in later version of `jest`. 
What it means is that we limit `jest` to be `^19.0` while latest jest is already `22.4.3` 

do you have a plan for that? @evocateur

codeNoobie:
Thanks @bochen2014 for quickly coming up with a change.

src/UpdatedPackagesCollector.js
```diff
@@ -101,6 +103,10 @@ export default class UpdatedPackagesCollector {
         .forEach(registerUpdated);
     }
 
+    if (Object.keys(updatedPackages).length > 0 && !this.repository.isIndependent()) {
```
evocateur: This isn't sufficient to determine the correct case, which is actually only necessary for _major_ version bumps. The current test is also insufficient for this case.

test/PublishCommand.js
```diff
@@ -166,6 +166,30 @@ describe("PublishCommand", () => {
...
+      GitUtilities.getLastTag.mockReturnValue("v1.0.0");
+      GitUtilities.diffSinceIn.mockImplementation((since, location) => {
+        if (location.endsWith("package-3")) {
+          return "packages/package-3/newfile.json";
+        }
+        return "";
+      });
+    });
+    it("should publish all packages in Fixed Version mode", async () => {
+      await run(testDir)("--cd-version", "patch");
```
evocateur: This isn't at all what was discussed in the issue, patch or minor bumps should remain partial based on actual changes, as the local sibling dependencies are `^` ranges (by default).

bochen2014: You are right. We haven’t reached 1.0 yet, so in our case, we need cater for minor bump as well

evocateur: Well, fun fact, below 1.0.0 the minor versions are treated as majors (e.g., `0.3.0` does not satisfy `^0.2.0`). So it's a bit more complicated than that.

bochen2014:
Hi @evocateur , I spent a while doing this one as we badly need this in our project. 

Please take a look and see if I'm on the right direction. If you agree with what I'm doing, I'll create unit test for it. thanks!

utils/collect-updates/collect-updates.js
```diff
@@ -65,6 +65,14 @@ function collectUpdates({ filteredPackages, packageGraph, options, execOpts, log
     return Array.from(candidates);
   }
 
+  if (!project.isIndependent() && options.exact && candidates.size > 1) {
```
bochen2014: if you run `lerna publish --exact` on a fixed version project, all packages should bump together regardless whether they have changed or not

commands/publish/index.js
```diff
@@ -146,8 +146,21 @@ class PublishCommand extends Command {
 
     const tasks = [
       () => this.getVersionsForUpdates(),
-      versions => {
-        this.updatesVersions = versions;
+      ({ versions, hasBreakingChange }) => {
+        let updatedVersions;
+        if (!this.project.isIndependent() && hasBreakingChange) {
```
bochen2014: if one package has a breaking bump, all packages should bump together in a fixed version project

core/conventional-commits/index.js
```diff
@@ -119,12 +119,31 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath }) {
...
+        resolve({
+          currentVersion: pkg.version,
+          nextVersion: semver.inc(pkg.version, data.releaseType),
+        });
       });
     });
   });
 }
 
+function isBreakingChange(currentVersion, nextVersion) {
```
evocateur: This doesn't belong here.

core/conventional-commits/index.js
```diff
@@ -119,12 +119,31 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath }) {
...
+  const releaseType = semver.diff(currentVersion, nextVersion);
+  switch (releaseType) {
+    case "major":
+      return true;
+    case "minor":
+      return semver.lt(currentVersion, "1.0.0");
```
evocateur: This will literally never work once a project has moved past v1.0.0. Same with the patch.

utils/collect-updates/collect-updates.js
evocateur: No, that's incorrect. `--exact` only affects the dependency range used, not the presence/absence of changes.

commands/publish/index.js
```diff
@@ -298,10 +311,18 @@ class PublishCommand extends Command {
...
-      Promise.resolve(getVersion(pkg)).then(version => versionMap.set(pkg.name, version));
+    const iterator = ({ versions, hasBreakingChange }, { pkg }) =>
+      Promise.resolve(getVersion(pkg)).then(({ currentVersion, nextVersion }) => {
+        versions.set(pkg.name, nextVersion);
+        /* eslint-disable  no-param-reassign */
+        hasBreakingChange = ConventionalCommitUtilities.isBreakingChange(currentVersion, nextVersion)
+          ? true
+          : hasBreakingChange;
+        /* eslint-enable */
+        return { versions, hasBreakingChange };
```
evocateur: The breaking change detection should be a post-hoc loop, there's no reason to pass along `currentVersion` everywhere, it's still available on each `pkg.version`.

evocateur:
All of these test failures should be caught locally. `npm test`, like literally any other npm package.

core/conventional-commits/index.js
bochen2014: Semver.diff(1.1.0,2.0.0) returns major. Why does it not work?

evocateur: But this is `semver.lt()`, how is that relevant?

bochen2014:
I didn’t run unit test locally, as I’m not sure if you want this change or not. If this change is something you want, I will fix all unit test and creat new ones

core/conventional-commits/index.js
bochen2014: If current version is less than 1.0.0, then a minor bump should be treated the same as major one

evocateur:
There _is_ a problem, but I don't see any evidence that your solution fixes it appropriately. How do you even know it fixes it without passing tests?

bochen2014:
npm link

I have a project to test my solution

core/conventional-commits/index.js
bochen2014: I tried to find a place to put it, any suggestion?

bochen2014:
PR updated per our discussion before. please take a review thanks
(added 2 unit tests and updated 1 which was a major bump itself) @evocateur

bochen2014:
all test passed in travis (appVeyor failed due to connection problems). 

I guess what this PR means is we don't have to run `--force-publish` whenever  a major bump happens (actually, for us we need to run force publish for minor bumps as we haven't reached 1.0.0 yet);

commands/publish/__tests__/publish-command.test.js
```diff
@@ -115,6 +120,49 @@ describe("PublishCommand", () => {
...
+      collectUpdates.mockImplementationOnce(() => [{ pkg: package1, localDependencies: [] }]);
+
+      await lernaPublish(testDir)();
+
+      const patch = await showCommit(testDir);
+      expect(patch).toMatchSnapshot("commit");
+      expect(updatedPackageVersions(testDir)).toMatchSnapshot("updated packages");
+    });
+
+    it("major bump shoudl bump all packages", async () => {
```
doug-wade: should*

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
     const tasks = [
       () => this.getVersionsForUpdates(),
       versions => {
-        this.updatesVersions = versions;
+        const packages =
+          this.filteredPackages.length === this.packageGraph.size
+            ? this.packageGraph
+            : new Map(this.filteredPackages.map(({ name }) => [name, this.packageGraph.get(name)]));
+        if (this.project.isIndependent() || versions.size === packages.size) {
+          // independent, force-pulibsh=*, or carnary
```
doug-wade: canary* and publish*

bochen2014:
how come I just updated some comment literals and all test now fail?

bochen2014:
do we have any update on this one?

If you think the change is too significant and potentially confusing to users, I can add a new flag,say, `--major-bump-all`, or whatever you like; 
We need this change because we don't our `package1@2.0.0` depends on `package2@0.1.0`;

like @evocateur mentioned somewhere else, once a pre-release 'graduates', we want to bump all  prerelease versions to normal versions -- and `lerna` is doing that now;  I think a major bump should be treated as an official graduation as well

commands/publish/__tests__/publish-command.test.js
```diff
@@ -70,6 +74,7 @@ describe("PublishCommand", () => {
   // we've already tested these utilities elsewhere
   npmPublish.mockResolvedValue();
   runLifecycle.mockImplementation(() => Promise.resolve());
+  collectUpdates.mockImplementation(collectUpdatesActual);
```
evocateur: oh man, this is great! I never thought about that...

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
+        case "patch":
+          return semver.lt(currentVersion, "0.1.0");
+        case "premajor":
+        case "preminor":
+        case "prepatch":
+        case "prerelease":
+          return false;
+        default: {
+          const err = { error: "not expected release type" };
+          throw err;
```
evocateur: This should throw a new Error instance, but I'm not really sure how we get to this case anyway, because `semver.diff()` already throws when its arguments are invalid versions.

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
         )
       : [this.packagesToPublish];
 
+    const isBreakingChange = (currentVersion, nextVersion) => {
```
evocateur: This should be a function external to the class, there's no reason to inline it here.

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
+
     const tasks = [
       () => this.getVersionsForUpdates(),
       versions => {
-        this.updatesVersions = versions;
+        const packages =
+          this.filteredPackages.length === this.packageGraph.size
+            ? this.packageGraph
+            : new Map(this.filteredPackages.map(({ name }) => [name, this.packageGraph.get(name)]));
+        if (this.project.isIndependent() || versions.size === packages.size || this.options.canary) {
```
evocateur: `versions.size === this.packages.length` is more than sufficient, you don't have to recreate the packages map above.

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
+          // independent, force-publish=*, or canary, we don't bump all
+          this.updatesVersions = versions;
+        } else {
+          // fixed mode, note that versions are now all set to highest version;
+          let nextVersion;
```
evocateur: `this.globalVersion` already has this value, there's no need to collect it (it's already the next major, in a fixed version scenario)

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
+          let hasBreakingChange = false;
+          versions.forEach((version, pkgName) => {
```
evocateur: This could be clearer as a for-of loop:

```js
let hasBreakingChange;

for (const [name, bump] of versions) {
  hasBreakingChange = hasBreakingChange || isBreakingChange(this.packageGraph.get(name).version, bump);
}
```

commands/publish/index.js
```diff
@@ -144,10 +144,58 @@ class PublishCommand extends Command {
...
+            const originVer = this.packages.find(p => p.name === pkgName).version;
+            if (isBreakingChange(originVer, version)) {
+              hasBreakingChange = true;
+            }
+          });
+          if (hasBreakingChange) {
+            this.updates = Array.from(packages.values());
+            this.updatesVersions = new Map();
+            this.updates.forEach(pkg => {
+              this.updatesVersions.set(pkg.name, nextVersion);
```
evocateur: nextVersion should just be `this.globalVersion`

commands/publish/index.js
bochen2014: we discussed it in last code review, I basically can't find a place to host it so I inlined it. Any suggestions?

evocateur:
```function isBreakingChange(currentVersion, nextVersion) {
  ...

evocateur:
Sorry, I meant that you don’t need to precalculate a new packages map just to check if it’s the same size as the versions map. The other parts of that conditional are still appropriate, I’m just not sure why you’re eagerly re-mapping the packages.

evocateur:
Looks great, thanks!

codeNoobie:
Thanks @bochen2014 @evocateur @doug-wade. Will this change be back ported to a non beta version? Or is it advisable to use the latest version?

evocateur:
@codeNoobie I use 3.x "in production" every day, it's fine. The beta tag is just there because I'm lazy and still plan a couple _technically_ breaking changes before 3.0 final.

codeNoobie:
Thanks @evocateur. Yeah I was concerned because of the beta tag.

bochen2014:
if https://github.com/lerna/lerna/issues/277 is the holdout that prevents `3.x` from going official,  I think you can take a look at how *semantic-release* bumps their versions. 
In short, in my understanding,  `semantic-release`  reads current version & `gitHead`  from `npm registry` rather than `git` which is what `lerna` is currently doing (this is the **only** difference between `lerna` and `semantic-release` regarding version bump)

if you want to do incremental version bumps for canary releases, you can take `semantic-release` way. i.e. inject a serial number in canary version, and next time you do canary publish, you read that number from `npm`  (not local `git` ) and increment it;

I'm pretty sure you have already got something in mind, but I think it's still worth mentioning it and maybe having a  think about it.

PR #1406, opened 2018-05-02
fix(collect-updates): Remove redundant short-circuit

bochen2014:
per our discussion before.

### DIFF
--- a/utils/collect-updates/collect-updates.js
+++ b/utils/collect-updates/collect-updates.js
@@ -65,12 +65,6 @@ function collectUpdates({
   const dependents = collectDependents(candidates);
   dependents.forEach(node => candidates.add(node));
 
-  if (canary || packages.size === candidates.size) {
-    logger.verbose("updated", "(short-circuit)");
-
-    return Array.from(candidates);
-  }
-
   const updates = [];
 
   packages.forEach((node, name) => {


commands/changed/__tests__/__snapshots__/changed-command.test.js.snap
```diff
@@ -52,8 +52,8 @@ exports[`ChangedCommand circular should list changes with --force-publish * 1`]
 
 exports[`ChangedCommand circular should list changes with --force-publish package-2 1`] = `
 "- package-2
-- package-3
-- package-4"
+- package-4
```
evocateur: So this is probably the reason there's still a separate loop. This is incorrect output for `lerna changed`, which should always report packages with changes in lexical (alphabetical) order. We might as well remove the short-circuit early return, because a mere `forEach` over the `packages` Map is _not_ an expensive proposition.

utils/collect-updates/collect-updates.js
```diff
@@ -65,21 +65,5 @@ function collectUpdates({
...
-  packages.forEach((node, name) => {
-    if (candidates.has(node)) {
-      logger.verbose("updated", name);
-
-      updates.push(node);
-    }
-  });
-
-  return updates;
+  return Array.from(candidates);
```
evocateur: Indeed, simply remove the short-circuit entirely and no tests need change at all:
```patch
diff --git a/utils/collect-updates/collect-updates.js b/utils/collect-updates/collect-updates.js
index 300655f2..e90f2191 100644
--- a/utils/collect-updates/collect-updates.js
+++ b/utils/collect-updates/collect-updates.js
@@ -65,12 +65,7 @@ function collectUpdates({
   const dependents = collectDependents(candidates);
   dependents.forEach(node => candidates.add(node));
 
-  if (canary || packages.size === candidates.size) {
-    logger.verbose("updated", "(short-circuit)");
-
-    return Array.from(candidates);
-  }
-
+  // The result should always be in the same order as the input
   const updates = [];
 
   packages.forEach((node, name) => {
```

bochen2014:
Can we do `return Array.from(candidates).sort()`?

evocateur:
We'd need to write a custom comparator, since `PackageGraph` nodes don't sort lexically by themselves. Same difference from always looping through packages (removing the short-circuit), like I requested.

bochen2014:
Got it. Thanks

PR #1408, opened 2018-05-02
feat(unit-test): use docker to run unit test

bochen2014:
I've noticed that the unit test is highly dependent on your host machine settings, e.g.
1. if you have a global `.gitignore` file, unit test will fail (`packages/**`, `bin/**`  by default are git ignored but are checked in jest snapshot)
2. if you use `yarn`  to run the test, it will fail because of `package.json` default registry field issue;
3. if your git version is old, it will fail because of `git diff` format issue in snapshot;
4. if your npm version is old, it will fail because of https://github.com/npm/npm/issues/9388#issuecomment-385962805

In short, it is very picky on the environment it runs in, so it's better to use a docker container for that

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+from node:latest
+label name="lerna/test-runner"
+label maintainer="https://github.com/lerna/lerna"
+
+copy . /opt/lerna
+run apt remove -y git &&\
+    echo "deb http://ppa.launchpad.net/git-core/ppa/ubuntu trusty  main" >> /etc/apt/sources.list &&\
+    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys A1715D88E1DF1F24 &&\
+    apt update &&\
+    apt install -y git &&\
+    rm -rf /var/lib/apt/lists/*
+
+workdir /opt/lerna
+run npm install
+
+cmd ["npm","test"]
+    


bochen2014:
https://asciinema.org/a/4XpFDDheMhqr1WhdpFlzHJcNC

evocateur:
> `packages/**`, `bin/**` by default are git ignored

Why? If this is something `git` does _by default_, I wonder why I've literally never been bitten by this?

> if you use `yarn` to run the test

Don't use yarn. It's right there in [CONTRIBUTING.md](https://github.com/lerna/lerna/blob/master/CONTRIBUTING.md#run-unit-tests)

We already make pains to make testing suck less on Windows, as well as disable certain global git configurations. I think it's fair to ask contributors to read the contributing guidelines and have reasonably up-to-date tools. If there's something missing from `CONTRIBUTING.md`, by all means let's fix that documentation.

bochen2014:
I think it’s because you don’t have `bin/**` in your global .gitignore file.

I have been bitten since day one, only figured out yesterday why I can’t pass the tests locally

evocateur:
Why do you have `bin/**` in your _global_ `.gitignore`? Who recommends that as a best practice?

bochen2014:
What happened to me was like, first it failed on my local machine, I then ran them  within a nodejs docker container, and it failed there as well.

I then dogged into it and find the unit test is very picky on environment it runs in, I so I built this tailored image

bochen2014:
Here you go
https://github.com/github/gitignore/blob/master/VisualStudio.gitignore


It was there since,, 4 or 5 years ago when I was using visual studio to do .net work. 
Bin/ is the binary folder so I think it makes since that visual studio wants to ignore it be default

evocateur:
Well, I don't know if you've noticed, this isn't a VisualStudio-based project. It's an npm-based project, where `bin` is an _extremely common_ directory name.

PR #1422, opened 2018-05-11
feat(amend): new --amend flag for the publish command

krotscheck:
## Description
For the CI/CD use case of releasing-per-commit, adding a commit to update
the version number and/or the CHANGELOG, for every merged patch, creates
unnecessary noise. This patch allows us to amend the existing commit with any
new changes, without replacing the message or generating a new commit.

Specific changes:
- Added the --amend commandline flag to the publish command
- Add an 'amend' flag to the git-commit helper, which translates into
  `--amend --no-edit`.
- Added tests for all the above.
- Updated the README.md flag with the new option.

## Motivation and Context
Our project is both an SPA and a library, with the former following a strict every-commit-goes-to-production workflow. As a result, we effectively generate a version commit for every feature commit, which gets really messy.

Additionally, we want jenkins to manage our commits for us, only merging those which have been tested against the current tip. Since we already have direct control over our git repo in CI/CD, all we really need to do is amend the existing feature commit with the version bump and CHANGELOG. Hence this commit.

## How Has This Been Tested?
I extended the existing unit tests in the publish library with the appropriate use cases. Changes appear to be isolated to `commands/publish`

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x ] My code follows the code style of this project.
- [ x ] My change requires a change to the documentation.
- [ x ] I have updated the documentation accordingly.
- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.


krotscheck:
It looks like there are some experimental features in Node 10 that dump to `stderr`, causing the tests to break. I've updated both the Appveyor and travis configuration to set the appropriate environment parameter to disable these warnings.

doug-wade:
I looked over this one, and the code looks pretty good and the tests seem to cover all the cases, so thanks for that @krotscheck -- great stuff.  However, the feature itself seems a little terrifying, what with the amending and force-pushing; I don't feel comfortable pushing the merge button without a second opinion on this one.  Whaddya think @evocateur?

krotscheck:
Yeah, it's not for the faint of heart. A slightly safer approach would be to disable the git push in this case, i.e,

```
    if (this.gitEnabled && !this.amend) {
      tasks.push(() => this.pushToRemote());
    }
```

krotscheck:
@evocateur Have you managed to take a look at this? Any opinion vis-a-vis the existing implementation and/or the alternative I proposed?

evocateur:
@krotscheck Sorry, I've been slammed at work. I agree with Doug, we shouldn't be automatically force-pushing.

commands/publish/lib/git-commit.js
```diff
@@ -7,11 +7,13 @@ const childProcess = require("@lerna/child-process");
 
 module.exports = gitCommit;
 
-function gitCommit(message, opts) {
+function gitCommit(message, amend, opts) {
```
evocateur: I'm not a huge fan of this extra argument (that is _usually_ `false`). Can we move this to the `opts` object instead?

`opts` is just `this.execOpts` from the PublishCommand instance. In the constructor, when `this.options.amend === true`, we can set `this.execOpts.amend = true`. We want to avoid adding the execOpts prop when the option is falsey because otherwise the tests are a bear to update.

commands/publish/index.js
```diff
@@ -52,9 +52,15 @@ class PublishCommand extends Command {
       tempTag: false,
       yes: false,
       allowBranch: false,
+      amend: false,
     });
   }
 
+  configureProperties() {
+    super.configureProperties();
+    this.execOpts.amend = this.options.amend === true;
```
evocateur: We don't need to override `configureProperties()`, just move this line near the top of `initialize()`.

evocateur:
I don't know why the Appveyor build is failing, it doesn't make any sense. :/

krotscheck:
The LC_ALL flag isn't the correct one, I've updated it, let's see if it works.

evocateur:
Still trying to get the dumb CI to pass, but this PR has waited long enough. Thanks for your patience, @krotscheck!

PR #1441, opened 2018-05-28
fix(tests): Special handling for Windows CI issues

honzajavorek:
I've noticed the `master` build [is failing](https://travis-ci.org/lerna/lerna/jobs/377976519). I'm trying to fix the problem.

## Description
1. When ran on Node 10, there is a warning about `fs.promises` being experimental. This warning fails two tests, which rely on exact string matching. The warning is irrelevant in the testing context. I turned off the warnings per advice under this PR.
2. Several tests failed on Windows, because they could not find git binaries in PATH (specifically `git-receive-pack`). I found a working solution in the [go-git's AppVeyor config](https://github.com/src-d/go-git/blob/master/appveyor.yml).
3. The Windows builds were failing on `integration\lerna-exec.test.js`. It is a super weird issue with environment variables and to me it seems like a bug in Node or AppVeyor. It took me several days to debug. Details are in the commit message.

## Motivation and Context
I want to contribute to this project and I want to be sure whether I broke something or not.

## How Has This Been Tested?
:)

## Types of changes

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ install:
   - ps: Install-Product node $env:nodejs_version
   - npm config set loglevel warn
   - npm install --global npm@latest
-  - set PATH=%APPDATA%\npm;%PATH%
+  - set PATH=%APPDATA%\npm;"C:\Program Files\Git\mingw64\bin";%PATH%
   - npm ci
 
 before_test:
--- a/helpers/cli-runner/index.js
+++ b/helpers/cli-runner/index.js
@@ -12,5 +12,5 @@ function runner(cwd, env) {
     env: Object.assign({ CI: true }, env),
   };
 
-  return (...args) => execa(LERNA_BIN, args, opts);
+  return (...args) => execa("node", [LERNA_BIN].concat(args), opts);
 }
--- a/helpers/cli-runner/index.js
+++ b/helpers/cli-runner/index.js
@@ -9,7 +9,7 @@ module.exports = runner;
 function runner(cwd, env) {
   const opts = {
     cwd,
-    env: Object.assign({ CI: true }, env),
+    env: Object.assign({ CI: "true" }, env),
   };
 
   return (...args) => execa("node", [LERNA_BIN].concat(args), opts);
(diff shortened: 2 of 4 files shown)


honzajavorek:
Hmm, AppVeyor tests are failing on following:

```
Command failed: git push -u origin master
    sh: warning: setlocale: LC_ALL: cannot change locale (en-US): No such file or directory
    git-receive-pack '/C:/Users/appveyor/AppData/Local/Temp/1/7be0080e0bb67f1219b76709764c4dd7': git-receive-pack: command not found
```

krotscheck:
Heya- you can get rid of all the experimental warnings by setting `NODE_NO_WARNINGS=1` in the appveyor and travis files. That doesn't get you off the hook, there's still some odd test failures on windows.

evocateur:
The errors now are dumb timing issues, not logging issues, as far as I can tell.

honzajavorek:
@evocateur So I got into extremely weird situation. The problem I'm fighting is with the `integration\lerna-exec.test.js` tests, i.e. `lerna exec`. They don't pass on Windows, the `stdout` is empty and it fails with various errors (readable in `stderr` of the child process). The `cli-runner` helper used to create the child process in the tests looks like this:

```js
const execa = require("execa");
const LERNA_BIN = require.resolve("../../core/lerna/cli");
module.exports = runner;

function runner(cwd, env) {
  const opts = {
    cwd,
    env: Object.assign({ CI: true }, env),
  };
  return (...args) => execa("node", [LERNA_BIN].concat(args), opts);
}
```

When I modify it like this:

```diff
- env: Object.assign({ CI: true }, env),
+ env: Object.assign({ CI: true, foo: "bar" }, env),
```

...the tests start to magically pass on Windows. I was trying to figure out why, but to no avail so far. With `foo: "bar"` added to the object it passes even without 146b413ad169dc2209da704ac3989abe9311ffaf in place! I'm afraid I'm not skilled enough in JavaScript to decipher such magic. Do you have any ideas?

honzajavorek:
@evocateur @krotscheck I think this is ready now! 🎉 😅

evocateur:
Apologies for the long delay in review, I really appreciate Windows help!

honzajavorek:
Thanks! This is great. Now I’ll try to finish the other PR I have here. I
didnt see activity so I paused my efforts for a moment.

I usually try to be humble in comments of my work, but the Windows
debugging was a week of hell I went through, so I’m really proud I was able
to finish and fix it :))

Hopefully its gonna make contributing easier for everyone involved as the
builds are once again informative. Maybe existing open PRs should rebase on
top of the current master to see if they actually pass.

Issue #1213, opened 2018-01-18
Packages using lerna are not distributed including LICENSE


kylef:
Packages which are distributed with lerna are often do not include a LICENSE file, this makes consuming packages that use lerna tricky as the license file and copyright notice are not included in the NPM distribution. Tools such as [license-checker](https://github.com/davglass/license-checker) won't be able to find the license file and respective copyright notice and author.

NPM would normally include files such as the LICENSE file in the NPM distribution. When running `npm pack` or such the license will be included in the created tarball.

> Certain files are always included, regardless of settings:
> - README
> - CHANGES / CHANGELOG / HISTORY
> - LICENSE / LICENCE
> - NOTICE

_From https://docs.npmjs.com/files/package.json#files._

The common pattern I've seen in libraries using lerna is that the LICENSE file is found in the root of the repository. Since lerna will publish each package in the `package/` directory, the package inside repository doesn't contain a copy of the LICENSE.

You can see this in babel, the repository contains a [LICENSE](https://github.com/babel/babel/blob/master/LICENSE) file, and package such as babel-runtime on NPM (tarball: https://registry.npmjs.org/babel-runtime/-/babel-runtime-7.0.0-beta.2.tgz) will not have any license file.

I am wondering if there is an automated way that this problem can be solved and the project LICENSE can be present in the distributed NPM tarball. Perhaps in `lerna publish` the license could be injected into the package?

evocateur:
Would symlinking the root LICENSE file into the package roots work? (just checked with `npm pack`, nope 😞)

So yeah, either copy the LICENSE file into each root manually, or lerna can do it I guess. fstream-npm's [implementation](https://github.com/npm/fstream-npm/blob/b0a7b6fd9381689eeef9147ea4a1c71c250b177f/fstream-npm.js#L123) is probably "good enough" in terms of identifying if a root license file exists.

honzajavorek:
It would be great if Lerna automatically dealt with this problem as it propagates to many packages using Lerna. Otherwise we who care about licensing need to go to every single monorepo maintainer and to ask them to add a custom mechanism to copy the licenses into the individual packages, which is, let's say, a long run.

evocateur:
I don’t disagree, I just don’t have time right now.

honzajavorek:
@evocateur I think I can work on this. If you have any advice on what to do or where to start, let me know.

honzajavorek:
My proposal would be:

[...]

PR #1465, opened 2018-06-13
feat(publish): Ensure published packages contain a LICENSE file

honzajavorek:
This PR is my attempt to implement changes proposed in https://github.com/lerna/lerna/issues/1213#issuecomment-392548883

## Description

- I added several local functions and tested them all: `createTempLicenses`, `getLicensePath`, `getPackagesWithoutLicense`, `removeTempLicenses`, `removeTempLicensesSync`
- In `initialize()` I added some lines to determine situation:
  - figuring out whether the root package has a license or not
  - figuring out packages without license
  - if the root license is present and there are packages without license, those are saved for later
  - if the root license is not present and there are packages without license, there is only a warning - nothing else can be done
- In `npmPublish()` I added creating temporary license files and removing them to the chain
  - if the chain fails, there is synchronous removal of the files, possible error is only printed
  - the licenses are added before prepublish scripts and removed after postpublish - it sort of made sense to me and it was also easier to fit into the current flow - let me know if you think otherwise
- I added `fs-extra` as a dependency for the `publish` package
- I added description of the feature to the README. Please let me know if there are other docs to update.

## Motivation and Context

Closes https://github.com/lerna/lerna/issues/1213

## How Has This Been Tested?

- I wrote several tests for both local lib additions and the publish command itself
- I modified the `npm-publish` mock to record information about license files at the time of publishing. This caused several other tests to fail, as they were relying on the simpler, previous version of the mock's registry. I updated the tests and snapshots.

## Types of changes

## Checklist:


honzajavorek:
The PR fails on Node 6. Although the failures seem to be completely unrelated to my changes and they look like timing issues, they fail **consistently**. This needs more investigation.

honzajavorek:
@evocateur That sounds a bit like a turn-down, although in https://github.com/lerna/lerna/issues/1213 you seemed to agree with the idea and we even agreed on how this could be implemented. After investing quite some time into the PR, I'm a bit confused now. Do you suggest a different implementation approach or are we revisiting whether this should be a part of Lerna at all?

Almost every project using Lerna (and that's most of the projects utilizing monorepos, and that's a lot of packages in the npm ecosystem) is suffering an issue where the sub packages are not distributed with a full license text when the maintainers only provide the full license text file in the root of the monorepo. While the intention is clear (all the packages are licensed under the root license), the result is debatable. The distributions of packages published to npm are without the license text, thus more strict company lawyers could prohibit using such pacakges or deem them as problematic and requiring special treatment. That brings burden to those wanting to use such packages.

Without full license text, the sub packages are not properly licensed, even though the `license` field contains a valid SPDX identifier. Some licenses, such as MIT, even explicitly require packing the text together with the distribution, it's written in the text of the license: https://github.com/babel/babel/pull/7308#issuecomment-367615152 Effectively, this way the source code of most Lerna-distributed packages on npm is unlicensed and cannot be used by anyone else (or, to be more precise, by anyone who cares about licensing).

npm does include the license file automatically in case it's present, so this problem wouldn't exist if the project were not monorepos. As Lerna is a tool to help maintainers to deal with specifics of having a monorepo, and especially with specifics of publishing the sub packages, I think this feature should be a core feature of the tool. Moreover, the behavior we discussed in the initial issue requires no additional work to the maintainers. It follows their intentions without having to care about anything, while it automatically solves the licensing problem for all the package users. Factoring the feature out to a separate tool doesn't solve the widespread issue as it requires the users of the packages to file issues on the projects they have problems to use, and it requires maintainers to make additional steps and to modify their project life cycle to accommodate the license distribution.

evocateur:
@honzajavorek Thanks for the push back, I understand a lot better now, and completely agree. Sorry for the scattered skepticism, I've been punishingly busy the last few weeks. I will rebase your changes, since I caused the conflicts.

Thanks again for all your efforts!

evocateur:
@StrahilKazlachev Definitely, I think we can tackle that in a separate PR.

honzajavorek:
@evocateur Thank *you* for maintaining Lerna despite the time constraints, for your initial guidance on how to implement this and for merging this! 🚀

PR #1479, opened 2018-06-19
Change default mutex port to 42444

gtomolillo:
## Description
The mutex option when using yarn now uses the port 42444

## Motivation and Context
The port 42424 previously used is also used by ASP.Net session state service on windows machines causing lerna to appear stuck unless changing it to a different port in the lerna.json file.

## How Has This Been Tested?
Tested on Windows and OSX environments

## Types of changes

## Checklist:


evocateur:
We already use a library ([get-port](https://github.com/sindresorhus/get-port)) that falls back to a random port if the preferred port isn't available.

PR #1482, opened 2018-06-22
README. The configuration for custom message for releases section is missing an s.

NicolaiSchmid:
## Description
The is an error in the readme, specifically in the [publish](https://github.com/lerna/lerna#publish) section.
The provided example for configuring the custom publish message is missing an s in `commands`:

> If the message contains `%s`, it will be replaced with the new global version version number prefixed with a "v".
> If the message contains `%v`, it will be replaced with the new global version version number without the leading "v".
> Note that this only applies when using the default "fixed" versioning mode, as there is no "global" version when using `--independent`.
> This can be configured in lerna.json, as well:
> ```json
> {
>  "commands": {
>     "publish": {
>       "message": "chore(release): publish %s"
>     }
>   }
> }

`command*s*`

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -486,7 +486,7 @@ This can be configured in lerna.json, as well:
 
 ```json
 {
-  "command": {
+  "commands": {
     "publish": {
       "message": "chore(release): publish %s"
     }


evocateur:
Actually, either `command` or `commands` is supported, but `command` is preferred and thus evident in the documentation. The `commands` namespace is deprecated in lerna v3.

PR #1507, opened 2018-07-19
fix(publish): move git sha to build metadata

### LINKED ISSUE #1118: Invalid canary versions for certain SHAs
## Expected Behavior
`--canary` should always generate valid version strings (in accordance with the generally accepted [semver rules](http://semver.org)).

## Current Behavior
`--canary` generates invalid versions if the current SHA's first eight characters are numeric and the first digit is zero. For example, `1.1.0-alpha.21636f26` is valid, but `1.1.0-alpha.08213312` isn't – see rule 9 of the specification ("Numeric identifiers MUST NOT include leading zeroes").

Note that this is not just an issue of just being pedantic – we're using JFrog's Artifactory, and without going into too much detail, it really, *really* doesn't like encountering a version string like that. We're publishing a lot of packages, multiple times a day, and we're hitting this issue quite regularly.

## Possible Solution
Change the template string in `PublishCommand`/`getCanaryVersion()` to be: `${nextVersion}-${preid}${hash}`, or `${nextVersion}-${preid}.commit${hash}`, or something similar.

## Steps to Reproduce
1. Commit until the latest SHA starts with a zero followed by seven digits. This might take a while :)
2. Publish using `--canary`

## Your Environment
| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.2.0 |
| `npm --version`  | 5.5.1 |
| `yarn --version` | 1.3.2 |
| `node --version` | 8.9.1 |

| OS | Version |
| --- | --- |
| macOS Sierra | 10.12.6 |

hbetts:
Move the git commit sha from the pre-release string of the semantic
version number to the build metadata string.

Git commit shas are not always valid pre-release strings according to
the Semantic Versioning Specification 2.0. For example, SemVer requires
that a numeric pre-release segment (a segment consisting of only
numbers) must not contain a leading zero. However, it's possible for
a git commit sha to be all numbers and for the first number to be zero.

The same rule does **not** apply to the build metadata string placed
after a `+` character. Therefore, to ensure the version contains git sha
information, but is still valid, we are moving the sha to the build
metadata string.

BREAKING CHANGE:

Git commit sha is now contained in the build metadata string of the
semantic version number.

So `1.1.0-alpha.81e3b443` now becomes `1.1.0-alpha+81e3b443` when you
run `lerna --canary`.

To increment the pre-release number after `alpha` please include it in
the string passed to the `--canary` flag:

```bash
lerna --canary=alpha.0
```

For convience, consider using the build number typically set in your
build environment by your CI platform:

```bash
lerna --canary=alpha.${BUILD_NUMBER}
```

Fixes #1118

---

## Description

Please see commit message.

## Motivation and Context

Please see #1118.

## How Has This Been Tested?

Please see snapshots.

## Types of changes

## Checklist:
  - I did not use the latest version of `npm`. I used a Docker container, and you can't upgrade older versions of `npm` in Docker because of an `npm` bug.

### DIFF
--- a/commands/create/__tests__/__snapshots__/create-command.test.js.snap
+++ b/commands/create/__tests__/__snapshots__/create-command.test.js.snap
@@ -109,7 +109,7 @@ new file mode 100644
 index SHA..SHA
 --- /dev/null
 +++ b/packages/my-cli/package.json
-@@ -0,0 +1,34 @@
+@@ -0,0 +1,26 @@
 +{
 +  "name": "my-cli",
 +  "version": "1.2.3",
@@ -129,20 +129,12 @@ index SHA..SHA
 +    "bin",
 +    "lib"
 +  ],
-+  "repository": {
-+    "type": "git",
-+    "url": "git+https://github.com/test/test.git"
-+  },
 +  "scripts": {
 +    "test": "echo \\"Error: run tests from root\\" && exit 1"
 +  },
 +  "dependencies": {
 +    "yargs": "^1.0.0-mocked"
-+  },
-+  "bugs": {
-+    "url": "https://github.com/test/test/issues"
-+  },
-+  "homepage": "https://github.com/test/test#readme"
++  }
 +}
 `;
 
@@ -206,7 +198,7 @@ new file mode 100644
 index SHA..SHA
 --- /dev/null
 +++ b/packages/my-es-cli/package.json
-@@ -0,0 +1,35 @@
+@@ -0,0 +1,27 @@
 +{
 +  "name": "my-es-cli",
 +  "version": "1.2.3",
@@ -227,20 +219,12 @@ index SHA..SHA
 +    "bin",
 +    "dist"
 +  ],
-+  "repository": {
-+    "type": "git",
-+    "url": "git+https://github.com/test/test.git"
-+  },
 +  "scripts": {
 +    "test": "echo \\"Error: run tests from root\\" && exit 1"
 +  },
 +  "dependencies": {
 +    "yargs": "^1.0.0-mocked"
-+  },
-+  "bugs": {
-+    "url": "https://github.com/test/test/issues"
-+  },
-+  "homepage": "https://github.com/test/test#readme"
++  }
 +}
 diff --git a/packages/my-es-cli/src/cli.js b/packages/my-es-cli/src/cli.js
 new file mode 100644
@@ -332,7 +316,7 @@ new file mode 100644
 index SHA..SHA
 --- /dev/null
 +++ b/packages/my-pkg/package.json
-@@ -0,0 +1,27 @@
+@@ -0,0 +1,19 @@
 +{
 +  "name": "my-pkg",
 +  "version": "1.2.3",
@@ -348,17 +332,9 @@ index SHA..SHA
 +  "files": [
 +    "lib"
 +  ],
-+  "repository": {
-+    "type": "git",
-+    "url": "git+https://github.com/test/test.git"
-+  },
 +  "scripts": {
 +    "test": "echo \\"Error: run tests from root\\" && exit 1"
-+  },
-+  "bugs": {
-+    "url": "https://github.com/test/test/issues"
-+  },
-+  "homepage": "https://github.com/test/test#readme"
++  }
 +}
 `;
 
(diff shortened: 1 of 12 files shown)


hbetts:
@evocateur I'm not sure why most of those snapshots were updated.

When I ran `npm test` it notified me that I needed to update my snapshots. So I did.

I understand why the `canary` snapshot was updated, but not the others.

My development environment is a Node 8 Docker container using the version of `npm` and `git` that comes with it.

I also don't understand why `npm` mangled the `package-lock.json` file either.

What do I need to do to get a working local build, and one that plays nicely with Lerna's CI setup?

evocateur:
There is a snapshot serializer that may need to be updated due to the version format change.

I’m still not a huge fan of passing in the numeric bit of the release ID. I’m in the process of major refactoring in this area, FYI (splitting version command out of publish).

evocateur:
v2.1.4 of git seems rather old, since the latest release is v2.17.x...

hbetts:
> The vast majority of the snapshots should not have changed (nor the lockfile).

I'll pull down the [Travis CI docker image](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally) so that I can have a development environment with the proper tools.

> There is a snapshot serializer that may need to be updated due to the version format change.

I'll keep an eye out for that.

> I’m still not a huge fan of passing in the numeric bit of the release ID.

Me neither, but the next option was to use the `canary` value, query the `npm` registry, find all package releases that match the version and canary identifier, then find the next available number. 

For example, if `1.0.0-beta.0` was published, I would expect Lerna to query the `npm` registry, see `beta.0`, increment the number, and append it to the canary string.

However, that seemed like _way_ too much complexity to introduce to address just this issue.

Would your refactor provide something like that?

> v2.1.4 of git seems rather old, since the latest release is v2.17.x...

Not unheard of though.

At my company we use Lerna 2.0 `publish` on RHEL 6 boxes rocking `git` version 1.8. 😉 (We'll have access to a newer version soonish, but even then, we use RHEL and LTS versions of Ubuntu using the versions of packages that come with those distributions. So we'll still be using a older version of `git` than what is added to the Travis CI build environment (Travis CI upgrades the version of git in their images beyond what comes by default with Ubuntu Trusty))

fenduru:
@evocateur @hbetts what needs to be done to get this merged and released? I currently cannot use Lerna canaries for my projects that publish to Artifactory which is quite unfortunate.

hbetts:
@fenduru I need to update this change so that unrelated snapshots are not updated. That will require a little time to build out a custom Docker image to provide a compatible development environment. (Longer story is copied below)

The other part of this is determining what @evocateur means by `I’m still not a huge fan of passing in the numeric bit of the release ID. I’m in the process of major refactoring in this area, FYI (splitting version command out of publish).`.

---

As known, most of the snapshots should not have changed. So, I spent this morning trying to get a reproducible development environment on my laptop. Using the official `node:8` docker image is not sufficient because it uses Debian Jessie (Version 8), which has [`git 2.1.4` installed](https://packages.debian.org/jessie/git). Lerna uses Travis CI as their build environment and Travis CI [maintains their own custom Docker images](https://docs.travis-ci.com/user/common-build-problems/#Running-a-Container-Based-Docker-Image-Locally), which are based on Ubuntu Trusty, with a newer version of `git` manually installed into the container image.

Therefore, to get a development environment that matches that of the core maintainer and their CI environment, I need to pull the Travis CI image. So, in addition to learning why `git` was too old in the standard Node images, I then started downloading the ~3GB Docker image for Travis CI.

However, Travis CI images do not contain Node, or any of the tooling needed to setup a Node development environment (such as `nvm`). Instead, anyone pulling their custom images are expected to install any tools or runtimes they require.

Therefore, I'm going to need to build a Docker image off of the Travis CI image, and manually install the required version of Node I need for Lerna development.

fenduru:
I was able to come up with a workaround for the time being:

```
yarn lerna publish --skip-git --cd-version=prerelease --preid=beta+`git rev-parse --short HEAD`--npm-tag next --exact --yes
```

Which yields a version like `2.2.1-beta+8e77993.0`

hbetts:
I believe @evocateur's answer is effectively #1522 

I'm closing this PR out because the underlying code has been rewritten by @evocateur.

Furthermore, #1522 actually addresses this issue.

evocateur:
@hbetts Thanks for all your contributions, they've been especially helpful lately!

Issue #1513, opened 2018-07-20
[Feature] Option to automatically 'publish' Releases on GitHub


ChrisBAshton:
When you push a tag to GitHub, it adds it to the list of Releases but is 'unpublished', therefore doesn't look like the most recent version:

![screen shot 2018-07-20 at 17 37 45](https://user-images.githubusercontent.com/5111927/43014306-a78c03f0-8c43-11e8-99de-34bbfcf0f447.png)

It would be really awesome if Lerna provided some option to automatically _publish_ each release with sensible defaults - perhaps e.g. with a prompt:

```
> Would you like to publish this release on GitHub releases? y/N
> Would you like to combine these updates into one release? y/N
# see https://github.com/lerna/lerna/issues/1512
> What should the release title be?
___
> What should the release description be?
____________
```

phun-ky:
Would also love to combine this with lerna-changelog and/or github-release-notes

ChrisBAshton:
Adding a comment to keep this 'fresh'. It's had a few +1s so looks like it's worth keeping open.

evocateur:
#1864 is in the process of implementing this, at least for users of `--conventional-commits`.

Issue #1091, opened 2017-10-30
Support 2FA `--otp=123456`


hzoo:
https://docs.npmjs.com/getting-started/using-two-factor-authentication

Need a prompt to be able to pass this for publishing, etc (right before since not that much time)

EDIT: current workaround is `NPM_CONFIG_OTP=123456 lerna publish`

evocateur:
I'm worried that we'll have to re-prompt if publishes start failing (due to expired TOTP). Really exposes how un-tested and gnarly our current publish failure handling is. :/

hzoo:
Yeah that's what happened, also since we published like 100 pkgs and it would expire in between (had to just turn it off)

djfarrelly:
To help people in the short term, perhaps npm's tokens could help. Creating a wrapper script to generate a token, adding it to your `.npmrc` then publishing using lerna then removing and deleting the token could work.

Docs: https://docs.npmjs.com/getting-started/working_with_tokens
Other helpful info: https://circleci.com/docs/1.0/npm-private-module-dependency/

rhysforyou:
For people finding this through Google, I'd like to point out that a good workaround is to set the `NPM_CONFIG_OTP` environment variable before running `lerna publish`, e.g.

```sh
NPM_CONFIG_OTP=123456 lerna publish
```

jthegedus:
The above solution doesn't seem to work with Yarn & lerna as a devDep, eg:

```shell
NPM_CONFIG_OTP=123456 yarn lerna publish
```

It still fails on the OTP during publication :disappointed:

jthegedus:
Fair point. I was under the impression that Yarn read all of npm's configs, wishful thinking. A note for others nonetheless.

dupski:
+1 for lerna to prompt for this, although `NPM_CONFIG_OTP` seems to work ok. Turning off 2FA doesn't seem like a good solution :)

rgbkrk:
As a result of the [recent compromise](https://status.npmjs.org/incidents/dn7c1fgrr7ng), I'd love to be able to enforce 2FA across our org. Is there a way to do the `npm pack` approach with lerna now or does this need work within lerna itself?

kachkaev:
[...]

PR #1522, opened 2018-07-26
feat: Split `lerna version` from of `lerna publish`

### LINKED ISSUE #277: [Feature] Canary should have incrementing version numbers.

### LINKED ISSUE #936: Lerna Publish pre-patch/minor/major behavior
First I just want to say this is a great project! Thanks for putting it out there for folks to use!

Here's my question: 
I don't know if there is a way to do this (not documented or I'm stupid) or if there isn't a way to do this, but if I publish a pre-patch/minor/major version, I would like a way for `lerna publish` to be able to figure out those packages still need a non-'pre' publish on subsequent `lerna publish`'s.

Example:
I have a feature branch which has several packages being updated. I want to be able to deploy that branch to an environment and test against it before merging it into my master branch. Therefore, I need to (I think) do a `lerna publish` which will get all those affected components since my build process relies on pulling in the actual npm packages. However, I obviously don't want to do a full version bump since this feature is still in testing. So I use the `pre-X` publish option. This works great!

Then I decide to merge that branch into master after some testing is done. However, now when I run `lerna publish`, it fails to publish anything because in `lerna`'s eyes, nothing has changed. I would still like to do an actual publish though on those packages that have a `pre-X` version number. Of course I can go through figure out which things changed and publish those packages manually, but in a project with a lot of packages, that can be pretty tedious.

Questions:
1. Is there a way to do this currently?
2. If there is, can that be documented in the readme (or if I just missed it, laugh at me and point me towards it)?
3. If there is not a way to do that, is it on a roadmap as a feature, or would you entertain putting that as a possible feature? I think it would be very useful for this kind of workflow.

Thanks again for the great project!

### LINKED ISSUE #956: publish --skip-npm && publish --skip-git never finds updates
Because the tag is written before the call to `publish --skip-git`, it compares the current state against itself, meaning there will never be updates to publish.

### LINKED ISSUE #961: Separate "version" and "publish" commands
Lerna needs a separate `version` command.

### Why
Folks who use npm a lot are probably comfortable with the separation between the `npm version` and `npm publish` commands. For me, this separation is important because it allows me to double- and triple-check things before publishing. When versioning and publishing a monorepo with dozens of packages, the "publish anxiety" factor is even higher. I've worked through some of that anxiety with various permutations of `--skip-git` and `--skip-npm` then `git diff`, but at some point you have to `git reset` and just trust that Lerna is going to do the right thing.

The other problem is that Lerna doesn't make it easy (possible?) to just publish all of the packages using the existing `package.json` versions. AFAICT, it's not possible to skip the increment step, which means that you have to run something like this instead:

```
lerna exec --bail=false npm publish
```

...which will output a lot of scary `npm publish` error messages for any of your packages that haven't been bumped since the last release.

We recently introduced a CI-driven automated publishing workflow in [Primer](/primer/primer-css), and to get around this we resorted to [running](https://github.com/primer/primer-css/blob/master/script/release#L9) a [`try-publish` script](https://github.com/primer/primer-css/blob/master/script/try-publish), which would only run `npm publish` if the version in each module's `package.json` hadn't already been published. Other teams that want to automate publishing will inevitably run into the same issue, and it would stink if they all ended up having to come up with their own solutions for it.

### How
There are a couple of ways to go about this:

1. To maintain backwards compatibility with the current CLI API, the `publish` command could introduce two new options:
    * one that does everything _but_ the git and npm steps, e.g. `lerna publish --skip-publish`
    * one that _only_ publishes without touching any `package.json` files, e.g. `lerna publish --skip-version`

1. If you're comfortable breaking the `publish` CLI for 2.x users, you could avoid adding any additional options to `publish` and just move all of the versioning logic over to a new `version` command.

### LINKED ISSUE #1056: lerna to publish packages with their current version
How do I tell `lerna` to publish packages with **their current version**?
Now `lerna` is always trying to increase the version of packages ...

I need this because I want to divide the publication process into 2 stages:
1. First I want to manually raise the package version only in git in dialog mode:
`lerna publish --skip-npm`
2. Only then with the help of CI I want to publish in npm:
`lerna publish --skip-git`

I do not need to raise the version twice.

## My Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.2.0|
| `npm --version`  | 5.4.2|
| `node --version` | v6.11.2|

| OS | Version |
| --- | --- |
| Windows 10 Pro | 1703 |

### LINKED ISSUE #1118: Invalid canary versions for certain SHAs

### LINKED ISSUE #1385: lerna with CI/CD pipeline

### LINKED ISSUE #1483: Ignore `v` in `--repo-version`
I want to publish with `circleci` and `git tag`. Normally I use `git tag` like `v1.0.0`, and publish in `circleci` like `yarn lerna publish --repo-version ${CIRCLE_TAG}`. As a result, `yarn lerna publish` will use `v1.0.0` to publish packages, but `--new-version` of `yarn publish` can not be a version with `v`. Maybe `v` should be ignored with `--repo-version`?

## Expected Behavior

```sh
yarn publish --repo-version v1.0.0
// publish packages with version `1.0.0`.
```

## Current Behavior

```sh
yarn publish --repo-version v1.0.0
// publish packages with version `v1.0.0`.
```

## Possible Solution

Add `repoVersion.replace(/^v/, '')` in [here](https://github.com/lerna/lerna/blob/master/commands/publish/index.js#L305).

```sh
yarn lerna publish --repo-version v1.0.0
```

```json
<!-- Please paste your `lerna.json` here -->
{
  version: 1.0.0,
  commands: { 
    publish: { 
      ignore: [ 
        *.md,
        __tests__/**
      ]
    }
  },
  lerna: 2.11.0,
  npmClient: yarn,
  npmClientArgs: [ 
    --no-lockfile
  ],
  packages: [ 
    packages/*
  ]
}
```
</p></details>

```txt
<!-- If you have a `lerna-debug.log` available, please paste it here -->
<!-- Otherwise, feel free to delete this <details> block -->
```
</p></details>

## Context

I hope this can publish in `circleci` with my `git tag`, and I want to still have `v` in my `git tag`. Like `standard-version`, this will ignore `v` when I use `--release-as`. Maybe `lerna` can ignore, too.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.11.0 |
| `npm --version`  | 6.1.0 |
| `yarn --version` | 1.7.0 |
| `node --version` | 10.2.1 |

| OS | Version |
| --- | --- |
| macOS High Sierra | 10.13.5 |

evocateur:
## Description

To facilitate further refactoring, the gigantic PublishCommand needs to be split in two. Several flags were deprecated and replaced with clearer options, but remarkably only two _technically_ breaking changes resulted from ~2 weeks of work.

BREAKING CHANGE:
* `--preid` now defaults to "alpha" during prereleases:

  The previous default for this option was undefined, which led to an awkward "1.0.1-0" result when passed to `semver.inc()`.

  The new default "alpha" yields a much more useful "1.0.1-alpha.0" result. Any previous prerelease ID will be preserved, just as it was before.

* `--no-verify` is no longer passed to `git commit` by default, but controlled by the new `--commit-hooks` option:

  The previous behavior was too overzealous, and the new option operates exactly like the corresponding [npm version](https://docs.npmjs.com/cli/version#commit-hooks) option of the same name.

  As long as your pre-commit hooks are properly scoped to ignore changes in package.json files, this change should not affect you. If that is not the case, you may pass `--no-commit-hooks` to restore the previous behavior.

## Motivation and Context

Fixes #277
Fixes #936
Fixes #956
Fixes #961
Fixes #1056
Fixes #1118
Fixes #1385
Fixes #1483
Fixes #1494

This PR lays the groundwork for #1091 (and #1513) by making `lerna publish` only _directly_ worrying about publish itself, not all those other complicated versioning decisions.

## How Has This Been Tested?

Ported unit tests, with "full-bleed" integration tests of `npm publish` thanks to the new `--dry-run` flag in npm@6.

## Types of changes
- [x] Breaking change (internal tools)

## Checklist:


morgs32:
@evocateur I salute you this looks so promising. I am curious about the bit of documentation that strongly discourages running `lerna version` from a branch. I kind of figured that I'd manually be incrementing version on a branch and in a PR, and then on merge to `master` I'd `lerna publish from-git`. But is that not a wise idea? What sort of process would you suggest or what did you have in mind?

evocateur:
@morgs32 My biggest concern with making tagged releases in feature branches is that the tags get mangled when merging.

I'm not sure if rebase merge would be any safer, since if it was rebasing onto upstream commits, the SHAs would change and thus the annotated tag(s) would be "broken", as the SHA it points to no longer exists. However, I haven't tried rebase merge in a long time, and I don't recall if `git rebase` itself preserves tags.

One of my goals in the near future is to improve the "auto-publish from CI" experience, perhaps with a subcommand `lerna publish release`. I envision some configurable branch awareness that would publish untagged prereleases on feature branches (or PRs only?) and then on `master` (or `latest`, whatever your "default" branch is configured to be) it would publish with tags, even creating github releases with changelogs (if you're using conventional commits).

So please let me know if `lerna publish from-git` works after a rebase merge, I would like to figure this out, even if it is "just" a clarification to the documentation!

commands/version/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+### `--allow-branch <glob>`
+
+A whitelist of globs that match git branches where `lerna version` is enabled.
+It is easiest (and recommended) to configure in `lerna.json`, but it is possible to pass as a CLI option as well.
+
+```json
+{
+  "command": {
+    "publish": {
```
danielo515: Is this correct @evocateur ? Do you configure `lerna version` by configuring publish on `lerna.json` ? Seems confusing

evocateur:
It's for backward-compatibility. The `lerna version` docs probably should be updated to use the correct namespace, though.

PR #1535, opened 2018-08-02
`git describe --tags --abbrev=0` instead of `git describe --abbrev=0`

btakita:
Fix for https://github.com/lerna/lerna/issues/16

## Description

Added `--tags` to `git describe` call

## Motivation and Context

There was an error that occurred when running `lerna updated`

## How Has This Been Tested?

Run `lerna updated` on a repo that has not been published by lerna.

## Types of changes

Add `--tags` option. Same solution as a previously merged fix - https://github.com/lerna/lerna/pull/17

### DIFF
--- a/utils/collect-updates/collect-updates.js
+++ b/utils/collect-updates/collect-updates.js
@@ -27,7 +27,7 @@ function collectUpdates(filteredPackages, packageGraph, execOpts, commandOptions
       committish = `${sha}^..${sha}`;
     } else if (!committish) {
       // attempt to find the last annotated tag in the current branch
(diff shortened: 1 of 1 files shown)


evocateur:
This isn't going to fix the issue.

evocateur:
Running `lerna updated` on a repo that has never been published will always fail? There are a zillion tests that validate `lerna publish` without existing tags of any kind will work correctly.

btakita:
@evocateur so the solution is to remove all tags from the repo? Why would this cause lerna to have an error?

Issue #1539, opened 2018-08-04
lerna add unable to retrieve package metadata from private repository


jd-carroll:
When attempting to install a package from a private repository (like Artifactory), lerna is unable to complete the installation because it cannot find the package.  The reason lerna is unable to find the package is because [`package-json`](https://www.npmjs.com/package/package-json) does not look at the repository configured via `npm config`.

The issue with `package-json` has been documented here: sindresorhus/package-json#45

However, there does not seem to be much movement on updating the package.

**Please use a different module to read package metadata from the configured npm repository**

evocateur:
Lerna doesn't install packages, that's what npm and yarn do. Strikes me as an Artifactory problem if they can't support public registry APIs?

PR #1540, opened 2018-08-04
fix: Install from Private Enterprise Repositories (#1539)

jd-carroll:
Fix installing packages when a private enterprise repository is configured.

(Note: I am using `npm@6.3.0` and I am having issues getting the `package-lock` to install correctly.  I believe lerna recently moved to a mono-repo style itself, but I wasn't able to figure out how to use the lerna tooling on itself.  Any guidance appreciated)

### DIFF
--- a/commands/add/index.js
+++ b/commands/add/index.js
@@ -2,7 +2,7 @@
 
 const dedent = require("dedent");
 const npa = require("npm-package-arg");
-const packageJson = require("package-json");
+const { RegistryClient } = require("package-metadata");
 const pMap = require("p-map");
 const path = require("path");
 const semver = require("semver");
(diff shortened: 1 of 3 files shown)


jd-carroll:
Another note: I believe the checks failed for the same reason I am having difficulty.  Happy to tweak as necessary.

evocateur:
Your package has _runtime_ dependencies on jest and ts-jest. That's not gonna fly.

Also, where is the source code? No `homepage` field in package.json means I can't locate its source, and I'm not in the habit of accepting dependencies I can't read first.

jd-carroll:
@evocateur Thanks for pointing that out, def a mistake and will get that fixed tonight!

jd-carroll:
@evocateur I've updated `package-metadata`, could you take a look?  I'm open to other thoughts on node module, most interested in getting the functionality in

evocateur:
This should be fixed in [v3.1.1](https://github.com/lerna/lerna/releases/tag/v3.1.1).

PR #1542, opened 2018-08-06
Document missing properties in lerna.json

XanderLuciano:
It seems that lerna.json is not fully documented anywhere? I found these settings throughout the github issues. Thought we should document them here for users to see along with a link to the example of how to use yarn's workspaces.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,8 @@ Running `lerna` without arguments will show all commands/options.
 ```json
 {
   "version": "1.1.3",
+  "useWorkspaces": true,
+  "npmClient": "yarn",
   "command": {
     "publish": {
       "ignoreChanges": [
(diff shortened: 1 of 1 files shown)


XanderLuciano:
NPM is default client as per stated here: https://github.com/lerna/lerna/tree/master/commands/bootstrap#options

> The default `--npm-client` is `npm`.

evocateur:
This is already documented here: https://github.com/lerna/lerna/tree/master/commands/bootstrap#--use-workspaces

Is that not sufficient?

XanderLuciano:
That's sufficient, just seems like the documentation is extremely spread out and it's really hard to just find all the information.

That said, I don't use lerna normally, I'm just trying to learn how to use lerna so I can submit a PR to a project that uses lerna and it took me a long time to figure out everything. Seems like commands are in the documentation that don't exist in the CLI (like lerna version) so my intentions were to just make the information a little more consolidated. 

IMO it doesn't make sense to document properties of the JSON file in a documentation page about the CLI.

evocateur:
I appreciate the effort to document, believe me.

I'm just not sure individual properties should necessarily be emphasized (especially non-universal ones) when it is the actual mechanism of "durable" configuration merging that is of interest. In other words, how CLI options are translated into a JSON configuration schema and merged with defaults.

XanderLuciano:
No problem! That makes sense and I trust you understand the project better than I.

I think it makes sense to keep these separated since some are universal and some are not. It's likely the documentation structure makes more sense when you are starting fresh. It's just difficult to pickup and understand an existing project configured with lerna.

Thanks for the clarification!

Issue #1559, opened 2018-08-13
Allow filter options for lerna changed in v3


bdwain:
## Expected Behavior
I should be able to say `lerna changed --since GITREF`. 

## Current Behavior
--since is an invalid option for the changed command

## Context
I wrote my own script to handle publishing of my packages because of limitations I've mentioned in [other](https://github.com/lerna/lerna/issues/1489) [issues](https://github.com/lerna/lerna/issues/1286). It relies on lerna changed to tell me what to build though. And I'm not always diffing since the last tagged release, so --since has been very useful (as well as other filter options).

The reasoning the docs give as to why filter options are not supported is 

> Unlike lerna ls, however, lerna changed does not support filter options, as filtering is not supported by lerna version or lerna publish.

I'm confused why that is relevant though. Those are completely separate commands. It makes more sense that they would be limited to the last tagged release because you wouldn't want to re-bump or re-publish a package. And if you are using them, you most likely don't need `lerna changed`. 

It seems like `lerna changed` is targeted more at people writing their own build scripts, and the reasoning for limiting version and publish doesn't necessarily apply to it.

## Your Environment
| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.0.3 |

bdwain:
Another benefit of this is that it seems like it would not require that the tags we publish be annotated tags. This seems to be a new requirement as of v3, but since lerna publish is not writing my tags, I was keeping them as lightweight because there was no benefit to making them annotated. 

If I was to say --since master or some other non-tag ref, there would be no need for me to make all of my tags annotated.

evocateur:
Lerna has created annotated tags (and required them to recognize previous releases) since late 2.x-beta. "Creating a lot of tags" is not a justification for keeping them lightweight, especially when they are used to demarcate releases.

Anyway, `lerna ls --since [ref]` works for your use case.

(One day I'd like to remove the entire `lerna changed` command in favor of a `--changed` flag for `lerna ls`; lerna has too many subcommands)

bdwain:
Ah didn't know lerna ls supports --since. Maybe that could be added to the documentation page? It only mentions --scope and --ignore when referencing filter flags.

bdwain:
Oh and thank you!

evocateur:
Yeah, that could be a good PR. The output of `lerna ls -h` does list it, but it tends to get lost in the giant list of options. (I need to figure out how to tell yargs to show a subset of the options without jumping through too many hoops...)

[...]

PR #1564, opened 2018-08-14
docs: Clarify that lerna list takes all filter flags

bdwain:
## Description
Clarify docs like mentioned in #1559 

## Motivation and Context
It wasn't clear that `list` accepted all filter options. Just changed the wording to make it more clear


evocateur:
Thanks!

PR #1565, opened 2018-08-14
feat(version): ask for confirmation before comitting or pushing

jantimon:
## Description

I was curious about what the `lerna publish  --conventional-commits` command would do in detail.  
So I checked if there is a `dry-run` and as lerna already asked me for changing the version numbers I expected lerna also to ask me for confirmation to publish my changes.

Unfortunately there were no confirmation and all my changes were pushed to npm.  

After googling I could find https://github.com/lerna/lerna/issues/51 so I thought maybe other would also like more control about destructive operations.

My change would provide the following publish experience:

```
$ lerna publish --conventional-commits

lerna info version 2.11.0
lerna info current version 0.0.0
lerna info Checking for updated packages...
lerna info Comparing with initial commit.
lerna info Checking for prereleased packages...

Changes:
 - foo: 0.0.1 => 1.1.0
 - bar: 0.0.1 => 1.1.0

? Are you sure you want to create these versions? Yes
? Are you sure you want to create a commit out of the above version updates? Yes
? Are you sure you want to push your git changes? Yes
```

Please note that this is only half of the feature. 

The combined output together with #1566 would be:

```
$ lerna publish --conventional-commits

lerna info version 2.11.0
lerna info current version 0.0.0
lerna info Checking for updated packages...
lerna info Comparing with initial commit.
lerna info Checking for prereleased packages...

Changes:
 - foo: 0.0.1 => 1.1.0
 - bar: 0.0.1 => 1.1.0

? Are you sure you want to create these versions? Yes
? Are you sure you want to create a commit out of the above version updates? Yes
? Are you sure you want to push your git changes? Yes
? Are you sure you want to publish these changes to npm? Yes
```

## How Has This Been Tested?

Actually I would love to test - but I would need some guidance how lerna would test confirms like the once I added.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #51, opened 2016-02-12
Feature request: lerna prepublish


gigabo:
Hi,

Really liking `lerna` so far.  However I've encountered some [unexpected behavior](https://github.com/kittens/lerna/issues/50) that wound up getting pushed to our public repo and published to npm.

I'd like to be able to run `lerna publish` to the point of creating the version commit and tag and stop it there so I can inspect what it did before proceeding.

My first thought for an interface for this is:

```bash
$ lerna prepublish
... make sure everything's as expected...
$ lerna publish
```

Where in this case `lerna publish` would see that the head revision is a version tag and skip the `prepublish` steps.

It might also be nice to have a `lerna revert` that's basically just `git reset --hard HEAD^` after verifying that the `HEAD` revision is a version commit.

I'd be happy to do the work for this and submit a PR if you're open to the idea.  How does it sound?  Ideas for a better interface?

Thanks!

Bo

kittens:
Sounds like you're looking for just `lerna updated` then?

gigabo:
I don't think `lerna updated` actually gives me what I want here.

Here's the output:
```bash
$ lerna updated

Lerna updated v1.1.0

Current version: 0.0.5
Version file location: /Users/bo/git/react-server/VERSION
Packages location: /Users/bo/git/react-server/packages

Checking packages...
- react-server
- react-server-gulp-module-tagger

```

What I'm looking for is: What exact changes will `lerna` make before publishing.

I just want the opportunity to review the commit before it's pushed and published in my name. :)

thejameskyle:
Oh like a `--dry-run` flag

gigabo:
Yes, exactly!  Though the partial work done during the dry run in general will be just fine and I wouldn't want to do it again for the actual publish.

gigabo:
So what about this for an interface:

```bash
$ lerna publish --dry-run # Create the version commit
$ lerna publish           # Push and publish (create commit if necessary)
$ lerna publish --abort   # Revert the version commit (if present)
```

hzoo:
Yeah ok, so you don't publish what you didn't intend?

gigabo:
Exactly!  If that interface looks okay I'd be happy to set it up and submit a PR.

thejameskyle:
[...]

PR #1566, opened 2018-08-14
feat(publish): ask for confirmation before npm publish

jantimon:
## Description

I was curious about what the `lerna publish  --conventional-commits` command would do in detail.  
So I checked if there is a `dry-run` and as lerna already asked me for changing the version numbers I expected lerna also to ask me for confirmation to publish my changes.

Unfortunately there were no confirmation and all my changes were pushed to npm.  

After googling I could find https://github.com/lerna/lerna/issues/51 so I thought maybe other would also like more control about destructive operations.

My change would provide the following publish experience:

```
$ lerna publish --conventional-commits

lerna info version 2.11.0
lerna info current version 0.0.0
lerna info Checking for updated packages...
lerna info Comparing with initial commit.
lerna info Checking for prereleased packages...

Changes:
 - foo: 0.0.1 => 1.1.0
 - bar: 0.0.1 => 1.1.0

? Are you sure you want to create these versions? Yes
? Are you sure you want to publish these changes to npm? Yes
```

Please note that this is only half of the feature. 

The combined output together with #1565 would be:

```
$ lerna publish --conventional-commits

lerna info version 2.11.0
lerna info current version 0.0.0
lerna info Checking for updated packages...
lerna info Comparing with initial commit.
lerna info Checking for prereleased packages...

Changes:
 - foo: 0.0.1 => 1.1.0
 - bar: 0.0.1 => 1.1.0

? Are you sure you want to create these versions? Yes
? Are you sure you want to create a commit out of the above version updates? Yes
? Are you sure you want to push your git changes? Yes
? Are you sure you want to publish these changes to npm? Yes
```

## How Has This Been Tested?

Actually I would love to test - but I would need some guidance how lerna would test confirms like the once I added.

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -14,6 +14,7 @@ const collectUpdates = require("@lerna/collect-updates");
 const npmDistTag = require("@lerna/npm-dist-tag");
 const npmPublish = require("@lerna/npm-publish");
 const { createRunner } = require("@lerna/run-lifecycle");
+const PromptUtilities = require("@lerna/prompt");
 const batchPackages = require("@lerna/batch-packages");
 const runParallelBatches = require("@lerna/run-parallel-batches");
 const versionCommand = require("@lerna/version");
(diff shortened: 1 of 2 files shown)


evocateur:
`lerna version --conventional-commits --no-git-tag-version` already exists for this kind of testing. Pass `--no-push` instead of `--no-git-tag-version` to see what kind of commit and tag(s) it would generate, and you can `git reset --hard origin/master && git tag -d v${x.y.z}` or whatever to throw away the test result.

I don't want a deluge of prompts that make users get into the habit of mashing the `y` key just to proceed with things they already expected to happen.

jantimon:
Well it’s not like people publish many times a day so it would give people more control and allow to edit changelog typos before publishing - other libraries like `standard-version` which is widely used act similar - they prepare everything but wait for the user to do the final push.

Is there any way to get closer to their approach?  
Maybe adding another option or command like `—guided-release`?

evocateur:
Sure, I think it's fair to restore at least one confirmation prompt in the publish flow when VersionCommand decisions haven't already been made (so the `lerna publish from-git` and `lerna publish --canary` cases). I don't think it needs dramatic restructuring of the existing control flow, however.

jantimon:
Unfortunately the publish actions which manipulate foreign systems are split into two commands:

  -version (git push)
  -publish (npm publish)

So inside the version command we can only prompt for `git push` and in the publish command we can only prompt for `npm publish`

How would you structure the flow?

evocateur:
I don't know why you need a prompt for everything? If you don't want to push _or_ publish, `lerna version --no-push` has you covered. Then `lerna publish from-git` after you've validated the commit and tag(s) and manually pushed to remote.

I literally did exactly that flow today, as I was publishing the lerna release that added a confirmation prompt to `lerna publish`...

jantimon:
Sorry I didn't see the commit.  
I am perfectly fine with only one confirmation for the entire publish process.  

Its probably even more important than confirming the version numbers.

But if I read that code correct your confirm would happen after the changes were already pushed to a foreign repository and would require force push permissions to repair that.

Correct?

evocateur:
The version command (which is called internally from the "legacy" `lerna publish`) will prompt for confirmation before making any changes, including git commit, tag, and push. This was already the case, btw, and now the confirmation message is clearer when run internally.

The other prompts I added today are for the `lerna publish from-git` case (`lerna version` already run seperately) and `lerna publish --canary`, neither of which makes any git commit/tag/push at all.

jantimon:
I really like the improved confirmation message texts.  
The problem right now is that the user has no confirmation **after** the files were changed by lerna.   
So for developers who don't know the lerna internals they have to trust lerna to publish everything exactly in their favor.

Other tools allow you to proof check that the versions they chose and the changes done by the tool are those you like to publish to foreign systems (be it git or npm)  

It would be a huge improvement to ask the user for confirmation to `publish` rather than for confirming that they want to execute the command they just typed into their terminal.

evocateur:
Each confirmation prompt presents the actions about to be taken, in terms of packages affected and versions to be published. That is the relevant user decision in the lerna workflow.

How exactly does it help to pepper a user with prompts to allow what are essentially implementation details to proceed?

jantimon:
You are absolutely right the decision is good as it is right now - only the timing could be improved.

If the prompt would happen right before `git push && npm publish` the user could check and even fix typos in the changelog, see which package.json would be changed and so on **before** hitting the enter key.

evocateur:
Sorry, not going to happen. If you're concerned about lerna's well-tested and robust git operations, call `lerna version --no-push` and inspect the result, etc, etc. I keep repeating myself.

jantimon:
Having all your knowledge about the code in lerna and about all the versions in your head you wouldn't need that change.  
So I respect your decision.  
But I really don't understand why you don't want to improving the dev experience for beginners who just want to get started with lerna.  
Imho you can't expect everyone to know all command line arguments and how they have to be mixed together from the start.  

And there will always be bugs - why do you insist not to allow developers to proof check that the changelog generation and file modification worked correctly **before** publishing?  
For example after upgrading from Lerna 2 to 3.  

I just can't understand why this would be a disadvantage over the current implementation.
It would completely solve the pain described in #51.

It's just the start but a better developer experience could like this:  
```
$ lerna publish --conventional-commits

lerna info version 2.11.0
lerna info current version 0.0.0
lerna info Checking for updated packages...
lerna info Comparing with initial commit.
lerna info Checking for prereleased packages...

Version bumps:
 - foo: 0.0.1 => 1.0.0
 - bar: 0.0.1 => 1.0.0

lerna info Added files to git
lerna info Created tag 1.0.0

? Are you sure you want to push these changes to git and release them to npm? Yes

lerna info Pushing version commit and tags to git remote...
lerna info Released "foo" 1.1.0 to npm
lerna info Released "bar" 1.1.0 to npm
lerna info 🎉
```

Issue #1214, opened 2018-01-18
Bug: Lerna prerelease publish menu provides incorrectly incremented version


pzrq:
## Expected Behavior

As per [node's implementation](https://www.npmjs.com/package/semver) of [prerelease semver](https://semver.org/#spec-item-9), Lerna should provide a `1.x.y-0`, `1.x.y-1`, ... version option in its publish menu.

## Current Behavior

As seen [here](https://github.com/mongodb-js/hadron-react/pull/10#issuecomment-358536006), lerna actually provided me with a `1.x.y+1-0` option instead of the `1.x.y-1` option I expected.

## Steps to Reproduce (for bugs)

1. (Might be optional) [Add 2FA](https://docs.npmjs.com/getting-started/using-two-factor-authentication) to an npm account
2. NPM_CONFIG_OTP=<YOUR_OTP> lerna publish --npm-tag=beta

<details><summary>lerna.json</summary><p>

```json
<!-- Please paste your `lerna.json` here -->
{
  "lerna": "2.6.0",
  "version": "1.13.1-0",
  "packages": [
    "packages/*"
  ]
}
```
</p></details>

## Context

As I was in the middle of using [`NPM_CONFIG_OTP`](https://github.com/lerna/lerna/issues/1137#issuecomment-353547051), had at most 30 seconds and was not sure if it would work at all, I decided to publish anyway.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.6.0 |
| `npm --version`  | 5.6.0 |
| `yarn --version` | 0.16.1 |
| `node --version` | v8.9.4 |

| OS | Version |
| --- | --- |
| macOS Sierra | 10.13.2 |

evocateur:
Yeah, we haven't adapted to OTP yet, sorry.

```bash
$ npx semver --increment prerelease --preid beta 1.13.0
1.13.1-beta.0
```

We literally call node's semver implementation [here](https://github.com/lerna/lerna/blob/6080e6b5b029714acc47a8489d35589e61542704/src/commands/PublishCommand.js#L503-L504), and that's what we get. The section on [prerelease identifiers](https://www.npmjs.com/package/semver#prerelease-identifiers) disagrees with your assertion that `1.x.y-1` was the expected result of a prerelease of `1.x.y`.

[...]

PR #1568, opened 2018-08-14
fix(version): Prioritize `--preid` over existing prerelease ID

moe971:
in order to switch a prerelease id from alpha version to rc I needed to change condition to prioritize preid insteed of an existing preid

## Description

I juste change order of OR statement to take first preid
 const resolvePrereleaseId = existingPreid => preid || (isPrerelease && existingPreid) || "alpha";

## Motivation and Context
On my continuous integration workflow version my app on alpha prerelease, then on it goes to release candidate, so I try tu use lerna version prerelease --preid rc but because my previous tag was  dam/library-client@0.1.1-alpha.10 lerna 3.0.4 used my existing preid.
The solution was to change order on the test statement.

## How Has This Been Tested?
It has been test on my project, it works 👍 

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/__tests__/version-bump-prerelease.test.js
+++ b/commands/version/__tests__/version-bump-prerelease.test.js
@@ -18,6 +18,7 @@ const showCommit = require("@lerna-test/show-commit");
 const gitAdd = require("@lerna-test/git-add");
 const gitTag = require("@lerna-test/git-tag");
 const gitCommit = require("@lerna-test/git-commit");
+const getCommitMessage = require("@lerna-test/get-commit-message");
 
 // test command
 const lernaVersion = require("@lerna-test/command-runner")(require("../command"));
@@ -77,4 +78,14 @@ test("version prerelease with previous prerelease supersedes --conventional-comm
 
   const patch = await showCommit(testDir);
   expect(patch).toMatchSnapshot();
+});
+
+test("version prerelease with existing preid bumps with the preid provide as argument", async () => {
+  const testDir = await initFixture("republish-prereleased");
+  // Version bump should have the new rc preid
+  await setupChanges(testDir);
+  await lernaVersion(testDir)("prerelease", "--preid", "rc");
+
+  const message = await getCommitMessage(testDir);
+  expect(message).toBe("v1.0.1-rc.0");
 });
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -201,7 +201,7 @@ class VersionCommand extends Command {
     const increment = bump && !semver.valid(bump) ? bump : "";
     const isPrerelease = increment.startsWith("pre");
 
-    const resolvePrereleaseId = existingPreid => (isPrerelease && existingPreid) || preid || "alpha";
+    const resolvePrereleaseId = existingPreid => preid || (isPrerelease && existingPreid) || "alpha";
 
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -741,9 +741,9 @@
       "integrity": "sha512-LAQ1d4OPfSJ/BMbI2DuizmYrrkD9JMaTdi2hQTlI53lQ4kRQPyZQRS4CYQ7O66bnBBnP/oYdRxbk++X0xuFU6A=="
     },
     "JSONStream": {
-      "version": "1.3.3",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.3.tgz",
-      "integrity": "sha512-3Sp6WZZ/lXl+nTDoGpGWHEpTnnC6X5fnkolYZR6nwIfzbxxvA8utPWe1gCt7i0m9uVGsSz2IS8K8mJ7HmlduMg==",
+      "version": "1.3.4",
```
evocateur: The lockfile should not have changed (`npm ci` should be sufficient). Please revert this file

commands/diff/__tests__/__snapshots__/diff-command.test.js.snap
```diff
@@ -44,10 +44,10 @@ index SHA..SHA 100644
 `;
 
 exports[`DiffCommand should diff packages from the most recent tag 1`] = `
-diff --git a/packages/package-1/package.json b/packages/package-1/package.json
+diff --git c/packages/package-1/package.json w/packages/package-1/package.json
 index SHA..SHA 100644
---- a/packages/package-1/package.json
-+++ b/packages/package-1/package.json
+--- c/packages/package-1/package.json
++++ w/packages/package-1/package.json
```
evocateur: I'm not sure why these snapshots needed to be updated. Are you using a non-English localization of git?

moe971: I'm on an english localization of git, just my OSX is in French. I revert my package-lock, ran `npm ci` the only diff is the new test. I have the same problems with snapshots.

evocateur:
Okay, well, I can assure you the snapshots do not need updating, and the CI build should pass once you revert the snapshot updates. We probably have to pass some flags to git or something, I don't know why it is trying to change the diff prefixes, it's completely unrelated to your change.

moe971:
As you said after revert everything is green on CI, thanks for your help.

evocateur:
I don't think this solves #1214, as it's actually a separate code path (but certainly accesses similar properties).

evocateur:
I think passing an explicit [`--src-prefix=a/` and `--dst-prefix=b/`](https://git-scm.com/docs/git-diff#git-diff---src-prefixltprefixgt) will solve your local snapshot problems. Can you pull from `master` and execute `npm run ci` to verify for me? (no rush, just not certain how to reproduce on my own machine)

moe971:
I sync my repo, 
My current commit match with 3.0.6 tag, I run `npm i` && `npm run ci` I've got 526 lint error for import/unresolved, did I miss something ?

evocateur:
Unless you're making changes to dependencies, I would run `npm ci` instead of `npm i`. All the lint errors are due to a bad node_modules state.

PR #1619, opened 2018-08-29
Remove misleading license name

janjur:
MIT license has it's defined form and meaning. Latest changes have violated the core of MIT license. 
Prohibiting anyone to use the software is not compatible with the MIT license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -21,7 +21,7 @@ Enforcement ("ICE"):
 - "LinkedIn Corporation"
 - "United Parcel Service Co"
 
-MIT License
+License
 
 Permission is hereby granted, free of charge, to any person obtaining
 a copy of this software and associated documentation files (the


jepler:
Remember to change the metadata which appears at https://www.npmjs.com/package/lerna

pluma:
The text says:

> The following license shall not be granted to the following entities (..):
> MIT License

So it's clearly saying "the license" granted to everyone else _is_ the "MIT license". But it's explicitly not granted to anyone on the list.

I think this is confusing but for different reasons. The way I read it is that the license of the library remains the MIT license but the license itself is wrapped in a meta-license that restricts who that license applies to.

So in other words, as far as anyone not on the list is concerned, the library is MIT licensed. But that would mean that they are granted all the rights of the MIT license and aren't themselves barred from distributing the library to any of the companies on the list.

This clearly isn't the intent. Instead I think this is meant as a restriction of the license, which indeed would change this to a new more restrictive license that merely resembles the MIT license. So it's probably more at home with other novelty licenses like the JSON License (which is heavily inspired by the MIT license but more permissive while at the same time containing a legally ambiguous ethical clause).

So I see two obvious solutions:

* keep calling it MIT and clarify that the blacklist is non-viral
* adjust the license text to explicitly include the blacklist as part of the license

StoneCypher:
No.

This is not the MIT license.  This is some other license.

Automated license compliance checkers are giving false results because you're using a different institution's name and making false representations about what's allowed.

The MIT license is chosen for its permissiveness.  You are in violation of the spirit of the license, as well as the law.

Please just change the license name.  This is dishonest, confusing, and purposeless.

Daniel15:
Why was this closed with no comment?

jamiebuilds:
Same reason as everyone else that opened this

Daniel15:
I agree with the premise of this particular PR though... Since the license is a custom license rather than the regular MIT license, saying "MIT" might confuse automated systems that check licenses. Just removing the word "MIT" seems like a reasonable change?

boydrh:
MIT's Lincoln Laboratory has an extensive history actively supporting ICE's mission. Their support ranges from collaboration on [DoD projects](https://archive.org/stream/DTIC_ADA419971/DTIC_ADA419971_djvu.txt) to developing [next-gen imaging systems](http://news.mit.edu/2017/foliage-penetrating-ladar-technology-may-improve-border-surveillance-0227) specifically for border enforcement operations.

Will you be adding MIT to the blacklist while still calling it an "MIT" license?

Issue #1649, opened 2018-09-04
Skip private packages with no version field in lerna version


HosseinAgha:
As described by https://github.com/lerna/lerna/issues/810#issuecomment-301781565 and https://github.com/lerna/lerna/issues/810#issuecomment-346123436 some npm packages don't have (or need) a npm version field because none of the package's code can be used by other packages. e.g. http servers.  
 
## Expected Behaviour
Lerna should skip *_private packages_* with *_no version field_* in `lerna version`.  
IMHO the package author's intention is pretty obvious when they don't have a version field and also marked their package as private.   
If the package was not private and there were no version field in package.json we can warn user about the issue and suggest a version `0.0.0` for example, instead of showing and setting a wrong `null` value.

## Current Behaviour
As shown in https://github.com/lerna/lerna/issues/810#issuecomment-301781565 Lerna prompts for new version and changes the package version to `null` incorrectly.

## Steps to Reproduce (for bugs)
1. create a npm package with no version field
2. run `lerna version`
3. press enter when prompted for version

<details><summary>lerna.json</summary><p>

```json
{
  "version": "independent",
  "lerna": "3.2.1",
  "packages": [
    "projects/*"
  ],
  "useWorkspaces": true,
  "npmClient": "yarn",
  "command": {
    "run": {
      "npmClient": "yarn"
    }
  }
}
```
</p></details>

## Context
One of the main goals of using semantic versioning is to reduce breaking changes to minimum. So users of your code can use and update their dependencies with the ease of mind.  
Some npm packages does not ship in shape of a library. They use `package.json` for dependency management and something like docker for shipping the application code. This may not make much sense for libraries but is a common practice in application development.

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.2.1 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS High Sierra | 10.13.X |

HosseinAgha:
@evocateur thank you for your response.   
Thats exactly what I'm trying to say here. I don't expect lerna to handle my applications let alone bundle them!   
I also happen to have some npm packages in my monorepo that need to be versioned and published.  
All I want from lerna is not to suggest incorrect `null` versions when there is not version field in private packages.

evocateur:
Yeah, you're right, sorry. When a package is explicitly private, we shouldn't force folks to assign a version if they choose to remove it.

[...]

PR #1654, opened 2018-09-05
fix(version): don't change packages with no version field in lerna version

HosseinAgha:
## Description
Checkout #1649  
I could add hasVersion code to `collect-updates` function but I was not sure about it's footprint.

## How Has This Been Tested?
I wish to add specific tests for this scenario if the general idea is OK with you.  
Is it ok if I add new fixtures in `publish/__fixtures__` directory for this purpose?  

## Types of changes

## Checklist:


commands/version/index.js
```diff
@@ -136,6 +137,8 @@ class VersionCommand extends Command {
       this.execOpts,
       this.options
     );
+    // only update packages that already have version field
+    this.updates = this.updates.filter(node => hasVersion(node));
```
evocateur: I would prefer we chained the filter to the `collectUpdates` call:

```js
    this.updates = collectUpdates(
      this.packageGraph.rawPackageList,
      this.packageGraph,
      this.execOpts,
      this.options
    ).filter(node => {
      if (!node.version) {
        // a package may be unversioned only if it is private
        if (node.pkg.private) {
          this.logger.info("version", "Skipping unversioned private package %j", node.name);
        } else {
          throw new ValidationError(
            "ENOVERSION",
            dedent`
              A version field is required in ${node.name}'s package.json file.
              If you wish to keep the package unversioned, it must be made private.
            `
          );
        }
      }

      return !!node.version;
    });
```

Also, instead of warning, we should definitely be throwing an error (which logs an error message) when a non-private package is missing a version field. I inlined the method because we really don't have a lot of logic to encapsulate.

evocateur:
Also, please move this change to a feature branch (not `master`) of your fork. Pretty sure you can modify the PR in-place, no need to close/re-open. Thanks!

HosseinAgha:
1. Added the tests for not versioned packages.
2. I found no way to change the source branch of the PR. I suppose only target branch is editable in github. I can create another PR if having a feature branch is necessary.
3. Just found out that `yarn` completely ignores installing dependencies for packages with no version in workspaces (private or not private) (https://github.com/yarnpkg/yarn/issues/6371) so this change is useless for _lerna + yarn workspaces_ users unless yarn fixes its behaviour.  
4. There seems to be a problem with installing lerna dependencies in development. I use latest npm version (`6.4.1`) and I have to `npm install` twice for all sub dependencies to install.

HosseinAgha:
@evocateur I think this is good to go.

jayhogan:
This feature would be useful to me. Any idea when this will land? Any assistance I can offer?

PR #1665, opened 2018-09-08
fix(publish): allow forcePublish on 'publish' command when used with canary option

diervo:
## Description
Allows publish command to support `force-publish` in conjunction with `--canary`

## Motivation and Context
From the documentation technically `force-publish` only is allowed on version, I think it should also be supported in publish

## How Has This Been Tested?
Tested it locally, changes are pretty trivial

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -195,7 +195,7 @@ class PublishCommand extends Command {
   }
 
   detectCanaryVersions() {
-    const { bump = "prepatch", preid = "alpha", tagVersionPrefix = "v", ignoreChanges } = this.options;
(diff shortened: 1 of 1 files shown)


evocateur:
Oh, and don't worry about the AppVeyor failures, I'm wrestling with line endings or something now in `master`.

dotansimha:
Thanks @diervo ! we really needs it!!

evocateur:
I ended up fixing this in v3.3.2, thanks for identifying the root cause.

diervo:
@evocateur I was about to add the tests today! :)
Thanks a lot!

PR #1704, opened 2018-09-28
Fix last tag detection to also include tags in branches

HazAT:
## Description
Lerna fails to detect tags that are not on the current branch. 
This results in that if you tag for example on a `release` branch, `lerna changed` returns all changes until the last tag of the current branch.

Let's say you have the following setup:
screenshot 2018-09-28 13 29 06

When running `lerna changed` results in this output:

```
yarn lerna changed
yarn run v1.3.2
$ /Users/haza/Projects/sentry-javascript/node_modules/.bin/lerna changed
lerna notice cli v3.4.0
lerna info Looking for changed packages since 4.0.5
@sentry/browser
@sentry/core
@sentry/hub
@sentry/minimal
@sentry/node
@sentry/types
@sentry/utils
lerna success found 7 packages ready to publish
✨  Done in 0.95s.
```
When clearly there is a merged release branch with version `4.0.6`.
It's still true that `4.0.5` was tagged on the master branch.
`lerna` runs following command under the hood: `git describe --always --long --dirty --first-parent` which returns: `4.0.5-9-g89885fae` which is correct but not taking tags on other branches into the calculation.

This fix takes the last `tag` from all branches into account. 
If I run `lerna changed` with this fix:
```
yarn lerna changed
yarn run v1.3.2
$ /Users/haza/Projects/sentry-javascript/node_modules/.bin/lerna changed
lerna notice cli v3.4.0
lerna notice Current HEAD is already released, skipping change detection.
lerna info No changed packages found
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Command running under the hood is: `git describe --tags $(git rev-list --tags --max-count=1) --always --long --first-parent` which returns `4.0.6-0-gf474d18d` (which is correct).

We are using `lerna` for our javascript SDKs https://github.com/getsentry/sentry-javascript
If you need anything let me know.

## Types of changes

## Checklist:

### DIFF
--- a/utils/describe-ref/lib/describe-ref.js
+++ b/utils/describe-ref/lib/describe-ref.js
@@ -7,15 +7,16 @@ module.exports = describeRef;
 module.exports.parse = parse;
 module.exports.sync = sync;
 
-function getArgs(options) {
+function getArgs(options, lastRev) {
   const args = [
     "describe",
+    // fetch it from the latest tag, no matter in which branch
+    "--tags",
+    lastRev,
     // fallback to short sha if no tags located
     "--always",
     // always return full result, helps identify existing release
     "--long",
-    // annotate if uncommitted changes present
-    "--dirty",
     // prefer tags originating on upstream branch
     "--first-parent",
   ];
@@ -27,8 +28,18 @@ function getArgs(options) {
   return args;
 }
 
+function getRevOfLastTag(syncCall = false, options = {}) {
+  const args = ["rev-list", "--tags", "--max-count=1"];
+  if (syncCall) {
+    return childProcess.execSync("git", args, options);
+  }
+  return childProcess.exec("git", args, options);
+}
+
 function describeRef(options = {}) {
-  const promise = childProcess.exec("git", getArgs(options), options);
+  const promise = getRevOfLastTag(false, options).then(({ stdout }) =>
+    childProcess.exec("git", getArgs(options, stdout), options)
+  );
 
   return promise.then(({ stdout }) => {
     const result = parse(stdout, options);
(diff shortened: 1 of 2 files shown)


utils/describe-ref/lib/describe-ref.js
```diff
@@ -7,15 +7,16 @@ module.exports = describeRef;
 module.exports.parse = parse;
 module.exports.sync = sync;
 
-function getArgs(options) {
+function getArgs(options, lastRev) {
   const args = [
     "describe",
+    // fetch it from the latest tag, no matter in which branch
```
evocateur: That's not what the `--tags` option of `git describe` [does](https://git-scm.com/docs/git-describe#git-describe---tags). You're enabling "lightweight" tags to be consulted, which is _not_ correct.

utils/describe-ref/lib/describe-ref.js
```diff
@@ -7,15 +7,16 @@ module.exports = describeRef;
...
+    "--tags",
+    lastRev,
     // fallback to short sha if no tags located
     "--always",
     // always return full result, helps identify existing release
     "--long",
-    // annotate if uncommitted changes present
-    "--dirty",
```
evocateur: Why have you removed this?

HazAT:
```
git describe --tags $(git rev-list --tags --max-count=1) --always --long --dirty --first-parent
fatal: --dirty is incompatible with commit-ishes
```

HazAT:
We found a workaround, using `lerna exec`
`$ lerna exec --since $(git rev-list --tags --max-count=1)`

I am not sure if I can provide enough insights / examples to justify my change.
I will go ahead and close this PR, not sure if this needs a fix then, maybe someone picks this up in the future.

Issue #1668, opened 2018-09-10
recovering from failed publish


pantoninho:
I'm having some troubles recovering from a failed publish.

After using `lerna version`, I called `lerna publish from-git`, which started publishing all my packages.
One of them failed (internet failure while uploading), and now I'm not able to resume the process.

If I call `lerna publish` it will fail as soon as it tries to re-publish one of the successfully published packages (due to publishing an already existing version).

In the past there was a `bail` option that allowed me to ignore failed commands and that was exactly what I was looking for, but looks like it was removed.

Is there a solution for this problem already?
Thank you

evocateur:
Actually, `lerna publish` has never had a `--no-bail` option, unlike `lerna exec` or `lerna run`. Unfortunately, the "recovery" right now is, after it fails and leaves a mostly-dirty git tree, you have to go into the package directories one-by-one and run `npm publish <packageName>-<version>.tgz`. If the git tree has been cleaned already, you might be able to run `npm publish` in the leaf directories of the affected packages. And that ... really sucks.

#1648 gets closer to a "recovery" scenario that might help make this suck less. In an ideal world, recovering from a failed publish wouldn't require _ignoring_ failures, the command would be smart enough to skip packages that already have matching versions available in the configured registry.

evocateur:
Yes, I would appreciate seeing what a solution could look like. I like the `from-package` positional argument I described in #1648 as a gate for the preflight version checks.

PR #1708, opened 2018-10-02
WIP: feat(publish): Add a "from-git" argument

### LINKED ISSUE #1648: publish packages with their current version (from package)
See #1056 and #1610 for background.

I want to divide the publication process into 2 stages, similar to how it is done when publishing single npm packages in CI:

First I want to manually raise the package version on a branch:

`lerna version [bump] --no-git-tag-version`

Then, after PR approval and merge to master, I want to publish the bumped versions in CI:

`lerna publish --from-package`

Here, `--from-package` is a new option that publishes the versions that are already in the package.json files. I imagine the implementation would only attempt to publish versions that have not already been published.

chriscasola:
Publish un-published releases by reading versions from
the package.json files and publishing any that are not
already available in the registry.

Fixes #1648

## Description
This is an initial attempt at implementing `lerna publish from-package`. The "from-package" option will inspect each package.json and determine if the version has been published to the registry. If it hasn't it will publish it.

I think it should also create tags, since if the workflow is something like the following, the tags will not have been created by `lerna version`:

```bash
lerna version [bump] --no-git-tag-version
# create and review PR
# commit release to master
lerna publish from-package
```

More things I need to figure out:
- [ ] How should I go about creating a tag for each version if one doesn't exist?
- [ ] Tests

## Motivation and Context
This change will allow users to prepare a release, review it and commit it, and then later publish it in CI.

The change will also make it much easier to attempt to publish again if some or all of the packages failed to publish to the registry in a prior run of `lerna publish`.

This may address #1648 and #1668

## How Has This Been Tested?
Change is still incomplete, I'll update this section later.

## Types of changes

## Checklist:


chriscasola:
I'm not sure why the CI job for window/node10 failed, but I think I finished everything that I had left WIP.

mlavina:
Hello @evocateur 

Any chance you can take a look at this PR anytime soon? 

My team is also running into similar issues and this PR would be super helpful. Currently, we are forced to basically roll back the git changes and remove git tags and what not on intermittent fails. This would definitely be super helpful for us.

Thank you,
Michael

StarpTech:
@chriscasola it was an installation issue by travis itself. Could you retrigger the build?

StarpTech:
@chriscasola rebase from master the .travis.yml was changed completly!

StarpTech:
Hi,
according to #1648 I expected that the new command `lerna publish from-package` will tag the versions but in the implementation it won't. If you check #1648 we want to archieve following workflow:

1. I want to manually raise the package version on a branch:
```
lerna version [bump] --no-git-tag-version
```
2. Then, after PR approval and merge to master, I want to publish the bumped versions in CI:
```
lerna publish from-package
```
3. **(Currently missing)** The versions from the package.json are used to create new git-tags

4. **(Impossible when the version are tagged before CI)** After the successfully publish `lerna changed` would return no changes.

commands/publish/lib/get-unpublished-packages.js
```diff
@@ -0,0 +1,41 @@
...
+  Accept: "application/vnd.npm.install-v1+json",
+};
+
+function getUnpublishedPackages(project, opts) {
+  log.silly("getPackageVersions");
+
+  let chain = Promise.resolve();
+
+  const mapper = (unpublished, pkg) =>
+    fetch.json(`/${pkg.name}`, Object.assign({}, opts, registryOptions)).then(
```
StarpTech: You can't pass the configuration object to the `npm-registry-fetch` package. In my examples it will always fallback to NPM registry.

StarpTech: It's because the config object isn't compatible with `Object.assign` you need to use
```js
Object.assign({}, opts.snapshot, registryOptions)
```

commands/publish/lib/get-unpublished-packages.js
```diff
@@ -0,0 +1,41 @@
...
+const pReduce = require("p-reduce");
+
+module.exports = getUnpublishedPackages;
+
+// Only the abbreviated package metadata is needed to
+// determine which versions have been published. This
+// saves transfer time for packages with a lot of
+// history.
+const registryOptions = {
+  Accept: "application/vnd.npm.install-v1+json",
```
evocateur: I think we should ultimately be using [`pacote.packument()`](https://github.com/zkat/pacote#--pacotepackumentspec-opts) instead of calling `npm-registry-fetch` directly. It caches stuff, as well as handling this "corgi doc" header.

commands/publish/lib/get-unpublished-packages.js
evocateur: Using `pacote` instead will resolve this.

commands/publish/lib/get-unpublished-packages.js
evocateur: Of course I left out the link I was trying to add in the request for changes: https://github.com/lerna/lerna/blob/ad649bf965a6452d5f4aad8026f4ed62a548cada/commands/add/index.js#L165

evocateur:
@StarpTech 

> (Currently missing) The versions from the package.json are used to create new git-tags

This would only apply in independent versioning mode, yes? In the default "fixed" versioning, the tag will already be present when a partial publish occurs.

StarpTech:
> This would only apply in independent versioning mode, yes? In the default "fixed" versioning, the tag will already be present when a partial publish occurs.

@evocateur No in both because it doesn't create another version "fixed" versioning in the current implementation. Please have a look at [lerna-workflow](https://github.com/StarpTech/lerna-workflow) to better understand the goals with #1648

Clarence-pan:
It's very useful. 👍
Does anyone know when will it be published to npm? And which version?

Issue #1691, opened 2018-09-19
Question: How can I publish only one package?


Gcaufy:
How can I publish only one package?

If I have 5 packages, and all those 5 packages are changed. They are independent, and I just want to publish one of them, because the others they are not ready yet.

## Possible Solution
Can we simply add a `Ignore` option when doing `leran publish`?

```
$ lerna publish
>   Patch (x.x.x)
    Minor (x.x.x)
    .....
    Ignore
```

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.4.0 |
| `npm --version`  | 6.0.1 |
| `node --version` | 8.9.4 |

| OS | Version |
| --- | --- |
| macOS Sierra | --- |

jimmytb:
I believe you can't, you have to do it manually by going to the folder and run:

``` npm publish ```

Than manually update the version of the package in the other packages.

This is the way i do it. Hope there is an other way

Gcaufy:
@jimmytb yup, this is what I am doing right now. I hope lerna can support publish one single package. then you don't care about the package.json and the git tag and git commits .

evocateur:
You can't. A Lerna-managed monorepo doesn't work that way.

Gcaufy:
@evocateur any reason why not to support it?

evocateur:
Because it breaks the fundamental mechanism Lerna uses to determine which packages need to be published? Absolutely nothing stopping you from _not_ placing several packages in the same monorepo that apparently shouldn't be.

PR #1710, opened 2018-10-02
feat(version): Add version SKIP to allow user skip the package

Gcaufy:
Add a "SKIP" choose when select a package version.

## Description
If user do not want to publish the package, he can choose "SKIP".

## Motivation and Context
For example. You have many independent packages, and you updated 10+ packages, and only one package need a emergency release, the others are not ready yet. Currently you have no way to publish the only one package except manually publish it.
So I added the "SKIP" option, then user can ignore the package, if they don't want to publish it for now.

FYI: https://github.com/lerna/lerna/issues/1691

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -100,7 +100,9 @@ class PublishCommand extends Command {
       this.updatesVersions = new Map(result.updatesVersions);
 
       this.runPackageLifecycle = createRunner(this.options);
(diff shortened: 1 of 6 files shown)


evocateur:
I'm sorry, this fundamentally breaks the contract of lerna: Every package that has changed since the most recent git tag will be published with a new version. Skipping a package during publish is an _enormous_ footgun, and will not be accepted.

PR #1712, opened 2018-10-02
feat(version): Add `--include-merged-tags` option

HazAT:
## Description
Disclaimer:
This is the successor PR to 
https://github.com/lerna/lerna/pull/1704

`lerna` fails to detect/consider tags from merged branches.
This results in that if you tag for example on a `release` branch, `lerna changed` returns all changes until the last tag of the current branch.

Let's say you have the following setup:
screenshot 2018-09-28 13 29 06

When running `lerna changed` results in this output:

```
yarn lerna changed
yarn run v1.3.2
$ /Users/haza/Projects/sentry-javascript/node_modules/.bin/lerna changed
lerna notice cli v3.4.0
lerna info Looking for changed packages since 4.0.5
@sentry/browser
@sentry/core
@sentry/hub
@sentry/minimal
@sentry/node
@sentry/types
@sentry/utils
lerna success found 7 packages ready to publish
✨  Done in 0.95s.
```
When clearly there is a merged release branch with version `4.0.6`.
It's still true that `4.0.5` was tagged on the master branch.

`lerna` runs following command under the hood: `git describe --always --long --dirty --first-parent` which returns: `4.0.5-9-g89885fae` which is correct but not taking tags on other branches into the calculation.

A new option `include-merged-tags` therefore was added to support tags from merged branches.
It changes the previous `git describe` command to 
`git describe --always --long --dirty --tags` which results in `4.0.6-8-g6fbeff9a` which in the example stated above is correct.

We are using `lerna` for our javascript SDKs https://github.com/getsentry/sentry-javascript

## Motivation and Context

We think that tagging in a release branch should be a valid scenario the `lerna` should be able to support. Right now the only way to work around this limitation is if we would work around lerna for bumping and releasing.

## How Has This Been Tested?

Locally in our repo + added test cases highlighting the new behavior.

## Types of changes

## Checklist:


HazAT:
@evocateur Sorry for the PR spam, while I stated in the PR before that we found a workaround with `--since` which is true to basically replace `lerna changed` but it's not allowing us to use `lerna version`.

**This PR is still WIP** I just want to know if you consider merging this new feature at all, if so please let me know I will finish the PR (write tests + docs for the new option).

Thanks for your time and the great project.

HazAT:
OK, added docs + tests, one existing test is failing (that I did not touch), not sure why though.

dzsodzso63:
> Sorry for the PR spam, while I stated in the PR before that we found a workaround with `--since` which is true to basically replace `lerna changed` but it's not allowing us to use `lerna version`.

@HazAT can you please explain this "workaround"? we have this issue since upgrading to lerna 3, that we cannot determine changed packages because of the `--first-parent` param (publish from release branches + PR), but `--since` switch doesn't work since lerna 3. What we should do?

HazAT:
@dzsodzso63 So we found a workaround for changed, using `--since` 
`yarn lerna exec --since $(git rev-list --tags --max-count=1) pwd`
But it doesn't work for `version` / `publish` so right now we are using my fork right now.

HazAT:
@dzsodzso63 I mean I don't see why this shouldn't be merged but it's not in my power ;)
I am happy to do anything to get this merged since we also don't want to rely on my fork for this ^^

danielcondemarin:
Hi @HazAT @dzsodzso63 Did you consider rebasing over true merges? Seems to work fine if you rebase your release branch onto master.

HazAT:
Rebased

dzsodzso63:
> Hi @HazAT @dzsodzso63 Did you consider rebasing over true merges? Seems to work fine if you rebase your release branch onto master.

Maybe I missing something here, but in my understanding, when I rebase the publisher branch to master, then the sha changes, so the tags created by lerna will point to non-existing commits. What do I miss?

HazAT:
@dzsodzso63 @danielcondemarin Right, we create the tag in our release branch, rebasing would create a new sha. 
We really want to have a "snapshot" of the tagged commit that we built. 
While rebasing and tagging afterward will probably work 99% of the time we run all tests on this specific commit/tag and want to release it without building it again.

dzsodzso63:
@HazAT ok, then we are on the same page (and we cannot even think about publishing after merge, as  we have protected master branch and we cannot commit to that directly)

HazAT:
@evocateur Is there anything left to do for this getting merged?

utils/describe-ref/lib/describe-ref.js
```diff
@@ -24,11 +24,18 @@ function getArgs(options) {
     args.push("--match", options.match);
   }
 
+  if (commandOptions.includeMergedTags) {
+    // we want to consider all tags, also from merged branches
+    args = args.filter(arg => arg !== "--first-parent");
+    // also include lightweight tags
+    args.push("--tags");
```
StarpTech: Did you consider these regarding lightweight tags?

* https://github.com/lerna/lerna/blob/aa4ae510b9c0a0935f0b1989859ba873a1e864cc/CHANGELOG.md#breaking-changes-2
* https://github.com/lerna/lerna/blob/f674f354f0260c57d885c181e16c9ce23ac252a8/doc/troubleshooting.md#publish-does-not-detect-manually-created-tags-in-fixed-mode-with-githubgithub-enterprise

HazAT: This doesn't really work for our case. As explained we want to tag in a release branch that gets merged into master. But the tagged commit is on the branch.

StarpTech: I only want to point to the fact that lightweight tags aren't coompatible with lerna and it seems that it will hard to land this.

HazAT: I see but it isn't really just a lightweight tag, right?

commands/changed/command.js
```diff
@@ -15,7 +15,13 @@ exports.describe = "List local packages that have changed since the last tagged
 exports.builder = yargs => {
   listable.options(yargs);
 
-  return versionOptions(yargs, "changed");
+  const newYargs = versionOptions(yargs, "changed");
+  newYargs.option("include-merged-tags", {
```
evocateur: Since this was already added in `commands/version/command.js`, you don't need to duplicate it here? `versionOptions()` was designed to avoid this duplication. It is hidden, true, but it still works.

commands/version/__tests__/version-command.test.js
```diff
@@ -563,4 +563,17 @@ describe("VersionCommand", () => {
...
+  describe("--include-merged-tags", () => {
+    it("accepts --include-merged-tags", async () => {
+      const testDir = await initFixture("normal");
+      await lernaVersion(testDir)("--include-merged-tags", "--yes", "patch");
+
+      expect(PromptUtilities.select).not.toBeCalled();
+      expect(PromptUtilities.confirm).not.toBeCalled();
+
+      const message = await getCommitMessage(testDir);
+      expect(message).toBe("v1.0.1");
```
evocateur: This isn't actually testing `--include-merged-tags`, though. I'd rather the tests you added to `changed-command.test.js` be moved to a separate test file under `commands/version/__tests__`, like `version-include-merged-tags.test.js`.

commands/changed/__tests__/changed-command.test.js
```diff
@@ -41,6 +44,26 @@ const setupGitChanges = async (cwd, filePaths) => {
   await gitCommit(cwd, "Commit");
 };
 
+const setupGitChangesWithBranch = async (cwd, masterPaths, branchPaths) => {
```
evocateur: This helper and the tests should be moved to a focused test file under the version command, as described previously. `lerna changed` is just a lightweight wrapper around the internals used by `lerna version`, it doesn't have any logic itself beyond the list output options.

commands/changed/README.md
```diff
@@ -27,3 +27,12 @@ package-2
 Unlike `lerna ls`, however, `lerna changed` **does not** support [filter options](https://www.npmjs.com/package/@lerna/filter-options), as filtering is not supported by `lerna version` or `lerna publish`.
 
 `lerna changed` also supports all the flags supported by [`lerna version`](https://github.com/lerna/lerna/tree/master/commands/version#options), but the only relevant one is [`--ignore-changes`](https://github.com/lerna/lerna/tree/master/commands/version#--ignore-changes).
+
+### `--include-merged-tags`
```
evocateur: This duplicated options block is not necessary. Adding a link to the docs under `commands/version` directly is fine.

helpers/git-checkout/index.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+const execa = require("execa");
+
+module.exports = gitCheckout;
```
evocateur: Thanks for these helpers, they are handy :)

utils/describe-ref/lib/describe-ref.js
evocateur: This is not going to fly. Lightweight tags are not intended to be used for tagging releases.

HazAT:
@evocateur Thanks for your review, I made the changes you requested.

PR #1727, opened 2018-10-12
fix(publish): Generate safer canary version

peterbakonyi05:
Previous way of generating canary version in lerna v3 (using `+`) is not compatible with all npm registries. We are using [ProGet](https://inedo.com/proget), and after upgradig to lerna 3, publishing a canary version fails because of the `+` sign. I didn't find any documentation on what characters are allowed in a prerelease version. Based on [npm's semver calculator](https://semver.npmjs.com/) both `+` and `.` should be fine. If you don't agree with the fix, could you kindly propose an alternative solution (maybe accepting a cli argument to override `+` if needed)? 

*Note: will update the specs once I get feedback if the change could be accepted*

## Motivation and Context
With `+` sign in package version, not all private registries will work with lerna v3 and this is blocking the upgrade.

## How Has This Been Tested?
Tested locally on Proget 4.7.14

## Types of changes

## Checklist:
- [ x] My code follows the code style of this project.

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -258,7 +258,7 @@ class PublishCommand extends Command {
 
       // semver.inc() starts a new prerelease at .0, git describe starts at .1
       // and build metadata is always ignored when comparing dependency ranges
-      return `${nextVersion}-${preid}.${Math.max(0, refCount - 1)}+${sha}`;
+      return `${nextVersion}-${preid}.${Math.max(0, refCount - 1)}.${sha}`;
     };
 
     if (this.project.isIndependent()) {


evocateur:
This is intentional, as the commit SHA is build metadata. It cannot use a `.` delimiter because then it would become a part of the prerelease specifier, which is inappropriate for the purpose of a canary version. Build metadata is necessary to ensure the uploaded tarball is unique, and is supported by the public registry.

Here's the spec: https://semver.org/#spec-item-10

Here's `semver`'s support: https://github.com/npm/node-semver/blob/master/range.bnf#L12

With all due respect, this isn't going to change. ProGet needs to fix their support. Presumably you're paying for it, so they have incentive enough.

peterbakonyi05:
@evocateur Thank you for the reply and pointing me to the right direction! I agree completely and will flag it to ProGet

PR #1751, opened 2018-10-30
docs: note some disadvantages with hoisting

joeytwiddle:
## Motivation and Context
Before enabling an option, I want to know what are the advantages and disadvantages of doing so.

## Description
Documented one possible disadvantage of hoisting:

- Accidentally require-ing a common dependency without adding it to `package.json`

## How Has This Been Tested?
I created three packages, two with a common dependency, and one without.

The third without was still able to require the common dependency, although it had not specified it in its `package.json`.

## Checklist:

## Additional
If there are any disadvantages which are not documented here, please let me know. I may be able to add them.

### DIFF
--- a/doc/hoist.md
+++ b/doc/hoist.md
@@ -32,7 +32,9 @@ When the `--hoist` flag is used:
   scripts continue to work unmodified.
 * Well-behaved Node-based software should continue to work unmodified.
 
-## Module resolution
+## Disadvantages with hoisting
+
+### Module resolution
 
 The [Node module resolution algorithm](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders)
 is recursive: When looking for package `A`, it looks in a local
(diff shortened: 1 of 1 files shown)


bertho-zero:
You can spot these errors with eslint.

joeytwiddle:
@bertho-zero Is this the plugin that can warn about that? [eslint-plugin-dependencies](https://github.com/zertosh/eslint-plugin-dependencies) If so I will add a link to it.

bertho-zero:
It's this plugin: https://github.com/benmosher/eslint-plugin-import

joeytwiddle:
That looks great, thank you~

wbern:
This is a very valid hint. I really want more information like this when it comes to lerna, so I know why the import works even though I feel like it shouldn't.

andrewcharnley:
Agree. Also many npm commands won't work - npm outdated and npm audit being two.

Also should compare 'lerna link convert'. See https://github.com/lerna/lerna/issues/1788

Issue #1357, opened 2018-04-04
Lerna will always publish all packages even if some are not changed


egoist:
## Expected Behavior
Only publish updated packages.

## Current Behavior

All packages will be published, run `lerna updated` it lists all packages even if none of them are updated.

## Possible Solution

Took a quick look at the code, not entirely sure how it works but seems it's related to:

https://github.com/lerna/lerna/blob/4c4aceee84b39eb399e968a0c87241ee59edce4f/utils/collect-updates/collect-updates.js#L46-L55

Does this mean the package will always be published if the current version is a prelease like `x.x.x-beta.x` and no `cdVersion` is present?

It works properly when running with `lerna updated --cd-version prerelease`, is this intended?

## Steps to Reproduce (for bugs)

1. Have some leaf packages whole versions are `x.x.x-beta.x`
2. Run `yarn lerna updated`

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.0.0-beta.14 (same for 2.7, 2.8)|
| `npm --version`  | 5.5.1 |
| `yarn --version` | 1.5.1 |
| `node --version` | 8.9.3 |

amkoehler:
> The assumption is that you are "graduating" the prerelease out of alpha/beta/whatever, and no "normal"-versioned package (in a lerna context) should ever depend on prereleased local dependencies.

This applies when using independent versioning too, correct? Not OP, but I similarly found that `lerna updated` always returns true if my packages were currently in a prerelease state (e.g. `4.0.0-2`). Since the first step of `lerna publish` is to run the equivalent of `lerna updated`, it makes sense why the packges in a prerelease state would always get published.

lili21:
same issue here. weird. I'm using `lerna@2.9.0`.

maybe `lerna` can output more information for debugging. like what make `lerna` think a **package** changed ?

```json
// lerna.json
{
  "lerna": "2.9.0",
  "packages": [
    "packages/*"
  ],
  "npmClient": "yarn",
  "useWorkspaces": "true",
  "version": "independent"
}
```

davidhouweling:
seems like a workspaces issue?

lili21:
`lerna updated` should tell me what exactly **updated** in code level. what I changed ?

`lerna updated --detail` ?

evocateur:
That’s `lerna diff`, but it does not show dependent bumps.

amkoehler:
[...]

PR #1778, opened 2018-11-13
fix(describe-ref): add --tags arg to git describe

Hoishin:
## Description

Run `git describe` with `--tags` arg to ensure it shows the most recent tag in history.

## Motivation and Context

When you manually put tags, but want the manual tags to be referenced in `lerna version`, this change is required.
[More information about annotated tags and unannotated tags](https://stackoverflow.com/a/11514139)

Fixes https://github.com/lerna/lerna/issues/1357

## How Has This Been Tested?

Changed locally installed lerna and confirmed that it shows the correct packages to version-up.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. (Don't think I have to)

### DIFF
--- a/utils/describe-ref/__tests__/describe-ref.test.js
+++ b/utils/describe-ref/__tests__/describe-ref.test.js
@@ -5,7 +5,7 @@ jest.mock("@lerna/child-process");
 const childProcess = require("@lerna/child-process");
 const describeRef = require("../lib/describe-ref");
 
-const DEFAULT_ARGS = ["describe", "--always", "--long", "--dirty", "--first-parent"];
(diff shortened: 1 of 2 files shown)


evocateur:
From the `git-tag` [docs](https://git-scm.com/docs/git-tag#_description):
> Annotated tags are meant for release while lightweight tags are meant for private or temporary object labels. For this reason, some git commands for naming objects (like `git describe`) will ignore lightweight tags by default.

Lerna uses annotated tags when versioning, uses `git describe` when determining previous releases, and neither of those will change.

Manual version tagging, if you need it, need only include a `-m` argument to remain compatible.

PR #1791, opened 2018-11-22
feat(publish): Add option `--no-git-reset` to leave unstaged changes in working tree

yoitsro:
## Description
Adds a flag which allows the user to specify whether or not the working tree should be reset or not. Defaults to true so the tree is always reset.

## Motivation and Context
When used in CI (specifically for my staging builds), I want lerna to be able to publish canary versions of my packages which can then be consumed in a subsequent docker build step. Right now, once the package versions are bumped, the working tree is then reset meaning that the docker build of the package which depends on the bumped package, refers to the original package version.

Relates to this comment https://github.com/lerna/lerna/issues/962#issuecomment-440826826 in #962.

## How Has This Been Tested?
 - [x] Unit tests
Added a unit test to ensure that when `--no-git-reset` is specified, `gitCheckout` isn't called.
 - [x] Integration tests
Added an integration test to ensure that when `--no-git-reset` is specified, the working tree is left dirty.

## Types of changes

## Checklist:
One test was already failing on my machine:
`isAnythingCommitted without and with a commit`

### DIFF
--- a/integration/lerna-publish-canary.test.js
+++ b/integration/lerna-publish-canary.test.js
@@ -5,6 +5,7 @@ const path = require("path");
 const cliRunner = require("@lerna-test/cli-runner");
 const commitChangeToPackage = require("@lerna-test/commit-change-to-package");
 const gitTag = require("@lerna-test/git-tag");
+const gitStatus = require("@lerna-test/git-status");
 const cloneFixture = require("@lerna-test/clone-fixture")(
   path.resolve(__dirname, "../commands/publish/__tests__")
 );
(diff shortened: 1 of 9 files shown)


integration/lerna-publish-canary.test.js
```diff
@@ -40,3 +41,42 @@ Successfully published:
...
+ - package-2 => 1.0.1-alpha.0+SHA
+ - package-3 => 1.0.1-alpha.0+SHA
+
+Successfully published:
+ - package-1@1.0.1-alpha.0+SHA
+ - package-2@1.0.1-alpha.0+SHA
+ - package-3@1.0.1-alpha.0+SHA
+`);
+
+  const { stdout: statusStdOut } = await gitStatus();
```
evocateur: I'm not sure we need an entire helper just to run `git status`, and in any case, a better snapshot would be tightly focused on _just_ the modified files, which is already used [elsewhere](https://github.com/lerna/lerna/blob/b69a7282b6841f2b40f94e6a7df5ecf565be97a1/commands/publish/__tests__/git-checkout.test.js#L15):
```suggestion
  const modified = await execa.stdout("git", ["ls-files", "--modified"], { cwd });
```

`modified` in this case is just a newline-separated list of relative file paths, easily snapshotted inline.

Issue #1799, opened 2018-11-28
'lerna add' creates dependency on '^undefined' tag if package.json is missing version


DLehenbauer:
If package 'foo' does not have a "version" field in it's package.json, ```lerna add foo --scope bar``` will add the following to bar's package.json:

```
  "dependencies": {
    "foo": "^undefined",
  }
```

...and then crash, leaving the project in a broken state until the invalid tag is manually cleaned up.
## Expected Behavior

Verify that version is valid before modifying package.json and emit a polite error if necessary.

## Current Behavior

```
$ npx lerna add foo --scope bar
lerna notice cli v3.5.0
lerna ERR! Error: Invalid tag name "^undefined": Tags may not have any characters that encodeURIComponent encodes.
lerna ERR!     at invalidTagName (.../node_modules/npm-package-arg/npa.js:87:15)
lerna ERR!     at fromRegistry (.../node_modules/npm-package-arg/npa.js:290:13)
lerna ERR!     at Function.resolve (.../node_modules/npm-package-arg/npa.js:77:12)
lerna ERR!     at Object.keys.forEach.depName (.../node_modules/@lerna/package-graph/index.js:109:30)
lerna ERR!     at Array.forEach (<anonymous>)
lerna ERR!     at PackageGraph.forEach (.../node_modules/@lerna/package-graph/index.js:103:38)
lerna ERR!     at Map.forEach (<anonymous>)
lerna ERR!     at new PackageGraph (.../node_modules/@lerna/package-graph/index.js:92:10)
lerna ERR!     at chain.then.packages (.../node_modules/@lerna/command/index.js:237:27)
lerna ERR! lerna Invalid tag name "^undefined": Tags may not have any characters that encodeURIComponent encodes.

$npx lerna bootstrap
(same as above)
```

## Possible Solution

Seems like a pretty easy fix to vet version before modifying bar/package.json

## Steps to Reproduce (for bugs)

1.  Start w/a simple lerna repro
2.  Remove "version" filed from one package's package.json
3.  Add a new dependency via 'npx lerna add foo --scope bar'
4.  Note that `npx lerna bootstrap' also now fails.

<details><summary>lerna.json</summary><p>

```json
{
  "packages": [
    "packages/*",
  ],
  "version": "0.1.0"
}
```
</p></details>

## Context

Small amount of time figuring out the cause of the error (foo/package.json), followed by a small amount of time figuring out that (bar/package.json) had been modified/corrupted.

## Your Environment

[...]

PR #1803, opened 2018-12-02
[#1799]: 'lerna add' creates dependency on '^undefined' tag if package.json is missing version

doug-wade:
Check if the tag would be undefined and throw if it would.

## Description

If the tag equals `"^undefined"`, I throw an error with the message "When adding a dependency, it must have a version."

## Motivation and Context

#1799 

## How Has This Been Tested?

I wrote a unit test.

## Types of changes

## Checklist:


commands/add/index.js
```diff
@@ -130,6 +130,10 @@ class AddCommand extends Command {
       const deps = this.getPackageDeps(pkg);
       const range = getRangeToReference(this.spec, deps, pkg.location, this.savePrefix);
 
+      if (range === "^undefined") {
+        throw new Error("When adding a dependency, it must have a version.");
```
evocateur: This should be an instance of `ValidationError` (already imported in this file), otherwise it will dump a log and echo a stack, which really isn't what we want.

evocateur:
#1804 was a bit closer to the origin of the problem. Thanks all the same!

PR #1853, opened 2019-01-08
fix(describe-ref): properly handle sha-like tag names

arahansen:
I ran into an issue trying to publish my changes. `describe-ref` would attempt to find the tags from the upstream, however it incorrectly treated the output as a "minimal sha" as if no tags were found.

This change more strictly checks for minimal sha output. I'm not 100% sure what the output is supposed to look like for these "minimal shas". However, the test I added matches the shape of my repo's tags, and the following is the output when trying to publish without these changes:

```
lerna ERR! Error: Command failed: git rev-list --count 20190107
[Step 2/3] lerna ERR! fatal: ambiguous argument '20190107': unknown revision or path not in the working tree.
[Step 2/3] lerna ERR! Use '--' to separate paths from revisions, like this:
[Step 2/3] lerna ERR! 'git <command> [<revision>...] -- [<file>...]'
[Step 2/3] lerna ERR! 
[Step 2/3] lerna ERR! 
[Step 2/3] lerna ERR!     at makeError (~/node_modules/execa/index.js:174:9)
[Step 2/3] lerna ERR!     at Function.module.exports.sync (~/node_modules/execa/index.js:338:15)
[Step 2/3] lerna ERR!     at Object.execSync (~/node_modules/@lerna/child-process/index.js:22:16)
[Step 2/3] lerna ERR!     at parse (~/node_modules/@lerna/describe-ref/lib/describe-ref.js:65:35)
[Step 2/3] lerna ERR!     at Function.sync (~/node_modules/@lerna/describe-ref/lib/describe-ref.js:50:18)
[Step 2/3] lerna ERR!     at collectUpdates (~/node_modules/@lerna/collect-updates/collect-updates.js:24:56)
[Step 2/3] lerna ERR!     at VersionCommand.initialize (~/node_modules/@lerna/version/index.js:144:20)
```

even though `git describe` seems to be producing a valid response.

## Types of changes

## Checklist:

### DIFF
--- a/utils/describe-ref/__tests__/describe-ref.test.js
+++ b/utils/describe-ref/__tests__/describe-ref.test.js
@@ -149,4 +149,16 @@ describe("describeRef.parse()", () => {
       sha: "a1b2c3d",
     });
   });
+
+  it("should return metadata for tag names that are sha-like", () => {
+    const result = describeRef.parse("20190104-5-g6fb4e3293");
+
+    expect(result).toEqual({
+      isDirty: false,
+      lastTagName: "20190104",
+      lastVersion: "20190104",
+      refCount: "5",
+      sha: "6fb4e3293",
+    });
+  });
 });
--- a/utils/describe-ref/lib/describe-ref.js
+++ b/utils/describe-ref/lib/describe-ref.js
@@ -56,10 +56,11 @@ function sync(options = {}, includeMergedTags) {
 }
 
 function parse(stdout, options = {}) {
+  const minimalShaRegex = /^([0-9a-f]{7,40})(-dirty)?$/;
   // when git describe fails to locate tags, it returns only the minimal sha
-  if (/^[0-9a-f]{7,40}/.test(stdout)) {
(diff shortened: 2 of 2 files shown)


utils/describe-ref/lib/describe-ref.js
```diff
@@ -56,10 +56,11 @@ function sync(options = {}, includeMergedTags) {
 }
 
 function parse(stdout, options = {}) {
+  const minimalShaRegex = /^([0-9a-f]{7,40})(-dirty)?$/;
```
evocateur: So basically all we're doing is adding a trailing `$` to the combined regex? Huh.

utils/describe-ref/__tests__/describe-ref.test.js
```diff
@@ -149,4 +149,16 @@ describe("describeRef.parse()", () => {
       sha: "a1b2c3d",
     });
   });
+
+  it("should return metadata for tag names that are sha-like", () => {
```
evocateur: Thanks so much for the test!

Issue #1767, opened 2018-11-07
RFC: Improving `lerna publish`


ThisIsMissEm:
Proposal for the future of lerna publish:

- deprecate automatic versioning, use `lerna version` first instead
  - this enables a more reliable & retryable `lerna publish` and removes functionality that’s duplicated between `lerna version` and `lerna publish`
- switch to libnpmpublish instead of shelling out to npm/yarn
  - This allows us to avoid a lot of the weird issues that can happen due to things like $PATH, different versions of npm/yarn, and the different capabilities of each.
- separate tagging as “latest” (dist-tag) from publishing the tarball, as to make “lerna publish” not give partial “latest” releases
  - Essentially whilst a publish is ongoing, installs should not fail
  - This also moves us towards a more “release” based model, which seems to be what people intend (looking at apollo/babel/etc)
- implement retries via checking the registry for each package at a given version, giving the option to abort the publish or continue
  - this allows for partial publishes to succeed, related to splitting versioning and publishing
- switch to serial/sequential/queue-based publishing of packages
  - This gives better error handling and OTP support
  - Concurrent publishes lead to a lot of error conditions resulting in complex code & hard to debug errors
- remove execScripts
  - this functionality is only used by a small subset of lerna users, and is probably better done via a custom “publish” command in your package.json at root level. Or via a custom wrapler around lerna

These are my initial thoughts, and I’m definitely interested to keep discussing these to inform development decisions.

The above list is based on some private conversation with various people, and opens up doors for future developments or improvements whilst greatly simplifying the codebase.

mjhenkes:
+1 to separating versioning from publishing.

In our lerna workflow many developers work on the various packages but very few have privileges to publish. This creates a situation where the publisher needs to know at publish time how to version a component. The developer pushing changes should know how their changes affect the version and ideally would be able to indicate that prior to publish time.

Secondly separating versioning from publishing would allow us to create an automated job to publish the packages. No versions choices would need to be made at publish time.

evocateur:
@mjhenkes It's important to note that versioning (`lerna version`) _is_ separate from publishing (`lerna publish`) today, it's just that `lerna publish` isn't quite quick enough on the uptake to realize this without a special `from-git` positional argument. You can pretty much call `lerna publish from-git` unconditionally in CI, as it will just no-op if the current revision lacks an annotated tag (or tags) created  by `lerna version`.

ThisIsMissEm:
[...]

PR #1862, opened 2019-01-09
feat(publish): Remove `--require-scripts`

evocateur:
## Description

In order to simplify publish logic, remove "sidecar" lifecycle support.

Holding as "WIP" until other breaking changes can be batched together.

## Motivation and Context

Refs #1767

## How Has This Been Tested?

Tests pass (there were none for the feature itself)

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -107,7 +107,14 @@ class PublishCommand extends Command {
     }
 
     if (this.options.requireScripts) {
-      this.logger.info("require-scripts", "enabled");
+      throw new ValidationError(
+        "removed",
+        [
+          "--require-scripts has been removed.",
+          "Execute these scripts from the appropriate package lifecycle.",
+          "See https://docs.npmjs.com/misc/scripts for documentation.",
+        ].join("\n")
+      );
     }
 
     // npmSession and user-agent are consumed by npm-registry-fetch (via libnpmpublish)
@@ -571,19 +578,6 @@ class PublishCommand extends Command {
     });
   }
 
-  execScript(pkg, script) {
-    const scriptLocation = path.join(pkg.location, "scripts", script);
-
-    try {
-      // eslint-disable-next-line import/no-dynamic-require, global-require
-      require(scriptLocation);
-    } catch (ex) {
-      this.logger.silly("execScript", `No ${script} script found at ${scriptLocation}`);
-    }
-
-    return pkg;
-  }
-
   removeTempLicensesOnError(error) {
     return Promise.resolve()
       .then(() =>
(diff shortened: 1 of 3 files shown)


ThisIsMissEm (dismissed):
LGTM, should we bail/error out if a user passes `--require-scripts`?

evocateur:
I believe yargs will throw an error when it fails to recognize `--require-scripts`, but it won't be a particularly helpful message by default.

commands/publish/index.js
```diff
@@ -62,7 +63,10 @@ class PublishCommand extends Command {
     }
 
     if (this.options.requireScripts) {
-      this.logger.info("require-scripts", "enabled");
+      throw new ValidationError(
+        "removed",
+        "--require-scripts has been removed.\nExecute these scripts from the appropriate package lifecycle."
```
evocateur: Hopefully this is clear enough?

screen shot 2019-01-11 at 09 50 18

ThisIsMissEm:
Perhaps we should also add a quick page into the docs explaining how to migrate, then we could link to it from the error message.

nicolo-ribaudo:
Note that Babel currently uses `--require-scripts`. I will check if/how we can migrate.

evocateur:
Only four places that I can find, briefly searching?

1. https://github.com/babel/babel/blob/master/packages/babel-polyfill/scripts/prepublish.js
2. https://github.com/babel/babel/blob/master/packages/babel-standalone/scripts/prepublish.js
3. https://github.com/babel/babel/blob/master/packages/babel-preset-env-standalone/scripts/prepublish.js
4. https://github.com/babel/babel/blob/master/packages/babel-polyfill/scripts/postpublish.js

None of these seem beyond the capabilities of a lifecycle script, as far as I can tell.

commands/publish/index.js
nicolo-ribaudo: If you want to make it clearer, you could add a link to https://docs.npmjs.com/misc/scripts

evocateur: Done!

mojavelinux:
I use this feature extensively in the release for Antora (see https://gitlab.com/antora/antora/-/blob/master/releasing.adoc#readme-dance). I use to update various files in the repository (to propagate the version number and update dates) as well as to feed a consolidated README file to the npm registry. Lerna would be substantially less powerful without this feature, which works very well.

evocateur:
@mojavelinux I don't see how a proper `prepack` and `postpack` in each leaf cannot replace the convoluted copying and pasting of script files currently employed.
```json
  "scripts": {
    "prepack": "node ../../scripts/prepublish.js",
    "postpack": "node ../../scripts/postpublish.js"
  }
```

(or, better yet, a root `prepack`/`postpack` that loops through all the leaves instead of a repetitive per-leaf `scripts` block)

PR #1863, opened 2019-01-09
feat(publish): Use temporary dist-tags by default

evocateur:
## Description

The previous default was _not_ to use temporary dist-tags during the `lerna publish` process, and thus a partially successful publish (due to authentication errors and the like) would result in a broken `latest` dist-tag.

Pass `--no-temp-tag` to restore the previous, slightly more dangerous behavior.

## Motivation and Context

#1862 should be merged first.

Refs #1767

## How Has This Been Tested?

moar tests

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/README.md
+++ b/commands/publish/README.md
@@ -55,13 +55,13 @@ This is useful when a previous `lerna publish` failed to publish all packages to
 - [`--ignore-scripts`](#--ignore-scripts)
 - [`--ignore-prepublish`](#--ignore-prepublish)
 - [`--no-git-reset`](#--no-git-reset)
+- [`--no-temp-tag`](#--no-temp-tag)
 - [`--no-verify-access`](#--no-verify-access)
 - [`--otp`](#--otp)
 - [`--preid`](#--preid)
 - [`--pre-dist-tag <tag>`](#--pre-dist-tag-tag)
 - [`--registry <url>`](#--registry-url)
 - [`--tag-version-prefix`](#--tag-version-prefix)
-- [`--temp-tag`](#--temp-tag)
 - [`--yes`](#--yes)
 
 ### `--canary`
@@ -170,6 +170,14 @@ To avoid this, pass `--no-git-reset`. This can be especially useful when used as
 lerna publish --no-git-reset
 ```
 
+### `--no-temp-tag`
+
+By default, `lerna publish` first publishes all changed packages to a temporary dist-tag (`lerna-temp`)
+and then after all are successful moves the new version(s) to the dist-tag configured by [`--dist-tag`](#--dist-tag-tag) (default `latest`).
+This is intended to help avoid "partial" releases (due to network, authentication, or other failure) corrupting consumer installs with conflicting metadata.
+
+If you do not need this insurance, pass `--no-temp-tag` (or set `command.publish.tempTag` to `false` in lerna.json) to disable this behavior.
+
 ### `--no-verify-access`
 
 By default, `lerna` will verify the logged-in npm user's access to the packages about to be published. Passing this flag will disable that check.
@@ -244,15 +252,6 @@ You could also configure this at the root level of lerna.json, applying to both 
   "version": "independent"
 }
 ```
-
-### `--temp-tag`
-
-When passed, this flag will alter the default publish process by first publishing
-all changed packages to a temporary dist-tag (`lerna-temp`) and then moving the
-new version(s) to the dist-tag configured by [`--dist-tag`](#--dist-tag-tag) (default `latest`).
-
-This is not generally necessary, as Lerna will publish packages in topological
-order (all dependencies before dependents) by default.
 
 ### `--yes`
 
@@ -360,4 +359,4 @@ Lerna will run [npm lifecycle scripts](https://docs.npmjs.com/misc/scripts#descr
 9. Run `publish` lifecycle in root
    - To avoid recursive calls, don't use this root lifecycle to run `lerna publish`
 10. Run `postpublish` lifecycle in root
-11. Update temporary dist-tag to latest, if [enabled](#--temp-tag)
+11. Update temporary dist-tag to latest, unless [disabled](#--no-temp-tag)
--- a/commands/publish/__tests__/publish-canary.test.js
+++ b/commands/publish/__tests__/publish-canary.test.js
@@ -15,7 +15,7 @@ const childProcess = require("@lerna/child-process");
 
 // mocked modules
 const writePkg = require("write-pkg");
-const npmPublish = require("@lerna/npm-publish");
+const npmDistTag = require("@lerna/npm-dist-tag");
 const PromptUtilities = require("@lerna/prompt");
 const checkWorkingTree = require("@lerna/check-working-tree");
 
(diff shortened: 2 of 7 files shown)


evocateur:
The primarily interesting commit is https://github.com/lerna/lerna/pull/1863/commits/4eff8d2280c39034b988bcccc8365e99cc97b02e, which does not include the changes from the "upstream" PR.

commands/publish/README.md
```diff
@@ -108,6 +108,13 @@ To avoid this, pass `--no-git-reset`. This can be especially useful when used as
 lerna publish --no-git-reset
 ```
 
+### `--no-temp-tag`
+
+By default, `lerna publish` first publishes all changed packages to a temporary dist-tag (`lerna-temp`)
+and then after all are successful moves the new version(s) to the dist-tag configured by [`--dist-tag`](#--dist-tag-tag) (default `latest`).
+
+If you like to live dangerously, pass `--no-temp-tag` (or set `command.publish.tempTag` to `false` in lerna.json) to disable this behavior.
```
ThisIsMissEm: Perhaps we need to explain why this is dangerous?

commands/publish/command.js
```diff
@@ -40,18 +40,15 @@ exports.builder = yargs => {
...
     "git-reset": {
       describe: "Reset the git working tree.\nPass --no-git-reset to disable.",
       type: "boolean",
       defaultDescription: "true",
     },
     "temp-tag": {
-      describe: "Create a temporary tag while publishing.",
+      describe: "Create a temporary dist-tag while publishing.\nPass --no-temp-tag to disable.",
       type: "boolean",
+      defaultDescription: "true",
```
ThisIsMissEm: How does this change affect the `lerna help` output? Should it actually default to `no-temp-tag` instead of `temp-tag`?

ThisIsMissEm: Also, isn't this capable of using a different string here?

commands/publish/README.md
evocateur: Yes, that's a better idea.

commands/publish/__tests__/get-unpublished-packages.test.js
```diff
@@ -27,8 +33,38 @@ getPackument.mockImplementation(async pkg => {
...
-  const project = new Project(cwd);
+  const packages = await getPackages(cwd);
+  const packageGraph = new PackageGraph(packages);
+
+  const opts = new Map();
+  const pkgs = await getUnpublishedPackages(packageGraph, opts);
 
-  const pkgs = await getUnpublishedPackages(project, {});
-  expect(pkgs.map(p => p.name)).toEqual(["package-1", "package-3", "package-4", "package-5"]);
+  expect(pkgs).toMatchInlineSnapshot(`
```
evocateur: These changes are just fallout from unrelated refactoring (prompted by removal of a dependency I thought was clear). In the end, I didn't make the effort to separate it.

commands/publish/__tests__/publish-canary-no-git-reset.test.js
```diff
@@ -73,24 +70,5 @@ test("publish --canary --no-git-reset", async () => {
...
-  expect(writePkg.updatedVersions()).toMatchInlineSnapshot(`
-Object {
-  "package-1": 1.0.1-alpha.0+SHA,
-  "package-2": 1.0.1-alpha.0+SHA,
-  "package-3": 1.0.1-alpha.0+SHA,
-  "package-4": 1.0.1-alpha.0+SHA,
-}
-`);
-  expect(gitCheckout).toHaveBeenCalledTimes(0);
+  expect(gitCheckout).not.toHaveBeenCalled();
```
evocateur: The majority of these assertions were removed because they were completely redundant and obscured the actual point of the test case.

commands/publish/__tests__/publish-canary.test.js
```diff
@@ -14,7 +14,7 @@ const childProcess = require("@lerna/child-process");
 
 // mocked modules
 const writePkg = require("write-pkg");
-const npmPublish = require("@lerna/npm-publish");
+const npmDistTag = require("@lerna/npm-dist-tag");
```
evocateur: Most assertions on the final dist-tag now need to use the `@lerna/npm-dist-tag` mock, which explains all the new stuff in the mock helper. (`@lerna/npm-publish` is almost always called with `lerna-temp` dist-tag, which isn't very useful as an assertion)

commands/publish/__tests__/publish-command.test.js
```diff
@@ -259,11 +256,10 @@ Map {
...
     it("publishes unpublished packages", async () => {
       const testDir = await initFixture("normal");
-      const project = new Project(testDir);
 
-      getUnpublishedPackages.mockImplementationOnce(async () => {
-        const pkgs = await project.getPackages();
-        return pkgs.slice(1, 3);
+      getUnpublishedPackages.mockImplementationOnce(packageGraph => {
+        const pkgs = packageGraph.rawPackageList.slice(1, 3);
+        return pkgs.map(pkg => packageGraph.get(pkg.name));
```
evocateur: another mostly-unrelated change due to refactoring `getUnpublishedPackages()`. it definitely exposes that my mock isn't abstract enough, and should probably be a network mock instead...

commands/publish/command.js
evocateur: We're using yargs' "auto-negation" feature here (any boolean option prefixed with `no-` is set to `false`), so it is a bit awkward, I agree. We can't use `"no-temp-tag"` as the key, otherwise it would need to be passed as `--no-no-temp-tag`.

(I could be wrong, it's been awhile since I tried to make this clearer. It could also require upstream changes in yargs)

commands/publish/lib/get-unpublished-packages.js
```diff
@@ -1,33 +1,30 @@
 "use strict";
 
 const log = require("libnpm/log");
-const pReduce = require("p-reduce");
+const pMap = require("p-map");
```
evocateur: I removed `p-reduce` from the main command script, forgot it was here, and then realized it could be implemented with more concurrency with `p-map`. (`p-reduce` executes serially)

integration/lerna-publish-canary.test.js
```diff
@@ -22,7 +22,7 @@ const env = {
 
 test("lerna publish --canary uses default prerelease id", async () => {
   const { cwd } = await cloneFixture("normal");
-  const args = ["publish", "--canary", "--yes"];
+  const args = ["publish", "--canary", "--yes", "--no-temp-tag"];
```
evocateur: I had to add `--no-temp-tag` to all the end-to-end tests because we don't have a good solution for local testing with an _actual_ registry, and none of the test packages are actually "live" on the public registry.

commands/publish/__tests__/get-unpublished-packages.test.js
evocateur: Actually, I think I _will_ move this commit out of this PR. It's very noisy.

ThisIsMissEm:
Hmm, okay. That is indeed a little bit awkward.

evocateur:
@ThisIsMissEm I think I figured out a pattern that makes the inverted booleans more ergonomic (documented as `--no-<thing>` in the `-h` output, even).

For example, `--no-progress` is the only option that makes sense, as [`--progress` is implied](https://github.com/lerna/lerna/blob/c79a82722095607940749f5c6fa1cdd90c56a437/core/global-options/index.js#L23-L27) (where progress is even used, getting kinda itchy in the "global" options). This change makes `--no-progress` output in the `--help`, while keeping the shape of the resulting `argv` object (mostly) consistent:
```patch
+    "no-progress": {
+      describe: "Disable progress bars. (Always off in CI)",
       type: "boolean",
     },
     progress: {
-      defaultDescription: "true",
-      describe: "Enable progress bars. (Always off in CI)\nPass --no-progress to disable.",
+      // proxy for --no-progress
+      hidden: true,
+      type: "boolean",
+    },
```

It does technically add new `noProgress` keys, _and_ can understand `--no-no-progress` flags, but I think we can ignore those, in practice. Dunno if I should port this upstream to yargs in a cleaner way, but figure this kinda-hacky solution is at least very... declarative? If a bit mysterious...

There's a bunch of other flag deprecations I want to remove in the semver major, so I'll be doing that soon enough...

evocateur:
this works globally across all commands:
```diff
diff --git a/core/lerna/index.js b/core/lerna/index.js
index a46caba..dbfd308 100644
--- a/core/lerna/index.js
+++ b/core/lerna/index.js
@@ -26,7 +26,7 @@ function main(argv) {
     lernaVersion: pkg.version,
   };
 
-  return cli()
+  const parser = cli()
     .command(addCmd)
     .command(bootstrapCmd)
     .command(changedCmd)
@@ -40,6 +40,23 @@ function main(argv) {
     .command(listCmd)
     .command(publishCmd)
     .command(runCmd)
-    .command(versionCmd)
-    .parse(argv, context);
+    .command(versionCmd);
+
+  invertBooleanOptions(parser);
+
+  return parser.parse(argv, context);
+}
+
+function invertBooleanOptions(parser) {
+  const opts = parser.getOptions();
+
+  for (const key of Object.keys(opts.key)) {
+    if (key.startsWith("no-")) {
+      // --no-foo => inverted --foo
+      parser.option(key.substring(3), {
+        hidden: true,
+        type: "boolean",
+      });
+    }
+  }
 }
```

✨

JamesHenry:
I will close this one because of its age, but we will look to weave it into the recoverable publish work as part of v7

PR #1864, opened 2019-01-10
feat(version): Create Github releases with `--github-release`

milesj:
Would solve https://github.com/lerna/lerna/issues/1513

## Description

This is a proof of concept so I'm looking for feedback. Is this something we want in Lerna? Would it be better as a future type of plugin? etc.

A bulk of the work already exists through conventional-commits + changelog generation. I updated the changelog logic to also return the new entry to use in the body of the GH release. I then wrapped everything in a `--github` command line option for opt-in purposes.

Other notes:

- The GitHub client is its own package that requires specific env vars to exist.
- I went with `--github` instead of `--release`, since the latter is a bit confusing, especially when terms like version and publish are thrown around. Furthermore, using `--github` permits additional features in the future, like adding labels to PRs.
- It would be nice to use [lerna-changelog](https://github.com/lerna/lerna-changelog) but the APIs aren't very compatible.

## Motivation and Context

I've been looking into using Lerna for auto-releases as a replacement for semantic-release (since it doesn't handle monorepos). The biggest drawback is the lack of GitHub releases, which most of my team wants, and is a requirement before we can switch over.

## How Has This Been Tested?

N/A. I'll add tests if this change is good to move forward.

## Types of changes

## Checklist:


commands/version/index.js
```diff
@@ -556,6 +578,32 @@ class VersionCommand extends Command {
...
+      this.releaseNotes.map(({ notes, name }) => {
+        const tag = name === "root" ? this.tags[0] : this.tags.find(t => t.startsWith(name));
+
+        if (!tag) {
+          return Promise.resolve();
+        }
+
+        return client.repos.createRelease({
+          owner: "", // TODO
+          repo: "", // TODO
```
evocateur: Could `owner` and `repo` be derived from project metadata? (root `package.json` `repository.url` field?)

commands/version/index.js
```diff
@@ -556,6 +578,32 @@ class VersionCommand extends Command {
...
   }
+
+  createGitHubReleases() {
+    this.logger.info("github", "Creating releases...");
+
+    const client = createGitHubClient();
+
+    return Promise.all(
+      this.releaseNotes.map(({ notes, name }) => {
+        const tag = name === "root" ? this.tags[0] : this.tags.find(t => t.startsWith(name));
```
evocateur: I don't think you need to create releases for every package when the versioning is _not_ independent (i.e., "fixed"). You would only need to create one release, the "root", in that case.

utils/github-client/github-client.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  if (GHE_VERSION) {
+    // eslint-disable-next-line
+    Octokit.plugin(require(`@octokit/plugin-enterprise-rest/ghe-${GHE_VERSION}`));
+  }
+
+  const client = new Octokit({
+    agent: undefined,
+    baseUrl: GHE_API_URL,
```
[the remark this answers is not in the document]
milesj: Yeah, will change it so it's a bit more apparent.

commands/version/index.js
```diff
@@ -556,6 +578,32 @@ class VersionCommand extends Command {
...
+        }
+
+        return client.repos.createRelease({
+          owner: "", // TODO
+          repo: "", // TODO
+          tag_name: tag,
+          name: tag,
+          body: notes,
+          draft: false,
+          prerelease: false,
```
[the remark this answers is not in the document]
milesj: Is there any flags/settings that I can dig into to determine a prerelease?

commands/version/index.js
milesj: When "fixed", this should only have 1 value in `releaseNotes` correct? Since the chain is only happening once, and not on every package.

milesj:
@evocateur Awesome to move forward on this! Made some adjustments based on feedback. I'll add tests once the implementation is done.

utils/github-client/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/lerna/lerna.git"
+  },
+  "scripts": {
+    "test": "echo \"Run tests from root\" && exit 1"
+  },
+  "dependencies": {
+    "@octokit/plugin-enterprise-rest": "^2.0.0",
+    "@octokit/rest": "^16.7.0"
```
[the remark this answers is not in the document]
milesj: Added.

commands/version/index.js
milesj: I ended up just calling `git` since that package field might not always exist.

commands/version/index.js
```diff
@@ -450,9 +461,15 @@ class VersionCommand extends Command {
...
           rootPath,
           tagPrefix: this.tagPrefix,
-        }).then(changelogLocation => {
+        }).then(({ logPath, newEntry }) => {
           // commit the updated changelog
-          changedFiles.add(changelogLocation);
+          changedFiles.add(logPath);
+
+          // add release notes
+          this.releaseNotes.push({
```
evocateur: This `push` needs to be guarded to only execute during `--independent` publish, otherwise "fixed" mode will be making too many releases (each package `CHANGELOG.md` is still updated with package-specific changes in fixed mode, but that isn't meant to be a separate release).

```js
if (independentVersions) {
  this.releaseNotes.push({
    name: pkg.name,
    notes: newEntry,
  });
}
```

The subsequent root push is still okay, that never runs in independent mode. And then the final loop to create github releases will be correct.

utils/github-client/github-client.js
```diff
@@ -0,0 +1,37 @@
...
+    throw new Error("A GH_TOKEN environment variable is required.");
+  }
+
+  if (GHE_VERSION) {
+    // eslint-disable-next-line
+    Octokit.plugin(require(`@octokit/plugin-enterprise-rest/ghe-${GHE_VERSION}`));
+  }
+
+  if (GHE_API_URL) {
+    options.baseUrl = GHE_API_URL;
```
evocateur: Much clearer, thanks!

commands/version/lib/git-repo-url.js
```diff
@@ -0,0 +1,16 @@
...
+
+const log = require("libnpm/log");
+const childProcess = require("@lerna/child-process");
+
+module.exports = getRepoUrl;
+
+function getRepoUrl(opts) {
+  log.silly("getRepoUrl");
+
+  const args = ["config", "--get", "remote.origin.url"];
```
evocateur: You'll need to account for the [`--git-remote`](https://github.com/lerna/lerna/blob/c79a82722095607940749f5c6fa1cdd90c56a437/commands/version/command.js#L45-L50) option here.

commands/version/index.js
```diff
@@ -556,6 +579,35 @@ class VersionCommand extends Command {
 
     return gitPush(this.gitRemote, this.currentBranch, this.execOpts);
   }
+
+  createGitHubReleases() {
+    this.logger.info("github", "Creating GitHub releases...");
+
+    const client = createGitHubClient();
+    const repo = parseGitUrl(gitRepoUrl());
```
evocateur: I think it would be clearer if `@lerna/github-client` exposed a `parseGitRepo()` method that returned this repo object, rather than embed the `git config --get ...` stuff in the version command. Does that make sense?

(it would also make unit testing easier with more expressive mocking)

milesj: Yup, totally can do.

commands/version/index.js
milesj: Instead of wrapping in independent mode, maybe we include all the release notes from each package as a single GH release?

evocateur: Well, coincidentally, that's been a feature request from some users of independent mode, but the current output of the changelog generator isn't quite clean enough (imo) for that to make sense. Also, since independent mode creates tags for each individual package, that means the GitHub tags/releases page gets spammed with a bunch of empty tags (now).

I'm not sure where we would put the "rolled-up independent" changelog. Much less what to call it. But the work you're doing lays the groundwork for future improvements in this area!

commands/version/command.js
```diff
@@ -42,6 +42,10 @@ exports.builder = (yargs, composed) => {
       describe: "Specify cross-dependency version numbers exactly rather than with a caret (^).",
       type: "boolean",
     },
+    github: {
```
evocateur: Do you mind if we rename this `--github-release`? `--github` just doesn't "scope" the flag enough for me (in the context of `lerna version`, "github"... what?). It also makes the option a bit more self-descriptive in the implementation.

milesj: Yup, of course.

milesj:
@evocateur Made changes. I think this is good to start adding tests.

commands/version/command.js
evocateur: Thanks!

evocateur:
The `showCommit` assertion pattern in https://github.com/lerna/lerna/blob/master/commands/version/__tests__/version-conventional-commits.test.js isn't sufficient? You might need to modify the mock in `commands/__mocks__/@lerna/conventional-commits.js` to make it more nuanced.

milesj:
I'll take a look at that, thanks.

sidharthachatterjee:
This looks great! We're looking at generating GitHub releases with change logs at https://github.com/gatsbyjs/gatsby/ and eagerly waiting for this to get in!

Thank you for this @milesj

evocateur:
yep, just waiting for the remainder of the tests, then we're good to go

milesj:
Yeah sorry, been a bit swamped and haven't had a change to dig back into it. Hoping this weekend.

evocateur:
@milesj No worries, I _completely_ understand.

evocateur:
I force-rebased your changes on top of recent work in `master`, so be sure to `git pull` before continuing.

milesj:
@evocateur Thanks for the heads up 👍

milesj:
@evocateur Added tests 👍

evocateur:
Hot damn, it worked [great](https://github.com/lerna/lerna/releases/tag/v3.11.0)!

milesj:
@evocateur AWW YESS! Thanks for accepting it, this is awesome.

ChrisBAshton:
Amazing work, thanks @milesj !

tunnckoCore:
That's great, but it also generates a `CHANGELOG.md`. Is it expected or a bug?
When pas `--no-changelog`, then it does not generate releases too, but in the docs is mentioned that it only requires `--conventional-commits`.

evocateur:
@tunnckoCore Yes, currently it is required to write out the changelog files due to the way the code is currently factored: https://github.com/lerna/lerna/blob/45a05acf2d52a46f553fc1efadf9a32312aca82b/core/conventional-commits/lib/update-changelog.js#L48-L68

The docs aren't _wrong_, per se, but they are certainly omitting a case. I can add an error when `--no-changelog` is also passed to make it extra clear.

tunnckoCore:
Yea, warning or directly error sounds good.

evocateur:
030de9d6 should do the trick.

azz:
Is there a issue to track supporting `--no-changelog` with `--github-release`?

evocateur:
Not that I'm aware of, no.

milesj:
@azz Curious how you think that would work?

azz:
I'm not completely across the technical details, but generating both CHANGELOG.md and GitHub releases is duplicate information.

azz:
Raised https://github.com/lerna/lerna/issues/1959 to discuss

PR #1885, opened 2019-01-23
docs(version): update to `version` command in examples

jenniesyip:
## Description
Was reading the [`lerna version` documentation](https://github.com/lerna/lerna/tree/master/commands/version#readme) and these parts confused me, are they supposed to be `publish` or `version`? Can we also configure lerna versioning also by configuring publish on `lerna.json`?

## Motivation and Context
Make documentation more clear

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -63,7 +63,7 @@ It is easiest (and recommended) to configure in `lerna.json`, but it is possible
 ```json
 {
   "command": {
-    "publish": {
+    "version": {
       "allowBranch": "master"
     }
   }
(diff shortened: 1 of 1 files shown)


evocateur:
You've identified one of the current wrinkles with Lerna, and the _even more_ confusing answer is that ...they're both right? For legacy reasons, `lerna version` _also_ reads the `command.publish` namespace, because often times users only call `lerna publish`, which automatically calls a nested `lerna version` when appropriate.

However! I agree with your change here, it is much clearer. It has the benefit of being forward-compatible with the next major version of Lerna, which will be _completely_ splitting `lerna publish` from `lerna version`, much like `npm publish` and `npm version` are separate commands today.

Please let me know if I can help answer any other questions you have about Lerna, especially if the docs need clarification. The docs themselves are probably the biggest "opportunity for improvement" in Lerna today, and I need all the help I can get. 😅 

Thank you!

jenniesyip:
@evocateur Thanks for the clarification! Appreciate it. :)

Issue #1886, opened 2019-01-23
Feature: lerna remove


tanhauhau:
Removes dependency from packages, the inverse of [@lerna/add](https://github.com/lerna/lerna/tree/master/commands/add)

Currently there's no way to remove a dependency from a lerna package,
the closest way we can do is to manually remove from `package.json`, `lerna clean --yes --scope=xxx` + `lerna bootstrap --scope=xxx`.

This command provides the remove dependency feature. Just like `npm uninstall` or `yarn remove`.

KaboomFox:
👍

einsqing:
👍

garrettmaring:
What does the `--yes` flag do?

One can also run `$ lerna exec 'yarn remove foo' --scope=xxx` to remove it.

a1300:
The following workaround worked for me:

1. Manually remove dependencies from `@org-name/package-name/package.json`
2. Run:
```bash
$ lerna bootstrap --scope @org-name/package-name --no-ci --force-local
```

mikhailrojo:
please add 
```
lerna remove package --scope=...
```

diminutivesloop:
What's the status of this feature? This issue is already over a year old.

danawoodman:
@evocateur is this something you plan to add? If not, maybe some documentation could be added as to the preferred method of removing packages?

chaddjohnson:
I am very surprised to find that there is an add but not a remove...

EricRabil:
Typed `add` instead of `create`... added a random package to all of my packages that I must now manually remove as `lerna exec 'yarn remove foo' --scope=xxx` spits out unrelated errors.

mesqueeb:
In Lerna still maintained?

a1300:
@mesqueeb last commit was 19 days ago. It seems to be still maintained.

geeeger:
i need it, too.

tkforce:
Seems like a reasonable feature request +1

lfre:
Apologies if this is already mentioned in another issue. I'm putting this out there for the next person running into the same issue I just did.

I noticed `lerna add` has an issue where the local package is removed from the lock files (package-lock.json, yarn.lock).
Re-adding or bootstrapping won't have effect, because `node_modules` is still there.

[...]

PR #1887, opened 2019-01-23
WIP: feat(remove): initial commit for lerna remove

tanhauhau:
I copied some of the code from `@lerna/add`, supporting glob and scope for filtering folders

## Description
Current state:
- [x] Filter packages to be changed, based on `filterOptions` and path matching with `glob`
- [x] Determine whether the dependencies is external or internal
  - [ ] Should determine this in package level, instead of project level, since if version doesn't match, it will not symlink
- [x] Update package.json
- if the dependency is external:
  - [x] Remove symlink
- else:
  - [x] call `npm uninstall` or `yarn remove`
  - [ ] taking care of hoist logic
 
- [ ] test
- [ ] documentation

## Motivation and Context

https://github.com/lerna/lerna/issues/1886

## Types of changes

## Checklist:


tanhauhau:
There's a lot yet to be done, would like to listen to your input on this @evocateur

commands/remove/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "type": "git",
+    "url": "git+https://github.com/lerna/lerna.git"
+  },
+  "scripts": {
+    "test": "echo \"Run tests from root\" && exit 1"
+  },
+  "dependencies": {
+    "@lerna/command": "file:../../core/command",
+    "@lerna/filter-options": "file:../../core/filter-options",
+    "@lerna/npm-uninstall": "file:../../utils/npm-uninstall",
```
evocateur: `@lerna/npm-uninstall` looks like a copypaste of `@lerna/npm-install`. Why do we need it? We're already removing the entry in the package.json, etc. Running `lerna bootstrap` (like the add command does) should be sufficient to clean up the other stuff.

commands/remove/index.js
```diff
@@ -0,0 +1,149 @@
...
+
+    chain = chain.then(() => this.removePackage());
+
+    // if (this.options.bootstrap !== false) {
+    //   chain = chain.then(() => {
+    //     const argv = Object.assign({}, this.options, {
+    //       args: [],
+    //       cwd: this.project.rootPath,
+    //       // silence initial cli version logging, etc
+    //       composed: "add",
```
evocateur: We should definitely still be doing this, instead of calling `npm uninstall`. Ensuring that `ci: false` is also passed makes sure we run `npm install` which will synchronize the lockfile and `node_modules` appropriately?

commands/remove/index.js
```diff
@@ -0,0 +1,149 @@
...
+        delete pkg.devDependencies[this.dependencyToRemove];
+      }
+      if (pkg.peerDependencies) {
+        delete pkg.peerDependencies[this.dependencyToRemove];
+      }
+      return pkg.serialize();
+    });
+  }
+
+  removeSymlink() {
```
evocateur: We're not removing the symlink, but the directory. In any case, I don't think we need this. Just bootstrap after serializing the modified dependencies.

commands/remove/yarn.lock
```diff
@@ -0,0 +1,2862 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
evocateur: Definitely remove this, please. Lerna itself uses npm, and in any case shouldn't be developing from a leaf dir.

commands/remove/package.json
tanhauhau: not really, the bootstrap command will not update the lockfile, at least that's not happening in `yarn`, due to https://github.com/lerna/lerna/blob/master/commands/bootstrap/index.js#L171

tanhauhau: i guess that's the main motivation of the MR, because removing dep in `package.json` and running lerna bootstrap will not remove the dep from `node_modules` nor updating the lock file.

evocateur: Err, that branch is only reached when you are attempting a filtered bootstrap. (scope, ignore, or since)

The chained bootstrap should have none of these options, thus the raw bootstrap should actually deal with this?

tanhauhau: oh, my bad.
still, bootstrap will have early return if all the stated dependencies have been satisfied.
for our case (lerna remove), all stated dependencies is satisfied, but we have extraneous dependencies in node_modules, should we 
1. reuse bootstrap, but add a new flag, (something like https://github.com/lerna/lerna/issues/1883#issuecomment-459580755) to bypass the early return of lerna bootstrap, thus able to trigger npm client to remove extraneous pacakges and update lockfile or
2. use `npm-uninstall` that focuses on removing packages, by triggering npm client's `remove`/`uninstall` command rather than `install`/`ci` in `lerna bootstrap`

evocateur: Oh, right, I forgot Yarn doesn't want you to remove stuff when you modify `package.json` and run `yarn install`...

So yeah, I guess we do need `npm-uninstall`. I'd prefer it was less copypasta, but that's more work that isn't blocking.

We also, however, need to account for the `--hoist` case (the removed dependency doesn't exist at the leaf, but in the root) and relative `file:` specifiers, which need only re-run `npm install` in the root. `lerna bootstrap` already accounts for the `file:` specifier case, but I really don't want to burden that poor command with any _more_ complexity...

tanhauhau: For the `hoist` case, how should we figure out whether the dependency to be removed is hoisted or installed in leaf node?

tanhauhau:
@evocateur 
- i've did some refactoring, extracting out `@lerna/use-temporary-manifest`, so that can be use when `npm install` and `npm uninstall`.
  - I might have break some test, I'll fix those later
- I've added some basic test for `lerna remove`
  - going to add accout to `--hoist` case, and `file:` case

But I would like to listen to your comments on the general directions of the change, and there will be definitely a lot of parts of code in `lerna remove` and `lerna bootstrap` & `npm-uninstall`, I am going to extract them out into packages, I am not sure the packages structure I have in mind align with your plans.

tanhauhau:
@evocateur what do you think?

lucaasrojas:
This PR's idea is be able to remove packages from the monorepo?

tanhauhau:
@lucaarojas yes

floriangosse:
It this PR still in development?

montogeek:
@evocateur What do you think?

evocateur:
I don't really see this going forward anymore, sorry. I appreciate all the effort spent on this PR, but modern monorepos should be using package managers with proper capabilities to deal with this sort of thing (pnpm recursive, npm v7, yarn i guess).

Given the minimal frequency of use for this sort of command, the maintenance burden far outweighs the benefit of inclusion in the core. Perfect opportunity for a third-party utility.

davidmurdoch:
> Given the minimal frequency of use for this sort of command, the maintenance burden far outweighs the benefit of inclusion in the core. Perfect opportunity for a third-party utility.


FWIW, I would have used this command every day this week.

Issue #1890, opened 2019-01-24
Don't check if private packages are published when using from-package


pshrmn:
Running `lerna publish from-package` triggers a published version check for every package, even private ones.

## Expected Behavior

Lerna detects that a package is private and automatically skips checking for its published version.

## Possible Solution

I'm not familiar enough with Lerna's internals to know how easy it is to determine if a package is private. I took a quick look at the `pkg` object that [Lerna uses to check if it is published](https://github.com/lerna/lerna/blob/b16451a2ed8b8535af6b3cb47289e3c4f6a1563f/commands/publish/lib/get-unpublished-packages.js#L14) and I didn't see a reference to the `private` property.

## Context

I have a package with a number of "private" example packages, so I end up spending a good bit of time waiting for all of those to be unsuccessfully checked.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.10.7 |
| `npm --version`  | 6.4.1 |
| `yarn --version` | n/a |
| `node --version` | 10.15.0 |

evocateur:
Yup, that's definitely a problem!

Quick sketch of the fix, would appreciate a PR with a simple test added [here](https://github.com/lerna/lerna/blob/master/commands/publish/__tests__/get-unpublished-packages.test.js):
```patch
diff --git a/commands/publish/lib/get-unpublished-packages.js b/commands/publish/lib/get-unpublished-packages.js
index 4e5cbf2..fe83bd9 100644
--- a/commands/publish/lib/get-unpublished-packages.js
+++ b/commands/publish/lib/get-unpublished-packages.js
@@ -11,6 +11,9 @@ function getUnpublishedPackages(packageGraph, opts) {
 
   let chain = Promise.resolve();
 
+  // don't bother attempting to get the packument for private packages
+  const graphNodesToCheck = Array.from(packageGraph.values()).filter(({ pkg }) => !pkg.private);
+
   const mapper = pkg =>
     getPackument(pkg.name, opts).then(
       packument => {
@@ -24,7 +27,7 @@ function getUnpublishedPackages(packageGraph, opts) {
       }
     );
 
-  chain = chain.then(() => pMap(packageGraph.values(), mapper, { concurrency: 4 }));
+  chain = chain.then(() => pMap(graphNodesToCheck, mapper, { concurrency: 4 }));
 
   return chain.then(results => results.filter(Boolean));
 }
```

PR #1905, opened 2019-01-30
fix(publish): Skip private package lookup

pshrmn:
Skip checking if a private package's current version is published since private packages are never published.

## Description

Filters out packages whose `private` value is `false`.

## Motivation and Context

#1890

## How Has This Been Tested?

I added a fixture with two packages, one public and one private and verify that only the public package shows up in the output from `getUnpublishedPackages`.

## Types of changes

## Checklist:


pshrmn:
Thanks for the outline in the issue. You made it so simple that I would have felt guilty not submitting a PR!

PR #1908, opened 2019-01-31
feat(version): Skip repository validation when git is unused

### LINKED ISSUE #1869: Lerna version needs code to be on a branch
I am trying to run `lerna version` in a non-git folder. More specifically I am skipping all git related actions by running `lerna version --no-git-tag-version --no-push --exact 0.0.1-SNAPSHOT` resulting in failure, due to [this](https://github.com/lerna/lerna/blob/713f72d1d5bd358147bb50df65b203b43a4667f8/commands/version/index.js#L91) check.

## Expected Behavior
I would expect this to be independent of git in the same way as npm is and just update the package.json versions:
[Only] [if run in a git repo, it will also create a version commit and tag.](https://docs.npmjs.com/cli/version)
This way the `--no-git-tag-version --no-push` arguments would be obsolete in a non-git project.

## Current Behavior
This currently fails with `ENOGIT`.

## Possible Solution
Check the branch only in case the git based features are used (`allow-branch`, `git-tag-version`, etc.).

1. Take any working lerna project, run `rm -rf **/.git`
2. run `lerna version <version>` 
3. failure 😱

## Context
I am using lerna to publish to our internal registry from a central build server. Most of the things happening on this server are out of my hands. There the code is checked out, then stripped of any potential "overhead" files (node_modules, .git and files created by the build), and `npm install`ed.
This is when I set the version.

Currently I was able to work around this issue by running `lerna exec npm version 0.0.1-SNAPSHOT`.

## Your Environment
Centrally managed build server 🔒

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.7.1 |
| `npm --version`  | 6.4.1 |
| `node --version` | 10.14.0 |

| OS | Version |
| --- | --- |
| some linux server | -- |

marikaner:
Do not fail version update when git features are not used.

## Description
I am just skipping the check on whether we are @ HEAD when git is not used.

## Motivation and Context
closes #1869 

## How Has This Been Tested?
I tried it with our project, but I did not add any automatic tests. It works in both fixed and locked mode and still throws an error in case git commands are not skipped.

## Types of changes

## Checklist:


marikaner:
Do you think you will accept this pull request if I fix what you mentioned? I would like to make sure it makes sense to invest time here before I actually do it ;)

marikaner:
Thank you for your review, @evocateur. I reverted the changes to the lockfile and will add tests as soon as we agree on the usefulness of this PR ;)

I finally understood your concerns. I misunderstood the purpose of the fixed mode (even though I read your documentation multiple times *) and am now aware that my issue is rather a corner case as you will basically only not need to be on a branch when using `--force-publish` as well as skipping tagging + pushing.
I still think it might be an improvement as it moves lerna's behaviour closer to npm's ([npm does not fail without git](https://docs.npmjs.com/cli/version.html)).

Please take a look at my current implementation and let me know whether this goes into the right direction in your opinion. I just overrode the `requiresGit` method and skipped all branch related actions accordingly.

\* "[Use this if you want to automatically tie all package versions together.](https://github.com/lerna/lerna#fixedlocked-mode-default)" confirmed my intuitive understanding...

marikaner:
@evocateur, I added some tests, one is still failing, if will try to get this fixed if I have time this evening, otherwise maybe you can help?

commands/version/__tests__/version-command.test.js
```diff
@@ -502,6 +502,89 @@ describe("VersionCommand", () => {
...
+        await lernaVersion(cwd)();
+      } catch (err) {
+        expect(err.prefix).toBe("ENOGIT");
+        expect(err.message).toBe("Detached git HEAD, please checkout a branch to choose versions.");
+      }
+
+      expect.assertions(2);
+    });
+
+    it("does not throw for version --force-publish --no-git-tag-version --no-changelog", async () => {
```
evocateur: In the event no tags are found, collect-updates acts as if you had passed `--force-publish` anyway.

marikaner: Ok, I guess this makes it much simpler. :)

utils/collect-updates/collect-updates.js
```diff
@@ -51,8 +49,10 @@ function collectUpdates(filteredPackages, packageGraph, execOpts, commandOptions
   let candidates;
 
   if (!committish || forced.has("*")) {
+    log.info("", `Assuming all packages changed`);
```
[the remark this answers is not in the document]
marikaner: 👍

marikaner:
@evocateur thank you. I removed `forcePublish` and `independent` mode from the conditions for `requiresGit`. Also, I moved the initialization of the command properties to `configureProperties` and adjusted the test conditions accordingly.

marikaner:
@evocateur What do you think?

utils/collect-updates/collect-updates.js
evocateur: I ported this file's changes in 9b8dd9c, please rebase when you can.

commands/version/index.js
```diff
@@ -43,7 +43,15 @@ class VersionCommand extends Command {
...
 
-  initialize() {
+  get requiresGit() {
+    return (
+      this.commitAndTag || this.pushToRemote || this.options.allowBranch || this.options.conventionalCommits
+    );
+  }
+
+  configureProperties() {
+    super.configureProperties();
```
evocateur: I must admit I had a bad reaction when I first saw this `super` call, but now... I really like it! It does a good job of compartmentalizing the operations, as you'd expect a class method to do.

evocateur:
I'll merge this once you've rebased.

commands/version/index.js
marikaner: I wasn't 100% sure about this either, but it seemed legit :)

utils/collect-updates/collect-updates.js
marikaner: I just merged the current master and resolved the hereby resulting merge conflict. Did you have anything else in mind? For rebasing I would have to either create a new branch or force push. Anyways, if you squash and merge this shouldn't make a difference.

evocateur:
Thanks!

marikaner:
Wohoo!

PR #1916, opened 2019-02-08
Add [ci skip] to all publish commit messages.

milesj:
## Description

When lerna is used for automatic releases in a CI using `--yes`, the publish commit triggers another auto-release (which is bad). This adds a `[ci skip]` to all publish commit messages to avoid it.

## Motivation and Context

Avoid unnecessary releases.

## How Has This Been Tested?

Just unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/__tests__/__snapshots__/version-bump-prerelease.test.js.snap
+++ b/commands/version/__tests__/__snapshots__/version-bump-prerelease.test.js.snap
@@ -3,6 +3,8 @@
 exports[`version patch with previous prerelease also graduates prereleased 1`] = `
 v1.0.1
 
+[ci skip]
+
 HEAD -> master, tag: v1.0.1
 
 diff --git a/lerna.json b/lerna.json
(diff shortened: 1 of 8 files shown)


evocateur:
This doesn't need to be built-in to lerna, one can already accomplish this by using the `--message` option or configuring `command.version.message` in lerna.json (same as the flag, but "durable" in the sense that nobody has to remember to add it, and the CI command becomes cleaner):

```bash
lerna publish --message "chore(release): Publish %s\n\n[ci skip]"
```
```json
{
  "command": {
    "version": {
      "message": "chore(release): Publish %s\n\n[ci skip]"
    }
  }
  "version": "1.2.3"
}
```

Note that when using independent versioning, no version placeholders are interpolated, and you might as well put the `[ci skip]` at the end of the subject, as the body of the commit message (a list of versions created) is appended to the `message` value.

milesj:
@evocateur The `message` *should* be enough, so that's cool with me. This doesn't necessarily have to be in Lerna.

Issue #1666, opened 2018-09-09
Missing license warning needs to be clarified, more actionable


patryk-bartoszek:
## Expected Behavior

Don't issue warring if license is provided in the package.json file. 
I shouldn't have to create LICENSE file if I'm using SPDX license identifier or expression.

## Current Behavior

Currently during publish I get the following message:
`lerna WARN ENOLICENSE Packages foo are missing a license`

I have following property set in package.json of foo package
```
  "license": "MIT"
```

## Possible Solution

This is the result of changes introduced in #1465
The current implementation of license check only looks for LICENSE / LICENCE files and ignores the license value provided in the package.json file.
https://github.com/lerna/lerna/blob/master/commands/publish/lib/get-packages-without-license.js

When using standard SPDX license string, there is really no need to provide license file. Tools such as license-checker work just fine with the default approach of defining license in package.json. 
I see no benefit to creating separate license file.

A possible solution would be to extend implementation so that it checks for license property in the package.json file. It would be nice to validate value according to rules provided in https://docs.npmjs.com/files/package.json#license
but that is optional from my point of view.

## Steps to Reproduce (for bugs)

1. Add licence field to package.json
2. run `lerna publish`

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.2.1 |
| `npm --version`  | 5.6.0 |
| `node --version` | v8.9.4 |

| OS | Version |
| ---: | :--- |
| Windows 10 | 1803 |

evocateur:
As [explained](https://github.com/lerna/lerna/pull/1465#issuecomment-405184876) in that PR, I don't think that is the case. More context [here](https://github.com/babel/babel/pull/7308#issuecomment-367615152). The fact that tools like `license-checker` work without the file being present in the package tarball doesn't change the requirements of various licenses, such as MIT.

The license [pattern](https://github.com/lerna/lerna/blob/ac0baa7b1d5c378f794c960ab13d70242d19ddc8/core/project/index.js#L195) is very broad, allowing any (English) spelling, case, or file extension (or none at all). It is very easy to [provide a license file](https://help.github.com/articles/adding-a-license-to-a-repository/) in the root of a monorepo, and in any case, this warning does not adversely affect the performance of `lerna publish`. If you want to hide it, `--loglevel error` will do the trick.

[...]

PR #1921, opened 2019-02-10
fix(publish): Make the missing license warning clearer and more actionable

ken-kenware:
Previously, the missing license warning just indicate that packages were missing a license. This change modifies the message to indicate it is looking for a root LICENSE file, making it easier for users to understand how to fix the problem.

## Description
Modified the error warning to be clearer. I mostly just used the language in the test description, since it seemed to make sense to me.

## Motivation and Context
Resolves issue #1666 (https://github.com/lerna/lerna/issues/1666)

## How Has This Been Tested?
`npm i` + `npm test`

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -380,11 +380,10 @@ class PublishCommand extends Command {
           this.packagesToBeLicensed = [];
           this.logger.warn(
             "ENOLICENSE",
-            `Packages ${packagesWithoutLicense
-              .map(pkg => pkg.name)
-              .join(", ")} are missing a root LICENSE file \n` +
-              "To fix this, add a LICENSE.md file in the root of this repository. " +
-              "See https://choosealicense.com for additional guidance."
+            `Packages ${packagesWithoutLicense.map(pkg => pkg.name).join(", ")} are missing a root LICENSE file`,
+            `\n`,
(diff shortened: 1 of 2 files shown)


commands/publish/index.js
```diff
@@ -380,7 +380,7 @@ class PublishCommand extends Command {
           this.packagesToBeLicensed = [];
           this.logger.warn(
             "ENOLICENSE",
-            `Packages ${packagesWithoutLicense.map(pkg => pkg.name).join(", ")} are missing a license`
+            `Packages ${packagesWithoutLicense.map(pkg => pkg.name).join(", ")} are missing a root LICENSE file`
```
evocateur: That's certainly one way to resolve the warning, but it's not strictly what the warning is concerned about.

How about we add a second line to the warning instead, saying something like 
```
"To fix this, add a LICENSE.md file in the root of this repository. See https://choosealicense.com for additional guidance."
```

ken-kenware: Sure! I kind of figured something like this wouldn't be _exactly_ right since, it really isn't that much more helpful. But I wanted to start somewhere. I'll make the change, let me know if the style isn't right.

evocateur: Looks great, thanks!

evocateur:
Thanks!

PR #1931, opened 2019-02-13
feat(global-options): Default concurrency to logical CPU count

### LINKED ISSUE #1929: Change default concurrency from 4 to available number of cpus
## Expected Behavior
It would be nice if concurrency defaulted to `os.cpus()`

## Current Behavior
Concurrency defaults to 4 no matter what.

## Possible Solution
Change the DEFAULT_CONCURRENCY variable to be `os.cpus()`.
Alternatively, enable this only with a flag to avoid a breaking change.

## Context
This would allow jobs to scale easily with different instance sizes when running in things like EC2. I have some tasks that run sometimes in a 4 core machine and sometimes in a 16 core machine or even larger.

Most local machines have 4 cores anyway, so they would remain unchanged. But it's hard to maximize concurrency using only the CLI right now.

```json
"run-stuff": "lerna run task --concurrency 8"
```

If I have a package.json task like above, it's hard to access the number of available CPUS to set the concurrency level. I need to instead run the lerna task from a node script, which is less convenient. Ideally, it would always maximize concurrency without needing to avoid the CLI. Or at least have that be available with a flag, like `--maximize-concurrency`. 

If this is something you'll consider, I'm happy to do a PR to add it.

romansh-wixcom:
## Description
The PR makes little changes to set concurrency option to count of logical CPU cores by default.

## Motivation and Context
`--concurrency` option is set to `4` by default what can be not the best value with different count of CPUs.
Setting the option to specific number isn't good options cause the same script can be run on different servers.

Fixes #1929 

## How Has This Been Tested?

Unfortunately, my computer have 4 logical CPU what was the value before my changes, so for now I didn't test what is effected. I'm going to run lerna tests in virtual environments with different count of CPU cores.

## Types of changes

## Checklist:

### DIFF
--- a/core/command/index.js
+++ b/core/command/index.js
@@ -4,6 +4,7 @@ const _ = require("lodash");
 const dedent = require("dedent");
 const execa = require("execa");
 const log = require("npmlog");
+const os = require("os");
 
 const PackageGraph = require("@lerna/package-graph");
 const Project = require("@lerna/project");
@@ -14,7 +15,7 @@ const cleanStack = require("./lib/clean-stack");
 const logPackageError = require("./lib/log-package-error");
 const warnIfHanging = require("./lib/warn-if-hanging");
 
-const DEFAULT_CONCURRENCY = 4;
+const DEFAULT_CONCURRENCY = os.cpus().length;
 
 class Command {
   constructor(argv) {
--- a/core/global-options/README.md
+++ b/core/global-options/README.md
@@ -6,7 +6,7 @@
 
 ### `--concurrency`
 
-How many threads to use when Lerna parallelizes the tasks (defaults to `4`)
+How many threads to use when Lerna parallelizes the tasks (defaults to count of logical CPU cores)
 
 ```sh
 $ lerna publish --concurrency 1
--- a/core/command/__tests__/command.test.js
+++ b/core/command/__tests__/command.test.js
@@ -9,6 +9,10 @@ const tempy = require("tempy");
 
 // partially mocked
 const ChildProcessUtilities = require("@lerna/child-process");
+const os = require("os");
+
+// normalize concurrency across different environments (localhost, CI, etc)
+jest.spyOn(os, "cpus").mockImplementation(() => new Array(42));
 
 // helpers
 const initFixture = require("@lerna-test/init-fixture")(__dirname);
@@ -70,14 +74,14 @@ describe("core-command", () => {
       const command = testFactory({ concurrency: "bla" });
       await command;
 
-      expect(command.concurrency).toBe(4);
+      expect(command.concurrency).toBe(42);
     });
 
     it("should fall back to default if concurrency given is 0", async () => {
       const command = testFactory({ concurrency: 0 });
       await command;
 
-      expect(command.concurrency).toBe(4);
+      expect(command.concurrency).toBe(42);
     });
 
     it("should fall back to 1 if concurrency given is smaller than 1", async () => {
--- a/core/global-options/index.js
+++ b/core/global-options/index.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const os = require("os");
+
 module.exports = globalOptions;
 
 function globalOptions(yargs) {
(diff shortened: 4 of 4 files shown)


core/global-options/README.md
```diff
@@ -6,7 +6,7 @@
 
 ### `--concurrency`
 
-How many threads to use when Lerna parallelizes the tasks (defaults to `4`)
+How many threads to use when Lerna parallelizes the tasks (defaults to count of logical CPU cores)
```
evocateur: You'll need to update the option configuration's [`defaultDescription`](https://github.com/lerna/lerna/blob/5c1b84c699d80dbd8c5cce009aae8b82569e32a8/core/global-options/index.js#L14) to reflect this, as well.

core/command/index.js
```diff
@@ -14,7 +15,7 @@ const cleanStack = require("./lib/clean-stack");
 const logPackageError = require("./lib/log-package-error");
 const warnIfHanging = require("./lib/warn-if-hanging");
 
-const DEFAULT_CONCURRENCY = 4;
+const DEFAULT_CONCURRENCY = os.cpus().length;
```
evocateur: To make the [tests](https://github.com/lerna/lerna/blob/5c1b84c699d80dbd8c5cce009aae8b82569e32a8/core/command/__tests__/command.test.js#L61-L89) pass, you'll need to mock out this result, otherwise it will always fail in environments other than your localhost (I have 12 logical cores on my laptop, for example).

```patch
diff --git a/core/command/__tests__/command.test.js b/core/command/__tests__/command.test.js
index 9cbbc60..0724038 100644
--- a/core/command/__tests__/command.test.js
+++ b/core/command/__tests__/command.test.js
@@ -9,6 +9,10 @@ const tempy = require("tempy");
 
 // partially mocked
 const ChildProcessUtilities = require("@lerna/child-process");
+const os = require("os");
+
+// normalize concurrency across different environments (localhost, CI, etc)
+jest.spyOn(os, "cpus").mockImplementation(() => new Array(42));
 
 // helpers
 const initFixture = require("@lerna-test/init-fixture")(__dirname);
@@ -70,14 +74,14 @@ describe("core-command", () => {
       const command = testFactory({ concurrency: "bla" });
       await command;
 
-      expect(command.concurrency).toBe(4);
+      expect(command.concurrency).toBe(42);
     });
 
     it("should fall back to default if concurrency given is 0", async () => {
       const command = testFactory({ concurrency: 0 });
       await command;
 
-      expect(command.concurrency).toBe(4);
+      expect(command.concurrency).toBe(42);
     });
 
     it("should fall back to 1 if concurrency given is smaller than 1", async () => {
```

evocateur:
(My concerns aside, I'm in favor of this change)

bdwain:
> Overall, I'm concerned that `os.cpus().length` isn't always the appropriate value for concurrent operations. Not that `4` is much better, it's pretty arbitrary.

I agree. But when it's not, my guess is they probably have a hard coded number in mind anyway, so they can just set the concurrency if they want. 

If people wanted to use half of the cpus, or all but one, they'd need a node script to do that. But that's the case today today, and if it was that popular of a request, there could be options such as `--use-half-cpus` or `--use-all-but-one-cpu`. I doubt they'd be necessary though.

core/command/index.js
romansh-wixcom: Thanks for the patch! I will apply it and test.

core/global-options/README.md
romansh-wixcom: Sure, I will update the `defaultDescription`

fyodorvi:
FYI, it seems that this feature caused us some grief in CircleCI where running a command across 30+ package would randomly fail on one of the packages with:

```
lerna ERR! yarn run build stderr:
error Command failed with signal "SIGKILL".
```

We fixed it by explicitly adding `--concurrency 2` (2 is the default number of core CircleCI gives you).

sidharthachatterjee:
CircleCI doesn't report number of cores correctly so this fails on it without concurrency being added explicitly 😞

evocateur:
Sounds like a Circle CI bug to me.

PR #1939, opened 2019-02-19
feat(run, exec): Adds `--force-local` flag to exec and run commands

Noam3kCH:
When executing run and exec commands, execution order was not per the dependancy graph in the case where local dependencies had unmatched versions. This PR adds `--force-local` flag support for these command to allow running them in topological order.

## Motivation and Context
in our environment we use `lerna bootstrap --force-local` which works as expected, however when executing `lerna run build` the build order is not per dependency chain, causing unexpected errors. 

## How Has This Been Tested?
We added a test, in coordination with the `bootstrap --force-local` test to check the specific case.
namely, mismatching package dependencies and expecting them to run in the correct topological order

## Types of changes

## Checklist:


amiryonatan:
hey @evocateur,
We added support for --force-local in a few more commands. 
Can you please give us your feedback ? 

Thanks!

utils/listable/lib/listable-format.js
```diff
@@ -37,12 +37,13 @@ function parseViewOptions(options) {
     showNDJSON: options.ndjson,
     showParseable: options.parseable,
     isTopological: options.toposort,
+    forceLocal: options.forceLocal,
```
evocateur: I'm not a fan of `--force-local` in the first place, and this isn't appropriate, as none of the listable commands accepts `--force-local` (nor should they).

commands/exec/README.md
```diff
@@ -79,3 +79,11 @@ Pass `--no-bail` to disable this behavior, executing in _all_ packages regardles
...
 Disable package name prefixing when output is streaming (`--stream` _or_ `--parallel`).
 This option can be useful when piping results to other processes, such as editor plugins.
+
+### `--force-local`
+
+```sh
+$ lerna exec --force-local <command>
+```
+
+When passed, this flag causes the `exec` command to always execute in topological order regardless of matching version range.
```
evocateur: That's not an accurate statement. When your local dependent ranges are not satisfied, they are not considered local dependencies, thus they are not topologically ordered, as they have no topological relationship.

Noam3kCH:
@evocateur
The motivation behind this PR came after we ran into unexpected behaviour where after you bootstrap with `--force-local` and have everything linked, the run command will execute in the same manner. 
This allows for example to test if some module dependencies which are behind could be updated to latest in single build command. Otherwise the use of the force local command is very limiting. 

Please let us know if you would like us to fix this pr, and in what approach, or drop it.
Thanks!

amiryonatan:
@evocateur please advise on how to proceed?

3y3:
Is this pr not approved only for `README.md` reason?
@evocateur can you help to approve pr? Any idea how to fix readme?

PR #2031, opened 2019-04-11
fix(exec): Handle node string error codes before setting process.exitCode

KonstantinSimeonov:
Thanks for the awesome project!

Handle node string error codes before setting process.exitCode in implementation of `lerna exec` command.
## Description

Added a switch statement based on the result exit code type that does the following:
- in case of a number exit code simply returns it
- in case of a string error code looks it up in `os.constants.errno` and returns a number from there
- in case of another type throws a `TypeError`

Also added a unit test, but since it can only fail under windows, and lerna does not seem to run tests under windows, I'm not sure how useful that test will be.

## Motivation and Context

Yesterday I ran into a problem with `lerna exec` under windows 10 (this does not seem to happen under linux). I do the following:

```bat
lerna exec --concurrency=1 --no-bail -- custom-command
```

Where `custom-command` could potentially fail with an `ENOENT`. The commands fail, but `lerna exec` reports success and `NaN` exit code:
GH SS

The exit code gets swallowed and some tools, like git hooks, successfully pass, when that shouldn't be the case.

Some investigation into your `child-process` module and `execa` led me to the observation that a certain code path returns such a `result` whose `result.code` property can be a node error string like `ENOENT`, `EEXIST` and so on.

I'm not a 100% sure that my way is the best way to address this, moving this logic to `child-process` might be better.

## How Has This Been Tested?

Added a unit test which fails under windows without my changes and passes when they're applied. Also gave it a spin on my linux machine.

## Types of changes

## Checklist:

### DIFF
--- a/commands/exec/index.js
+++ b/commands/exec/index.js
@@ -1,5 +1,7 @@
 "use strict";
 
+const os = require("os");
+
 const ChildProcessUtilities = require("@lerna/child-process");
 const Command = require("@lerna/command");
 const batchPackages = require("@lerna/batch-packages");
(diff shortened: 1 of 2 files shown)


evocateur:
I ended up moving the logic into `@lerna/child-process` (covering `lerna run` and the single-package `exec` failure cases), but I appreciate the original effort!

KonstantinSimeonov:
That's what I think is more appropriate as well, to potentially handle the case in all places, but I didn't feel like going through all the cases to see whether I would break something with the change :P

PR #2034, opened 2019-04-12
Fix typo in publish docs

msrose:


evocateur:
Thanks so much for this PR, #2030 has already dealt with it.

msrose:
Ah sorry for the noise then, should have scanned the open PRs before opening a new one 😅

PR #2041, opened 2019-04-16
fix(version): Resolve prerelease for version without bump

### LINKED ISSUE #2025: Detect current prerelease in lerna version custom prerelease
When using the `lerna version` CLI, selecting the `Custom Prerelease` option defaults to `alpha`. If the package is already in `beta`, it would be nice if it defaulted to `beta`.

## Expected Behavior

When selecting `Custom Prerelease`, the default matches the current prerelease. If the package is currently in `beta`, the `Custom Prerelease` should select `beta` as the default (no user input required).

## Current Behavior

The default prerelease is `alpha`. If a package is currently in `beta`, the default will revert it to an `alpha` release`.

## Context

Automatically detecting the current prerelease would be a nice quality of life change to prevent accidentally reverting the prerelease type.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.13.1 |
| `npm --version`  | 6.7.0 |
| `node --version` | 11.13.0 |

pshrmn:
Fixes #2025

## Description

This PR modifies the `resolvePrereleaseId` function to use the package's prerelease identifier as the default identifier when it exists, not only when when the `bump` CLI arg is an increment keyword.

## Motivation and Context

Currently, the existing prerelease identifier is only used when `lerna version` is provided an increment bump. This means that if the current version is `1.0.0-beta.2`, the "Custom Prerelease" prompt's default value is `alpha`.

## How Has This Been Tested?

I ran the existing tests, but haven't added any new ones yet. I was looking at the tests in [`prompt-version.test.js`](https://github.com/lerna/lerna/blob/c599c64b11f06afe2fa16ed2c448b16de7da147a/commands/version/__tests__/prompt-version.test.js), but those use a custom `resolvePrereleaseId` function. I'd be happy to add a test, but I'm not sure what the best way to test the `resolvePrereleaseId` function created in `getVersionsForUpdates` is.

I'm not actually sure why the `isPrerelease` check exists, so I'm not sure if removing it will break anything. I didn't have any luck tracing the code's history (it was introduced in #1522) to see why it was originally added.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -278,10 +278,9 @@ class VersionCommand extends Command {
     const { bump, conventionalCommits, preid } = this.options;
     const repoVersion = bump ? semver.clean(bump) : "";
     const increment = bump && !semver.valid(bump) ? bump : "";
-    const isPrerelease = increment.startsWith("pre");
 
     const getExistingPreId = version => (semver.prerelease(version) || []).shift();
-    const resolvePrereleaseId = existingPreid => preid || (isPrerelease && existingPreid) || "alpha";
+    const resolvePrereleaseId = existingPreid => preid || existingPreid || "alpha";
 
     const makeGlobalVersionPredicate = nextVersion => {
       this.globalVersion = nextVersion;


commands/version/index.js
```diff
@@ -278,10 +278,9 @@ class VersionCommand extends Command {
     const { bump, conventionalCommits, preid } = this.options;
     const repoVersion = bump ? semver.clean(bump) : "";
     const increment = bump && !semver.valid(bump) ? bump : "";
-    const isPrerelease = increment.startsWith("pre");
 
     const getExistingPreId = version => (semver.prerelease(version) || []).shift();
-    const resolvePrereleaseId = existingPreid => preid || (isPrerelease && existingPreid) || "alpha";
```
evocateur: It is certainly distressing that removing this inline conditional doesn't break _any_ tests. The prompting _is_ very poorly tested, I'll be the first to admit.

I would expect a modification to `makePromptVersion`'s arguments to be necessary to establish the correct `preid` in the dialog, at the very least. I think the best way to write the tests is to do them first, before changing this code, just so we can establish what _fails_ in the current context. Without that, I really can't say if this is the right approach.

pshrmn: Sounds good. I'll revert locally and poke around with the tests to try to find the use case that relies on the `isPrerelease` conditional.

pshrmn: I've been looking at this a bit and I think that the `isPrerelease` condition is a no-op.

The result of `resolvePrereleaseId` function is the identifier passed as the third argument to `semver.inc`. The identifier is only used when the second argument is a `pre___` type; for `major`, `minor`, and `patch` releases, the identifier is ignored.

If the condition is removed, a failing test would happen when `isPrerelease` is false and `existingPreid` is not `undefined`; the third argument to `semver.inc` would be `existingPreid` instead of `alpha`. However, since the value returned by `resolvePrereleaseId` is never used if `isPrerelease` is false, the value doesn't actually matter.

evocateur: You're right, of course. Thanks for taking the time to explain it in detail. I verified it locally with this change:
```diff
--- a/commands/version/__tests__/version-bump-prerelease.test.js
+++ b/commands/version/__tests__/version-bump-prerelease.test.js
@@ -152,8 +152,8 @@ test("independent version prerelease does not bump on every unrelated change", a
   // simulate choices for pkg-a then pkg-b
   prompt.mockChoices("patch", "PRERELEASE");
   prompt.input.mockImplementationOnce((msg, cfg) =>
-    // a convoluted way of entering "bumps" at the prompt
-    Promise.resolve(cfg.filter("bumps"))
+    // the _existing_ "bumps" prerelease ID should be preserved
+    Promise.resolve(cfg.filter())
   );
 
   await lernaVersion(cwd)();
```

PR #2073, opened 2019-05-06
Add gitlab releases to version

LukeeeeBennett:
* Deprecate `version --github-release` in favour of `version --create-release=[gitlab|github]`.
* Add `GitLabClient` w/ `createReleases` method to create releases using GitLab API. 
* Add `createRelease` function to DRY GitLab and GitHub releases.

## Description

I think it's pretty much all there. Let me know if you need any more info.

### Links

https://docs.gitlab.com/ee/user/project/releases/index.html
https://docs.gitlab.com/ee/api/releases/index.html

### Screenshot

Screenshot 2019-05-06 at 04 11 37

## Motivation and Context

I use https://gitlab.com and self-hosted GitLab instances to host my favouritestest repos. I would like the releases feature to work for them.

## How Has This Been Tested?

There are tests for deprecation, `GitLabClient` and `--create-release=gitlab`. _(`--create-release=github` is tested in existing `--github-release` tests)_

Tested locally with:

```
$ GL_TOKEN=[TOKEN] npx lerna version --create-release=gitlab --conventional-commits
# AND
$ GL_TOKEN=[TOKEN] GL_API_URL="https://gitlab.selfhosted.io/api/v4" npx lerna version --create-release=gitlab --conventional-commits
```

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

**Note:** `--github-release` has been appropriately deprecated.

## Checklist:

### DIFF
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -51,7 +51,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--exact`](#--exact)
 - [`--force-publish`](#--force-publish)
 - [`--git-remote`](#--git-remote-name)
-- [`--github-release`](#--github-release)
+- [`--create-release`](#--create-release-type)
 - [`--ignore-changes`](#--ignore-changes)
 - [`--include-merged-tags`](#--include-merged-tags)
 - [`--message`](#--message-msg)
@@ -194,19 +194,25 @@ lerna version --git-remote upstream
 
 When run with this flag, `lerna version` will push the git changes to the specified remote instead of `origin`.
 
-### `--github-release`
+### `--create-release <type>`
 
 ```sh
-lerna version --github-release --conventional-commits
+lerna version --conventional-commits --create-release github
+lerna version --conventional-commits --create-release gitlab
 ```
 
-When run with this flag, `lerna version` will create an official GitHub release based on the changed packages. Requires `--conventional-commits` to be passed so that changelogs can be generated.
+When run with this flag, `lerna version` will create an official GitHub or GitLab release based on the changed packages. Requires `--conventional-commits` to be passed so that changelogs can be generated.
 
 To authenticate with GitHub, the following environment variables can be defined.
 
 - `GH_TOKEN` (required) - Your GitHub authentication token (under Settings > Developer settings > Personal access tokens).
 - `GHE_API_URL` - When using GitHub Enterprise, an absolute URL to the API.
 - `GHE_VERSION` - When using GitHub Enterprise, the currently installed GHE version. [Supports the following versions](https://github.com/octokit/plugin-enterprise-rest.js).
+-
+To authenticate with GitLab, the following environment variables can be defined.
+
+- `GL_TOKEN` (required) - Your GitLab authentication token (under User Settings > Access Tokens).
+- `GL_API_URL` - An absolute URL to the API, including the version. (Default: https://gitlab.com/api/v4)
 
 > NOTE: When using this option, you cannot pass [`--no-changelog`](#--no-changelog).
 
@@ -288,7 +294,7 @@ lerna version --conventional-commits --no-changelog
 
 When using `conventional-commits`, do not generate any `CHANGELOG.md` files.
 
-> NOTE: When using this option, you cannot pass [`--github-release`](#--github-release).
+> NOTE: When using this option, you cannot pass [`--create-release`](#--create-release-type).
 
 ### `--no-commit-hooks`
 
--- a/commands/version/command.js
+++ b/commands/version/command.js
@@ -40,9 +40,10 @@ exports.builder = (yargs, composed) => {
       requiresArg: true,
       defaultDescription: "origin",
     },
-    "github-release": {
-      describe: "Create an official GitHub release for every version.",
-      type: "boolean",
+    "create-release": {
+      describe: "Create an official GitHub or GitLab release for every version.",
+      type: "string",
+      choices: ["gitlab", "github"],
     },
     "ignore-changes": {
       describe: [
@@ -171,6 +172,11 @@ exports.builder = (yargs, composed) => {
       hidden: true,
       type: "boolean",
     })
+    .option("github-release", {
+      // TODO: remove in next major release
+      hidden: true,
+      type: "boolean",
+    })
     .check(argv => {
       /* eslint-disable no-param-reassign */
       if (argv.ignore) {
@@ -200,6 +206,12 @@ exports.builder = (yargs, composed) => {
         delete argv.skipGit;
         delete argv["skip-git"];
         log.warn("deprecated", "--skip-git has been replaced by --no-git-tag-version --no-push");
+      }
+
+      if (argv.githubRelease) {
+        argv.createRelease = "github";
+        delete argv.githubRelease;
+        log.warn("deprecated", "--release has been replaced by --create-release=github");
       }
       /* eslint-enable no-param-reassign */
 
--- a/core/project/lib/deprecate-config.js
+++ b/core/project/lib/deprecate-config.js
@@ -6,6 +6,10 @@ const path = require("path");
 
 module.exports = compose(
   // add new predicates HERE
+  remap("command.version.githubRelease", "command.version.createRelease", {
+    alsoRoot: true,
+    toValue: value => value && "github",
+  }),
   remap("command.publish.npmTag", "command.publish.distTag", { alsoRoot: true }),
   remap("command.publish.cdVersion", "command.publish.bump", { alsoRoot: true }),
   remap("command.publish.ignore", "command.publish.ignoreChanges"),
@@ -21,9 +25,10 @@ module.exports = compose(
  * @param {String} target Path of renamed option
  * @param {Object} opts Optional configuration object
  * @param {Boolean} opts.alsoRoot Whether to check root config as well
+ * @param {Function} opts.toValue Return the new config value given the current value
  * @return {Function} predicate accepting (config, filepath)
  */
-function remap(search, target, { alsoRoot } = {}) {
+function remap(search, target, { alsoRoot, toValue } = {}) {
   const pathsToSearch = [search];
 
   if (alsoRoot) {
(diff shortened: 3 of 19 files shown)


commands/version/lib/Releases.js
```diff
@@ -0,0 +1,81 @@
...
+    if (this.options.conventionalCommits !== true) {
+      throw new ValidationError("ERELEASE", "To create a Release, you must enable --conventional-commits");
+    }
+
+    if (this.options.changelog === false) {
+      throw new ValidationError("ERELEASE", "To create a Release, you cannot pass --no-changelog");
+    }
+  }
+
+  release(tags, releaseNotes, execOpts) {
```
LukeeeeBennett: Self-review: Can instantiate with these so might as well?

commands/version/lib/Releases.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const repo = parseGitRepo(this.options.gitRemote, execOpts);
+
+    log.silly("Releases repo", repo);
+
+    const client = this.client(repo);
+
+    return Promise.all(
+      releaseNotes.map(({ notes, name }) => {
+        const tag = name === "fixed" ? tags[0] : tags.find(t => t.startsWith(`${name}@`));
```
LukeeeeBennett: Self-review: Opportunity to improve readability here now.

commands/version/lib/Releases.js
```diff
@@ -0,0 +1,81 @@
...
+        );
+    }
+  }
+
+  host(repo) {
+    if (process.env.GL_TOKEN || repo.source.includes("gitlab")) {
+      return "gitlab";
+    }
+
+    if (process.env.GH_TOKEN || repo.source.includes("github")) {
```
LukeeeeBennett: Self-review: `repo` isn't required as tokens are required in each case.

commands/version/lib/Releases.js
LukeeeeBennett: Actually, should think about just using a factory method in 1 location 🤔

commands/version/command.js
```diff
@@ -40,8 +40,8 @@ exports.builder = (yargs, composed) => {
       requiresArg: true,
       defaultDescription: "origin",
     },
-    "github-release": {
-      describe: "Create an official GitHub release for every version.",
+    "release": {
```
evocateur: So this is a breaking change, as-is. I would suggest renaming the flag `create-release`, requiring a string argument from the enum `["gitlab", "github"]`, so we can be explicit about _which_ type of release to create. We then need to provide a backward-compatible hidden alias of `--github-release` to `--create-release=github`, similar to how `--skip-git` is [handled](https://github.com/lerna/lerna/blob/ea41fe91328858af0b5f5b8c715b17f1af637436/commands/version/command.js#L196-L198).

Just in case anyone has set `"githubRelease": true` in their `lerna.json`, you'll _also_ need to [deprecate the config](https://github.com/lerna/lerna/blob/ea41fe91328858af0b5f5b8c715b17f1af637436/core/project/lib/deprecate-config.js#L8)... although, argh, there's no way as-written in that file to change a boolean prop to a string val...

Here's a sketch of how I think it might look:
```patch
 module.exports = compose(
   // add new predicates HERE
+  remap("command.version.githubRelease", "command.version.createRelease", { alsoRoot: true, toValue: "github" }),
   remap("command.publish.npmTag", "command.publish.distTag", { alsoRoot: true }),
```
(`toValue` would default to the existing field value)

utils/gitlab-client/lib/GitLabClient.js
```diff
@@ -0,0 +1,35 @@
+const path = require("path");
+
+const log = require("npmlog");
+const request = require("request-promise-native");
```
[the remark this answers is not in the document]
LukeeeeBennett: No no that's great, I was trying to look for the best option and found this in a deps list somewhere. 😂

commands/version/command.js
LukeeeeBennett: Thank you, I was hoping there would be some direction on deprecation and this sounds great.

> there's no way as-written in that file to change a boolean prop to a string val...

We can take a look at that too, happy for it to be a separate PR if you would prefer.

LukeeeeBennett: @evocateur How would you like to handle the `"githubRelease": false` case?

Should we make `toValue` a function `toValue = (value) => val && 'github'`?

commands/version/command.js
```diff
@@ -201,6 +207,12 @@ exports.builder = (yargs, composed) => {
         delete argv["skip-git"];
         log.warn("deprecated", "--skip-git has been replaced by --no-git-tag-version --no-push");
       }
+
+      if (argv.release) {
+        argv.createRelease = "github";
+        delete argv.githubReleases;
```
LukeeeeBennett:
```suggestion
        delete argv.githubRelease;
```

LukeeeeBennett: and the deprecation message below, `release` -> `github-release`

core/project/lib/deprecate-config.js
```diff
@@ -36,12 +41,17 @@ function remap(search, target, { alsoRoot } = {}) {
...
 
+        const fromVal = dotProp.get(obj.config, searchPath);
+        const toVal = toValue ? toValue(fromVal) : fromVal;
+        const from = stringify({ [searchPath]: fromVal });
+        const to = stringify({ [target]: toVal });
+
         log.warn(
           "project",
-          `Deprecated key "${searchPath}" found in ${localPath}\nPlease rename "${searchPath}" => "${target}"`
+          `Deprecated key "${searchPath}" found in ${localPath}\nPlease update ${from} => ${to}`
```
evocateur: I think you could factor out the update message generation into a method of its own, it would be clearer. Also, if the values haven't changed, we should probably omit them from the message, it's just confusing.

utils/gitlab-client/lib/GitLabClient.js
```diff
@@ -0,0 +1,41 @@
...
+    return fetch(releaseUrl, {
+      method: "post",
+      body: JSON.stringify({ name, tag_name: tagName, description: body }),
+      headers: {
+        "PRIVATE-TOKEN": this.token,
+        "Content-Type": "application/json",
+      },
+    }).then(({ ok, status, statusText }) => {
+      if (!ok) {
+        log.error("create release", `Failed to create release\nRequest returned ${status} ${statusText}`);
```
evocateur: The first argument here, the topic, should be `"gitlab"`

utils/gitlab-client/lib/GitLabClient.js
```diff
@@ -0,0 +1,41 @@
...
+      } else {
+        log.silly("Created release successfully.");
```
evocateur: There should be a log topic argument (`"gitlab"`) before the message.

commands/version/command.js
evocateur: yes, we should not be introducing `--release`, as we're renaming `--github-release` -> `--create-release <type>`

commands/version/command.js
```diff
@@ -171,6 +172,11 @@ exports.builder = (yargs, composed) => {
       hidden: true,
       type: "boolean",
     })
+    .option("release", {
```
evocateur:
```suggestion
    .option("github-release", {
```

commands/version/command.js
```diff
@@ -201,6 +207,12 @@ exports.builder = (yargs, composed) => {
         delete argv["skip-git"];
         log.warn("deprecated", "--skip-git has been replaced by --no-git-tag-version --no-push");
       }
+
+      if (argv.release) {
```
evocateur:
```suggestion
      if (argv.githubRelease) {
```

commands/version/README.md
```diff
@@ -43,7 +43,7 @@ If you have any packages with a prerelease version number (e.g. `2.0.0-beta.3`)
 - [`--exact`](#--exact)
 - [`--force-publish`](#--force-publish)
 - [`--git-remote`](#--git-remote-name)
-- [`--github-release`](#--github-release)
+- [`--release`](#--release)
```
evocateur:
```suggestion
- [`--create-release`](#--create-release-type)
```

commands/version/README.md
```diff
@@ -161,19 +161,24 @@ lerna version --git-remote upstream
 
 When run with this flag, `lerna version` will push the git changes to the specified remote instead of `origin`.
 
-### `--github-release`
+### `--release`
```
evocateur:
```suggestion
### `--create-release <type>`
```

commands/version/README.md
```diff
@@ -161,19 +161,24 @@ lerna version --git-remote upstream
...
 
 ```sh
-lerna version --github-release --conventional-commits
+lerna version --release --conventional-commits
```
evocateur: Let's reverse the order of the flags and show both option values, diffing the end-of-line is easier to read
```suggestion
lerna version --conventional-commits --create-release github
lerna version --conventional-commits --create-release gitlab
```

commands/version/README.md
```diff
@@ -255,7 +260,7 @@ lerna version --conventional-commits --no-changelog
 
 When using `conventional-commits`, do not generate any `CHANGELOG.md` files.
 
-> NOTE: When using this option, you cannot pass [`--github-release`](#--github-release).
+> NOTE: When using this option, you cannot pass [`--release`](#--release).
```
evocateur:
```suggestion
> NOTE: When using this option, you cannot pass [`--create-release`](#--create-release-type).
```

commands/version/__tests__/version-github-release.test.js
```diff
@@ -19,19 +19,19 @@ const initFixture = require("@lerna-test/init-fixture")(__dirname);
 // test command
 const lernaVersion = require("@lerna-test/command-runner")(require("../command"));
 
-test("--github-release does not create a release if --no-push is passed", async () => {
+test("--release does not create a release if --no-push is passed", async () => {
```
evocateur: Might as well revert this file, as the tests _should_ still work if you've configured the hidden option fallback correctly.

LukeeeeBennett:
@evocateur ready for another pass :) Thanks for your help so far.

wldcordeiro:
🎉 This looks so amazing thanks for doing this @LukeeeeBennett I had seen this feature and made the request and hadn't yet found the time to try this.

commands/version/README.md
```diff
@@ -51,7 +51,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--exact`](#--exact)
 - [`--force-publish`](#--force-publish)
 - [`--git-remote`](#--git-remote-name)
-- [`--github-release`](#--github-release)
+- [`--create-release`](#--create-release-type)
```
wldcordeiro: This would force a 4.x wouldn't it as it's changing the argument entirely. Would we want to keep the old one and add the new one and deprecate later in a 4.x with more changes?

LukeeeeBennett: This is correct. The old option can be used but it is deprecated.

https://github.com/lerna/lerna/pull/2073/files#diff-9577e51e2f56e073be1dcdf9bf503198R214

and

https://github.com/lerna/lerna/pull/2073/files#diff-0450c85a73670cf460e9ce9c771f100fR9

Thanks for have a look :)

Issue #955, opened 2017-08-04
[Feature]: Update peer dependency versions at publish time


nicojs:
Given this package structure:

```
root
| -- packages
     | -- foo
     | -- bar
```

And `bar` is a peerDependency of `foo`.

When we publish a new version of `foo` and `bar` also update of the peerDependency in `foo` to the latest version.

We should only update the peer dependency on `bar` if the following conditions are met:

* When the `peerDependency` on `bar` was an exact version. So `"bar": "^0.14 || ^15.0.0"` should **not** be updated, but `"bar": "0.5.6"` should be updated. **Question**: what about `"bar": "^0.5.6"`?
* When the `foo` dependency is being published. i.e. we should not force a release of `foo` if `bar` is updated and we're running in independent versioning mode. **Question**: or should we force it?

evocateur:
That line in PublishCommand _would_ seem to indicate the intention, but I don't see any tests validating it.

[Package##dependencies()](https://github.com/lerna/lerna/blob/master/src/Package.js#L39) is being iterated over in [PublishCommand##updatePackageDepsObject()](https://github.com/lerna/lerna/blob/master/src/commands/PublishCommand.js#L486), which among other things notably lacks `devDependencies`. I'm not sure `Package##allDependencies()` is needed here (it's used in `BootstrapCommand` and `PackageGraph`, exclusively), and in any case, that doesn't include `peerDependencies`.

A test of PublishCommand that makes assertions about sibling peer dependency updates seems in order.

nicojs:
Hmm ok. So it works, but isn't tested or documented. What about my 2 questions? 

Earlier today, when releasing [stryker](https://github.com/stryker-mutator/stryker) (independent versioning with conventional commits), we noticed that the peer dependencies were updated (yay!), but not in line with what we would expect with conventional commits.

For example: `bar` gets a major update. For example `0.5.0` -> `0.6.0`. I would `foo` to also get a `major` update, because it is now peer dependent on the next major release of `bar`.

Shepless:
So just to clarify - does lerna bump the peerDependencies between packages?

evocateur:
@Shepless As I noted above, there are no tests to validate this assumption, and I don't use `peerDependencies` very much, myself. So, "maybe"?

[...]

Issue #995, opened 2017-08-29
Limit updated packages only to those that have actual changes


darrylhodgins:
## Expected Behavior

For a project with a large number of packages, `lerna` should be able to run a _unit test_ script on a limited set of packages, in order for CI to quickly identify candidates for merge.

## Current Behavior

Consider a simplified set of packages:

```
- root package
    - child package 1
    - child package 2
        - grandchild package
```

Suppose there are changes only to the `grandchild package`.  Currently, `lerna run <somescript>` will run `<somescript>` on `root package`, `child package 2` and `grandchild package`.

Suppose also, that this project has _proper_ unit tests which are not mistakenly _integration_ tests.  In that case, since there are no changes to `child package 2`, it should not be necessary to run _unit_ tests on `root package`.

## Expected Behavior

It should be possible to run a variant of `lerna run <somescript>` that only runs `<somescript>` on `child package 2` and `grandchild package`.

## Possible Solution

TBD

## Context
I have a project with 70+ packages.  We use `--since` to limit packages to those with changes.  In some scenarios, more than 50 of the packages are caught up in the limited CI build, for a trivial change to a "core" dependency.

PR #2075, opened 2019-05-07
fix(package): update local peerDependencies

raymondfeng:
Fixes https://github.com/lerna/lerna/issues/955

## Description

As described in #995, `peerDependencies` is not updated for local packages.

## Motivation and Context

`lerna version` should update local packages in `peerDependencies`.

## How Has This Been Tested?

I couldn't find existing test suite for `Package.updateLocalDependencies()`.

## Types of changes

## Checklist:

### DIFF
--- a/core/package/index.js
+++ b/core/package/index.js
@@ -185,6 +185,11 @@ class Package {
       depCollection = this.devDependencies;
     }
 
+    // fall back to peerDependencies
+    if (!depCollection || !depCollection[depName]) {
+      depCollection = this.peerDependencies;
+    }
+
     if (resolved.registry || resolved.type === "directory") {
       // a version (1.2.3) OR range (^1.2.3) OR directory (file:../foo-pkg)
       depCollection[depName] = `${savePrefix}${depVersion}`;


core/package/index.js
```diff
@@ -185,6 +185,11 @@ class Package {
       depCollection = this.devDependencies;
     }
 
+    // fall back to peerDependencies
+    if (!depCollection || !depCollection[depName]) {
+      depCollection = this.peerDependencies;
```
evocateur: As @ljharb [noted](https://github.com/lerna/lerna/issues/955#issuecomment-332074663) in the linked issue, unconditionally updating the minimum range of peer dependencies like this would constitute a breaking change. You'd need a completely separate branch of logic to properly update it, IFF `semver.satisfies(depVersion, peerDeps[depName])` returns `false`, and _then_ you'd need to preserve whatever the current value was (if any) and either append a new alternation (`|| ^{depVersion}`, e.g., for a new major) or set a new value (I'm not even sure that's valid, since _adding_ new peers should _always_ be explicit).

JamesHenry:
Hi @raymondfeng 👋 

Thank you very much for taking the initiative to contribute this PR!

You may or may not know that lerna is now under the stewardship of Nrwl (announcement here https://github.com/lerna/lerna/issues/3121), a company with a long history of not just producing valuable open-source software (OSS), but also backing others (at the time of writing, Nrwl has donated over $50,000 to OSS it _hasn't_ created, see https://opencollective.com/nx for full details).

Quite simply, Nrwl ❤️  OSS, and is committed to making lerna the best it can be. We use it ourselves.

We hope you will continue to be a part of this community as we look to take things forward from here!

Please see https://github.com/lerna/lerna/discussions/3140 for more details on our plans for 2022.

In this case of this PR, because there has been no activity on it for over 3 years, we are going to close it for now. Further discussion on how to take this forward can take place on the linked issue.

Many thanks again 🙏

PR #2076, opened 2019-05-07
Support 2FA for npm via --otp

### LINKED ISSUE #1091: Support 2FA `--otp=123456`

rbuckton:
## Description
This adds an `--otp` CLI argument to `lerna publish` to support publishing to NPM registries that require two-factor authentication.

## Motivation and Context
It is currently possible to publish using 2FA *without* the `--otp` argument, however this requires setting an environment variable *prior* to calling `lerna publish`, and the process of setting an environment variable differs in different shells (i.e. bash, cmd, PowerShell, etc.). Providing this via an option simplifies the process and documents the capability.

## How Has This Been Tested?
I've added tests to verify the `--otp` option is passed to `npmPublish` as a configuration option and have also verified that the `--otp` option works by publishing a monorepo to NPM using an account that requires 2FA to publish.

## Types of changes

## Checklist:
- [x] All new and existing tests passed. (NOTE: There are tests that currently fail *without* this change when run on Windows).

Fixes #1091

### DIFF
--- a/commands/publish/README.md
+++ b/commands/publish/README.md
@@ -51,6 +51,7 @@ This is useful when a previous `lerna publish` failed to publish all packages to
 - [`--git-head <sha>`](#--git-head-sha)
 - [`--no-git-reset`](#--no-git-reset)
 - [`--no-verify-access`](#--no-verify-access)
+- [`--otp`](#--otp)
 - [`--preid`](#--preid)
 - [`--registry <url>`](#--registry-url)
 - [`--temp-tag`](#--temp-tag)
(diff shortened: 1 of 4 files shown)


rbuckton:
@evocateur, do you have time to take a look?

commands/publish/README.md
```diff
@@ -140,6 +141,16 @@ If you are using a third-party registry that does not support `npm access ls-pac
...
 
+### `--otp`
+
+When publishing to a registry that requires two-factor authentication, you can specify a "one-time password" using `--otp`:
+
+```sh
+lerna publish --otp 123456
+```
+
+> Please keep in mind that one-time passwords often expire within a few minutes of their generation. Please ensure adequate time for the publish process to complete.
```
evocateur: They actually expire within 30 seconds, so it's not clear that this solution would be guaranteed to work in all cases (imagine per-package builds that take 2 seconds each in a repo with 30 packages to publish...).

rbuckton: While this is true, until lerna has true OTP support, the only alternative is to set the NPM_CONFIG_OTP environment variable (which differs per system) and has the same downsides.

rbuckton:
Per-package OTP isn't terribly complicated with otplease in your branch, the problem is having to repeatedly enter the same OTP until a new one is needed. One way might be to maintain an in-memory cache of OTPs per registry and have otplease update that cache prior to a retry. I can tinker with something and put up a separate PR for that.

rbuckton:
Closing in favor of #2084

PR #2081, opened 2019-05-10
feat(bootstrap): Allow to specify custom folder for packages symlinks

paul-roman:
## Description

This allows to specify a `symLinkDir` in a package `package.json` to create the symbolic link from a specific folder of the package.

## Motivation and Context

This was motivated by the need to access to the folder structure of the transpiled source of one of our packages, thus we needed that only the `dist` folder was used in the symbolic link.

## How Has This Been Tested?

We added a `"symLinkDir": "dist"` field in one of our packages `package.json`, we ran `lerna bootstrap` and used another package which depends on the first one. Only the `dist` folder was used for the symbolic link, allowing us to access to the whole folder structure of the `dist` folder using
```js
import '@repo/child-package/directory/subdirectory/feature';
```

## Types of changes

## Checklist:

### DIFF
--- a/core/package/index.js
+++ b/core/package/index.js
@@ -53,6 +53,9 @@ class Package {
       rootPath: {
         value: rootPath,
       },
+      symLinkDir: {
+        value: pkg.symLinkDir,
+      },
       // internal state is "private"
       [PKG]: {
         configurable: true,
--- a/utils/symlink-dependencies/symlink-dependencies.js
+++ b/utils/symlink-dependencies/symlink-dependencies.js
@@ -77,7 +77,10 @@ function symlinkDependencies(packages, packageGraph, tracker) {
         });
 
         // create package symlink
-        chain = chain.then(() => createSymlink(dependencyNode.location, targetDirectory, "junction"));
+        const dependencyLocation = dependencyNode.pkg.symLinkDir
+          ? path.resolve(dependencyNode.location, dependencyNode.pkg.symLinkDir)
+          : dependencyNode.location;
+        chain = chain.then(() => createSymlink(dependencyLocation, targetDirectory, "junction"));
 
         // TODO: pass PackageGraphNodes directly instead of Packages
         chain = chain.then(() => symlinkBinary(dependencyNode.pkg, currentNode.pkg));


core/package/index.js
```diff
@@ -53,6 +53,9 @@ class Package {
       rootPath: {
         value: rootPath,
       },
+      symLinkDir: {
```
evocateur: `symlinkDir`, if you must. We don't call symlinks "sym-links", after all.

core/package/index.js
```diff
@@ -53,6 +53,9 @@ class Package {
...
+        value: pkg.symLinkDir,
```
evocateur: Can you point me to an example of _any_ other usage in the ecosystem? I don't want to add arbitrary fields to package.json just to satisfy a vanishingly small percentage of use cases.

utils/symlink-dependencies/symlink-dependencies.js
```diff
@@ -77,7 +77,10 @@ function symlinkDependencies(packages, packageGraph, tracker) {
         });
 
         // create package symlink
-        chain = chain.then(() => createSymlink(dependencyNode.location, targetDirectory, "junction"));
+        const dependencyLocation = dependencyNode.pkg.symLinkDir
+          ? path.resolve(dependencyNode.location, dependencyNode.pkg.symLinkDir)
```
evocateur: Wouldn't this be better served by the `lerna bootstrap` equivalent of [`--contents`](https://github.com/lerna/lerna/tree/master/commands/publish#--contents-dir)? This smells to high heaven of Angular-isms.

kamranayub:
Looks like this is halfway there to closing #2209.

The `publishConfig.directory` works well for publishing. In our monorepo we have two packages, `core` and `baked`. `baked` uses components from `core`. Since we use Rollup + Babel + TypeScript (these are libraries), we have a `dist` directory that we publish as the package. While we _can_ publish new versions of the `core` components, for local development we're looking at using `lerna link` to make that easier to have a nicer workflow.

In a file within `baked` you may have:

```ts
import { MyComponent } from '@org/core-components`
```

This works for a published package but since `lerna link` symlinks the root directory and _not_ the `dist` directory, for local development you must change the path:

```ts
import { MyComponent } from '@org/core-components/dist`
```

This PR and #2209 would make that workflow much easier. @paul-roman are you planning to add tests/make any additional changes? Otherwise I could possibly take a stab at contributing this.

Based on these discussions, what I'm seeing is some other changes:

- Change `symLinkDir` to use `contents` instead
- Make some other changes outlined in https://github.com/lerna/lerna/issues/2209#issuecomment-517788971

kamranayub:
I have opened a new PR that continues these changes: https://github.com/lerna/lerna/pull/2274

evocateur:
Closing in favor of #2274

PR #2104, opened 2019-05-23
fix(bootstrap): respect --force-local option

bajtos:
## Description

Before this change, `--force-local` option was not respected and monorepo-local packages outside of the defined --scope were still installed from the npm registry.

This commit fixes the code handling `--force-local` logic, so that:
- Local packages are resolved as links within the monorepo, as one would expect.
- The following warning IS NOT printed anymore:  _Installing local packages that do not match filters from registry_

## Motivation and Context

In https://github.com/strongloop/loopback-next, we have a test that runs `lerna bootstrap --scope {pkg}` to bootstrap a single package. I noticed that our test starts to fail when the code in the package depends on changes made elsewhere in the monorepo and those changes were not released to the npm registry yet.

After further digging I found that lerna is installing monorepo-local dependencies from the npm registry, instead of creating symlinks.

Eventually, I found https://github.com/lerna/lerna/commit/71174e4709 which changed the behavior of `lerna bootstrap` command in what I would consider a backwards-incompatible way, but fortunately added `--force-local` flag.

Except `--force-local` does not work right now!

## How Has This Been Tested?

It turns out the existing snapshot-based test for `--force-local` was using snapshot values describing the wrong behavior. I have updated these snapshots to match the desired outcome.

## Types of changes

## Checklist:

@evocateur You are the author of `--force-local` option, could you please review my pull request? Thanks! 🙏

### DIFF
--- a/commands/bootstrap/__tests__/__snapshots__/bootstrap-command.test.js.snap
+++ b/commands/bootstrap/__tests__/__snapshots__/bootstrap-command.test.js.snap
@@ -319,9 +319,6 @@ Object {
     "bar@^2.0.0",
     "foo@^1.0.0",
   ],
-  "packages/package-4": Array [
-    "package-3@^1.0.0",
-  ],
 }
 `;
 
@@ -331,6 +328,21 @@ Array [
     "_src": "packages/package-1",
     "dest": "packages/package-4/node_modules/@test/package-1",
     "type": "junction",
+  },
+  Object {
+    "_src": "packages/package-3",
+    "dest": "packages/package-4/node_modules/package-3",
+    "type": "junction",
+  },
+  Object {
+    "_src": "packages/package-3/cli1.js",
+    "dest": "packages/package-4/node_modules/.bin/package3cli1",
+    "type": "exec",
+  },
+  Object {
+    "_src": "packages/package-3/cli2.js",
+    "dest": "packages/package-4/node_modules/.bin/package3cli2",
+    "type": "exec",
   },
 ]
 `;
--- a/commands/bootstrap/index.js
+++ b/commands/bootstrap/index.js
@@ -161,7 +161,7 @@ class BootstrapCommand extends Command {
     chain = chain.then(filteredPackages => {
       this.filteredPackages = filteredPackages;
 
-      if (filteredPackages.length !== this.targetGraph.size) {
+      if (filteredPackages.length !== this.targetGraph.size && !this.options.forceLocal) {
         this.logger.warn("bootstrap", "Installing local packages that do not match filters from registry");
 
         // an explicit --scope, --ignore, or --since should only symlink the targeted packages, no others
--- a/commands/bootstrap/__tests__/bootstrap-command.test.js
+++ b/commands/bootstrap/__tests__/bootstrap-command.test.js
@@ -313,6 +313,24 @@ describe("BootstrapCommand", () => {
       expect(symlinkedDirectories(testDir)).toMatchSnapshot();
     });
 
+    it("should respect --force-local when a single package is in scope", async () => {
+      const testDir = await initFixture("basic");
+
+      await lernaBootstrap(testDir)("--scope", "package-4", "--force-local");
+
+      // no packages were installed from the registry
+      const installed = installedPackagesInDirectories(testDir);
+      expect(installed["packages/package-4"] || []).toEqual([]);
+
+      // package-3 was resolved as a local symlink
+      const symlinked = symlinkedDirectories(testDir);
+      expect(symlinked).toContainEqual({
+        _src: "packages/package-3",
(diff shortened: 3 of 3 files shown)


commands/bootstrap/__tests__/__snapshots__/bootstrap-command.test.js.snap
```diff
@@ -319,9 +319,6 @@ Object {
     "bar@^2.0.0",
     "foo@^1.0.0",
   ],
-  "packages/package-4": Array [
-    "package-3@^1.0.0",
-  ],
```
bajtos: Here!

This snapshot shows which dependencies were resolved from the npm registry. 

Before my change, the dependency of `packages/package-4` on `package-3` was resolved from the npm registry.

After my change, `package-3` is resolved as a link within monorepo. (See changes in the second snapshot below.)

commands/bootstrap/index.js
```diff
@@ -161,7 +161,7 @@ class BootstrapCommand extends Command {
     chain = chain.then(filteredPackages => {
       this.filteredPackages = filteredPackages;
 
-      if (filteredPackages.length !== this.targetGraph.size) {
+      if (filteredPackages.length !== this.targetGraph.size && !this.options.forceLocal) {
```
evocateur: Theoretically, this is what the `targetGraph` on line 168 was supposed to be doing:
```js
this.targetGraph = new PackageGraph(filteredPackages, "allDependencies", this.options.forceLocal);
```

I'm not sure this is the appropriate change.

bajtos:
@evocateur thank you for a quick response ❤️ 

> Can you write a (failing) test that illustrates the original problem, instead of changing the snapshot of an existing test?

Sure. If we wanted to keep using snapshot-based assertions, then I could pretty much copy `should respect --force-local` test as-is. 

https://github.com/lerna/lerna/blob/959a00a73bb0d72b1c277a1ea8d0143f42ecdd0f/commands/bootstrap/__tests__/bootstrap-command.test.js#L307-L314

I don't see much value in that, therefore in e0b0f991 I am adding a new test with explicit assertions. This test fails with the following message:

```
FAIL  commands/bootstrap/__tests__/bootstrap-command.test.js (5.061s)
  ● BootstrapCommand › with local package dependencies › should respect --force-local when a single package is in scope

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

    - Array []
    + Array [
    +   "@test/package-1@^0.0.0",
    +   "package-3@^1.0.0",
    + ]
```

commands/bootstrap/index.js
bajtos: > Theoretically, this is what the targetGraph on line 168 was supposed to be doing:

Here is the problem: `filteredPackages` contains only packages in scope (as determined via `--scope` flag).

When I run the new test I have added:
- Original `targetGraph` contains entries for the following packages: `package-3`, `@test/package-1`, `package-2`, `package-4`
- `filteredPackages` contains entries for `package-4` only
- When `new PackageGraph` is called, it creates entries for `package-4` only. This looks reasonable to me, because we gave the constructor a list of packages containing `package-4` only.

Another problem I see in the current implementation: even when I provide `--force-local`, `lerna bootstrap` prints the warning _Installing local packages that do not match filters from registry_. That does not make sense to me - as I understand `--force-local`, it's purpose is to ensure that local packages are symlinked. So either `--force-local` is not doing what it advertises, or the warning is lying.

> I'm not sure this is the appropriate change.

I see.

Here is the problem I need to fix: when I run `lerna bootstrap --scope `package-4`, I want lerna to symlink all local packages (`@test/package-1`, `package-3`). This used to be the behavior before 71174e4 was landed and I want to get it back :)

Now my assumption is that `--force-local` is supposed to address my needs but does not work as intended right now. If that's right and you still think my change is not appropriate, then could you please advise me how to proceed to get the issue fixed? It's enough to give me pointers to other places where to look.

If my assumption is not correct and the current behavior of `--force-local` is intended, then could you please advise me what would be an acceptable solution/change in lerna that will give me:
 - all local dependencies installed as symlinks, regardless of `--scope` arguments
 - no warnings printed to console

bajtos: Conceptually:

The block `if (filteredPackages.length !== this.targetGraph.size) { /* use scoped targetGraph */ }` was introduced by 71174e4. 

In my change, I am skipping this block entirely when `options.forceLocal` is enabled, thus reverting back to behavior before 71174e4.

bajtos:
@evocateur ping, have you have a chance to look at my response to your review comments?

bajtos:
@evocateur thank you for accepting my contribution ❤️ I totally understand your situation, my kids were toddlers just few years ago 👶 I really appreciate the time you took away from your family to help me solve the issue we are experiencing in our project 🙇

Issue #1282, opened 2018-02-23
Link and publish not the package directory but it's child `dist` folder instead


Kamilius:
Currently `$ lerna publish` as well as `$ lerna bootstrap` is publishing whole package folders with source files. The problem with this is that I need to publish not the package folders, but their `dist` folders instead, as well as using those `dist` folders as symlink source while using `$ lerna link/bootstrap`

## Expected Behavior
Ability to configure `publish` and `link` sources to make those operations specifically over built `dist` folders, but not over packages themselves. Ability to run `bootstrap run npm-script` should still be executed where it is now.

## Current Behavior
`publish` and `link` are publishing and linking whole packages with all of development related files

## Steps to Reproduce (for bugs)
1. go to `lerna` root
2. run `$ lerna bootstrap`
3. symlinks for packages are leading to other packages folders, not to those package's `dist`

<details><summary>lerna.json</summary><p>

```json
{
  "lerna": "2.8.0",
  "packages": [
    "packages/*"
  ],
  "version": "2.0.0-beta-1",
  "npmClient": "yarn"
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.8.0 |
| `npm --version`  | 5.6.0 |
| `yarn --version` | 1.3.2 |
| `node --version` | v8.9.3 |

| OS | Version |
| --- | --- |
| MacOS High Sierra | 10.13.3 (17D47) |

evocateur:
This is what the package.json [`files`](https://docs.npmjs.com/files/package.json#files) and [`main`](https://docs.npmjs.com/files/package.json#main) fields are for.

```json
{
  "name": "my-leaf-package",
  "version": "1.0.0",
  "main": "dist/index.js",
  "files": [
    "dist"
  ],
  "dependencies": {}
}
```

skidding:
@evocateur do you know of any reason why Lerna would publish files outside `package.json#files`? I'm experiencing this but only in some cases.

- [`files` NOT including src](https://github.com/react-cosmos/react-cosmos/blob/master/packages/react-cosmos-test/package.json#L7-L11)
- `src` gets published
  
![image](https://user-images.githubusercontent.com/250750/37393727-7b84adf6-27a4-11e8-8e33-d0327b7a2aee.png)

Any pointer to how to debug this is very appreciated! Related: https://github.com/react-cosmos/react-cosmos/issues/611

evocateur:
[...]

Issue #901, opened 2017-06-28
[Feature] Allow custom publish subdirectory


esakal:
Many libraries in the Angular ecosystem publish to NPM from a subdirectory that was cooked during the build process. The process is usually as followed:
- build the library in a subdirectory (i.e `dist` directory)
- copy package.json, license and readme to the `dist` directory
- do some package.json cleanups in the `dist` directory (delete devDepedencies, move dependencies to peerDependencies, remove scripts...)
- npm publish from the `dist` directory.

I understand that it is not a common technique for node.js package developers but i think it is common for web package developers and my libraries uses the same technique..

While embedding `lerna` into our shared packages monorepo - [kaltura-ng](https://github.com/kaltura/kaltura-ng) I read a lot of issues in lerna and googled about this topic. I read carefully the conversation of issue #91 and even used the same subject with my issue. Unless I missed new issues that address this feature it was marked as 'wontfix' with a recommendation to use the 'package.json:files' array instead.

The reasons for using this approach instead of `package.json:files` array are:
- Many known libraries in the Angular ecosystem does it ([angular/angular](https://github.com/angular/angular), [ReactiveX/rxjs](https://github.com/ReactiveX/rxjs), [cyclejs/cyclejs](https://github.com/cyclejs/cyclejs)) so they must have a reason.
- There is a lot of hoo-ha/complexity with the way node.js resolve modules  for the **web projects** since during the bundling process you **must** refer to the same instance of the library. Unless the bundler (typescript, webpack etc...) provide a hack/workaround/solution to force the library to use its' own node_modules, it will not work. Publishing from sub-directory works just because the `dist/node_modules` not exists.
- During development if the symlink is done against the root, when you import nested class which was not exported in the main index, you will need to refer to the `dist` as part of the path `import { something } from 'my-package/dist/something`. but once you publish from `dist` folder directly, you should somehow fix the pass by removing the `dist` during the tranpiling which is not a valid option.
- The libraries being used as dependencies during development should be assigned as peerDependencies at runtime because you want the application to provide them.

so to recap, we cannot just publish the package with a `dist` folder, we need to publish from `dist` directly

There are some caveats that I could think about with my suggested approach:
1. The dist folder must exists with a package.json inplace before the `lerna bootstrap` process symlink the folders. 
2. The build process should not delete the 'dist' folder, instead it should just clear its content otherwise the symlink of dependent libraries will be broken.

IMO those two caveats are manageable as:
[...]

PR #2109, opened 2019-05-27
Enable per-leaf pack directory via "packageDir" key in package.json

heavypennies:
## Description

Some packages produce a build directory and contents that are different than the root/source folder structure in the project.  This addition to lerna enables those packages to specify a different base directory for the Pack function.  This is akin to enabling e.g. `npm publish <dir>` on a per-leaf basis.  

To use this feature, add `"packageDir", "dist"` to package.json

and you'll get this:
`lerna notice packed @your-org/your-package packages/your-package/dist`

If you do not add `packageDir` to your package.json, you'll continue to get this:
`lerna notice packed @your-org/your-package packages/your-package`

## Motivation and Context

When we have lerna managed modules that transpile and otherwise construct a package structure outside the root of the source tree, it's ideal to be able to run e.g. `npm publish dist`.  The "files" directive in package.json DOES enable targeting only the dist contents, and for flat packages where index.js only includes relative files, this works.  However when publishing a library exposing multiple folders e.g. `import * from "@your-org/your-package/feature"` having to include a `dist` dereference e.g. `import * from "@your-org/your-package/dist/feature"` is not ideal.

I also noticed `--contents dist` would have a similar effect, but I was in need of a per-leaf solution to this issue.

This will fix https://github.com/lerna/lerna/issues/1282

## How Has This Been Tested?

First, I added a test to publish-command to ensure the pack-directory command would include the proper folder name if the package.json key was included.  

`npm run test -- publish-command` 

Then I tried it out in a local project: 
First I ran `lerna publish --force-publish` on my package without `packageDir` included in package.json.  Result:

`lerna notice packed @your-org/your-package packages/your-package`

Next I ran `lerna publish --force-publish` on my package with `"packageDir" : "dist"` included in package.json.  Result:

`lerna notice packed @your-org/your-package packages/your-package/dist`

```
$ node --version
v10.15.2
$ npm --version
6.4.1
$ lerna --version
3.14.1
```

## Types of changes

## Checklist:


evocateur:
I really appreciate the PR, my sincere apologies for the delay in review.

Inspired by your review, I implemented ea861d9c to fix this. Published in [v3.14.2](https://github.com/lerna/lerna/releases/tag/v3.14.2).

ithinkihaveacat:
Also fixes #901 (which is locked—trying to add a reference from that to this feature for people who come across the issue via search).

Issue #2125, opened 2019-06-06
[Feature] Ignore local dependents that define the dependency as a range for version bumps


julianburr:
## Expected Behavior

Alright, this might be controversial, so no hard feelings if you think this is a use case lerna should not be supporting. We're currently using lerna for our util libs and component library. This includes one core library pretty much all other libraries depend on. This core library might be changing quite frequently, so we would like to avoid having to bump all packages every time we do changes to the core lib.

Right now we define the dependency as a range, so e.g. in `some-lib` we do:

```json
{
  "devDependencies": {
    "core-lib": "2.x"
  }
}
```

The ideal behavior would be that lerna detects this when collecting the packages it needs to bump and ignores packages that use a range like above. Imo when using a range like this, the developer explicitly says that they want a specific version and don't care about minor/patch version.

This way we would still get the benefit of being able to use the symlinked local version (while the major version matches) without having to bump all other packages frequently just because of changes to `core-lib`.

Side note: we used to exploit `lerna publish --scope` for the desired behavior in the past 😅 

## Current Behavior

Lerna (more specifically `@lerna/collect-updates`) always includes all dependents that are currently depending on the local version of any of the changed packages.

## Possible Solution

I made a quick PR for it: https://github.com/lerna/lerna/pull/2126 ... again, if you think its not something lerna should support thats fine, just thought a PR might make things easier :blush:

danoc:
This is related to https://github.com/lerna/lerna/issues/718.

evocateur:
You should read #706 and #707. This isn't happening, sorry.

julianburr:
No worries, thanks for pointing to the related issues/prs :blush:

PR #2126, opened 2019-06-06
Ignore local dependents for version bumps, if they define the dependency as a range

julianburr:
## Description

We have one core library in our repo that uses lerna, which a lot of the other libraries depend on and which might change quite frequently. In order to avoid having to bump all packages every time we only change this core library, it would be ideal to be able to define the dependency as a range (see issue for details). This PR adds a check to the `collect-updates` util for the above.

## Motivation and Context

https://github.com/lerna/lerna/issues/2125

## How Has This Been Tested?

I `yarn link`ed the package within our repo. Also ran `yarn test utils/collect-updates`. Can add a new test case for the added behavior if needed.

## Types of changes

Not sure tbh, it changes the current behavior when you use ranges, but before ranges would basically automatically be overwritten by the bumped version, so they couldn't really be used anyway?!

## Checklist:

### DIFF
--- a/utils/collect-updates/lib/collect-dependents.js
+++ b/utils/collect-updates/lib/collect-dependents.js
@@ -25,6 +25,12 @@ function collectDependents(nodes) {
         return;
       }
 
+      const version = dependentNode.localDependencies.get(currentNode.name);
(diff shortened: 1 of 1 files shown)


evocateur:
#707 explains why this won't happen.

Issue #2113, opened 2019-05-29
[Feature] Allow custom publish subdirectory


aleclarson:
Related: #91 #901 #1282 #1817

I'm setting up `react-spring` to use Lerna, and I want `lerna publish` to use `npm publish ./dist` instead of `npm publish ./`, but Lerna's maintainers have declined such a feature request in the past.

For each package in `packages/` directory, there exists a stub `package.json` with only a little bit of metadata. The rest of the metadata is copied from the root `package.json` using a custom build script. I want the published contents to be excluded from source control (via `.gitignore`), so I don't want to mutate the stub `package.json` before publish. I *could* use the `prepublish` and `postpublish` scripts, but that can interfere with uncommitted changes and I need to keep the published contents around for local development.

One workaround is to put my package stubs in the `.packages/` directory, but I'd like to avoid that, because some platform-specific code *will* exist in these package stubs, and I want that code to be easily discoverable by contributors. It's better to put the package stubs in `packages/` since that's the "standard" for mono repos.

LMK if you have any questions. 👍

aleclarson:
> Especially nowadays with ES module exports and whatnot combined with tree-shaking module bundlers, there’s really no fundamental _necessity_ for false-basedir publishing. https://github.com/lerna/lerna/issues/901#issuecomment-350423290

@evocateu It's a useful feature for cross-platform libraries, I think. Tree-shaking can't solve the issue of platform differences in TypeScript definitions, for example.

aleclarson:
Ah, I just noticed the `--contents` flag for the `publish` command.

https://github.com/lerna/lerna/tree/master/commands/publish#--contents-dir

aleclarson:
@evocateur It looks like Lerna doesn't use the `package.json` in the `--contents` directory. Should that be fixed?

evocateur:
> Tree-shaking can't solve the issue of platform differences in TypeScript definitions

That sounds like a TypeScript problem, not a Lerna problem.

evocateur:
Also, I think I have a true fix for the underlying issue.

aleclarson:
Which issue exactly? I've had so many while setting up this monorepo (not all Lerna related). 😅 

Anyway, the monorepo is up and running now. Note: It's still using `patch-package` for Lerna.
See here: https://github.com/react-spring/react-spring/tree/v9

evocateur:
[...]

PR #2130, opened 2019-06-10
fix(publish): respect "pkg.publishConfig.directory"

aleclarson:
This is a quick fix for https://github.com/lerna/lerna/issues/2113#issuecomment-500380715

I assume you'll want to do it a different way :)

### DIFF
--- a/commands/publish/lib/get-unpublished-packages.js
+++ b/commands/publish/lib/get-unpublished-packages.js
@@ -12,7 +12,7 @@ function getUnpublishedPackages(packageGraph, opts) {
   let chain = Promise.resolve();
 
   // don't bother attempting to get the packument for private packages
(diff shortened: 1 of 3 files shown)


aleclarson:
> Failing tests is a non-starter.

Yeah sorry, I would fix/write tests if I had time.

> What specific issue does this solve?

Running `lerna publish` does *not* respect `pkg.publishConfig.directory` when it filters the list of packages by their "privateness". It should be checking the `package.json` in the publish directory, instead. The package root is marked private to avoid accidental `npm publish`. When the publish directory has its `package.json` generated, the `private: true` field is removed.

evocateur:
cc https://github.com/lerna/lerna/issues/2113#issuecomment-512963596

PR #2156, opened 2019-06-27
feat(publish): add `--registry-scope` option to override registry urls set with scope

lkuechler:
## Description

Allow overriding of scoped registry urls.

For example if you have a `.npmrc` file like this:
```
@scope:registry=https://npm.pkg.github.com/
```
And you execute a publish like this:
```
yarn lerna publish from-git --yes --registry https://custom.registry.com/
```
Lerna will currently always try to publish too `https://npm.pkg.github.com/`.

## Motivation and Context

It is currently not possible to override scoped registry urls from a `.npmrc` file in the CLI.

## How Has This Been Tested?

This was tested manually with linking this package in to the project where we needed the ability to override scoped registries.

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/index.js
+++ b/commands/publish/index.js
@@ -101,6 +101,20 @@ class PublishCommand extends Command {
       registry: this.options.registry,
     });
 
+    if (this.options["registry-scope"]) {
+      const registryScope = Array.isArray(this.options["registry-scope"])
+        ? this.options["registry-scope"]
+        : [this.options["registry-scope"]];
+
+      registryScope.forEach(scope => {
+        const registryScopeParts = scope.split(/(^@.*?):/);
+
+        if (registryScopeParts && registryScopeParts.length === 3) {
+          this.conf.set(`${registryScopeParts[1]}:registry`, registryScopeParts[2], "cli");
+        }
+      });
+    }
+
     this.conf.set("user-agent", userAgent, "cli");
 
     if (this.conf.get("registry") === "https://registry.yarnpkg.com") {
--- a/commands/publish/command.js
+++ b/commands/publish/command.js
@@ -51,6 +51,11 @@ exports.builder = yargs => {
       type: "string",
       requiresArg: true,
     },
+    "registry-scope": {
+      describe: "Use the specified registry for all npm client operations.",
+      type: "string",
+      requiresArg: true,
+    },
     "require-scripts": {
       describe: "Execute ./scripts/prepublish.js and ./scripts/postpublish.js, relative to package root.",
       type: "boolean",


evocateur:
Is [`publishConfig.registry`](https://docs.npmjs.com/files/package.json#publishconfig) in the leaf `package.json` insufficient for this use case? (I personally use this config successfully every day)

commands/publish/index.js
```diff
@@ -101,6 +101,20 @@ class PublishCommand extends Command {
       registry: this.options.registry,
     });
 
+    if (this.options["registry-scope"]) {
```
evocateur: Yargs automatically camelCases option names, so you can use `this.options.registryScope`.

commands/publish/index.js
```diff
@@ -101,6 +101,20 @@ class PublishCommand extends Command {
...
+      const registryScope = Array.isArray(this.options["registry-scope"])
```
evocateur: Where is this documented that it can accept more than one value? How does one discriminate?

commands/publish/index.js
```diff
@@ -101,6 +101,20 @@ class PublishCommand extends Command {
...
+        ? this.options["registry-scope"]
+        : [this.options["registry-scope"]];
+
+      registryScope.forEach(scope => {
+        const registryScopeParts = scope.split(/(^@.*?):/);
+
+        if (registryScopeParts && registryScopeParts.length === 3) {
+          this.conf.set(`${registryScopeParts[1]}:registry`, registryScopeParts[2], "cli");
```
evocateur: If your repo-root `.npmrc` already has these scoped registry configs, why is it necessary to re-assign them here?

matoilic:
I am facing the same issue in a project with the lates version 3.22.1 of Lerna. I have tried to set the registry in `lerna.json`, in the root `package.json`, in the `package.json` of each package in the monorepo, and I have also tried passing the registry explicitly using the `--registry` CLI option but nothing works. It always uses the registry set for the specific scope in `.npmrc`.

Since it's normal to have a separate publishing endpoint with repository management tools like Artifactory and Nexus, is there a possibility to get this fixed in the near future? Or is there a workaround in the current version of Lerna?

lkuechler:
I did not have time to continue working on this PR but I found a workaround.

Before publishing to the other registry you can rewrite the `.npmrc` like this:
```
echo "@scope:registry=https://registryurl/" >> .npmrc
git update-index --assume-unchanged .npmrc
```

SalimBensiali:
HI all. I got bitten by this one recently too. I have a `local-publish` npm script in a project's `package.json` where I test publishing of packages to a local [verdaccio](https://verdaccio.org/) registry first. The script looks like so:

`package.json`
```js
{
  ...
  "scripts": {
    "local-publish": "yarn lerna publish --no-push --no-git-tag-version --registry http://localhost:4873"
  },
  ...
}
```
and our `.npmrc` file had
```
@scope:registry=https://npm.pkg.github.com
```

And so despite the `--registry http://localhost:4873` being passed, `lerna publish` happily published the package to the Github registry instead of the local verdaccio registry.

@evocateur I would be happy to work on this PR if you are still considering contributions re the issue?

is the current approach using `--registry-scope` the way to go or do you think the `--registry` flag should also override the registry for `@scope`s defined inside `.npmrc`?

Cheers

Issue #1998, opened 2019-03-18
feat(version): Update 'version' field of package-lock.json


larsgw:
## Expected Behavior
I would expect `lerna version` to update the `version` field of `package-lock.json` files in packages that it updates the version of.

## Current Behavior
Currently it seems like `lerna version` only updates `package.json`.

## Solution
Update the `version` field in `package-lock.json` and `npm-shrinkwrap.json`. Edit: and `yarn.lock` I guess, and who knows what else, which also might explain why this isn't currently implemented.

## Steps to Reproduce (for bugs)
1. `lerna init`
2. `lerna create foo`
3. `cd packages/foo && npm install && cd ../..` (generate `package-lock.json`)
4. `lerna version`
5. `grep version packages/a/package{,-lock}.json`

<details><summary>lerna.json</summary><p>

```json
{
  "packages": [
    "packages/*"
  ],
  "version": "0.0.0"
}
```
</p></details>

## Context
I haven't encountered any actual problems with it, but it just looks weird (and it's wrong too).

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | `3.13.0` |
| `npm --version`  | `6.9.0` |
| `yarn --version` | N/A |
| `node --version` | `v11.11.0` |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| Debian | 8.2.0-14 |

hugomallet:
Related : #1415

larsgw:
Oh, I hadn't found that one, sorry.

revelt:
**Should we close this issue?** It seems #1415 had a working workaround...

To add 2p., I use Lerna to manage bunch of npm packages and lockfiles within npm packages are neither published to npm nor [desired](https://github.com/sindresorhus/ama/issues/479). I block lockfiles inside packages using `.npmrc` per-package, in each package's root file, single line inside `package-lock=false`. But I do keep the lockfile in the Lerna's root. With all this said, I don't care about lockfiles at all.

I guess it depends, what are those packages you use the Lerna for, if it's front-end pieces maybe lockfiles are desired.

hugomallet:
The role of lerna is to maintain easily consistency in the monorepo.
If it breaks the consistency between package.json and package-lock.json i think it is a problem. Even more serious if it is committed and pushed automatically. My opinion is that it should not be resolved as a "workaround" only.

(discussing about the usage or utility of package-lock.json is not the purpose i think).

revelt:
[...]

Issue #1415, opened 2018-05-08
Publish does not update package-lock


Robinfr:
## Expected Behavior
When doing a publish, the version number should not only be bumped in the `package.json` but also in the `package-lock.json` when available.

## Current Behavior
The package-lock version does not get updated. This causes the version to be updated the next time an install is run creating an extra (useless) commit.

## Possible Solution
`package-lock.json` version number should be made equal to the `package.json` version number when publishing. 

## Context
I'm doing a Lerna publish automatically. This pulls the latest version and publishes all of the packages to NPM. However, now I constantly have to create extra commits for the package-lock.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.2.0 |
| `npm --version`  | 5.8.0 |
| `node --version` | 8.11.1 |

evocateur:
The root `package-lock.json`, or the leaf package-locks (`packages/*/package-lock.json`)?

Robinfr:
The leaf package locks.

Op 8 mei 2018 6:27 p.m. schreef Daniel Stockman <notifications@github.com>:

The root package-lock.json, or the leaf package-locks (packages/*/package-lock.json)?

evocateur:
We're not calling `npm version`, just modifying the `package.json` directly and then calling `npm publish`. I try to avoid leaf package lockfiles, much like lerna itself does (requires lerna 3.x).

evocateur:
> Is there any reason not to?

For lerna 2.x, it's because I don't want to deal with that old codebase anymore.

For lerna 3.x (current `master`), it's a bandwidth issue. I'm only one person, and I have a day-job.

Robinfr:
I'm not asking whether there is a reason for you not to, I'm asking in general. One of the pros of open source is that you don't have to be solely responsible for developing the software. If you agree with the approach I could try taking a look at implementing it when I have some time to help you out.

Op 8 mei 2018 9:29 p.m. schreef Daniel Stockman <notifications@github.com>:

Is there any reason not to?

For lerna 2.x, it's because I don't want to deal with that old codebase anymore.

For lerna 3.x (current master), it's a bandwidth issue. I'm only one person, and I have a day-job.

evocateur:
Sure, a PR would be great.

tizmagik:
I'd be happy to assist with this as well if you could point me in the right direction 😄

Robinfr:
[...]

PR #2160, opened 2019-06-28
feat(package): .serialize() also writes package-lock version

mikemimik:
## Description

The `@core/package` class has a method called `.serialize()` which writes the in-memory package to disk. When running `lerna version` the `.version` property of the package is updated, written, and committed (depending on flags). However, the `package-lock.json` file _(which also has a top level `version` property)_ doesn't get updated, and is then out of sync with the `package.json` file.

This change updates the `serialize()` function to check if the package has a `package-lock.json` file in the directory, and if it does: read it in, update the version, then write it to disk.

## Motivation and Context

Any monorepo project that has `package-lock.json` files in it's package folders need to run another step to get the `package-lock.json` files to stay in sync with the `package.json` files. The script usually looks something like:

```bash
lerna exec -- npm install --no-audit --no-progress
git add packages/**/package-lock.json
git commit -m "chore(package-lock): bump package lock files [skip ci]"
git push
```
_(or some such variation of the above)_

Solves: https://github.com/lerna/lerna/issues/1998
Linked: https://github.com/lerna/lerna/issues/1415

## How Has This Been Tested?

Tests where added to the `core/package/__tests__/core-package.test.js` file.
- `describe("get .lockFileLocation")` was **added**
  - contains `it("should return the package-lock location")`
- `describe(".serialize()")` was **updated**
  - contains `it("writes changes to package.json to disk")` **updated**
  - contains `it("writes changes to package-lock to disk")` **added**

## Types of changes

## Checklist:

### DIFF
--- a/core/package/index.js
+++ b/core/package/index.js
@@ -4,6 +4,9 @@ const npa = require("npm-package-arg");
 const path = require("path");
 const loadJsonFile = require("load-json-file");
 const writePkg = require("write-pkg");
+const fs = require("fs");
+const writeJsonFile = require("write-json-file");
+const pWaterfall = require("p-waterfall");
 
 // symbol used to "hide" internal state
 const PKG = Symbol("pkg");
@@ -28,6 +31,12 @@ function shallowCopy(json) {
 
     return obj;
   }, {});
+}
+
+// determine if a package-lock.json file exists
+function hasPackageLock(location) {
+  const packageLock = path.resolve(location);
+  return fs.existsSync(packageLock);
 }
 
 class Package {
@@ -76,6 +85,9 @@ class Package {
       },
       manifestLocation: {
         value: path.join(location, "package.json"),
+      },
+      lockFileLocation: {
+        value: path.join(location, "package-lock.json"),
       },
       nodeModulesLocation: {
         value: path.join(location, "node_modules"),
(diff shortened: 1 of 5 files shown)


mikemimik:
It occurs to me that while this update will bump the version property in the `package-lock.json` file to be the same as the version property in the `package.json` file, the `package-lock.json` tree structure does not reflect the changes in package versions. This is a fundamental flaw in this feature work, and thusly should not be merged in yet. I will work to remedy this.

core/package/index.js
```diff
@@ -77,6 +86,9 @@ class Package {
       manifestLocation: {
         value: path.join(location, "package.json"),
       },
+      lockFileLocation: {
+        value: path.join(location, "package-lock.json"),
```
evocateur: this could be `npm-shrinkwrap.json`, or `yarn.lock`, or, or, or...

core/package/index.js
```diff
@@ -181,7 +193,27 @@ class Package {
...
+    const lockChanges = { version: this[PKG].version };
+
+    const tasks = [
+      () => loadJsonFile(this.lockFileLocation),
+      lockFileData => Promise.resolve(Object.assign({}, lockFileData, lockChanges)),
+      updatedLockFileData =>
+        writeJsonFile(this.lockFileLocation, updatedLockFileData, { detectIndent: true }),
+    ];
+
+    return hasPackageLock(this.lockFileLocation) ? pWaterfall(tasks) : Promise.resolve();
```
evocateur: I don't think we need `p-waterfall` just to chain promises. This check should be done way earlier, too.

package.json
```diff
@@ -7,7 +7,7 @@
     "ci": "npm test -- --ci --maxWorkers=2 && npm run integration -- --ci",
     "fix": "npm run lint -- --fix",
     "integration": "jest --config jest.integration.js --maxWorkers=2",
-    "lint": "eslint . --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint",
+    "lint": "eslint . --ignore-path .gitignore --ignore-pattern \"**/node_modules/\" --cache --cache-location ./node_modules/.cache/eslint",
```
evocateur: Unclear why this is necessary, as it is already ignored.

core/package/index.js
```diff
@@ -30,6 +33,12 @@ function shallowCopy(json) {
   }, {});
 }
 
+// determine if a package-lock.json file exists
```
evocateur: This isn't the appropriate place for lockfile maintenance. It's a function of the version command, not integral to the nature of a Package.

evocateur:
[v3.18.2](https://github.com/lerna/lerna/releases/tag/v3.18.2)

PR #2174, opened 2019-07-11
docs(add): Highlight single-argument limitation

raybooysen:
Addition to docs

## Description
lerna add only supports adding one package at a time, compared to yarn or NPM.  Makes sense to note this in the docs

## Motivation and Context
It's a common mistake from developers to run code like:
```
lerna add --scope=axy express webpack
```
and then see it fail.  The docs can now show that only 1 package can be added at a time

## How Has This Been Tested?
N/A

## Checklist:


evocateur:
Thanks!

Issue #2107, opened 2019-05-23
3.14.x: Prerelease publish failed to bump `version` in most `package.json` files.


abernix:
This behavior is a scripted behavior which we've used for some time now, but seems to have broken with 3.14.1 (presumably `3.14.0` too, but we'd been using `3.13.4` before).  Essentially, Lerna didn't bump the `version`s for the packages it said it was going to bump, committed the one that it did update (the very first one), and then went on to try and publish — ultimately failing because it attempts to publish over the previously published version.

## Expected Behavior

When running the following, I expect it to suggest bumping all the appropriate packages by a prerelease faction, prompt with proposed versions, and when confirmed, publish them to their appropriate `prerelease` versions (e.g. `.alpha.n`, in this case):

```
lerna publish --exact \
	--force-publish=apollo-server-core \
	--include-merged-tags \
	--preid alpha \
	--dist-tag next \
	prerelease 
```

As I expected, this yielded a prompt of suggested version bumps, as it typically does:

```
Changes:
 - apollo-datasource-rest: 0.4.0 => 0.4.1-alpha.0
 - apollo-engine-reporting: 1.2.0-alpha.3 => 1.2.0-alpha.4
 - apollo-server-azure-functions: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-cloud-functions: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-cloudflare: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-core: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-express: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-fastify: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-hapi: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-integration-testsuite: 2.6.0-alpha.3 => 2.6.0-alpha.4 (private)
 - apollo-server-koa: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-lambda: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-micro: 2.6.0-alpha.3 => 2.6.0-alpha.4
 - apollo-server-plugin-base: 0.5.0-alpha.3 => 0.5.0-alpha.4
[...]

Issue #2165, opened 2019-07-05
Lerna run isn't running on packages after upgrading to v.3.14.x and above


alcferreira:
## Expected Behavior

It seems that `lerna run` isn't running on packages from a monorepo.
Until v3.13.4 it used to work fine

## Current Behavior

It is not working after v3.14.x

## Steps to Reproduce (for bugs)

1. Clone this [repo](https://github.com/CraveFood/farmblocks)
2. Install dependencies `yarn`
3. Run [`yarn build`](https://github.com/CraveFood/farmblocks/blob/master/package.json#L11) > which will create a [lib directory](https://github.com/CraveFood/farmblocks/blob/master/packages/alert/package.json#L21) on every package 
4. Check out if the lib directory has been created for the alert package: `/farmblocks/packages/alert/lib`
5. Remove this directory
6. Upgrade lerna to the latest version (or v3.14.x) `yarn add --dev lerna@latest -W` 
7. Run `yarn build` again. The directory `/farmblocks/packages/alert/lib` should've been created but it was not.

<details><summary>lerna.json</summary><p>

```json
<!-- Please paste your `lerna.json` here -->
{
  "lerna": "3.13.1",
  "packages": ["packages/*"],
  "version": "independent",
  "command": {
    "publish": {
      "conventionalCommits": true,
      "message": "chore(release): [ci skip] publish"
    }
  },
  "npmClient": "yarn",
  "useWorkspaces": true
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.15.0 |
| `yarn --version` | 1.16.0 |
| `node --version` | 12.1.0 |

| OS | Version |
| --- | --- |
| macOS Mojave | 10.14 |

danez:
Probably the same issue as #2107

evocateur:
@alcferreira [v3.16.0](https://github.com/lerna/lerna/releases/tag/v3.16.0) contains the fix in #2107, can you try it and verify if this is fixed? Thanks!

alcferreira:
@evocateur and @danez fix confirmed, thank you all!

PR #2185, opened 2019-07-16
fix(package-graph): Flatten cycles to avoid skipping packages

nicolo-ribaudo:
## Description

To correctly handle cycles, I wrote an algorithm which flattens cycles into single nodes. Then every cycled, when it becomes a "leaf node", can be normally executed.

Example:
```
          +---+
   /----> | B |
  |       +---+
  |
+---+     +---+ <----\
| A | --> | C |       |
+---+     +---+ --\   |
                   |  |
          +---+ <-/   |
          | D |       |
          +---+ -----/
            |
            v
          +---+
          | E |
          +---+
```
Becomes
```
          +---+
   /----> | B |
  |       +---+
  |
+---+     +-------+
| A | --> | C + D |
+---+     +-------+
              |
              v
            +---+
            | E |
            +---+
```

Now that there aren't anymore nodes, `E` and `B` can be executed since they are leaf nodes. The graph becomes:

```
+---+     +-------+
| A | --> | C + D |
+---+     +-------+
```

`C + D` now is a leaf node, so `C` and `D` can be executed. The cycle is then removed from the graph.

At this point, `A` becomes a leaf node and can be executed.

Note that two cycles might have an intersection (or be nested): in that case they are handled as a single cycle.

## Context

Fixes https://github.com/lerna/lerna/issues/2107
Need to check: https://github.com/lerna/lerna/issues/2165 (@alcferreira)

## How Has This Been Tested?

It fixed our problem in the Babel publishing process.

## Types of changes

## Checklist:


utils/pack-directory/__tests__/pack-directory.test.js
```diff
@@ -175,41 +175,41 @@ Object {
...
+            "mode": "MODE",
+            "path": "index.js",
+            "size": 25,
+          },
+        ],
+        "id": "package-3@1.0.0",
+        "integrity": "INTEGRITY",
+        "name": "package-3",
+        "shasum": "SHASUM",
+        "size": 449,
```
nicolo-ribaudo: I don't understand this change.

utils/listable/__tests__/listable-format.test.js
```diff
@@ -242,8 +242,8 @@ pkg-1
       expect(loggingOutput("warn")).toContainEqual(expect.stringContaining("pkg-2 -> pkg-3 -> pkg-2"));
       expect(text).toMatchInlineSnapshot(`
 pkg-2
-pkg-1
 pkg-3 (PRIVATE)
+pkg-1
```
nicolo-ribaudo: This should be ok. 2-3 are a cycle, and 1 depends on 2.

commands/run/__tests__/run-command.test.js
```diff
@@ -217,23 +217,45 @@ describe("RunCommand", () => {
...
       expect(output.logged().split("\n")).toEqual([
         "package-dag-1",
         "package-standalone",
         "package-dag-2a",
         "package-dag-2b",
-        "package-dag-3",
         "package-cycle-1",
         "package-cycle-2",
         "package-cycle-extraneous",
+        "package-dag-3",
```
nicolo-ribaudo: This should be ok: "dag" and "cycle" are two separate graphs, so there isn't a specified order between them.

commands/run/__tests__/run-command.test.js
```diff
@@ -217,23 +217,45 @@ describe("RunCommand", () => {
       const [logMessage] = loggingOutput("warn");
       expect(logMessage).toMatch("Dependency cycles detected, you should fix these!");
       expect(logMessage).toMatch("package-cycle-1 -> package-cycle-2 -> package-cycle-1");
-      expect(logMessage).toMatch("package-cycle-2 -> package-cycle-1 -> package-cycle-2");
-      expect(logMessage).toMatch(
-        "package-cycle-extraneous -> package-cycle-1 -> package-cycle-2 -> package-cycle-1"
```
nicolo-ribaudo: Here it only removed duplicated warnings about the same cycle.

core/package-graph/index.js
```diff
@@ -190,7 +299,10 @@ class PackageGraph extends Map {
   }
 
   /**
-   * Return a tuple of cycle paths and nodes, which have been removed from the graph.
+   * Return a tuple of cycle paths and nodes.
+   *
+   * @deprecated Use collapseCycles instead.
```
nicolo-ribaudo: I couldn't change the signature of this method for backward compatibility. Instead of adding a new boolean switch, I decided to create a new method. The old and new `visits` are similar but incompatible.

core/package-graph/index.js
```diff
@@ -238,19 +350,71 @@ class PackageGraph extends Map {
...
+
+      if (walkStack.indexOf(topLevelDependent) === -1) {
+        // eslint-disable-next-line no-use-before-define
+        visitWithStack(baseNode, topLevelDependent);
+      }
+    };
+
+    const visitWithStack = (baseNode, currentNode = baseNode) => {
+      walkStack.push(currentNode);
+      currentNode.localDependents.forEach(visits.bind(null, baseNode));
```
nicolo-ribaudo: The old implementation used an arrow function (`dependentNode => visits(baseNode, dependentNode)`), but `.bind` makes it way easier to understand the stack while debugging since the intermediate function is elided.

core/package-graph/index.js
```diff
@@ -238,19 +350,71 @@ class PackageGraph extends Map {
...
+      }
+
+      if (walkStack.indexOf(topLevelDependent) === -1) {
+        // eslint-disable-next-line no-use-before-define
+        visitWithStack(baseNode, topLevelDependent);
+      }
+    };
+
+    const visitWithStack = (baseNode, currentNode = baseNode) => {
+      walkStack.push(currentNode);
```
nicolo-ribaudo: In the old implementation, this array was cloned each time before adding a new element.
I used push/pop in an intermediate implementation which was incompatible with the `walk.concat(newNode)` method, but at the end I decided to leave it here since it is faster.

core/package-graph/index.js
```diff
@@ -238,19 +350,71 @@ class PackageGraph extends Map {
...
+        });
+
+        cycles.add(cycle);
+        cyclePaths.push(cycle.toString());
+
+        return;
+      }
+
+      if (walkStack.indexOf(topLevelDependent) === -1) {
+        // eslint-disable-next-line no-use-before-define
```
nicolo-ribaudo: :sweat_smile:

Those are two mutually recursive functions, please don't hate me

nicolo-ribaudo:
Oh this PR makes lerna warn about 6 cycles in Babel (instead of about 30), making it way more actionable to resolve :smile:

evocateur:
I have no idea why the exec+run tests are flaking (inconsistently!) in CI. I'm fixing that now, then I'll merge.

__fixtures__/cycle-parent/graph.txt
```diff
@@ -0,0 +1,13 @@
+https://github.com/lerna/lerna/issues/2107#issuecomment-511987691
+
+          +---+
+   /----> | B |
+  |       +---+
+  |
++---+     +---+ <----\
+| A | --> | C |       |
++---+     +---+ <-\   |
```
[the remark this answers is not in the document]
evocateur: Good catch, I've fixed this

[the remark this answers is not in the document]
nicolo-ribaudo: Thanks!

nicolo-ribaudo:
@evocateur Thanks for your new commits ☺️

evocateur:
ok ok i'm done tweaking i swear

evocateur:
@nicolo-ribaudo @bweggersen Huge thanks for all your help!

Issue #2232, opened 2019-08-16
Lerna import is too opinionated


anand-sundaram-zocdoc:
## Expected Behavior
Suppose my lerna.json has ```"packages": ["packages/**"],```.
If I run:
```lerna import ../original_foobar --dest="packages/components/foobar"```

Ideally I would like this to import all the code with commit history from the original_foobar directory into a newly created directory called components/foobar.

## Current Behavior
Instead of the desired behavior I described, lerna import will:
1) complain that "packages/components/foobar" is not a packages directory, even though it is within the packages directory
2) not support moving the code except into a destination directory that is named the same as the source directory

## Workaround
To make it work how I wanted, I had to:
1) rename original_foobar to foobar on my machine first before running lerna import
2) Change lerna.json to have "packages/components/foobar/**" as an additional entry in the packages list
3) Run lerna import lerna import ../foobar --dest="packages/components"
4) Change the lerna.json back because that's just silly

## Possible Solution
We could:
1) Change the implementation of the package directory validation check in https://github.com/lerna/lerna/blob/bb048cb306b5cfcb039aa98f667cf3751cf0ad20/commands/import/index.js#L65 to allow subdirectories of a package directory to be chosen as targets
2) change https://github.com/lerna/lerna/blob/bb048cb306b5cfcb039aa98f667cf3751cf0ad20/commands/import/index.js#L71 to make the target dir name customizable instead of just forcing it to be the name of the source directory

## Steps to Reproduce (for bugs)
N/A; see expected behavior above.

<details><summary>lerna.json</summary><p>

```json
{
  "packages": [
    "packages/**"
  ],
  "version": "independent",
  "command": {
    "publish": {
        "allowBranch": "master"
    },
    "version": {
        "allowBranch": "master",
        "conventionalCommits": true,
        "message": "chore(publish): update versions"
    }
  }
}

```
</p></details>

## Context
I want to move code into a monorepo, but instead of moving it from the source repo into the monorepo as one package, I want to move different subfolders of a single source repo into the monorepo as separate subpackages. Lerna import does not handle this use case so I have been hacking together a combination of git subdirectory filters on the source repo followed by lerna imports. Trying to make Lerna support handling all the work for my desired code migration strategy cleanly would be way more ambitious; all I'm looking for here is buy-in to make the import command a little easier to use.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.16.4 |
| `npm --version`  | 6.9.0 |
[...]

PR #2234, opened 2019-08-17
feat(import): allow more flexibility

anand-sundaram-zocdoc:
## Description
This changes the lerna import command to:
1. Support any directory that is contained in one of the lerna packages directories, not just those packages directories themselves, as destination base paths
2. Support naming the new destination directory that will be created by running lerna import inside the base path mentioned above whatever you want, instead of forcing it to be the name of the directory from which the files are being imported.

## Motivation and Context
See https://github.com/lerna/lerna/issues/2232 for a description of the really convoluted workflow that is currently required without these changes for me to use lerna import as desired to achieve the outcomes I want.

## How Has This Been Tested?
I added unit tests for the two new behaviors I wish to support.

## Types of changes

## Checklist:
_NOTE: this is my first PR to lerna so if I checked something here incorrectly please let me know how I can improve_

### DIFF
--- a/commands/import/README.md
+++ b/commands/import/README.md
@@ -47,10 +47,17 @@ $ lerna import ~/Product --flatten
 
 ### `--dest`
 
-When importing repositories, you can specify the destination directory by the directory listed in lerna.json.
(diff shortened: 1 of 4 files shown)


evocateur:
As I wrote in the issue, this is far too complicated for a silly command that shouldn't even exist in lerna core.

Issue #2269, opened 2019-09-16
Alternative option for .json format


acao:
Hey friends! Love working with lerna.

So. I'm curious if theres any plans to add more entries here:
https://github.com/lerna/lerna/blob/bb048cb306b5cfcb039aa98f667cf3751cf0ad20/core/project/index.js#L22

Our innersourcing pattern relies on the hope for a commonjs format for all configs, however I know, being a tooling maintainer as well, the limitations of non-json configuration formats.

Possibly there could be a global argument for specifying a full file path including an extension that cosmiconfig supports transparently? I tried just monkey-patching it myself locally and a `lerna.config.js` entry worked here flawlessly.

acao:
If y'all are open to this, I'd be glad to open a PR

evocateur:
`lerna.json` already supports a [yargs-like "extends" keyword](https://github.com/lerna/lerna/commit/0b28ef5fe8e4fc376870848acc8f6ccc0905a7b2), which supports dependency packages as well as local paths. Regrettably, not documented.

Is that not sufficient?

acao:
@evocateur works for our needs! I can agree that `<workspace-root>/lerna.json` is an important convention to keep (i've worked on a good many CI pipelines that depend on it), rewriting version would be impossible as you mention, and `extends` allows for everything we need for this purpose and more.

PR #2270, opened 2019-09-16
feat: support common js config files

acao:
## Description
Support a commonjs format config file as `lerna.config.js`, as described in #2269 

## Motivation and Context
- more flexible shared tooling
- provide more flexible options depending on git context (i.e. conventional pre-release for specific branches, etc)
- in our particular case, we want to share `lerna version`, `lerna publish`, etc configurations (conventional commits is big factor) across a few dozen monorepos in a clean and overrideable fashion
- the `lerna.config.js` filename takes inspiration from `jest.config.js` and `babel.config.js`, but I'm open to other conventions, and cosmic config was a great choice :)

## How Has This Been Tested?
- I've tested it locally using yarn link and a complicated lerna config file
- I've added a simple unit test that loads and asserts a fixture `lerna.config.js`

## Types of changes
  - downstream tools such as dependabot, renovatebot, etc may expect a lerna.json file still yet?

## Checklist:

### DIFF
--- a/core/project/README.md
+++ b/core/project/README.md
@@ -4,7 +4,7 @@
 
 ## Configuration Resolution
 
-Lerna's file-based configuration is located in `lerna.json` or the `lerna` property of `package.json`.
+Lerna's file-based configuration is located in `lerna.json`, a commonjs module exported from `lerna.config.js` or the `lerna` property of `package.json`.
 Wherever this configuration is found is considered the "root" of the lerna-managed multi-package repository.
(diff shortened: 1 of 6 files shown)


acao:
@evocateur good point, and as you point out on the original issue #2269, `extends` is an option that suffices for our needs, and I think i will close this PR in favor of a PR that documents the `extends` feature itself. 

this approach I propose in this PR raises too many issues, breaks fixed versioning and inevitably breaks with the very useful convention of `lerna.json` always being the file to look for in the monorepo base. For our needs `extends` will get the job done.

PR #2274, opened 2019-09-17
feat(link): Use `publishConfig.directory` as symlink source if it exists to allow linking sub-directories

### LINKED ISSUE #2209: Publish and link package subdirectory (like lib/dist/build)
Re-raising this issue: #1282 (also #901 and #91)

The main use case for publishing the build artifact staging directory (usually named `dist`/`lib`/`build`) is to enable deep imports without including the build directory in the path. The `main`/`module`/`browser` package properties only enable *bare* imports like `require('package')` instead of deep imports like `require('package/directory/module')`.

Deep imports are a common pattern and have the advantages of being more efficient and removing the need for workarounds like tree shaking. Eventually this issue should be solved by [import maps](https://github.com/WICG/import-maps), but publishing the build directory is the a reasonable option until import maps are available.

I'd be willing to contribute a PR if there were some pointers about the required changes for this feature.

kamranayub:
## Description

This is a fork of https://github.com/lerna/lerna/pull/2081 but uses the suggestions from https://github.com/lerna/lerna/issues/2209#issuecomment-517788971

- Use `pkg.contents` to resolve symlink source path, allowing usage of `publishConfig.directory` to link to a sub-directory
- Add support for `--contents` to `bootstrap` and `link` with docs

## TODO

- [x] Update documentation to mention semantics of `directory` with regards to `link` command

## Open Questions

- [x] ~~❓I believe this may be a breaking change, as anyone using `publishConfig.directory` and who've run `lerna link` before will experience a change in behavior. I could enhance this PR to include a flag for `link`, suggestions welcome.~~
- [x] ~~❓  I'm unclear what the comment "And probably need to make the `pkg.bin` field resolvable from contents, as well (it's currently hard-code)" refers to. In the Jest snapshots, it _looks_ correct for bin files (I think) but wondering if @evocateur had some more context for that comment that I should add to the PR.~~

## Motivation and Context

Closes #2209

There was an existing PR already https://github.com/lerna/lerna/pull/2081 but it needed tests and I believe the idea was to use `pkg.contents` instead of introducing a new config property.

I explained my motivation/need in my comment (https://github.com/lerna/lerna/pull/2081#issuecomment-532233745):

---

The `publishConfig.directory` works well for publishing. In our monorepo we have two packages, `core` and `baked`. `baked` uses components from `core`. Since we use Rollup + Babel + TypeScript (these are libraries), we have a `dist` directory that we publish as the package. While we _can_ publish new versions of the `core` components, for local development we're looking at using `lerna link` to make that easier to have a nicer workflow.

In a file within `baked` you may have:

```ts
import { MyComponent } from '@org/core-components'
```

This works for a published package but since `lerna link` symlinks the root directory and _not_ the `dist` directory, for local development you must change the path:

```ts
import { MyComponent } from '@org/core-components/dist'
```

## How Has This Been Tested?

- Added Jest tests with snapshots that test the generated src/dest paths
- Linked `lerna` with `npm link` and tested within our own monorepo for our use case
  - Specifically I tested depending from package `baked` -> `core` and successfully consumed TypeScript-based components from the symlinked `core` package.

## Types of changes

## Checklist:


commands/publish/README.md
```diff
@@ -303,7 +303,7 @@ You can customize the dist-tag on a per-package basis by setting [`tag`](https:/
 
 ### `publishConfig.directory`
 
-This _non-standard_ field allows you to customize the published subdirectory just like [`--contents`](#--contents-dir), but on a per-package basis. All other caveats of `--contents` still apply.
+This _non-standard_ field allows you to customize the published subdirectory just like [`--contents`](#--contents-dir), but on a per-package basis. All other caveats of `--contents` still apply. When symlinking using `lerna link` or `lerna bootstrap`, this will be the _source_ directory of the symlink, just like how the published package would be consumed.
```
evocateur: Details for `lerna link` and `lerna bootstrap` should be documented in their respective READMEs, not `lerna publish`'s.

commands/link/README.md
```diff
@@ -10,7 +10,7 @@ Install [lerna](https://www.npmjs.com/package/lerna) for access to the `lerna` C
 $ lerna link
 ```
 
-Symlink together all Lerna `packages` that are dependencies of each other in the current Lerna repo.
+Symlink together all Lerna `packages` that are dependencies of each other in the current Lerna repo. When using `publishConfig.directory`, that will be the source directory of the link (e.g. `package-1/dist => node_modules/package-1`).
```
evocateur: You should document this separately, as it is done in the [publish README](https://github.com/lerna/lerna/tree/master/commands/publish#publishconfigdirectory). It should also be documented in the bootstrap README, if it is intended to be used there as well (I assume it is?). The `publishConfig.directory` namespace is lamentable for these cases, but it is what it is.

You'll also need to copy over the `--contents` [config](https://github.com/lerna/lerna/blob/afb04146796acf66611baefd06da4170ca1162b6/commands/publish/command.js#L27-L32) from the publish command into bootstrap and link (with appropriate modifications to the `description` field, obviously).

evocateur:
I don't see this as a breaking change, really. If one is already using `--contents` or `publishConfig.directory`, then this is somewhat of a "return" to expected behavior.

commands/link/__tests__/__fixtures__/with-contents/packages/package-3/cli1.js
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env node
+console.log("Hello, world!");
```
evocateur: To make this more interesting, you might want to move this (and other bin targets) into a source location that makes sense for packages that would use `--contents`/`publishConfig.directory`. Then the snapshots would be more useful.

commands/link/__tests__/__fixtures__/with-contents/packages/package-1/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@test/package-1",
+  "version": "1.0.0",
+  "publishConfig": {
+    "directory": "dist"
```
evocateur: Also, I think you'll also want to run a test on the _existing_ fixtures with `--contents build` (or something), to distinguish it from the `publishConfig.directory` behavior.

kamranayub: Thanks, this exposed some missing functionality with `link` and `bootstrap` so they respected `--contents`

commands/bootstrap/__tests__/__fixtures__/with-contents/lerna.json
```diff
@@ -0,0 +1,4 @@
+{
+  "ci": false,
```
evocateur: I don't see any bootstrap tests using this fixture? In any case, I think you can skip it (and remove this fixture tree), because aside from the `--contents` flag (thank you!), I don't think bootstarp should care about what is an implementation detail of the symlinking lib.

commands/bootstrap/__tests__/bootstrap-command.test.js
```diff
@@ -345,6 +345,14 @@ describe("BootstrapCommand", () => {
       });
     });
 
+    it("should respect --contents argument during linking step", async () => {
```
kamranayub: @evocateur You're right, I thought I was using the _with-contents_ fixture for this test but I wasn't! This was meant to test the new `--contents` argument I added.

Should be fixed now.

commands/bootstrap/__tests__/__fixtures__/with-contents/lerna.json
kamranayub: Left a comment, this was a mistake, for some reason I thought I needed the fixture for the test 🤷‍♂ Thanks!

evocateur:
Published in [v3.17.0](https://github.com/lerna/lerna/releases/tag/v3.17.0)

PR #2280, opened 2019-09-21
feat(filter-options): Allow command to continue if no packages are matched

debugmaster:
## Description
- Added new param `continueIfNoMatch` in `filterPackages()` to prevent an exception from being thrown if all filters did not match any package.
- Added `--continue-if-no-match` as a Filter option to enable/disable new param.
- Added documentation to the method and param. Because this option is hidden by design (because it is not really a filter), I'd like to hear what the reviewers may suggest.

## Motivation and Context
My motivation is being able to execute the same command in one or more Lerna-managed mono repos, which may or may not contain the same scopes. As long as every team follows the definition of scopes, the same CI can be applied to every team.

For example:
```
lerna-repo-1/
  package.json
  team-a/
    lib-1/
      package.json
    service-1/
      package.json
  team-b/
    service-1/
      package.json
    service-2/
      package.json
lerna-repo-2/
  team-c/
    component-1/
      package.json
    lib-1/
      package.json
    service-1/
      package.json
```

The CI tries to create a scope like `--scope <team>-<type-of-artifact>-*`. With this new option, we don't need to conditionally select which artifacts to build. For example, building libs of Team B would fail in the current behavior.

Imagine also some people may have suggested adding empty packages only to bypass CI.

## How Has This Been Tested?
I have added unit tests to cover the new param. They contain the same scenarios by the tests that throw the exception (plus the new param);

## Types of changes

## Checklist:

### DIFF
--- a/core/filter-options/__tests__/get-filtered-packages.test.js
+++ b/core/filter-options/__tests__/get-filtered-packages.test.js
@@ -90,6 +90,20 @@ test.each`
   expect.hasAssertions();
 });
 
+test.each`
+  argv
+  ${["--scope", "not-a-package", "--continue-if-no-match"]}
+  ${["--ignore", "package-*", "--continue-if-no-match"]}
+  ${["--scope", "package-@(1|2)", "--ignore", "package-{1,2}", "--continue-if-no-match"]}
+`("no errors and no packages $argv", async ({ argv }) => {
+  const packageGraph = await buildGraph(cwd);
+  const execOpts = { cwd };
+  const options = parseOptions(...argv);
+
+  const result = await getFilteredPackages(packageGraph, execOpts, options);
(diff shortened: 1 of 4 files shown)


debugmaster:
@evocateur how often Lerna has a round of reviews?

JabbyPanda:
`continue-if-no-match` flag name is such not obvious name to memorize, `--allow-empty-scope` would be better choice, IMHO

debugmaster:
I agree that it is not a trivial name, but does it require to be memorized? This is not meant to be necessary all the time. That's why I decided to go with a more explanatory flag. 

Regarding your suggestion, `--allow-empty-scope` seems ambiguous to me because of the parameter `--scope`. Of course, this is a matter of stating to the user what is expected when used.

PR #2287, opened 2019-10-01
added ex

uxitten:
Improved README


evocateur:
The existing documentation is more than sufficient, in this case.

Issue #1683, opened 2018-09-17
--scope is not working


acostaf:
Reopen please --scope is not working on "lerna": "^3.4.0"

npm run this "build:dep": "lerna run --scope common build"

I get lerna ERR! EFILTER No packages remain after filtering [ 'common' ]

my Lerna file   
"packages": [
    "common",
...
  ],

phun-ky:
You've to follow this syntax, which sadly is poorly documented:

    lerna --scope @yourprojectname/commin exec  -- npm run build

evocateur:
It's notable that `--scope` matches package _names_, not directories. I highly doubt you are the owner of the "common" package on npm.

evocateur:
(Scope applies perfectly well to `lerna run`, but it won't execute in packages that lack the target script)

acostaf:
@evocateur common it is my own project, the script name exist but you are right about package name, I was calling the folder, thanks

PR #2301, opened 2019-10-15
CHG: Update guides.md to correct syntax

teripaquitinho:
Add @my-project part to the args when calling my-package

## Motivation and Context
Related to this... https://github.com/lerna/lerna/issues/1683 

## How Has This Been Tested?
Locally I could not run Jest debugger and I had to modify the launch file accordingly

### DIFF
--- a/doc/guides.md
+++ b/doc/guides.md
@@ -15,7 +15,7 @@ It is possible to debug [Jest](https://facebook.github.io/jest/) tests in a Lern
     "runtimeArgs": [
         "exec",
         "--scope",
-        "my-package",
+        "@my-project/my-package",
         "--",
         "node"
     ],


doc/guides.md
```diff
@@ -15,7 +15,7 @@ It is possible to debug [Jest](https://facebook.github.io/jest/) tests in a Lern
     "runtimeArgs": [
         "exec",
         "--scope",
-        "my-package",
+        "@my-project/my-package",
```
evocateur: In this example, the package is named `my-package`, not `@my-project/my-package`. They need to match (the package name, not the directory name).

teripaquitinho: true... i understand this... however having  a project fooproj with packages pkg1, pkg2, pkg3 

I had to run this with

```
runtimeArgs": [
        "exec",
        "--scope",
        "@fooproj/pkg1",
        "--",
        "node"
    ],
    "args": [
        "${workspaceRoot}/node_modules/jest/bin/jest.js",
        "--runInBand",
        "--no-cache",
        "packages/pkg1"
    ]
```

Please note `pkg1` is fine on the args but i had to specify the proj in the `runtimeArgs` ... i just want to reflect this in the guide. I'm not talking about directories... how do you suggest I can do that?

evocateur: This example is making an awful lot of assumptions about package structure and directory structure, true. The `--scope` argument requires the package name, and the Jest positional argument requires the directory path. There is no requirement with Lerna that the directory name matches the package name, so this is kinda... expected?

The current example is using a very common pattern (directory and package name match), but it is out of scope for this example to document [`--scope`](https://github.com/lerna/lerna/blob/master/core/filter-options/README.md#--scope-glob).

Issue #2267, opened 2019-09-16
dist folder not getting ignored inside scoped package folder


kushalmahajan:
I have a library package having `.gitignore` inside specifying dist. Despite that it doesn't picks up. Strangely it picks up node_modules for ignoring.

I'm using lerna with yarn workspaces. Never had this issue before. Either it's yarn workspaces or lerna. Help me out please!

kushalmahajan:
Temporary issue.

Issue #2141, opened 2019-06-18
Lerna throws error when publishing dynamically generated packages which are in .gitignore


darthtrevino:
## Expected Behavior
In react-dnd, I have a script that generates npm packages for CommonJS build variants of the core libraries (https://github.com/react-dnd/react-dnd/blob/master/packages/alternative_builds/cjs/create_cjs_modules.js) - the dynamically generated packages are in the .gitignore.

When I run `lerna run build`, this will execute the `build` script in all of the packages as expected, including the dynamically generated packages.

When I run `lerna publish`, I expect the same kind of behavior - the dynamic packages are published along with the regular ones. 

## Current Behavior
The `lerna run build` step works as expected .
The `lerna publish` step skips over packages in the .gitignore.

## Possible Solution
Possibly check for a flag in the dynamically generated modules package.json. Maybe something like `publishConfig.publishIfGitIgnored: true`

</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.15.0 |
| `npm --version`  | 6.9.0 |
| `yarn --version` | 1.16.0 |
| `node --version` | 12.4.0 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Mojave | 10.14.5 |

darthtrevino:
Derp, this one was my bad

Issue #1096, opened 2017-11-01
Using workspaces for install/bootstrap but excluding certain packages from `lerna publish`


skyl:
I have a similar situation as what is described in #216. I have lerna 2.4.0 and I'm using yarn workspaces (yarn 1.2.0). In package.json, I have:

```
  "workspaces": [
    "build/*",
    "packages/*",
    "projects/*"
  ],
```

For `yarn install`, I want all of the package.json's for each of these workspaces taken into consideration and dependencies hoisted/bootstrapped etc (I'm using yarn install instead of lerna bootstrap). This is nice, works great.

I don't have the luxury of working in a total monorepo though. So, I clone other repos in eg `projects` and then add the new dir to `projects/.gitignore`. Yarn install finds these external projects and installs their dependencies. I have `"private": true` in these projects' package.jsons.

## Expected Behavior

I would expect that `lerna publish` would skip over these external, .gitignore'd, private repos and not try to publish them at all.

## Current Behavior

When I run `lerna publish`, these private external repos are taken into account and because they are git ignored, I get weird behavior that I'm not expecting.

```
lerna publish
lerna info version 2.4.0
lerna info current version 0.3.2
lerna info Checking for updated packages...
lerna info Comparing with v0.3.2.
? Select a new version (currently 0.3.2) Patch (0.3.3)

Changes:
 - i-want-to-publish-this: 0.3.2 => 0.3.3
 - dont-publish-this: 4.2.0 => 0.3.3 (private)

? Are you sure you want to publish the above changes? Yes
lerna ERR! execute caught error
lerna ERR! Error: The following paths are ignored by one of your .gitignore files:
lerna ERR! projects/dont-publish-this/package.json
lerna ERR! Use -f if you really want to add them.
lerna ERR! 
lerna ERR!     at Function.module.exports.sync (/.../node_modules/execa/index.js:303:26)
lerna ERR!     at Function.execSync (/.../node_modules/lerna/lib/ChildProcessUtilities.js:59:30)
lerna ERR!     at Function.addFile (/.../node_modules/lerna/lib/GitUtilities.js:68:39)
lerna ERR!     at /.../node_modules/lerna/lib/commands/PublishCommand.js:653:41
lerna ERR!     at Array.forEach (native)
lerna ERR!     at PublishCommand.updateUpdatedPackages (/.../node_modules/lerna/lib/commands/PublishCommand.js:652:22)
[...]

PR #2303, opened 2019-10-16
Update git-add.js

### LINKED ISSUE #2151: Cannot Publish Dynamically Generated Packages
## Expected Behavior

In this repository: https://github.com/react-dnd/react-dnd, there are a handful of packages that are dynamically created when running `yarn build`. These are the CommonJS variants of the core packages.

When I run a release, I expect that my packages are all published and that the repository is tagged appropriately.

## Current Behavior
When the release command is run, the semver questions are asked and the versions are bumped, but action fails when creating github tags. Logs are pasted below.

## Possible Solution
If package.json files are .gitignored, then don't add them in the git add step.

</p></details>

```txt
107 error Error: Command failed: git add -- packages/core/dnd-core/package.json packages/core/react-dnd/package.json packages/testing/test-backend/package.json packages/alternative_builds/cjs/dnd-core/package.json packages/core/html5-backend/package.json packages/alternative_builds/cjs/react-dnd/package.json packages/alternative_builds/cjs/test-backend/package.json packages/testing/test-utils/package.json packages/alternative_builds/cjs/test-utils/package.json packages/documentation/examples-decorators/package.json packages/alternative_builds/cjs/html5-backend/package.json packages/alternative_builds/umd/package.json packages/documentation/examples-hooks/package.json packages/documentation/docsite/package.json lerna.json
107 error The following paths are ignored by one of your .gitignore files:
107 error packages/alternative_builds/cjs
107 error Use -f if you really want to add them.
107 error
107 error     at makeError (/Users/christrevino/Workspace/oss/react-dnd/node_modules/execa/index.js:174:9)
107 error     at /Users/christrevino/Workspace/oss/react-dnd/node_modules/execa/index.js:278:16

```
</p></details>

## Your Environment
| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.15.0 |
| `npm --version`  | 6.9.0 |
| `yarn --version` | 1.16.0 |
| `node --version` | 12.4.0 |

| OS | Version |
| --- | --- |
MacOS Mojave

rozbo:
tested this way is the simplest to resolve 
#2151 
#1075 
#2267 
#2141 
#2113 
#1096 
#216 

etc....

```
       -u, --update
          Update the index just where it already has an entry matching <pathspec>. This removes as well as modifies index entries to match the working tree, but adds no new files.     
If no <pathspec> is given when -u option is used, all tracked files in the entire working tree are updated (old versions of Git used to limit the update to the current directory and its subdirectories).
```

It has a flaw in that it cannot track the newly created files,  but no new files are created in our scenario, so it is suitable for us.

### DIFF
--- a/commands/version/lib/git-add.js
+++ b/commands/version/lib/git-add.js
@@ -12,5 +12,5 @@ function gitAdd(files, opts) {
 
   const filePaths = files.map(file => slash(path.relative(opts.cwd, path.resolve(opts.cwd, file))));
 
-  return childProcess.exec("git", ["add", "--", ...filePaths], opts);
+  return childProcess.exec("git", ["add", "-u","--", ...filePaths], opts);
 }
--- a/commands/version/lib/git-add.js
+++ b/commands/version/lib/git-add.js
@@ -12,5 +12,5 @@ function gitAdd(files, opts) {
 
   const filePaths = files.map(file => slash(path.relative(opts.cwd, path.resolve(opts.cwd, file))));
 
-  return childProcess.exec("git", ["add", "-u","--", ...filePaths], opts);
+  return childProcess.exec("git", ["add", "-u", "--", ...filePaths], opts);
 }


rozbo:
Omm, it's my fault, I didn't look at the pull request notice. I am so sorry!
But as you can see, this case happens when a package to be published is in a .gitignore file. In this case, it should be logical that the check is no longer performed. but the actual situation is no difference to execution `git add`, this cause the error.
I really think this is a common scenario. For example, in a multi-person development scenario, a local development package doesn't want to be shared with others. At this point, obviously I should put it in gitignore. Another example is a dynamically generated package
the issues above is search the related words , it's scenario is related this problem.
In my point, the most thorough solution is to add an option to completely ignore the packages I don't want to publish, rather than relying on the private option. just like `command.bootstrap.ignore`.
Anyway, I wish you all the best, thank you for everything you have done!

evocateur:
Thanks for understanding.

Issue #2004, opened 2019-03-22
Add multiple packages in a single command


jordanbtucker:
## Expected Behavior

The following command would allow a user to add the `bar` and `baz` packages to the `foo` package.

```
lerna add --scope=foo bar baz
```

## Current Behavior

Currently you can only add one package at a time. To achieve the goal of the previous example, you must run the following.

```
lerna add --scope=foo bar
lerna add --scope=foo baz
```

## Context

It's quite common to install multiple npm packages as dependencies.

```
npm i -D eslint eslint-config-standard eslint-plugin-standard eslint-plugin-node eslint-plugin-import eslint-plugin-promise eslint-plugin-react
npm i -D @babel/core @babel/preset-env @babel/preset-react
npm i -D react react-dom redux react-redux react-router-dom
npm i -D webpack webpack-cli webpack-dev-server babel-loader file-loader css-loader style-loader
```

Intuitively, lerna should also allow this behavior. 

#1608 was opened to request this feature, but no discussion came from it. The result was essentially "lerna doesn't do this" and the reason given was vague.

> The ambiguity of targets forced a simplification of logic.

I'm not sure what that is supposed to mean, but this behavior caught me out the first time I tried to use lerna because I just expected it to work like npm.

JabbyPanda:
Supported glob patterns for `--scope` and its opposite counterpart `--ignore` can be looked upon in existing tests for `get-filtered-packages` utility 

https://github.com/lerna/lerna/blob/master/core/filter-options/__tests__/get-filtered-packages.test.js#L56-L64

![image](https://user-images.githubusercontent.com/255913/65586067-d0e6b080-df83-11e9-94ae-f522953bac0b.png)

jordanbtucker:
@JabbyPanda This issue has nothing to do with `--scope`.

JabbyPanda:
`--scope` does not support passing value as array 

https://github.com/lerna/lerna/issues/1846
https://github.com/lerna/lerna/issues/689

jordanbtucker:
@JabbyPanda This still has nothing to do with `--scope`. This issue is about `add`ing multiple packages in the same command, regardless of the scope. Basically, this should work:

```bash
# This works with npm:
npm install bar baz

# This does not work with lerna:
lerna add bar baz
```

I just added `--scope` to the command since it's a common option used. It's more common to add a package to only one of your monorepo packages than it is to add it to all of your packages.

georgyfarniev:
[...]

PR #2311, opened 2019-10-19
add --destination option to lerna add command

wingleung:
## Description
Moved the destination argument to a separate option.

Added a `--destination (-d)` option to the `lerna add` command which accepts globs to set the destination where we should add the packages, for example... 
```
lerna add package-1 --destination packages/*
```

## Motivation and Context
This allows for multiple package installs in one command (#2004)

## How Has This Been Tested?
Unit tests updated to match the new use of the `lerna add` command, currently broken because you can't throw errors in nested asynchronous callbacks of a promise. Will fix it as soon as we get more clarity around this new `--destination` approach.

## Types of changes

## Checklist:

### DIFF
--- a/commands/add/__tests__/add-command.test.js
+++ b/commands/add/__tests__/add-command.test.js
@@ -160,10 +160,10 @@ describe("AddCommand", () => {
     expect(pkg1).not.toDependOn("@test/package-1");
   });
 
-  it("filters targets by optional directory globs", async () => {
+  it("filters targets by optional directory destination globs", async () => {
     const testDir = await initFixture("basic");
 
-    await lernaAdd(testDir)("@test/package-1", "packages/package-2");
+    await lernaAdd(testDir)("@test/package-1", "--destination packages/package-2");
     const [, pkg2, pkg3, pkg4] = await getPackages(testDir);
 
     expect(pkg2).toDependOn("@test/package-1");
--- a/commands/add/command.js
+++ b/commands/add/command.js
@@ -5,7 +5,7 @@ const filterable = require("@lerna/filter-options");
 /**
  * @see https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module
  */
-exports.command = "add <pkg> [globs..]";
+exports.command = "add <pkg..>";
 
 exports.describe = "Add a single dependency to matched packages";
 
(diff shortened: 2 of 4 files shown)


commands/add/index.js
```diff
@@ -17,7 +17,25 @@ const getRangeToReference = require("./lib/get-range-to-reference");
...
+    chain = chain.then(
+      () =>
+        new Promise((resolve, reject) => {
+          const pkgArgv = Object.assign({}, argv, {
+            pkg,
+            onResolved: resolve,
+            onRejected: reject,
+          });
+
+          result.push(new AddCommand(pkgArgv));
```
evocateur: This isn't how you make multiple additions. A command is always unitary.

commands/add/index.js
```diff
@@ -28,7 +46,11 @@ class AddCommand extends Command {
   initialize() {
     this.dependencyType = this.options.dev ? "devDependencies" : "dependencies";
     this.spec = npa(this.options.pkg);
-    this.dirs = new Set(this.options.globs.map(fp => path.resolve(this.project.rootPath, fp)));
+    this.dirs = new Set(
+      this.options.destination
+        ? this.options.destination.map(fp => path.resolve(this.project.rootPath, fp))
+        : []
```
evocateur: This breaks all existing patterns. Not likely to be merged as-is.

commands/add/command.js
```diff
@@ -5,21 +5,24 @@ const filterable = require("@lerna/filter-options");
...
     .positional("pkg", {
       describe: "Package name to add as a dependency",
-      type: "string",
-    })
-    .positional("globs", {
-      describe: "Optional package directory globs to match",
       type: "array",
     })
     .options({
+      d: {
```
evocateur: This alias is too similar to the existing `-D`/`--dev` option. Unclear why it is necessary, or what precedent.

commands/add/__tests__/add-command.test.js
```diff
@@ -160,10 +160,10 @@ describe("AddCommand", () => {
     expect(pkg1).not.toDependOn("@test/package-1");
   });
 
-  it("filters targets by optional directory globs", async () => {
+  it("filters targets by optional directory destination globs", async () => {
```
evocateur: This would need more tests, if this were to be accepted.

evocateur:
This really isn't the way forward for Lerna, long-term. I'd rather let package managers manage workspace packages (i.e., `pnpm`/`yarn`/ maybe someday `npm`).

Issue #2333, opened 2019-11-08
Customize changelog and additional git add files


matzeeable:
## Context

I summarize two feature requests into this issue because I need it for my Open Source project https://github.com/matzeeable/wp-reactjs-starter which I want to restructure into a multi-package repository. I started to use `lerna` today and have successfully implemented it. Unfortunely there are two things which are missing so `lerna` can be used together with WordPress plugin development.

#### 1. It should be possible to pass `parserOpts` and `writerOpts` to `conventionalChangelog()`:

https://github.com/lerna/lerna/blob/eeacb92a4878ed8002f95359dc6d0b8ec14df12e/core/conventional-commits/lib/update-changelog.js#L58

In WordPress.org description changelog I do not want to show the links to associated GitLab / GitHub issues so I want to disable the option [`linkReferences`](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-writer#linkreferences). Another customization would be to disable the static `CHANGELOG.md` header, as already mentioned here: https://github.com/lerna/lerna/issues/2272 

There is a similar request https://github.com/lerna/lerna/issues/1471:
> You would need to create your own conventional-changelog preset

This can be very expensive just to change one option `linkReferences`. And why not customize good presets that already provide such options?

#### 2. Allow to `git add` other files

https://github.com/lerna/lerna/blob/eeacb92a4878ed8002f95359dc6d0b8ec14df12e/commands/version/index.js#L602

In WordPress the version is read from the [Plugin header fields](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields) so there is an additional file `index.php` which needs to be `git add`'ed. The version is adjusted via [`version`](https://github.com/matzeeable/wp-reactjs-starter/blob/6aa27807cf6ae92b94acbb880bc4f539ed407f5f/package.json#L40) automatically.

## Current Behavior
Currently the above features are not yet implemented and no such customizations / options are available, yet.

## Possible Solution
1. `parserOpts` and `writerOpts` are not needed as arguments via CLI. Perhaps it is possible to allow something like

```
{
    "command": {
        "version": {
            "conventionalCommits": true,
            "conventionalChangelog": {
                "writerOpts": {
                    "linkReferences": false
                }
            }
        }
    }
}
```

2. Introduce a new CLI and `lerna.json` option: `git-add-include` which allows an array of glob patterns (if needed).

## Your Environment

| Executable | Version |
| ---: | :--- |
[...]

PR #2335, opened 2019-11-08
feat(version): add `--git-add-include` parameter

matzeeable:
## Description
Introduce a new CLI and `lerna.json` option: `git-add-include` which allows an array of glob patterns (if needed).

## Motivation and Context
I summarize two feature requests in #2333 because I need it for my Open Source project https://github.com/matzeeable/wp-reactjs-starter which I want to restructure into a multi-package repository. I started to use `lerna` today and have successfully implemented it. Unfortunely there are two things which are missing so `lerna` can be used together with WordPress plugin development.

In WordPress the version is read from the [Plugin header fields](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/#header-fields) so there is an additional file `index.php` which needs to be `git add`'ed. The version is adjusted via [`version`](https://github.com/matzeeable/wp-reactjs-starter/blob/6aa27807cf6ae92b94acbb880bc4f539ed407f5f/package.json#L40) automatically.

## How Has This Been Tested?
It does not affect other areas of lerna because it is only used in `updatePackageVersions()`. I tested the changes locally on my lerna example repository.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/package.json
+++ b/commands/version/package.json
@@ -49,6 +49,7 @@
     "@lerna/validation-error": "file:../../core/validation-error",
     "chalk": "^2.3.1",
     "dedent": "^0.7.0",
+    "globby": "^9.2.0",
     "load-json-file": "^5.3.0",
     "minimatch": "^3.0.4",
     "npmlog": "^4.1.2",
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -51,6 +51,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--exact`](#--exact)
 - [`--force-publish`](#--force-publish)
 - [`--git-remote`](#--git-remote-name)
+- [`--git-add-include`](#--git-add-include-glob)
 - [`--create-release`](#--create-release-type)
 - [`--ignore-changes`](#--ignore-changes)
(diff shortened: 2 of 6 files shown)


commands/version/package.json
```diff
@@ -49,6 +49,7 @@
     "@lerna/validation-error": "file:../../core/validation-error",
     "chalk": "^2.3.1",
     "dedent": "^0.7.0",
+    "globby": "^9.2.0",
```
milesj: I would suggest `fast-glob` over all alternatives.

milesj:
Just needed this feature last week and was surprised to see it here. Would love this.

commands/version/package.json
matzeeable: Used `globby` because it is used in the other core packages, too. :-)

evocateur:
This is a job for [version lifecycle scripts](https://github.com/lerna/lerna/tree/master/commands/version#lifecycle-scripts).

milesj:
@evocateur I'll try that.

I tried `git add` before but it failed on dirty working directory (example: https://travis-ci.com/airbnb/lunar/jobs/254257044). Had to `git add + git commit` for it to work. But I'm guessing this happened during the wrong phase.

PR #2355, opened 2019-11-19
chore: update lock file

daern91:
PR updates lock file, particularly for issue #2348 where older versions of `yargs` would not work. More info inside the issue and this comment: https://github.com/lerna/lerna/issues/2348#issuecomment-555476487

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -122,9 +122,9 @@
       }
     },
     "@babel/parser": {
-      "version": "7.7.2",
-      "resolved": "https://registry.npmjs.org/@babel/parser/-/parser-7.7.2.tgz",
-      "integrity": "sha512-DDaR5e0g4ZTb9aP7cpSZLkACEBdoLGwJDWgHtBhrGX7Q1RjhdoMOfexICj5cqTAtpowjGQWfcvfnQG7G2kAB5w==",
+      "version": "7.7.3",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -9520,9 +9614,9 @@
       "integrity": "sha512-a4UGQaWPH59mOXUYnAG2ewncQS4i4F43Tv3JoAM+s2VDAmS9NsK8GpDMLrCHPksFT7h3K6TOoUNn2pb7RoXx4g=="
     },
     "yargs": {
-      "version": "14.2.0",
-      "resolved": "https://registry.npmjs.org/yargs/-/yargs-14.2.0.tgz",
-      "integrity": "sha512-/is78VKbKs70bVZH7w4YaZea6xcJWOAwkhbR0CFuZBmYtfTYF0xjGJF43AYd8g2Uii1yJwmS5GR2vBmrc32sbg==",
+      "version": "14.2.2",
```
daern91: 14.2.2 resolves the issue, more info: https://github.com/babel/babel/pull/10733#discussion_r347712415

evocateur: > we should not upgrade to v15 of yargs yet as it drops support for node 6

Yep, that's on my radar.

evocateur:
Thanks for the effort, but a monorepo root lockfile has no effect on its child modules. I'll be fixing this properly momentarily.

Issue #2067, opened 2019-05-03
Do not tag version changes with flag like "--no-tags" but do commit version changes


rambabusaravanan:
## Expected Behavior
Needs an option **not to tag but to commit** in git.

## Current Behavior
We have "--no-git-tag-version" option which will nethier commit nor tag version changes." Looking for a similar one

## Possible Solution
Would be good if we have option like "--no-tags" in cli or in lerna.json

## Context
Tagging versions in git is a better practice. But we need to do this only for actual release (patch, minor and major). For prerelease versions, we don't want to flood our repo with tags during development.

This is how we currently (planned to) use

```sh
# Developers can run this anytime in local 
"npm run bump:prerelease" 
> "lerna version prerelease --no-tags -y" 
```
```sh
# Can run this only when needed (upon release branches)
"npm run bump -- patch"
> "lerna version patch" 
```
These are npm scripts are to be executed at root.

evocateur:
I'm confused why you're running a `version` script in every leaf package to call ...Lerna? `lerna version prerelease` is what I _think_ you're trying to do.

Why not use [`--canary`](https://github.com/lerna/lerna/tree/master/commands/publish#--canary) for the dev prereleases? It doesn't change anything, but it does include the `gitHead` in the published packages.

rambabusaravanan:
That's a mistake. They are root level npm scripts which include `lerna version` and not `lerna run version`. Edited the issue. Thanks for this.

And --canary is also a good option and I'll try.

Still looking for `--no-tags` argument as a feature

evocateur:
You must understand, git tags are fundamental to how Lerna operates. If you're not willing to pass `--no-git-tag-version`, I don't think you should be committing these prerelease versions at all. `--no-tag` is an enormous footgun for the vast majority of use cases.

rambabusaravanan:
Okay well. I got that git tags are fundamental for Lerna to operate.

We have the below consideration
- only CI have access to publish. so jenkins runs publish command (usually for every commit)
- only developers should bump the version from his local and changes will be commited

Since the `--canary` option is only available in `lerna publish` command and not in `lerna version` command, only Jenkins can publish canary build. But for every commit, new build will be published which we usually don't prefer.

Currently we are doing manually as below (for development build)
1. `lerna version prerelease --no-push --preid 'dev'` bump version without pushing anything
2. `git tag -d v0.0.1-dev.0` delete the tag in local
3. `git push` and push commits to remote

[...]

Issue #1993, opened 2019-03-17
lerna version --no-tag is either needs readme clarification or is buggy


revelt:
## Expected Behavior

`lerna version` documentation says:

> By default, lerna version will push the committed and tagged changes to the configured git remote. Pass --no-push to disable this behavior.
> https://github.com/lerna/lerna/tree/master/commands/version#--no-push

I noticed that when this behaviour is disabled, tags are not created at all.

Now, from logical standpoint, either documentation needs to say something like:

> By default, lerna version will **create and** push the committed and tagged changes to the configured git remote. Pass --no-push to **neither create not push** any tags at all.

Or, **this is a bug** (which I hope it is because I want those git tags, I just don't want them pushed at that time).

## Some context

I want to push my monorepo to a remote (on GitLab) and its CI should run `lerna version`, update all files and publish to npm. Then, CI should commit whatever was changed, to the same remote, as if somebody did the release for me. What's left is only `git pull` on my computer.

I achieved all the above except git tags are missing.

Now, obvious direction is to remove `--no-push`, but since action is happening on CI already, there's no need to push **at this step**. I will push everything in the end anyway, and importantly, with `[skip ci]` commit message which will skip that push causing an infinite loop (every push starts a build).

Maybe I'm doing something not right, but documentation is scarce and I don't see many people publishing npm libraries using lerna monorepos from CI. If you know any, please leave a comment, I'd gladly reference their setup.

## Current Behavior
Currently, `lerna version --no-push` will not create any tags at all.

## Possible Solution
If possible, either clarify the documentation to state that git tags will not be created and will not be pushed, OR fix the lerna to create the tags but not push them.

## Steps to Reproduce (for bugs)

I called the following npm script from GitLab CI:

```
lerna version --conventional-commits --no-commit-hooks --no-push --yes
```

For reference, here's my repo illustrating the situation, from this morning's release: https://gitlab.com/codsen/codsen/commit/c00a72a42dcbedefccb122bdd4396f03117ef245
Git tags are missing.

<details><summary>lerna.json</summary><p>

```json
{
  "command": {
    "publish": {
      "conventionalCommits": true,
      "message": "chore: Release"
    }
  },
  "lifecycle": {
    "version": "npm run build && git add ."
  },
  "packages": [
    "packages/*"
  ],
  "version": "independent"
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.13.1 |
[...]

Issue #1813, opened 2018-12-06
Abort git push on publish error


timkinnane:
## Expected Behavior
When npm publish is unsuccessful, due to config or version issues, the publish process should abort before adding tags and pushing.

## Current Behavior
On errors in publish config etc, the npm packages aren't published, but the version tags are added to the repo. This adds steps when debugging release flow.

I'm still learning Lerna and getting my config right, practicing with local npm registry etc. I'm often needing to roll back my HEAD and go to github and delete tags because the publish action failed but the git changes proceeded as if it succeeded.

## Possible Solution
Put git actions after publish, so if it exits early they don't happen.

## Steps to Reproduce (for bugs)
Release with a bad config, e.g. no token or a version that already exists.

## Version
Lerna 3.5.0

evocateur:
The git operations are independent of the npm operations, though. npm has literally no opinion whether or not your code is actually versioned, tagged, or pushed to the remote.

In any case, you have a couple options:
1. `--no-push` with explicit `git push`:
```
lerna publish --no-push && git push
```
You could configure this in `lerna.json` like this:
```json
{
  "command": {
    "publish": {
      "push": false
    }
  },
  ...
}
```
2. Recover from completely failed npm publish with `from-git` keyword:
```
lerna publish from-git
```
This will skip the git commit and tagging and just rerun the `npm publish` steps.

evocateur:
"A version already exists" seems like a drastically bad state that _should_ fail?

"No token" we could probably do with better preparation during `lerna publish`, but is rather hampered by the use of a third-party registry that likely doesn't implement anything useful in terms of validating endpoints (e.g., `npm ping`, `npm whoami`, or `npm profile get`).

timkinnane:
Thanks @evocateur I assumed that an error from npm could cause the lerna publish process to exit, thereby preventing the git actions from occurring. When you say they're independent, do you mean it's two parallel processes? So they both run regardless of each other's success?
Just wanted to clarify that if you could. I'm happy to close the issue and use the `--no-push` approach, thanks for pointing that out.

PR #2358, opened 2019-11-21
Adds "--no-git-tag" and "--no-git-commit" version options

artem-zakharchenko:
This pull request allows to granularly control the tagging and committing steps during the `lerna version` command.

- Closes https://github.com/lerna/lerna/issues/2067

## Description

This pull request introduces two new options to the `lerna version` command:

- `--no-git-tag` - when provided, skips the tagging of the release;
- `--no-git-commit` - when provided, skips the release commit.

> Internally, `this.commitAndTag` value is rewritten to align with these two newly introduced options.

## Motivation and Context

Our Git workflow requires each release to have a dedicated Pull request and be reviewed before landing in `master`. This implies that the release process is split into two steps:

1. Release pull request. Contains a release commit **without tags**.
1. Tagging on master, after the release pull request is merged.

With the current implementation of `lerna version` it's possible to either:

1. commit + tag (default).
1. Skip both commit and tag.

The flow described above would require to preserve the git commit, but skip tagging, as tags are set once the release pull request is merged into `master`.

## How Has This Been Tested?

Please see the added integration tests (c02f9130ae58b3102e7ce6a209de8f5a3dc7e5b7).

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/__tests__/version-no-git-commit.test.js
+++ b/commands/version/__tests__/version-no-git-commit.test.js
@@ -1,0 +1,40 @@
+"use strict";
+
+jest.mock("../lib/git-push");
+jest.mock("../lib/is-anything-committed");
+jest.mock("../lib/is-behind-upstream");
+jest.mock("../lib/remote-branch-exists");
+
+const path = require("path");
+
+// helpers
+const loggingOutput = require("@lerna-test/logging-output");
+const initFixture = require("@lerna-test/init-fixture")(path.resolve(__dirname, "../../publish/__tests__"));
+const showCommit = require("@lerna-test/show-commit");
+
+// test command
+const lernaVersion = require("@lerna-test/command-runner")(require("../command"));
+
+describe("--no-git-commit", () => {
+  it("does not create a git commit", async () => {
+    const testDir = await initFixture("normal");
+    await lernaVersion(testDir)("--no-git-commit");
+
+    const logMessages = loggingOutput("info");
+    expect(logMessages).toContain("Skipping commit");
+
+    const commit = await showCommit(testDir);
+    expect(commit).not.toBe("v1.0.1");
+  });
+
+  it("tags the release", async () => {
+    const testDir = await initFixture("normal");
+    await lernaVersion(testDir)("--no-git-commit");
+
+    const logMessages = loggingOutput("info");
+    expect(logMessages).not.toContain("Skipping tag");
+
+    const commit = await showCommit(testDir);
+    expect(commit).toContain("tag:");
+  });
+});
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -72,7 +72,7 @@ class VersionCommand extends Command {
 
     this.gitRemote = gitRemote;
     this.tagPrefix = tagVersionPrefix;
-    this.commitAndTag = gitTagVersion;
+    this.commitAndTag = this.options.gitTag && this.options.gitCommit && gitTagVersion;
     this.pushToRemote = gitTagVersion && amend !== true && push;
     // never automatically push to remote when amending a commit
 
@@ -277,6 +277,14 @@ class VersionCommand extends Command {
       );
     } else {
       this.logger.info("execute", "Skipping releases");
+    }
+
+    if (!this.options.gitTags) {
+      this.logger.info("execute", "Skipping tags");
+    }
+
+    if (!this.options.gitCommit) {
+      this.logger.info("execute", "Skipping commit");
     }
 
     return pWaterfall(tasks).then(() => {
@@ -637,8 +645,10 @@ class VersionCommand extends Command {
     const message = tags.reduce((msg, tag) => `${msg}${os.EOL} - ${tag}`, `${subject}${os.EOL}`);
 
     return Promise.resolve()
-      .then(() => gitCommit(message, this.gitOpts, this.execOpts))
-      .then(() => Promise.all(tags.map(tag => gitTag(tag, this.gitOpts, this.execOpts))))
+      .then(() => this.options.gitCommit && gitCommit(message, this.gitOpts, this.execOpts))
+      .then(
+        () => this.options.gitTag && Promise.all(tags.map(tag => gitTag(tag, this.gitOpts, this.execOpts)))
+      )
       .then(() => tags);
   }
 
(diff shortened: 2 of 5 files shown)


artem-zakharchenko:
Tests seem to pass except Windows. Getting the next exception:

```
Worker information
Secret environment variables are not obfuscated on Windows, please refer to our documentation: https://docs.travis-ci.com/user/best-practices-security
0.05s0.06s3.32sDisabling Windows Defender
$ powershell -Command Set-MpPreference -DisableArchiveScanning \$true
$ powershell -Command Set-MpPreference -DisableRealtimeMonitoring \$true
$ powershell -Command Set-MpPreference -DisableBehaviorMonitoring \$true
git.checkout
1.74s$ git clone --depth=3 https://github.com/lerna/lerna.git lerna/lerna
0.09s
Setting environment variables from .travis.yml
$ export NO_UPDATE_NOTIFIER=1
$ export NODE_NO_WARNINGS=1
$ export LERNA_CI_SCRIPT=ci
$ export LERNA_CI_SCRIPT="ci:windows"
$ export LERNA_CI_TYPE="non-publish"
nvs.setup
Using NVS for managing Node.js versions on Windows (BETA)
22.76s$ nvs add 12
Downloading bootstrap node from https://nodejs.org/dist/v10.12.0/node-v10.12.0-win-x64.7z
######################################################################## 100.0%
The command "nvs add 12" failed and exited with 1 during .
```

Looks like Travis CI cannot properly install NodeJS version 12 during the CI.

pksunkara:
@evocateur Would love if you could give this a look.

commands/version/index.js
```diff
@@ -650,13 +662,17 @@ class VersionCommand extends Command {
...
-      .then(() => gitCommit(message, this.gitOpts, this.execOpts))
-      .then(() => gitTag(tag, this.gitOpts, this.execOpts))
+      .then(() => this.shouldCommit && gitCommit(message, this.gitOpts, this.execOpts))
+      .then(() => this.shouldTag && gitTag(tag, this.gitOpts, this.execOpts))
       .then(() => [tag]);
   }
 
   gitPushToRemote() {
-    this.logger.info("git", "Pushing tags...");
+    if (this.shouldTag) {
```
artem-zakharchenko: Since `gitPushToRemote()` method is called when either `this.shouldTag` or `this.shouldCommit` is true, if inside the method, and `!this.shouldTag`, means only `this.shouldCommit` equals `true`. Thus, it's safe to fallback to log the commit message.

artem-zakharchenko:
@honzajavorek, hi 👋Please, any chance you could look into this? Thank you.

honzajavorek:
I could, but I only previously contributed to Lerna, I'm no maintainer here. I cannot provide an authoritative review nor merge the PR. That's up to @evocateur.

commands/version/index.js
```diff
@@ -72,7 +72,9 @@ class VersionCommand extends Command {
 
     this.gitRemote = gitRemote;
     this.tagPrefix = tagVersionPrefix;
-    this.commitAndTag = gitTagVersion;
+    this.shouldCommit = this.options.gitCommit && gitTagVersion;
+    this.shouldTag = this.options.gitTag && gitTagVersion;
```
evocateur: How exactly does one tag a release commit that doesn't exist? I'm very confused by this.

commands/version/__tests__/version-no-git-commit.test.js
```diff
@@ -0,0 +1,40 @@
...
+describe("--no-git-commit", () => {
+  it("does not create a git commit", async () => {
+    const testDir = await initFixture("normal");
+    await lernaVersion(testDir)("--no-git-commit");
+
+    const logMessages = loggingOutput("info");
+    expect(logMessages).toContain("Skipping commit");
+
+    const commit = await showCommit(testDir);
+    expect(commit).not.toBe("v1.0.1");
```
evocateur: This isn't actually verifying what you think it is. How does `git tag` target a revision that doesn't exist?

darekkay:
> I'm still confused why this even makes sense. I'm not inclined to add (more) footguns to Lerna in support of an extremely convoluted release process.

@evocateur I understand your point of not adding unnecessary options, as they make the implementation more complex. However, personally I was happy to see this PR, because it would solve an issue we are facing for a long time.

First, there are some use cases, like #2067 and #1993 
However, I am facing another use case, similar to #1813

Here's the short version of it:

- Everytime a Pull Request is merged into `master`, our CI (Jenkins) triggers `lerna publish`
- By default, Lerna will create a new version, tag the commit and push it back to the `master` branch
- If another Pull Request is merged in between, the push will fail ("commit behind upstream"). However, the Tags do still exist, as they can be pushed without an error
- The new Pull Request will trigger another `lerna publish` (after the first job is finished / has failed)
- This job will also fail at the  `git push` part, but for another reason - the Git tags already exist (previous step)

Our fix is to delete the "wrongfully" created git tags and re-trigger the job.
The workaround is to make sure that no Job is currently running before a Pull Request is merged - which is not great.

It might be possible that there are other ways to solve it (and we are open to suggestions), but this PR would solve our issue without any other required changes. We do not need the git tags at all. However, the current `--no-git-tag-version` option does more than the name suggests, so we would have to make further adjustements to make it work.

joebowbeer:
`publish from-package` was added to assist CI workflows. I'm curious to know how your desired workflow differs from the one described in https://github.com/lerna/lerna/issues/1648#issuecomment-439689466

commands/version/index.js
artem-zakharchenko: Thank you for the review. Your concern is valid. I've reviewed the logic and discussed it in a team and I'll update the pull request based on your comments.

FransBosuil:
We would like to use the --no-git-commit flag allowing us to add the tags without adding the additional commit message. 

As far as we understand, this is not possible with the currently provided functionality. But, with the above changes, it will be. Can you please indicate when this change is expected to be delivered?

commands/version/index.js
FransBosuil: The functionality should indeed not tag the release commit. It is indeed not possible as it does not exist. However, it can tag the last commit, no?

JamesHenry:
Hi Folks, in light of its age and the lack of buy in from @evocateur in the past threads, I am going to close this one.

If you have any issues or feature requests for the latest lerna please do feel free to open a fresh issue with maximal context and any relevant reproductions for issues.

Many thanks!

PR #2376, opened 2019-12-10
feat: Add `--profile` option to `lerna exec` and `lerna run`

bweggersen:
Adds an option to profile the run and exec commands. It produces a performance profile which can be analyzed using DevTools in a Chromium-based browser.

## Description
Profiles the run and exec commands and produces a performance profile which can be analyzed using DevTools in a Chromium-based browser (direct url: `devtools://devtools/bundled/devtools_app.html`). The profile shows a timeline of the operations where each operation is assigned to an open slot. The number of slots is determined by the `--concurrency` option and the number of open slots is determined by `--concurrency` minus the number of ongoing operations. The end result is a visualization of the parallel execution of your scripts.

image

The default location of the performance profiles is at the root of your project.

## Motivation and Context
We have over 200 packages in our repository and we see build time regressions from time to time. Visualizing `lerna run build` helps us uncover obvious mistakes or opportunities for improvement. For instance, in the example above, the final green box in thread 0 depends on the green box immediately preceding it on thread 1 (expanding the threads in the DevTool reveals the name of each package). By splitting the green package on thread 1, we can allow the final package on thread 0 to start much sooner, giving us a performance improvement of ~5 mins on all builds.

## How Has This Been Tested?
I have tested this locally with all the argument permutations. If we should have some more formal testing of this feature, I would greatly appreciate some guidance on what that could look like.

## Types of changes

## Checklist:


package-lock.json
```diff
@@ -1299,7 +1299,10 @@
       "requires": {
         "@lerna/query-graph": "file:utils/query-graph",
         "figgy-pudding": "^3.5.1",
-        "p-queue": "^4.0.0"
+        "fs-extra": "^8.1.0",
+        "npmlog": "^4.1.2",
+        "p-queue": "^6.2.1",
```
evocateur: Unfortunately, p-queue 5+ does not support node v6, and thus this is a breaking change. Any specific reason this needed to be upgraded, anyway?

(I plan on dropping support for node v6 + v8 real soon now, but I'd rather avoid it in this PR.

package-lock.json
```diff
@@ -1299,7 +1299,10 @@
...
+        "upath": "^1.2.0"
```
evocateur: What does `upath` provide that the builtin `path` doesn't? Cross platform normalization is available via `path.normalize()`.

utils/run-topologically/profiler.js
```diff
@@ -0,0 +1,86 @@
...
+};
+
+const generateOutputname = () => {
+  const now = new Date(); // 2011-10-05T14:48:00.000Z
+  const datetime = now.toISOString().split(".")[0]; // 2011-10-05T14:48:00
+  const datetimeNormalized = datetime.replace(/-|:/g, ""); // 20111005T144800
+  return `Lerna-Profile-${datetimeNormalized}.json`;
+};
+
+class Profiler {
```
evocateur: This really feels like a separate, general purpose module completely independent of Lerna. I'm also surprised it doesn't exist already, accepting a callback and output options, et cetera.

utils/run-topologically/profiler.js
```diff
@@ -0,0 +1,86 @@
...
+    if (!this.profile) {
+      return;
+    }
+
+    const outputFolder = this.profileLocation
+      ? upath.join(this.rootPath, this.profileLocation)
+      : this.rootPath;
+    const outputPath = upath.join(outputFolder, generateOutputname());
+
+    fs.outputJsonSync(outputPath, this.events);
```
evocateur: Seems like you could just as easily return the promise version of this method, as it is always being called inside the queue's `onIdle()` chain.

utils/run-topologically/profiler.js
```diff
@@ -0,0 +1,86 @@
...
+  return `Lerna-Profile-${datetimeNormalized}.json`;
+};
+
+class Profiler {
+  constructor({ concurrency, log, profile, profileLocation, rootPath }) {
+    this.events = [];
+    this.profile = profile;
+    this.log = log;
+    this.profileLocation = profileLocation;
+    this.rootPath = rootPath;
```
evocateur: I'm not sure why you need `profileLocation` _and_ `rootPath`, the defaulting would be wherever the config is created, not in this class's `output` method.

utils/run-topologically/run-topologically.js
```diff
@@ -27,8 +36,11 @@ const TopologicalConfig = figgyPudding({
  * @returns {Promise<Array<*>>} when all executions complete
  */
 function runTopologically(packages, runner, opts) {
-  const { concurrency, graphType, rejectCycles } = TopologicalConfig(opts);
+  const { concurrency, graphType, log, profile, profileLocation, rejectCycles, rootPath } = TopologicalConfig(
```
evocateur: Cool thing about figgy-pudding is that you don't need to make `TopologicalConfig` aware of the profiler options, you can separate those into a `ProfilerConfig` factory that you pass the raw `opts` to instead. (which helps the argument that the profiler itself should be a separate package)

In Profiler.js:
```js
const ProfilerConfig = figgyPudding({
  concurrency: {},
  log: { default: npmlog },
  // alias 'profile' to 'enabled', reads better
  profile: "enabled",
  // a boolean does not need a default, it's implied
  enabled: {},
  profileLocation: {},
  rootPath: {},
};

class Profiler {
  constructor(opts) {
    const { concurrency, log, enabled, profileLocation, rootPath } = ProfilerConfig(opts);
    // ...
```

in run-topologically:
```js
  const profiler = new Profiler(opts);
```

commands/exec/index.js
```diff
@@ -127,6 +127,10 @@ class ExecCommand extends Command {
     return runTopologically(this.filteredPackages, runner, {
```
evocateur: It strikes me that it should be the consumers of `runTopologically()` that wrap the `runner` (that they're already creating) with the profiler and output the profiler results, _not_ `@lerna/run-topologically` itself. It's a bit repetitive, but that's nothing new as far as `lerna exec` vs `lerna run` is concerned...

```js
  runCommandInPackagesTopological() {
    let profiler;
    let runner = this.options.stream
      ? pkg => this.runCommandInPackageStreaming(pkg)
      : pkg => this.runCommandInPackageCapturing(pkg);

    if (this.options.profile) {
      profiler = new Profiler(this.options);
      runner = pkg => profiler.run(() => runner(pkg), pkg.name);
    }

    let chain = runTopologically(this.filteredPackages, runner, {
      concurrency: this.concurrency,
      rejectCycles: this.options.rejectCycles,
    });

    if (profiler) {
      chain = chain.then(results => {
        profiler.output(); // changes if it's async, i guess
        return results;
      });
    }

    return chain;
  }
```

commands/exec/README.md
```diff
@@ -79,3 +79,27 @@ Pass `--no-bail` to disable this behavior, executing in _all_ packages regardles
...
+Profiles the command executions and produces a performance profile which can be analyzed using DevTools in a
+Chromium-based browser (direct url: `devtools://devtools/bundled/devtools_app.html`). The profile shows a timeline of
+the command executions where each execution is assigned to an open slot. The number of slots is determined by the
+`--concurrency` option and the number of open slots is determined by `--concurrency` minus the number of ongoing
+operations. The end result is a visualization of the parallel execution of your commands.
+
+The default location of the performance profiles is at the root of your project.
+
+```sh
+$ lerna run build --profile
```
evocateur: This should be `lerna exec` (wish i had an easier way of sharing the obviously duplicated guts of `lerna exec` & `lerna run`...)

commands/exec/README.md
```diff
@@ -79,3 +79,27 @@ Pass `--no-bail` to disable this behavior, executing in _all_ packages regardles
...
+```
+
+> **Note:** Lerna will only profile when topological sorting is enabled (i.e. without `--parallel` and `--no-sort`).
+
+### `--profile-location <location>`
+
+You can provide a custom location for the performance profiles. The location is relative to the root of your project.
```
evocateur: Technically, path options are relative to the PWD when the command is called, but I suppose this is good enough. (They can also be passed as absolute paths, which means the `path.join(rootPath, profilePath)` in Profiler should be `path.resolve(rootPath, profilePath)`, at least)

commands/exec/command.js
```diff
@@ -57,6 +57,16 @@ exports.builder = yargs => {
         hidden: true,
         type: "boolean",
       },
+      profile: {
+        group: "Command Options:",
+        describe: "Profile command executions and output performance profile to default location.",
+        type: "bolean",
```
evocateur: `s/bolean/boolean/`

package-lock.json
bweggersen: No, I can revert it. I was playing with the new custom QueueClass which was introduced in p-queue 6 in an earlier iteration of the PR. When I abandoned that strategy, I just kept the upgrade.

utils/run-topologically/profiler.js
bweggersen: I couldn't find any. If you have any suggestions, I'm totally open to use one. Having said that, I do feel that the part that could be generalized is pretty small compared to the parts that are specific for our use case.

In any case, I'm happy to create a separate profiler package in Lerna.

utils/run-topologically/profiler.js
bweggersen: Good point!

utils/run-topologically/run-topologically.js
bweggersen: Awesome! Thanks!

commands/exec/index.js
bweggersen: Also a good point. I'll make the adjustments.

commands/exec/README.md
bweggersen: I'm indifferent on whether to make it relative to the root or to the pwd. I thought it made some sense since if you run `lerna something` from a package, it will run as if it was run from root (i.e. `lerna run build` from `packages/foo` runs `lerna run build` from the root). What would you prefer?

utils/run-topologically/profiler.js
bweggersen: I can move the logic of creating the output path out of the `output` method and place it in the `constructor`. Does that sound better?

package-lock.json
bweggersen: If `profileLocation` is provided with `\\` as separators, and lerna is run in a posix environment, or if `profileLocation` is provided with `/` as separators, and lerna is run in a Windows environment, there will be a mismatch path separator mismatch, as far as I can see. And that will result in runtime errors.

```
> upath.join("c:\\windows\\nodejs\\path", "packages/foo")
"c:/windows/nodejs/path/packages/foo"

> path.join("c:\\windows\\nodejs\\path", "packages/foo")
"c:\\windows\\nodejs\\path/packages/foo"

> upath.join("/Users/root/nodejs", "packages\\foo")
"/Users/root/nodejs/packages/foo"

> path.join("/Users/root/nodejs", "packages\\foo")
"/Users/root/nodejs/packages\\foo"

> path.normalize(path.join("/Users/root/nodejs", "packages\\foo"))
"/Users/root/nodejs/packages\\foo"
```

commands/exec/README.md
evocateur: I think for CLI consistency (and "principle of least surprise") we should treat it as relative to the CWD. In the vast majority of cases, that will be identical to the root of the project.

package-lock.json
evocateur: I looked into it, `upath` seems pretty solid. Thanks for introducing me to the package. :)

utils/run-topologically/profiler.js
evocateur: We could even [coerce](https://github.com/yargs/yargs/blob/master/docs/api.md#coercekey-fn) the `--profile-location` option with `upath.resolve` (proxied from `path` with internal normalization) and accept `opts.profileLocation` as-is.

PR #2396, opened 2019-12-28
WIP: Add automatic scope resolution with --currentScope

alexpusch:
Add --currentScope filter option. Work in progress - if you find this change desirable I will add necessary tests. 

## Description

When adding --currentScope in commands, lerna will use the current working directory package as scope

## Motivation and Context

Running a lerna command for a specific package is common. For example in order to add an npm to a package one will `lerna add --scope={current package name}`. This is quite verbose is foreign to the old fashioned `npm install` in package directory. I suggest adding an option, and in the future maybe even making this a default behavior, to use the current working directory as scope.

## How Has This Been Tested?

TBD

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My change requires a change to the documentation.

### DIFF
--- a/core/filter-options/index.js
+++ b/core/filter-options/index.js
@@ -15,6 +15,11 @@ function filterOptions(yargs) {
       type: "string",
       requiresArg: true,
     },
+    currentScope: {
+      describe: "Include only package in current working directory.",
+      type: "boolean",
+      conflicts: "scope",
+    },
     ignore: {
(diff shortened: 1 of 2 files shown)


evocateur:
Sorry, no. You should only run `lerna` commands from the project root, not package (leaf) directories.

Issue #2422, opened 2020-01-21
Really strange bootstrapping behaviour with --nohoist; *only* nohoist'd package is installed in subpackages


ELLIOTTCABLE:
## Expected Behavior
When running `lerna bootstrap` with a `--nohoist` argument, subpackages should have *all* of their dependencies installed.

## Current Behavior
With the package `bs-platform` selected for `--nohoist`, `lerna bootstrap` *only*(!?) installs `bs-platform` in the subpackages, instead of following the normal behaviour of installing all of the subpackages' dependencies after hoisting.

As a side-effect, this also destroys the package-lockfile for each subpackage, replacing it with one containing, well, only `bs-platform`:

    $ git diff --stat
     packages/bs-excmd/package-lock.json | Bin 508867 -> 416 bytes
     packages/excmd/package-lock.json    | Bin 586183 -> 435 bytes
    $ cat packages/excmd/package-lock.json
    {
       "name": "bs-excmd",
       "version": "1.0.0-secondary-rollup-package.0",
       "lockfileVersion": 1,
       "requires": true,
       "dependencies": {
          "bs-platform": {
             "version": "5.2.1",
             "resolved": "https://registry.npmjs.org/bs-platform/-/bs-platform-5.2.1.tgz",
             "integrity": "sha512-3ISP+RBC/NYILiJnphCY0W3RTYpQ11JGa2dBBLVug5fpFZ0qtSaL3ZplD8MyjNeXX2bC7xgrWfgBSn8Tc9om7Q=="
          }
       }
    }

Note in the debugging-output below that, at some point, Lerna appears to be writing a `package.json` including *only* `bs-platform`, though I don't understand why it's doing that ...

## Steps to Reproduce (for bugs)
1. `git clone https://github.com/ELLIOTTCABLE/excmd.js.git`
2. `cd excmd.js && git checkout wtf-lerna`
3. `npx lerna bootstrap`
4. `git diff --stat` — see the package-lockfiles?
5. `(cd packages/bs-excmd/ && npm ls)`
6. `(cd packages/bs-excmd/ && ls node_modules)`

<details><summary>lerna.json</summary><p>

```json
{
   "packages": ["packages/*"],
   "command": {
      "bootstrap": {
         "strict": true,
         "hoist": true,
         "nohoist": "bs-platform"
      },
      "run": {
         "stream": true
      }
   },
   "version": "1.0.0-secondary-rollup-package.0"
}
```
</p></details>

[...]

PR #2424, opened 2020-01-23
Leave external dependencies that only exist in a single leaf-package unhoisted

ELLIOTTCABLE:
> (Notably `npm test` on HEAD was producing broken snapshots for me, so this changeset isn't ready to merge — I need to rebase it without my temporaray-fix commit, once master is building and passing tests.)

## Description
At the moment, in behaviour that I found extremely surprising (see #2422), `lerna bootstrap --hoist` will remove ... basically *all* dependencies of leaves, which seems to disagree with `hoist.md`:

> **Common dependencies** will be installed only to the top-level `node_modules`, and omitted from individual package `node_modules`.

This patch ensures that `lerna bootstrap` will only ever hoist dependency/version instances that *appear in more than one leaf-project*.

## Motivation and Context
At least in my projects, I've got quite a few dependencies (note: not devDependencies) that stop working when hoisted. I didn't expect this to be a problem; because I was only expecting to hoist JavaScript-tooling common to several projects — Prettier, Jest, Babel, Typescript, etc.

More importantly, the current behaviour just seems a little ... unintuitive. Why hoist a peculiar build-tool specific to only a single project, into the monorepo root?

## How Has This Been Tested?
I've variously updated and/or expanded upon the related `__tests__` related to `lerna bootstrap`; but I'm happy to offer more testing, if this change is desired.

## Types of changes

I'm not actually sure that this is a breaking change — I would suppose that this can only take an already-hoisted package and *return* it to normal npm semantics, so theoretically it shouldn't break any already-hoisted packages?

That said, it's a change in behaviour, and I don't see it as critical; so it makes sense to defer it until a major bump, if y'all are interested in this patch. (=

## Checklist:

- [x] My code follows the code style of this project: *I think! Feedback welcome.*
- [ ] All new and existing tests passed: *See note at the top.*

### DIFF
--- a/commands/diff/__tests__/__snapshots__/diff-command.test.js.snap
+++ b/commands/diff/__tests__/__snapshots__/diff-command.test.js.snap
@@ -1,59 +1,52 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`DiffCommand passes diff exclude globs configured with --ignored-changes 1`] = `
-diff --git a/packages/package-1/package.json b/packages/package-1/package.json
-index SHA..SHA 100644
---- a/packages/package-1/package.json
-+++ b/packages/package-1/package.json
-@@ -1,5 +1,5 @@
- {
-   "name": "package-1",
--  "changed": 0,
-+  "changed": 1,
-   "version": "1.0.0"
- }
+"[0;31m------ [0m[0;1mpackages/package-1/package.json 100644[0m
+[0;32m++++++ [0m[0;1mpackages/package-1/package.json 100644[0m
+[0;100;30m@|[0m[0;1m-1,5 +1,5[0m ============================================================
+[0;100;30m |[0m{
+[0;100;30m |[0m  \\"name\\": \\"package-1\\",
+[0;41;30m-|[0m[0;0m[0;2m  \\"changed\\":[0m[0;31m 0[0m[0;2m,[0m[0m
+[0;42;30m+|[0m[0;0m  \\"changed\\":[0;32m 1[0m,[0m
+[0;100;30m |[0m  \\"version\\": \\"1.0.0\\"
+[0;100;30m |[0m}"
 `;
 
 exports[`DiffCommand should diff a specific package 1`] = `
-diff --git a/packages/package-2/package.json b/packages/package-2/package.json
-index SHA..SHA 100644
---- a/packages/package-2/package.json
-+++ b/packages/package-2/package.json
-@@ -1,6 +1,6 @@
- {
-   "name": "package-2",
--  "changed": 0,
-+  "changed": 1,
-   "version": "1.0.0",
-   "dependencies": {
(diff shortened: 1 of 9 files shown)


commands/diff/__tests__/__snapshots__/diff-command.test.js.snap
```diff
@@ -1,59 +1,52 @@
...
---- a/packages/package-1/package.json
-+++ b/packages/package-1/package.json
-@@ -1,5 +1,5 @@
- {
-   "name": "package-1",
--  "changed": 0,
-+  "changed": 1,
-   "version": "1.0.0"
- }
+"[0;31m------ [0m[0;1mpackages/package-1/package.json 100644[0m
```
evocateur, line 4: Your (global) git config seems to be enforcing colors no matter what? Should probably switch that to `auto` instead.

evocateur:
> Why hoist a peculiar build-tool specific to only a single project, into the monorepo root?

Because a single leaf that has a completely unique build tool is antithetical to a monorepo.

If your dependencies are too brittle to survive `--hoist`, then I wouldn't recommend using `--hoist`. (Also (not) incidentally what [`--nohoist`](https://github.com/lerna/lerna/tree/master/commands/bootstrap#--nohoist-glob) was meant for)

Installing as many dependencies as possible in the root ("common" or not) is dramatically faster than installing zillions of "singleton" packages in multiple leaf nodes.

If `--nohoist` is broken, I would rather fix that instead of fundamentally altering the behavior of `--hoist` (yes, this is a breaking change).

(I would _rather_ nuke `lerna bootstrap` entirely, it is a gloriously heinous hack largely replaced by a modern package manager's "workspaces" support)

ELLIOTTCABLE:
I am unlikely to have time or energy to revisit this patch anytime soon; I'm happy whether you decide to leave it open and wait until that day arrives, or close it as inactive. )'=

Contextually: this was relevant to my work, as my monorepos include *multiple languages*. There's not much Lerna, or really any tooling, can do to make my OCaml tooling for the OCaml-components of my work "hoist" in the npm sense; so "hoisting" the OCaml components up to the root is painfully unhelpful; whereas multiple projects share JavaScript-universe tooling, so hoisting *those* components is very helpful.

> Because a single leaf that has a completely unique build tool is antithetical to a monorepo.

I don't know, I think that's a bit conservative / constrictive on what "a monorepo" *is*, and who it can be useful for — it may be antithetical to what *you've* used monorepos for, but I'd assert that it's far from true across-the-board. ¯\\_(ツ)\_/¯

I hope this feedback was helpful! Nice catch on the git-config, lmao, that's almost certainly what was screwing up the test-suite for me back then. 🤨

srisuk2519:
เมื่อ อา. 21 มิ.ย. 2563 เวลา 16:26 ELLIOTTCABLE <notifications@github.com>
เขียนว่า:

PR #2441, opened 2020-02-08
fix(import): Support non-ASCII characters in file paths

WeiAnAn:
## Description
When importing, if any file path contain non-ascii char like Chinese, it will occur an error.

```
ValidationError: Failed to apply commit f60aec4.
    Command failed: git am -3 --keep-non-patch
    error: mode change for packages/91b633aa89e0fab2957cc60ec58fe8c5/"\346\252\224\346\241\210", which is not in current HEAD
    error: could not build fake ancestor
    Applying: rename
    Patch failed at 0001 rename
    Use 'git am --show-current-patch' to see the failed patch
    When you have resolved this problem, run "git am --continue".
    If you prefer to skip this patch, run "git am --skip" instead.
    To restore the original branch and stop patching, run "git am --abort".
```

Because git will wrap non-ascii file path by `"`, the path replacement does not handle this.

## Motivation and Context

Fix by checking `"` in regular expression in replacement.

## How Has This Been Tested?

Add a unit test.
The test include git file add, rename and copy action.
I think it cover all regexp.

## Types of changes

## Checklist:

### DIFF
--- a/commands/import/__tests__/__fixtures__/files-with-non-ascii-char/package.json
+++ b/commands/import/__tests__/__fixtures__/files-with-non-ascii-char/package.json
@@ -1,0 +1,3 @@
+{
+  "name": "files-with-non-ascii-char"
+}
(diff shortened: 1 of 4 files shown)


tomyam1:
Tried this patch after I couldn't import repo (same error as above) with some Hebrew file names and it worked great. Thanks

PR #2450, opened 2020-02-14
feat: Support the `workspace:` protocol

nicolo-ribaudo:
## Description

Yarn 2 and pnpm (I know that it isn't supported, but saw that you are working on it) support the `workspace:` protocol for monorepo package versions.

When used, it explicitly marks a dependency to assert that it is never downloaded from the registry. Both the package managers allow making the `workspace:` protocol required: if it isn't specified, then the dependency is downloaded from the registry even if a local package would match the requested version.
[yarn docs](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace), [pnpm docs](https://pnpm.js.org/en/workspaces#workspace-ranges-workspace).

This PR contains two different commits: the first one makes lerna not fail when someone uses the `workspace` protocol (which is what made me open this PR).

The second one implements a `--local-dependencies` option which makes `lerna` actually use the `workspace` protocol. I'm not sure that this is needed since people would probably directly enable the `useWorkspaces` option and delegate this task to yarn, but I added the commit just in case you could be interested in it. I would have no problems with dropping it.

Please don't review them as a single unit :pray:

## Motivation and Context

I didn't open an issue about it, but this is currently blocking Babel from upgrading to yarn 2 (https://github.com/babel/babel/pull/11142).

## How Has This Been Tested?

I added tests to the affected packages.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. (I will do it depending on what gets accepted)

### DIFF
--- a/commands/link/__tests__/__fixtures__/explicit-local/lerna.json
+++ b/commands/link/__tests__/__fixtures__/explicit-local/lerna.json
@@ -1,0 +1,8 @@
+{
+  "commands": {
+    "link": {
+      "localDependencies": "explicit"
+    }
+  },
+  "version": "1.0.0"
+}
--- a/core/package-graph/index.js
+++ b/core/package-graph/index.js
@@ -13,11 +13,19 @@ const { reportCycles } = require("./lib/report-cycles");
  * @param {String} graphType ("allDependencies" or "dependencies")
  *    Pass "dependencies" to create a graph of only dependencies,
  *    excluding the devDependencies that would normally be included.
- * @param {Boolean} forceLocal Force all local dependencies to be linked.
  * @param {String} localDependencies Force all local dependencies to be linked.
  */
 class PackageGraph extends Map {
-  constructor(packages, graphType = "allDependencies", forceLocal) {
+  constructor(
+    packages,
+    graphType = "allDependencies",
+    localDependencies = "auto" // | "force" | "explicit"
+  ) {
+    // For backward compatibility
+    if (localDependencies === true || localDependencies === "forceLocal") {
+      localDependencies = "force"; // eslint-disable-line
+    }
+
     super(packages.map(pkg => [pkg.name, new PackageGraphNode(pkg)]));
 
     if (packages.length !== this.size) {
@@ -75,8 +83,13 @@ class PackageGraph extends Map {
           return currentNode.externalDependencies.set(depName, resolved);
         }
 
-        if (forceLocal || resolved.fetchSpec === depNode.location || depNode.satisfies(resolved)) {
-          // a local file: specifier OR a matching semver
+        if (
+          explicitWorkspace ||
+          localDependencies === "force" ||
+          resolved.fetchSpec === depNode.location ||
+          (localDependencies !== "explicit" && depNode.satisfies(resolved))
+        ) {
+          // a local file: specifier, a matching semver or a workspace: version
           currentNode.localDependencies.set(depName, resolved);
           depNode.localDependents.set(currentName, currentNode);
         } else {
--- a/core/package-graph/index.js
+++ b/core/package-graph/index.js
@@ -14,6 +14,7 @@ const { reportCycles } = require("./lib/report-cycles");
  *    Pass "dependencies" to create a graph of only dependencies,
  *    excluding the devDependencies that would normally be included.
  * @param {Boolean} forceLocal Force all local dependencies to be linked.
+ * @param {String} localDependencies Force all local dependencies to be linked.
  */
 class PackageGraph extends Map {
   constructor(packages, graphType = "allDependencies", forceLocal) {
@@ -57,8 +58,17 @@ class PackageGraph extends Map {
         // Yarn decided to ignore https://github.com/npm/npm/pull/15900 and implemented "link:"
         // As they apparently have no intention of being compatible, we have to do it for them.
         // @see https://github.com/yarnpkg/yarn/issues/4212
-        const spec = graphDependencies[depName].replace(/^link:/, "file:");
+        let spec = graphDependencies[depName].replace(/^link:/, "file:");
+
+        // npa doesn't support the explicit workspace: protocol, supported by
+        // pnpm and Yarn.
+        const explicitWorkspace = /^workspace:/.test(spec);
+        if (explicitWorkspace) {
+          spec = spec.replace(/^workspace:/, "");
+        }
+
         const resolved = npa.resolve(depName, spec, currentNode.location);
+        resolved.explicitWorkspace = explicitWorkspace;
 
         if (!depNode) {
           // it's an external dependency, store the resolution and bail
--- a/core/package/__tests__/core-package.test.js
+++ b/core/package/__tests__/core-package.test.js
@@ -7,6 +7,7 @@ const os = require("os");
 const path = require("path");
 const loadJsonFile = require("load-json-file");
 const writePkg = require("write-pkg");
+const npa = require("npm-package-arg");
 
 // file under test
 const Package = require("..");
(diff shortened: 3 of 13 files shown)


core/package/__tests__/core-package.test.js
```diff
@@ -273,6 +274,37 @@ describe("Package", () => {
       );
     });
   });
+
+  describe(".updateLocalDependency()", () => {
+    it("works with workspace: protocols", () => {
```
evocateur, line 279: I didn't have any direct tests of this method here because it is repeatedly tested in the publish and version tests. This is good to have, though.

core/package/__tests__/core-package.test.js
```diff
@@ -273,6 +274,37 @@ describe("Package", () => {
...
+
+      const resolved = npa.resolve("a", "^1.0.0", ".");
+      resolved.explicitWorkspace = true;
+
+      pkg.updateLocalDependency(resolved, "2.0.0", "^");
+
+      expect(pkg.toJSON()).toMatchInlineSnapshot(`
+        Object {
+          "dependencies": Object {
+            "a": "workspace:^2.0.0",
```
evocateur, line 298: One wrinkle, however, is that _sometimes_ `pkg.updateLocalDependency()` is intended to overwrite any protocol (such as `file:`) with the resolved version. Existing code has cheesy conditionals that exclude the call when the `resolved.type === "directory"`, so we're gonna have to make this method smarter somehow (or split the intent into two methods...)

core/package/index.js
```diff
@@ -209,6 +209,9 @@ class Package {
     if (resolved.registry || resolved.type === "directory") {
       // a version (1.2.3) OR range (^1.2.3) OR directory (file:../foo-pkg)
       depCollection[depName] = `${savePrefix}${depVersion}`;
+      if (resolved.explicitWorkspace) {
+        depCollection[depName] = `workspace:${depCollection[depName]}`;
```
evocateur, line 213: As I mentioned in the newly-added test for this method, this isn't always the case (thinking specifically of prepublish resolution of `${protocol}:` versions)...

utils/query-graph/query-graph.js
```diff
@@ -16,14 +18,15 @@ class QueryGraph {
...
+   * @param {String} [opts.localDependencies="auto"] Valid values are "force" or "explicit"
    * @param {Boolean} [opts.rejectCycles] Whether or not to reject cycles
    * @constructor
    */
   constructor(packages, opts) {
     const options = QueryGraphConfig(opts);
 
     // Create dependency graph
-    this.graph = new PackageGraph(packages, options.graphType);
+    this.graph = new PackageGraph(packages, options.graphType, options.localDependencies);
```
evocateur, line 29: [run-topologically](https://github.com/lerna/lerna/blob/master/utils/run-topologically/run-topologically.js#L33) also needs this option passed through, as unfortunately most of its callsites are passing custom config, not the whole command-resolved `options` object...

commands/link/__tests__/__fixtures__/explicit-local/lerna.json
```diff
@@ -0,0 +1,8 @@
+{
+  "commands": {
+    "link": {
+      "localDependencies": "explicit"
```
evocateur, line 4: We're missing CLI option config for `--local-dependencies`, as well as README documentation. It's especially burdensome for an option like this because it is such a _broad_ thing. Perhaps instead of a flag, we need to centralize the "am i using a custom protocol" logic better?
* https://github.com/lerna/lerna/blob/448f2aee7258febc15c131c1128688326a52778f/commands/publish/index.js#L519-L522
* https://github.com/lerna/lerna/blob/448f2aee7258febc15c131c1128688326a52778f/commands/version/index.js#L522-L524
* https://github.com/lerna/lerna/blob/448f2aee7258febc15c131c1128688326a52778f/commands/bootstrap/index.js#L240-L248

After all, in an ideal world, the `workspace:` protocol would "just work", similar to how `file:`/`link:` is handled today.

ChiefORZ:
why isn't this feature getting any more attention? we are waiting hard for this one...

nicolo-ribaudo:
@ChiefORZ I won't have time to work on this PR for 1-2 months. If you want to take my PR and address the review comments it would be appreciated :pray:

arcanis:
@evocateur if by any chance you saw a way to land this PR in any way I'd appreciate - it's currently difficult for third-party projects to use the `workspace:` protocol without having to drop Lerna, which makes the adoption ever so slightly harder for us 🙁

Carl-Foster:
I'm happy to take over this PR and address the comments.

Carl-Foster:
I've put some time into this PR and trying to add tests to address the comments however I'm finding it hard to navigate. I feel like `lerna` needs to detect if `yarn berry` is being used and transparently pass through most of the logic down to the `@yarnpkg/core` package. This would require a fairly significant rework of the implementation and I don't know enough about the architecture to make that decision.

okikio:
Can someone please merge this pull request, I have been waiting for this feature for literally 9 months....

nicolo-ribaudo:
@okikio I don't have time to work on this PR anymore, but @evocateur already reviewed it (https://github.com/lerna/lerna/pull/2450#pullrequestreview-417310713).
If you want to cherry-pick my commits, handle @evocateur's review and open a new PR, that would probably help a lot moving this forward.

okikio:
> @okikio I don't have time to work on this PR anymore, but @evocateur already reviewed it ([#2450 (review)](https://github.com/lerna/lerna/pull/2450#pullrequestreview-417310713)).
> If you want to cherry-pick my commits, handle @evocateur's review and open a new PR, that would probably help a lot moving this forward.

I thought the PR was finalized and just not yet merged. Ok, I'll open a new PR but how do I get @evocateur to review.

okikio:
@nicolo-ribaudo I have gone through your commits, I don't think they would adequately accomplish the task required. When discussing the workspace protocol, most devs using lerna would most likely want it to act like if the workspaces protocol wasn't used. By default, lerna can't set packages to use the workspace protocol, so why bother building it into lerna, why not just let users continue with the tools they were using prior, so, for example, yarn and pnpm, if devs are already using these tools it would be less work to implement, all that would be required would be to just add an option to ignore the workspace protocol and let either yarn or pnpm handle the publishing, so, lerna would handle versioning and changelog generation. What do you think @evocateur does this sound good?

core/package-graph/index.js
```diff
@@ -57,16 +66,30 @@ class PackageGraph extends Map {
...
         // @see https://github.com/yarnpkg/yarn/issues/4212
-        const spec = graphDependencies[depName].replace(/^link:/, "file:");
+        let spec = graphDependencies[depName].replace(/^link:/, "file:");
+
+        // npa doesn't support the explicit workspace: protocol, supported by
+        // pnpm and Yarn.
+        const explicitWorkspace = /^workspace:/.test(spec);
+        if (explicitWorkspace) {
+          spec = spec.replace(/^workspace:/, "");
+        }
```
matzeeable, line 76: Just a sidenote: The same should be done for the [`:patch`](https://yarnpkg.com/features/protocols#patch) protocol.

merceyz: That one can't really be replaced

evocateur:
I've rebased this onto the latest in main: https://github.com/lerna/lerna/compare/test/workspace-protocol

I'm going to do some local validation, maybe write some more tests. We're still lacking option config for `--local-dependencies` (and documentation), so this still isn't ready to go as-is.

I can't promise any timeline, but this is increasing in priority for me.

brianespinosa:
For anyone else still needing this, it seems like the right approach might just be implementing publishing features yourself on top of yarn as they are doing in the [above referenced PR](https://github.com/storybookjs/storybook/pull/15413) from Storybook.

ghiscoding:
hey everyone happy to share that `workspace:` protocol is now supported in Lerna-Lite with this new release [1.2.0](https://github.com/ghiscoding/lerna-lite/releases/tag/v1.2.0). 
Give it a try and happy coding 🎉

fahslaj:
To all it may concern, this PR is being closed for clarity since Lerna already supports the `workspace:` protocol. It was added a while ago as part of the `pnpm` support included in Lerna 6.

Lerna recognizes the `workspace:` prefix in Yarn 1, Yarn 2+ (with node_modules linker) and pnpm workspaces and correctly converts them to the appropriate dependency spec when publishing.

Thank you to everyone for your participation in the Lerna community!

binaryunary:
Does this mean #2564 can be closed?

JamesHenry:
Thanks for the reminder @binaryunary, yes!

cumt-robin:
Couldn't find any versions for "@jview/utils" that matches "workspace:*"

I dont know why...

it's lerna v7.

image

JamesHenry:
@cumt-robin You are using yarn v1/classic which never supported the workspace protocol. If you want to use it you migrate to latest yarn v3+

cumt-robin:
您的邮件我已收到！  我会在第一时间内回复您！非常感谢！

PR #2459, opened 2020-02-19
[feat] Provide with support for rate limited otps

antoine-malliarakis:
Add automatic wait-and-regenerate mechanism for rate limited otp.

## Description
The `publish` command now automatically wait and regenerates otp whenever rate limit errors are being seen in order to support 2FA with large number of public packages.

## Motivation and Context

We have a repository with around 70 public packages and whenever a "core" package is updated we, obviously, publish 70 new artifacts. It turns out that with the 2FA mechanism being used through otp we have rate limits of around 4/5 artifacts "at a time" which means that a typical scenario would be to do:

1. `lerna publish -y --otp=test` => failure after having published 4 artifacts (recompiled + run tests on all 70 artifacts)
2. `git restore`
3. `lerna publish from-package -y --otp=test` => failure after having published 4 artifacts (recompiled + run tests on 66 artifacts)
4. repeat 2 & 3 until there is nothing more to publish

Here the idea was to basically provide with a mechanism which would, when a rate limit error is being raised, request for another otp after having waited for some time configured through command line.

## How Has This Been Tested?

Manually, by running `lerna publish --otp=X --otp-rate-limit-delay=5` on a lerna enabled repository plugged on npmjs.

## Types of changes

## Checklist:


commands/publish/__tests__/publish-command.test.js
```diff
@@ -239,6 +241,24 @@ Map {
...
+
+    it("Passes otp delay if specified.", async () => {
+      const testDir = await initFixture("normal");
+
+      const otp = "123456";
+      npmPublish.mockImplementation(() => Promise.resolve());
+
+      getTwoFactorAuthRequired.mockResolvedValueOnce(true);
+
+      await lernaPublish(testDir)("--otp", otp, "--otp-rate-limit-delay", 5e-5);
```
antoine-malliarakis: 5e-5 is actually 3 milliseconds.

commands/publish/command.js
```diff
@@ -68,6 +68,13 @@ exports.builder = yargs => {
       type: "string",
       requiresArg: true,
     },
+    "otp-rate-limit-delay": {
```
antoine-malliarakis: I tried to find a better name but I must confess I was at a loss here.

commands/publish/index.js
```diff
@@ -126,7 +126,15 @@ class PublishCommand extends Command {
     });
 
     // cache to hold a one-time-password across publishes
-    this.otpCache = { otp: this.conf.get("otp") };
+    const otpRateLimitDelayCliParameter = this.options["otp-rate-limit-delay"];
```
evocateur, line 129: Yargs automatically camel-cases this to `otpRateLimitDelay` for us. Also, it's not _only_ provided by the CLI, as it can also be configured in lerna.json (`command.publish.otpRateLimitDelay`).

commands/publish/__tests__/publish-command.test.js
```diff
@@ -239,6 +241,24 @@ Map {
       );
       expect(otplease.getOneTimePassword).toHaveBeenLastCalledWith("Enter OTP:");
     });
+
+    it("Passes otp delay if specified.", async () => {
+      const testDir = await initFixture("normal");
+
+      const otp = "123456";
+      npmPublish.mockImplementation(() => Promise.resolve());
```
evocateur, line 249: This shouldn't be necessary? It's already [manually mocked](https://github.com/lerna/lerna/blob/master/commands/__mocks__/@lerna/npm-publish.js).

commands/publish/__tests__/command.test.js
```diff
@@ -0,0 +1,23 @@
+"use strict";
+
+const command = require("../command");
+const yargs = require("yargs");
```
evocateur, line 4: We're not testing yargs, please remove this file. The test added to publish-command.test.js is sufficient to verify the option config was recognized.

core/otplease/otplease.js
```diff
@@ -66,7 +88,18 @@ function attempt(fn, opts, otpCache) {
           semaphore.release();
           return attempt(fn, opts.concat({ otp: otpCache.otp }), otpCache);
         }
-        return getOneTimePassword()
+        let basePromise = Promise.resolve();
+        let promptForPassword = () => getOneTimePassword();
+        if (rateLimitedOtp) {
+          basePromise = new Promise(resolve => {
+            const delay = opts.otpRateLimitDelay;
```
evocateur, line 95: Since we know it's rate limited, couldn't we prompt for a timeout period, perhaps providing a default value that would work fine in most cases? I'm concerned that folks may not always know that they might be rate-limited, as it might vary depending on the changeset. An explicitly-configured `--otp-rate-limit-delay` (or `command.publish.otpRateLimitDelay` in lerna.json) would skip this prompt and just start the timeout.

core/otplease/__tests__/otplease.test.js
```diff
@@ -69,6 +71,22 @@ describe("@lerna/otplease", () => {
     expect(otpCache.otp).toBe("654321");
   });
 
+  it("request new otp if cache value is blocked because of rate limits.", async () => {
```
evocateur, line 74: Excellent tests, thank you.

commands/publish/command.js
evocateur: It's an awkward concept to express, I think you've done well enough. :)

core/otplease/otplease.js
antoine-malliarakis: - [ ] Add prompt for otp-rate-limite-delay if option is not passed *and* the error reported is related to rate limitations (with a default value being 5 minutes ?)

commands/publish/__tests__/command.test.js
antoine-malliarakis: - [ ] Remove useless file

core/otplease/otplease.js
antoine-malliarakis: While implementing the prompt I stumbled upon a tricky usability open question.

Since all of this may be in concurrent cases, I've already been facing the case where

* a first "thread" faces an OTP rate limit related error, starts locking and says he'll be waiting
* other threads continue their work until their reach the same stage (emitting all corresponding logs)
* the first thread which was stuck "waiting" now prompts for the otp, once it's done it releases its lock and everybody resumes

Implementing the prompt would mean that we'd have 

* a user being asked for some time value while other threads' logging would be stuck (due to the pausing)
* once the user enters a value all other threads' logging would resume (which could result in a *lot* of logs ;) ) and its wait would start (obviously I'm guessing the time reference would be before the lock is obtained, right ?)

Would that seem right ? (This is really an open question)

JamesHenry:

If you or anybody else would like to implement this functionality against the codebase as it stands to day then we would grateful accept and review a PR.

Thanks again!

Issue #1283, opened 2018-02-23
Document all the options that can be configured on lerna.json


danielo515:
Hello.

I've been using lerna for almost a year, and I'm really enjoying all the features that it provides which deserves suffering some the not-so-importang bugs.
However an experience that is always a nightmare is reading the documentation of lerna.json.
The information is spread across the entire readme, presenting partials/parts of each section on different places. There is a discrete line instructing that you can configure any parameter lerna accepts as a command on the lerna.json file, and there is a (at least to me) meaningless example of how to do it. In practice however, some options that are supposed to be supported are not supported at all, and some options that you will never thing that can be configured there can. Also some options should be configured with the same name, other needs a different syntax and so on. 
Figuring out where each option should be placed is also very hard, and leads to enormous amount of confusion. I several times ended adding the options to the call of the lerna command (which is way worse) just to not have to investigate how to configure that on lerna.json.

I know that I am not alone because there are plenty of questions about this here:

- #1064 
- https://github.com/lerna/lerna/issues/1064#issuecomment-348826267
- #1165

Here is an example of something that @evocateur has pointed out that I would never figured out:

```js
{
  "lerna": "2.9.0",
  "command": {
    "bootstrap": {
      "npmClientArgs": ["--no-package-lock"]
    }
  },
  "version": "1.2.3"
}
```

Before that I tried `package-lock: false`, `no-package-lock:true` and some other crazy ideas until  I got frustrated enough to open an issue.

My point is, that the documentation I'm requesting has been mostly already wrote, but it is burried on issues, people asking questions, some stack overflow question here and there...  I really think that documenting this all once in one place will reduce the number of questions here and people reading the documentation dozens of times.

 Thanks and regards

evocateur:
I completely agree! Would you be willing to drive this documentation refactoring? I was hoping to get to something like this, but I’ve got a lot of things on my plate (don’t we all, heh). I will make an effort to be responsive to anything you can find the time to contribute.

danielo515:
Yes of course!
How would you to like to approach this? Just adding a small concise section about it or a complete refactor, removing old references and planning  everything with a different style.
To be honest I think the first approach is more realistic :smile:

evocateur:
Practically speaking, the first approach is the only approach. 😭

Let's start with the small, concise section (with maybe some deletions of *really* bad old examples). Thanks!

light24bulbs:
[...]

PR #2465, opened 2020-02-25
Add JSON schema for lerna.json, built from declaration file

cspotcode:
## Description

Generates a JSON schema for lerna.json, providing tab completion and inline documentation, examples, and default values in modern code editors.

## Motivation and Context

Addresses #1283

## How Has This Been Tested?

Manually, by using it in lerna's own lerna.json.  The schema is incomplete, and currently is a proof-of-concept.  I'm hoping others will be interested and help to complete it.

## Types of changes

## Checklist:

### DIFF
--- a/lerna-config.d.ts
+++ b/lerna-config.d.ts
@@ -1,0 +1,101 @@
+/** lerna.json config file */
+export interface LernaConfig extends RootAndPerCommandOptions, IgnoreChangesOption {
+  /** the current version of the repository */
+  version: string;
+
+  /**
+   * Per-command options.
+   */
+  command?: {
+    bootstrap?: BootstrapCommand;
+    create?: CreateCommand;
+    publish?: PublishCommand;
+    version?: VersionCommand;
+  };
+
+  /**
+   * Array of globs to use as package locations.
+   */
+  packages?: string[];
+
+  /**
+   * Originally used to indicate the current version of Lerna. Made obsolete and removed in v3
+   * @deprecated
+   */
+  lerna?: string;
+}
+
+/**
+ * Commands
+ */
+
+interface Command extends RootAndPerCommandOptions {}
+
+interface BootstrapCommand extends Command {
+      /**
+       * an array of globs that won't be bootstrapped when running the lerna bootstrap command.
+       */
+      ignore?: string;
+      /**
+       * array of strings that will be passed as arguments directly to npm install during the lerna bootstrap command.
+       */
+      npmClientArgs?: string[];
+      /**
+       * an array of globs that restricts which packages will be bootstrapped when running the lerna bootstrap command.
+       */
+      scope?: string[];
+}
+
+interface CreateCommand extends Command {
+  homepage?: string;
+  license?: string;
+}
+
+interface PublishCommand extends Command, VersionMessageOption, IgnoreChangesOption {
+  /**
+   * use it to set a custom registry url to publish to instead of npmjs.org, you must already be authenticated if
+   * required.
+   */
+  registry?: string;
+}
+
+interface VersionCommand extends Command, VersionMessageOption {
+  allowBranch?: string;
+  conventionalCommits?: boolean;
+  exact?: boolean;
+  gitRemote?: string;
+}
+
+/*
+ * Options
+ *
+ * Declared as interfaces, then mixed-in the `*Command` interfaces above.
+ * Like JSON schema's "allOf" functionality.
+ */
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -193,110 +193,6 @@
         }
       }
     },
-    "@evocateur/libnpmaccess": {
```
evocateur, line 196: There are way too many removals from this lockfile.

lerna.schema.json
```diff
@@ -0,0 +1,170 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
evocateur, line 2: This is a generated schema, yes? Why does it need to be committed?

evocateur: Does it need to be published somewhere?

lerna-config.d.ts
```diff
@@ -0,0 +1,101 @@
+/** lerna.json config file */
+export interface LernaConfig extends RootAndPerCommandOptions, IgnoreChangesOption {
```
evocateur, line 2: This is very interesting, but I'm a little leery of an external declaration file separate from the code that defines these options. Is it possible to transform Yargs config into type declarations, for instance? That would go a long way toward making this more maintainable.

lerna.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "./lerna.schema.json",
```
evocateur, line 2: How do others outside of this repo consume the schema?

package.json
```diff
@@ -10,7 +10,9 @@
     "integration": "jest --config jest.integration.js --maxWorkers=2",
     "lint": "eslint . -c .eslintrc.yaml --no-eslintrc --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint",
     "pretest": "npm run lint",
-    "test": "jest"
+    "test": "jest",
+    "build": "npm run build:schema",
```
evocateur, line 14: When does the build need to run? `preversion`?

As it is now, this will never be run by any automated process, and humans (me) will inevitably forget to run it.

lerna.schema.json
cspotcode: It's been a little while so I don't remember, but I think I did this for code review purposes.

lerna-config.d.ts
cspotcode: Ideally yargs comes up with a way to declare yargs config as decorators on an interface. (on a class)  They're finally porting to TypeScript so they might do something like this.

lerna.json
cspotcode: Via relative path or URL: "./node_modules/lerna/leran.schema.json" or "unpkg.com/..."

package.json
cspotcode: Not sure where it should go exactly.  The schema should be bundled into the published tarball but it's not needed by the tests.  Could go in a `prepare` script.

cspotcode:
@evocateur I responded to all the comments.  Unfortunately I don't have time allocated to work on this.  But I think collaboration is switched on, so others can push to this branch.

package-lock.json
cspotcode: I don't know what the removals are.  Probably someone can `git checkout origin/master -- package-lock.json && npm install` to clean that up.

evocateur: In my experience it's usually because of a manual addition to a root package.json that didn't already have a `node_modules` installed. Or possibly the other way around. Anyway, fairly easy to fix.

lerna-config.d.ts
evocateur: Yes, that would be nice.

package.json
evocateur: Seems like the schema should be located under the `core/lerna` directory, then, generated as a `prepublishOnly` script.

lerna-config.d.ts
kylemh: If I wanted to push this over the finish line, it seems like this might be the only hold up.

I definitely get not wanting to lose parity between code and this typing... Is there file or folder of examples I could run a type-check against to test for parity?

JamesHenry:
Hi @cspotcode thanks a lot for contributing to lerna and its community!

We actually added a schema for lerna.json already, so this one can be closed.

You can leverage it by setting:

```json
"$schema": "node_modules/lerna/schemas/lerna-schema.json",
```

In your **lerna.json**

Many thanks 🙏

PR #2473, opened 2020-02-29
fix(create-symlink): Replace @zkochan/cmd-shim with cmd-shim

### LINKED ISSUE #2472: Symlinks for binaries do not work for yarn users on windows
The lerna release v3.16.2 introduces a bug for yarn users on windows. Symlinks for binaries (in node_modules/.bin) are no longer created. This bug is not present in v3.16.1 but exists in all newer versions.

Interestingly this behavior only occurs for yarn users, it works fine with npm.
Please see my minimal reproducible example at  https://github.com/TimHolzherr/reproduce-lerna-bug

## Expected Behavior
- Lerna creates symlinks for binaries also for yarn users on windows.
- [minimal reproducible example](https://github.com/TimHolzherr/reproduce-lerna-bug) succeeds

## Current Behavior
- Lerna does not create symlinks for binaries for yarn users on windows.
- [minimal reproducible example](https://github.com/TimHolzherr/reproduce-lerna-bug) fails

## Possible Solution
- It is likely that this regression was introduced either with commit [c3f7998](https://github.com/lerna/lerna/commit/c3f7998) or with [60d1100](https://github.com/lerna/lerna/commit/60d1100)

1. git clone https://github.com/TimHolzherr/reproduce-lerna-bug.git 
2. yarn && yarn bootstrap && yarn test-binary-symlink

```json
{
  "packages": ["bar", "foo"],
  "npmClient": "yarn",
  "version": "0.0.0"
}
```
</p></details>

## Your Environment
- yarn@1.22.0
- node@v12.13.0
- Windows 10 (19041.84)

TimHolzherr:
## Description
There seems to be an issue for windows users with @zkochan/cmd-shim.
The issue is described in #2472.
In commit [60d1100](https://github.com/lerna/lerna/commit/60d1100) cmd-shim@2.0.2 was replaced with @zkochan/cmd-shim because cmd-shim was no longer maintained.
In the meantime cmd-shim is again actively maintained and does not have the above mentioned issue.
I thus suggest to switch back to cmd-shim@4.0.1

## Motivation and Context
Fixes #2472

## How Has This Been Tested?
- Tested in minimal reproducible example from #2472 on windows with yarn and npm.
- Unit and integration tests passed locally

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -931,7 +931,7 @@
     "@lerna/create-symlink": {
       "version": "file:utils/create-symlink",
       "requires": {
-        "@zkochan/cmd-shim": "^3.1.0",
+        "cmd-shim": "^4.0.1",
         "fs-extra": "^8.1.0",
         "npmlog": "^4.1.2"
       }
@@ -1648,6 +1648,7 @@
       "version": "3.1.0",
       "resolved": "https://registry.npmjs.org/@zkochan/cmd-shim/-/cmd-shim-3.1.0.tgz",
       "integrity": "sha512-o8l0+x7C7sMZU3v9GuJIAU10qQLtwR1dtRQIOmlNMtyaqhmpXOzx1HWiYoWfmmf9HHZoAkXpc9TM9PQYF9d4Jg==",
+      "dev": true,
       "requires": {
         "is-windows": "^1.0.0",
         "mkdirp-promise": "^5.0.1",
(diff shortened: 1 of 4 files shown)


TimHolzherr:
Hi @evocateur, thanks for responding so fast to my issue #2472.
In my opinion the easiest solution to the problem would be to just ditch @zkochan/cmd-shim and go back to cmd-shim but if you have another idea I am open for suggestions. :smiley:

bponomarenko:
There is another motivation to do so – currently used `"@zkochan/cmd-shim": "^3.1.0"` depends on `mkdirp-promise`, which have been deprecated.

evocateur:
[v4.0.0](https://github.com/lerna/lerna/releases/tag/v4.0.0) (now available as `lerna@next`) should resolve this, as we switched to the latest `cmd-shim`. Apologies for the delay, thanks for the PR!

Issue #1195, opened 2018-01-13
[conventional-commits] Add a flag to disable conventional-recommended-bump


apapko:
Context: lerna publish --conventional-commits

As a developer i would like to have an ability to disable _conventional-recommended-bump_ when using --conventional-commits flag during release.
For less-mature projects, it is beneficial to have control over how release versions get incremented. However, lerna _--conventional-commits_ uses _conventional-recommended-bump_ by default to increment release version. It would be great to have an ability to disable it and use interactive version prompt instead ( as it is now when --conventional-commits are not used ).

## Expected Behavior
Allow to run --conventional-commits without recommended-bump ( interactive version prompt instead )
example:
```
lerna publish --conventional-commtis --disable-recommended-bump
```
or perhaps this:
``` 
lerna publish --conventional-commit --version-prompt
```

evocateur:
Why bother with `--conventional-commits` at all if you don't want it to compute the bump?

Incidentally, you can pass `--cd-version <semver-keyword>` (using some unstable-sounding keyword like "preminor" or "prepatch" or "prerelease") to pre-empt the angular fix/feat/breaking calculation. The changelog generation should still work, if that's the only reason you're using `--conventional-commits` in this context.

| Starting Version | `lerna publish` Flags | Result |
| ---------------- | --------------------- | ------ |
| v0.1.0 | `--cd-version preminor --preid beta` |  v0.2.0-beta.0 |
| v0.2.0-alpha.0 | `--cd-version prerelease` | v0.2.0-alpha.1 |

diegohaz:
@evocateur Please, correct me if I'm wrong. The problem I see with your suggestion is that whether you publish all packages as `alpha` or none of them.

In https://github.com/reakit/reakit I wanted to publish only one of my packages as alpha. My solution was to [patch](https://github.com/reakit/reakit/blob/next/patches/%40lerna%2Bversion%2B3.13.1.patch) `@lerna/version` to disable the recommended version bump.

It seems to work fine, but it would be better if lerna supported it.

PR #2486, opened 2020-03-09
feat: preserve major version zero on breaking changes (when using conventional commits)

bajtos:
## Description

According to [semver](https://semver.org), major version zero (`0.y.z`) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable. The version `1.0.0` defines the (initial stable) public API.

In my opinion formed by my experience building https://github.com/strongloop/loopback-next, in order to allow monorepos to use major version zero meaningfully, the transition from `0.x` to `1.x` must be explicitly requested by the user. Breaking changes MUST NOT automatically bump the major version from `0.x` to `1.x`.

The usual convention is to use semver-patch bumps for bugfix releases and semver-minor for everything else, including breaking changes. This matches the behavior of `^` operator as implemented by `npm` (see [docs of the semver module](https://www.npmjs.com/package/semver#caret-ranges-123-025-004)).

This commit implements the convention described above for packages in major version zero:
- a patch-level change bumps semver-patch version (NO CHANGE)
- a minor-level change bumps semver-minor version (NO CHANGE)
- a major-level (breaking) change bumps semver-minor version (NEW)

## Motivation and Context

In LoopBack, we have been several times bitten by `lerna version` bumping up our packages from 0.x to 1.0 (see https://github.com/strongloop/loopback-next/issues/1068). Most recently, we dropped support for Node.js 8.x because it reached end of life. We described this change as breaking, and so in the next release, lerna wanted to graduate our experimental packages with unstable API to stable 1.0. That was not desired at all! 

Related discussions and alternative solutions:

- https://github.com/lerna/lerna/issues/1195 _[conventional-commits] Add a flag to disable conventional-recommended-bump_
- https://github.com/conventional-changelog/conventional-changelog/issues/239 _Do not bump to 1.0.0 when issuing a breaking change < 1.0.0_
- https://github.com/conventional-changelog/conventional-changelog/issues/294 _Recommend different bumping strategy for 0.x versions_
- https://github.com/npm/node-semver/issues/177 _Incorrect Increment for versions < 1.0_ (rejected)
- https://www.npmjs.com/package/shifted-semver-increment (no longer maintained)

## How Has This Been Tested?

I added a new unit test, verified that it fails against the current implementation and passes with my patch applied.

I am not sure if a single test is enough for a change like this. Please suggest what other scenarios to cover by tests to give us enough confidence, I am happy to add more tests as needed.

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation. -- I am not sure about this?
- [ ] I have updated the documentation accordingly. -- I think no doc changes are needed.
- [ ] All new and existing tests passed. (`lerna bootstrap --npm-client yarn` is failing for me, the error seems unrelated to my changes.)

### DIFF
--- a/core/conventional-commits/lib/recommend-version.js
+++ b/core/conventional-commits/lib/recommend-version.js
@@ -51,7 +51,7 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath, tagPrefix, pre
 
         // result might be undefined because some presets are not consistent with angular
         // we still need to bump _something_ because lerna saw a change here
-        const releaseType = data.releaseType || "patch";
+        let releaseType = data.releaseType || "patch";
 
         if (prereleaseId) {
           const shouldBump = shouldBumpPrerelease(releaseType, pkg.version);
@@ -59,6 +59,26 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath, tagPrefix, pre
           log.verbose(type, "increment %s by %s", pkg.version, prereleaseType);
           resolve(semver.inc(pkg.version, prereleaseType, prereleaseId));
         } else {
+          if (semver.major(pkg.version) === 0) {
+            // According to semver, major version zero (0.y.z) is for initial
+            // development. Anything MAY change at any time. The public API
+            // SHOULD NOT be considered stable. The version 1.0.0 defines
+            // the (initial stable) public API.
+            //
+            // To allow monorepos to use major version zero meaningfully,
+            // the transition from 0.x to 1.x must be explicitly requested
+            // by the user. Breaking changes MUST NOT automatically bump
+            // the major version from 0.x to 1.x.
+            //
+            // The usual convention is to use semver-patch bumps for bugfix
+            // releases and semver-minor for everything else, including
+            // breaking changes. This matches the behavior of `^` operator
+            // as implemented by `npm`.
+            //
+            if (releaseType === "major") {
+              releaseType = "minor";
+            }
+          }
           log.verbose(type, "increment %s by %s", pkg.version, releaseType);
           resolve(semver.inc(pkg.version, releaseType));
         }
--- a/core/conventional-commits/__tests__/__fixtures__/major-zero/lerna.json
+++ b/core/conventional-commits/__tests__/__fixtures__/major-zero/lerna.json
@@ -1,0 +1,8 @@
+{
+  "command": {
+    "publish": {
+      "conventionalCommits": true
+    }
+  },
+  "version": "independent"
+}
--- a/core/conventional-commits/__tests__/__fixtures__/major-zero/package.json
+++ b/core/conventional-commits/__tests__/__fixtures__/major-zero/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "conventional-commits-major-zero",
+  "repository": "lerna/conventional-commits-major-zero",
+  "version": "0.0.0-root"
+}
--- a/core/conventional-commits/__tests__/__fixtures__/major-zero/packages/package-0/package.json
+++ b/core/conventional-commits/__tests__/__fixtures__/major-zero/packages/package-0/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "package-0",
+  "repository": "lerna/conventional-commits-major-zero",
+  "version": "0.1.0"
+}
--- a/core/conventional-commits/__tests__/__fixtures__/major-zero/packages/package-1/package.json
+++ b/core/conventional-commits/__tests__/__fixtures__/major-zero/packages/package-1/package.json
@@ -1,0 +1,5 @@
+{
+  "name": "package-1",
+  "repository": "lerna/conventional-commits-major-zero",
+  "version": "1.0.0"
+}
--- a/core/conventional-commits/__tests__/conventional-commits.test.js
+++ b/core/conventional-commits/__tests__/conventional-commits.test.js
@@ -217,6 +217,21 @@ describe("conventional-commits", () => {
       ).rejects.toThrow("Unable to load conventional-changelog preset 'conventional-changelog-garbage/pail'");
     });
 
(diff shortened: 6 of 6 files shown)


bajtos:
Discussion points:

- At high-level, is it acceptable to have different bumping behavior for 0.x versions implemented inside lerna?
- Should we introduce a new CLI option allowing users to choose between the old or the new behavior? We already have similar options `--conventional-prerelease` and `--conventional-graduate`, see https://github.com/lerna/lerna/tree/master/commands/version#prerelease
- How and where to document the new behavior (and the new option if it's introduced)?

core/conventional-commits/lib/recommend-version.js
```diff
@@ -51,14 +51,34 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath, tagPrefix, pre
...
+            // by the user. Breaking changes MUST NOT automatically bump
+            // the major version from 0.x to 1.x.
+            //
+            // The usual convention is to use semver-patch bumps for bugfix
+            // releases and semver-minor for everything else, including
+            // breaking changes. This matches the behavior of `^` operator
+            // as implemented by `npm`.
+            //
+            if (releaseType === "major") {
+              releaseType = "minor";
```
raymondfeng: I guess if `--conventional-graduate` is `true`, we will bump the release to the new major, such as `0.1.0` -> `1.0.0`?

bajtos: Personally, I would find such rule as possibly confusing. Also, how would such rule work in large monorepos having packages both in pre-release versions (`2.0.0-beta.1`) and major zero versions (`0.2.5`)? 

I am currently envisioning the following process for graduating from 0.x to 1.0.0:

1. Add a commit changing the version in package.json to `1.0.0-1` (or a similar pre-release). This will be just a commit, no release/tag is associated with it. The commit can have a nice descriptive message for the changelog, e.g. `feat: graduate to 1.0`.

2. Run `lerna version` with `--conventional-graduate` to bump the temporary version `1.0.0-1` to final `1.0.0`.

raymondfeng: Please note `--conventional-graduate` allows a list of packages. Whether it's passed in as a CLI option or configured in `lerna.json`, we probably should honor it. See https://github.com/lerna/lerna/blob/master/commands/version/README.md#--conventional-graduate.

tpluscode:
@bajtos please refer to #2549 I've just created.

NPM treats minor change to 0.x versions as breaking, ie. will not update 0.4 to 0.5. This can be tried on https://semver.npmjs.com

Thus, for 0.x packages only breaking change should bump minor and any other change should bump patch

bajtos:
Thank you @tpluscode for the link to the issue you have created. I am confused about why are you posting this here? My pull request have been waiting for review by project maintainers for more than a month by now. I don't see how filling a new issue is going to help progress forward?

Are you perhaps trying to say that my pull request is implementing incorrect logic for determining which part of the 0.x version to bump up? If yes, then could you please be more specific about what should I change?

tpluscode:
I'm sorry, I thought you would tell me.

> The usual convention is to use semver-patch bumps for bugfix releases and semver-minor for everything else, including breaking changes

Here is where our understanding differs on incrementing 0.x packages. You say

> semver-patch bumps for bugfix releases
> semver-minor for everything else, including breaking changes

Where I think it should be

> server-minor for breaking changes
> semver-patch for anything else

I say that because npm treats minor bump on 0.x as breaking change and will not update automatically. I will however update for example `1.1 -> 1.2`.

bajtos:
Thank you @tpluscode  for clarification!

> Here is where our understanding differs on incrementing 0.x packages. You say
> 
> > semver-patch bumps for bugfix releases
> > semver-minor for everything else, including breaking changes
>
> Where I think it should be
>
> > semver-minor for breaking changes
> > semver-patch for anything else
>
> I say that because npm treats minor bump on 0.x as breaking change and will not update automatically. I will however update for example 1.1 -> 1.2.

You have raised a valid point :+1: 

Personally, I don't care that much whether new features bump semver-minor or semver-patch version. I am fine to implement whichever rule that gets widest consensus.

Anything would be better than the current behavior, where breaking changes are bumping 0.x to 1.0 :(

tpluscode:
> Anything would be better than the current behavior, where breaking changes are bumping 0.x to 1.0 :(

Agreed. This is why I create a separate issue. In the event that your PR gets merged at its current state and it'd still need tweaking to align 100% with npm's treatment of 0.x version.

Unless you'd agree to adjust right here and we can kill two birds with one stone

core/conventional-commits/lib/recommend-version.js
```diff
@@ -51,14 +51,34 @@ function recommendVersion(pkg, type, { changelogPreset, rootPath, tagPrefix, pre
...
           resolve(semver.inc(pkg.version, prereleaseType, prereleaseId));
         } else {
+          if (semver.major(pkg.version) === 0) {
+            // According to semver, major version zero (0.y.z) is for initial
+            // development. Anything MAY change at any time. The public API
+            // SHOULD NOT be considered stable. The version 1.0.0 defines
+            // the (initial stable) public API.
+            //
+            // To allow monorepos to use major version zero meaningfully,
+            // the transition from 0.x to 1.x must be explicitly requested
```
evocateur, line 69: I think some info-level logging indicating what is happening would be helpful. I appreciate the detailed comment.

core/conventional-commits/lib/recommend-version.js
evocateur: I don't see `--conventional-graduate` affecting 0.x ranges, as they are not "prereleases" in the sense that option is intended to modify. 0.x ranges, as I understand them, mean "API unstable", not "this is a prerelease of a specified quality".

evocateur:
Thanks for your patience!

bajtos:
@evocateur thank you for landing my pull request! ❤️

PR #2505, opened 2020-03-21
support only Node LTS versions or >=10.13.0

dulmandakh:
## Description
This PR bumps Node version requirements to >=10.13.0 or supported LTS.

## Motivation and Context
Node 6.x is EOL for a while now, and even dropped from Lerna CI in 2018 (https://github.com/lerna/lerna/commit/97429bbf6fe45dda3b9cd06bdb4c3d03a6f63847). Also some dependencies, for example fs-extra ^8.0.0, dropped support for it.

Node 8 is already EOL.

Supporting only LTS versions then explicitly requiring it would allow developers to bump dependencies, even remove some.

## How Has This Been Tested?
Run tests on Node 10.x and Node 12.x.

## Types of changes
- [x ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ x] All new and existing tests passed.


evocateur:
I already have a commit staged for this sort of change: https://github.com/lerna/lerna/commit/634092534e74bdd81be56a259453931f8fde77fb

There are many other things to do before we can proceed with the next major, however, chief among them ...defining those things. 😅 I would appreciate any help to that end you can contribute.

Thank you for taking the time to propose this change!

PR #2507, opened 2020-03-22
Add ignorePackages option

Marcosld:
## Description
Create a way to ignore certain packages when versioning or publishing with lerna.

## Motivation and Context
In https://github.com/lerna/lerna-changelog/issues/128#issuecomment-597693311 was mentioned if there was any way to ignore packages when versioning. I would also like to ignore some packages and not be prompted about versioning them when I run `publish` and `version`, even if these ignored packages or their dependecies have changes.

## How Has This Been Tested?
I added some tests to the project to check that my change was working. I also linked the lerna package and tested the changes locally in a dummy repository.

My changes shouldn't be affecting any other areas than `publish` and `version` commands.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -53,6 +53,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--force-publish`](#--force-publish)
 - [`--git-remote`](#--git-remote-name)
 - [`--create-release`](#--create-release-type)
+- [`--ignore-packages`](#--ignore-packages)
 - [`--ignore-changes`](#--ignore-changes)
 - [`--ignore-scripts`](#--ignore-scripts)
 - [`--include-merged-tags`](#--include-merged-tags)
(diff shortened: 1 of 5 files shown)


Codermar:
Why was this not merged yet?

Codermar:
I really need this feature... It is absolutely annoying how Lerna keeps me from publishing because it "finds" changes in a private package linked via submodule

achingbrain:
This would be really useful. Please can it be merged?

dilame:
I'm voting to merge it

c10b10:
This is very useful.

evocateur:
[v3.21.0](https://github.com/lerna/lerna/releases/tag/v3.21.0) added the [`--no-private` option](https://github.com/lerna/lerna/commit/a9b9f97457e4e4b0cac7f4ce562458d921a1f9be), is this not sufficient?

Other filtering of non-private version/publish is not going to fly, as this violates one of the core tenets of Lerna.

If a non-private package is still being published against your wishes, you have options such as making it private (`"private": true` in package.json), publishing more often so as to limit the scope of any given `lerna publish`, or removing the offending package from the monorepo entirely, as it clearly doesn't belong there if you're convinced it should not be versioned in coordination with the other monorepo siblings.

Marcosld:
`--no-private` option serves the purpose on my end :)

Codermar:
Sorry for the delay. It seems like the --no-private flag works to an extent for me.

gdlm91:
What option do I have if I'm using Lerna with yarn workspaces, but I want to release one of the packages independently from the others? 

I can't seem to find a way to tell Lerna: "use workspaces,  except for this package/pattern"

PR #2528, opened 2020-04-08
fix: do not require conventional commits for releases

### LINKED ISSUE #2521: lerna publish with pre-v1 modules
I have a monorepo with several modules, some of which are pre-v1 and others post-v1.

I want to have changelogs and GitHub releases generated so I use `createRelease: true` in my `lerna.json` which then tells me I have to use `--conventional-commits`.

When I do this, lerna tries to bump all `0.x.0` modules with `BREAKING CHANGES:` in their commit messages to `1.0.0` - is there a way to continue to release `0.x.0` modules instead and still have GitHub releases and changelogs?

The project has decided not to release `v1` of these modules for reasons out of my control.

achingbrain:
## Description

Removes the requirement to use conventional commits for releases because it prevents you from releasing 0.x.x of a module.

## Motivation and Context

Some projects (such as the one I'm working on at the moment) want to use lerna to coordinate releases across multiple modules but cannot release v1 due to reasons out of our control - there is no code requirement to use version numbers generated by the conventional commits module, a manually chosen version number works just as well.

Fixes #2521 

## How Has This Been Tested?

I did a [release](https://applitools.com/blog/wp-content/uploads/2019/05/pasted-image-0.png) of https://github.com/ipfs/js-ipfs - it prompted me to choose versions for the various packages, I did, they were published to npm, github releases and changelogs were created as expected.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -204,7 +204,7 @@ lerna version --conventional-commits --create-release github
 lerna version --conventional-commits --create-release gitlab
 ```
 
-When run with this flag, `lerna version` will create an official GitHub or GitLab release based on the changed packages. Requires `--conventional-commits` to be passed so that changelogs can be generated.
+When run with this flag, `lerna version` will create an official GitHub or GitLab release based on the changed packages.
 
 To authenticate with GitHub, the following environment variables can be defined.
 
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -79,10 +79,6 @@ class VersionCommand extends Command {
     this.createRelease = this.pushToRemote && this.options.createRelease;
     this.releaseNotes = [];
 
-    if (this.createRelease && this.options.conventionalCommits !== true) {
-      throw new ValidationError("ERELEASE", "To create a release, you must enable --conventional-commits");
-    }
-
     if (this.createRelease && this.options.changelog === false) {
       throw new ValidationError("ERELEASE", "To create a release, you cannot pass --no-changelog");
     }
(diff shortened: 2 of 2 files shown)


achingbrain:
I see lots of unrelated looking test failures, lots of:

```
    ValidationError: Branch 'master' doesn't exist in remote 'origin'.
    If this is a new branch, please make sure you push it to the remote first.
```

Might be because I'm running the tests in a fork?

achingbrain:
@evocateur do you have any interest in this feature? I'm more than willing to investigate the test failures to finish this off but could you give me a nod as to whether this will be merged?

evocateur:
@achingbrain Sorry for the delay.

I don't see this happening, chiefly because I don't understand how you proposed to read the minds of developers to create meaningful release notes _without_ `--conventional-commits` enabled. I just merged https://github.com/lerna/lerna/pull/2486, does this not cover your use case? It enables 0.x. ranges to stay 0.x despite breaking changes when using `--conventional-commits`.

The tests are failing because your proposed patch is now generating changelogs by default for every release, implicitly _depending_ on conventional commits to produce anything of value, yet not guaranteeing the intent with the required `--conventional-commits` flag.

achingbrain:
You don't have to read their minds, just their commit messages.  Maybe this PR has a bad title, the idea was to still use conventional commits, just don't try to guess the new version number.

Anyway the point is moot - it looks like #2486 solves my problem, I'll give that a go.

Ahmeed2m:
@evocateur There is no workaround or any flag to still generate changelogs according to the detected or defined convention **and** prompting the user for each package bump instead of relying on the logic of the conventional commits/semVer?

I get the logical reasoning behind using the conventional commits in the first place is doing everything smoothly without developer interaction but there should be an option for an edge case or for someone who don't want to follow the semVer approach.

Issue #1094, opened 2017-10-31
[Feature] Preview changelog for projects that use `--conventional-commits`


StyleT:
I want to be able to preview changes that lerna will do to the `CHANGELOG.md` during `publish` command. 

## Possible Solution
Add `--conventional-commits` flag to the `updated` command

evocateur:
Oooh, that sounds interesting. It would also promote better factoring of the `conventional-changelog-cli` code. Also, a `--dry-run` flag, instead of actually writing it out...

I'm not sure `updated` is the right place for it, but it does sound like the easiest, at first glance.

ntwb:
It's a shame the bot keeps closing (what I consider anyway) valuable issues 😢 

p.s. Unfortunately, I've not the time to contribute to this project even though I'd love to

rishabhgupta:
Opened a PR for adding support for changelog preview here #2555

PR #2555, opened 2020-04-24
Feat/changelog preview

rishabhgupta:
## Description
Adds a way to preview and modify the generated changelog file when using conventional commits. 

When the version command is run with `--changelog-preview` option, once the change log is written it will open in VIM where the user can preview and make changes to the log file. Once the user exists out of vim the file is committed. This will happen for every change log generated. 

## Motivation and Context

At times you want to preview or make changes to the generated change log. For eg:
1. Correct a spelling or a grammatical error in a commit message.
2. Remove certain commit messages from the log.
3. Add additional documentation links (styleguidist or storybooks) with the change log.

ISSUE: https://github.com/lerna/lerna/issues/1094

## How Has This Been Tested?

Switched to forked copy and verified the publish and version command on existing lerna repo.

My changes shouldn't be affecting any other areas than publish and version commands.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -371,13 +371,11 @@ lerna publish from-git --tag-version-prefix=''
 
 ### `--changelog-preview`
 
-This option allows to preview and make changes in the changelog before its committed.
+This option allows to preview and make changes in the changelog before its committed. Should only be run locally and not in CI environment.
 
 ```bash
 # locally
 lerna version --changelog-preview
-# on ci
-lerna publish --changelog-preview
 ```
 
 ## Deprecated Options
--- a/commands/version/README.md
+++ b/commands/version/README.md
@@ -66,6 +66,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--sign-git-tag`](#--sign-git-tag)
 - [`--yes`](#--yes)
 - [`--tag-version-prefix`](#--tag-version-prefix)
+- [`--changelog-preview`](#--changelog-preview)
 
 ### `--allow-branch <glob>`
 
@@ -366,6 +367,17 @@ Keep in mind that currently you have to supply it twice: for `version` command a
 lerna version --tag-version-prefix=''
 # on ci
 lerna publish from-git --tag-version-prefix=''
+```
+
+### `--changelog-preview`
+
+This option allows to preview and make changes in the changelog before its committed.
+
+```bash
+# locally
+lerna version --changelog-preview
+# on ci
+lerna publish --changelog-preview
 ```
 
 ## Deprecated Options
--- a/commands/version/command.js
+++ b/commands/version/command.js
@@ -140,6 +140,10 @@ exports.builder = (yargs, composed) => {
       alias: "yes",
       type: "boolean",
     },
+    "changelog-preview": {
+      describe: "Peview and make changes in changelog before its committed.",
+      type: "boolean",
+    },
   };
 
   if (composed) {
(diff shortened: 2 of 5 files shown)


commands/version/README.md
```diff
@@ -66,6 +66,7 @@ Running `lerna version --conventional-commits` without the above flags will rele
 - [`--sign-git-tag`](#--sign-git-tag)
 - [`--yes`](#--yes)
 - [`--tag-version-prefix`](#--tag-version-prefix)
+- [`--changelog-preview`](#--changelog-preview)
```
evocateur, line 69: This (and its target block) should be alphabetized after `--changelog-preset`.

commands/version/command.js
```diff
@@ -140,6 +140,10 @@ exports.builder = (yargs, composed) => {
       alias: "yes",
       type: "boolean",
     },
+    "changelog-preview": {
```
evocateur, line 143: This should be alphabetized after the `changelog-preset` key (yes, _behind_ conventional-changelog, ordering is... challenging)

core/conventional-commits/lib/update-changelog.js
```diff
@@ -66,8 +71,23 @@ function updateChangelog(pkg, type, { changelogPreset, rootPath, tagPrefix = "v"
       const content = [CHANGELOG_HEADER, newEntry, changelogContents].join(BLANK_LINE);
 
       return fs.writeFile(changelogFileLoc, content.trim() + EOL).then(() => {
-        log.verbose(type, "wrote", changelogFileLoc);
+        // if change log preview option is passed open the file in vim before commiting
+        if (changelogPreview) {
+          return new Promise((resolve, reject) => {
+            const vim = spawn("vim", [changelogFileLoc], { stdio: "inherit" });
```
evocateur, line 77: This is a bit specific, what if I don't like `vim`, and prefer `code` instead? What if my platform doesn't have `vim`? (i mean, yeah, "don't use `--changelog-preview`" is a reasonable answer to the latter question)

evocateur:
I'm not sure motivation #2 is valid, if the conventional commit addressed something, isn't removing it violating the spirit of the convention? (It's not a law, sure, and I'm not your dad, but it just feels weird to explicitly endorse by virtue of accepting this feature)

rishabhgupta:
@evocateur if we need to preview the change log, cant we just run `lerna version --conventional-commits  --no-git-tag-version` this will show the changelog file which will be generated right? or am i missing something here?

JamesHenry:
Hi @rishabhgupta 👋 

Thank you very much for taking the initiative to contribute this PR!

You may or may not know that lerna is now under the stewardship of Nrwl (announcement here https://github.com/lerna/lerna/issues/3121), a company with a long history of not just producing valuable open-source software (OSS), but also backing others (at the time of writing, Nrwl has donated over $50,000 to OSS it _hasn't_ created, see https://opencollective.com/nx for full details).

Quite simply, Nrwl ❤️  OSS, and is committed to making lerna the best it can be. We use it ourselves.

We hope you will continue to be a part of this community as we look to take things forward from here!

Please see https://github.com/lerna/lerna/discussions/3140 for more details on our plans for 2022.

In this case of this PR, because there has been no activity on it for over 2 years, we are going to close it for now. If you would like to discuss the possibility of adding new features to lerna, please kindly open a new issue (or upvote and comment on an existing one) so we can cover all the details: https://github.com/lerna/lerna/issues/new/choose

Many thanks again 🙏

PR #2572, opened 2020-05-06
chore: Use git describe correctly

ChiefORZ:
When automatically detecting the next canary version with `lerna publish --canary` somehow lerna was not detecting the most recent canary version of the package or the packages automatically... This commit fixes it in this, but i dont know if something else breaks with this change.

## Types of changes

## Checklist:

### DIFF
--- a/utils/describe-ref/lib/describe-ref.js
+++ b/utils/describe-ref/lib/describe-ref.js
@@ -10,6 +10,8 @@ module.exports.sync = sync;
 function getArgs(options, includeMergedTags) {
   let args = [
     "describe",
+    // return latest tag - otherwise the ref is incoorrect
+    "--tags",
     // fallback to short sha if no tags located
     "--always",
     // always return full result, helps identify existing release
--- a/commands/publish/__tests__/publish-canary.test.js
+++ b/commands/publish/__tests__/publish-canary.test.js
@@ -384,7 +384,7 @@ test("publish --canary --include-merged-tags calls git describe correctly", asyn
   expect(spy).toHaveBeenCalledWith(
     "git",
     // notably lacking "--first-parent"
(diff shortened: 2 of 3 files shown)


utils/describe-ref/lib/describe-ref.js
```diff
@@ -10,6 +10,8 @@ module.exports.sync = sync;
 function getArgs(options, includeMergedTags) {
   let args = [
     "describe",
+    // return latest tag - otherwise the ref is incoorrect
```
evocateur, line 13: By "correct" you mean... detects broken tags that aren't properly annotated? There should be no situation in which Lerna itself creates release tags that are undetectable _without_ the `--tags` option here.

azu: I have met same issue.

I've release packages and tagged git tag via GitHub API.
The tag is lightweight(non-annotated) tag.

- https://git-scm.com/docs/git-describe

`git descript` can not get lightweight(non-annotated) tag without `--tags` flag.
This commit will fix this issue that I met.

Example repository that used  GitHub API for tagging `git tag`

https://github.com/azu/lerna-monorepo-github-actions-release

```
❯ git describe
v1.3.0-68-ge149829

❯ git describe --tags
v2.0.1-3-ge149829
```

evocateur:
Lerna is using `git describe` correctly for its use case. External tools need to produce correctly-annotated tags to be compatible.

brandones:
@evocateur Looks like there's about three different issues, each with a few people giving reasons why this PR is a necessary fix. The existing behavior of `--canary` is insane. See #2622 and its duplicates, #2780 and #2778 .

brandones:
You should be able to get a repository into a state such that it needs this change simply by using `lerna publish --no-git-tag-version`.

Issue #2592, opened 2020-05-21
Topographical sort should respect local tarballs


JacobLey:
Hello, I've started using Lerna and want to maintain all my packages locally.

That is to say I want to `npm pack` each package into a tarball, and add that tarball as a dependency in another package.
The packing I am handling manually, although some existing issues appear to address that https://github.com/lerna/lerna/issues/2134

An example simple repo:
```
.
├── lerna.json
├── package.json
├── packages
│   ├── bar
│   │   ├── package.json
│   └── foo
│       ├── package.json
│       └── my-scope-foo-1.0.0.tgz
```

`packages/bar/package.json` will look something like:
```
{
  "name": "@my-scope/bar",
  "dependencies": {
    "@my-scope/foo": "file:../foo/my-scope-foo-1.0.0.tgz"
  }
}
```

## Expected Behavior
`lerna ls --toposort` should print:
> @my-scope/foo v1.0.0 packages/foo
> @my-scope/bar v1.0.0 packages/bar
> lerna success found 2 packages

Note that foo comes before bar (despite coming second in the default alphabetical sort).
That is because foo is acknowledged as a dependency of bar.

## Current Behavior
`lerna ls --toposort` should print:
> @my-scope/bar v1.0.0 packages/bar
> @my-scope/foo v1.0.0 packages/foo
> lerna success found 2 packages

Lerna displays bar before foo. This behavior impacts scripts such as `lerna run` and `lerna exec` which might be dependent on topographical sort.

If the dependency is changed to 
```
"@my-scope/foo": "file:../foo
```
order is correct (although that logically produces a different dependency with links)

## Possible Solution
https://github.com/lerna/lerna/pull/2593
check if the local file is a tarball that would have been generated by that package. If so, acknowledge it as a local dependency.

I don't suspect this logical change would be backwards incompatible for most users (i.e. that means someone is using tarball and explicitly _not_ respecting order), but if it were, could adjust the `lerna.json` to include some `allowLocalTarballs: true` argument.

## Context
As noted in the intro, I want to locally pack and install dependencies.

"Alternatives" I don't want to do:
* npm link (lernas default for local dependencies)
    * There is some weirdness with accessing and de-duping sub-dependencies, and acknowledging `.npmignore`.
* publish packages
    * repository is private, can't use public npm
    * Don't want to pay for private npm or self-host the registry.

[...]

PR #2593, opened 2020-05-21
feat(add): Filepath dependencies should be treated as local dependencies agnostic of installation method

JacobLey:
Adding support for https://github.com/lerna/lerna/issues/2592 to create a local dependency when a tarball is used

closes https://github.com/lerna/lerna/issues/2592

See issue for in-depth discussion of motivation.

## Description
When calculating the package-graph, do an additional check to see if the file is actually a tarball, in which case it should be added as a local dependency (existing functionality is that is it is tracked as external)

## Motivation and Context
https://github.com/lerna/lerna/issues/2592

## How Has This Been Tested?
Updated test suite for that file to check that tarballs are included as local.

Applied these changes in a private repo and topographical sort is as expected now.

## Types of changes

## Checklist:

### DIFF
--- a/core/package-graph/index.js
+++ b/core/package-graph/index.js
@@ -65,7 +65,12 @@ class PackageGraph extends Map {
           return currentNode.externalDependencies.set(depName, resolved);
         }
 
-        if (forceLocal || resolved.fetchSpec === depNode.location || depNode.satisfies(resolved)) {
+        if (
+          forceLocal ||
+          resolved.fetchSpec === depNode.location ||
+          depNode.satisfies(resolved) ||
+          resolved.type === "file"
+        ) {
           // a local file: specifier OR a matching semver
           currentNode.localDependencies.set(depName, resolved);
           depNode.localDependents.set(currentName, currentNode);
--- a/core/package-graph/__tests__/package-graph.test.js
+++ b/core/package-graph/__tests__/package-graph.test.js
@@ -92,6 +92,43 @@ describe("PackageGraph", () => {
       expect(pkg1.localDependents.has("pkg-2")).toBe(true);
       expect(pkg2.localDependencies.has("pkg-1")).toBe(true);
     });
+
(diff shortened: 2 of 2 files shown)


JacobLey:
It appears my test fails on windows non-publish (I am on MacOS) https://travis-ci.org/github/lerna/lerna/jobs/689764449

Would appreciate any guidance on how to debug...

JacobLey:
This led to my solution: https://stackoverflow.com/questions/13922086/regex-working-in-linux-but-not-in-windows-node-js

I can appreciate the [frustration](https://github.com/lerna/lerna/blob/master/core/package/__tests__/core-package.test.js#L39) with windows

core/package-graph/index.js
```diff
@@ -65,7 +66,26 @@ class PackageGraph extends Map {
           return currentNode.externalDependencies.set(depName, resolved);
         }
 
-        if (forceLocal || resolved.fetchSpec === depNode.location || depNode.satisfies(resolved)) {
+        let isLocal = forceLocal || resolved.fetchSpec === depNode.location || depNode.satisfies(resolved);
```
evocateur, line 69: Why not reuse `npm-package-arg`'s existing [`resolved.type === 'file'](https://github.com/npm/npm-package-arg#result-object)? Seems a lot easier than all of this complicated regular expression juggling.

JacobLey: Ah yea that is a lot easier.

I was trying to solve a very specific case, but any local file is by definition a "local" dependency...

JacobLey:
Now there are lots of noisy, unrelated commits... can squash if desired

JamesHenry:
Hi @JacobLey, I am sorry for how long this PR sat in limbo. Do you have time to bring it up to date with the latest?

JamesHenry:

If you or anybody else would like to implement this functionality against the codebase as it stands to day then we would grateful accept and review a PR.

Thanks again!

Issue #2056, opened 2019-04-26
Tag prefix ignored during checkWorkingTree() call for lerna version and publish


treywashington:
Calling `lerna version` on a tagged HEAD commit errors regardless of whether the tag matches `--tag-version-prefix`:
 ```
ERR! ERELEASED The current commit has already been released. Please make new commits before continuing.
```

## Expected Behavior

Expected behavior is that `lerna version` and `lerna publish` will version the commits since the last release tag matching the `--tag-version-prefix` set or default 'v'.

Example: 
```bash
> git describe --long // otherTagScheme.10.0.0-0-gc9a9d91
> git describe --match v* --long // v0.0.1-50-gc9a9d91
> lerna version patch // Changes: @scope/pkg: 0.0.1 => 0.0.2
```

## Current Behavior

Current behavior is to treat any tag returned from `git describe --long` as the tag since the last lerna release. 

Example:
```bash
> git describe --long // otherTagScheme.10.0.0-0-gc9a9d91
> git describe --match v* --long // v0.0.1-50-gc9a9d91
> lerna version patch // ERR! ERELEASED The current commit has already been released. Please make new commits before continuing.
```

## Possible Solution

Suggested fix is to pass through the match option to `describeRef()` in  [checkWorkingTree()](https://github.com/lerna/lerna/blob/86ba7a03fe8aa86733d4dd1de8136ee918fe698b/utils/check-working-tree/lib/check-working-tree.js#L13)

The match option can be derived from the execOption `--tag-version-prefix` for both the `publish` and `version` commands.

Corollary to this is to make sure that `lerna version` updates the packages changed since `git describe --match ${tag-version-prefix}* ` as well, not if its a tag that doesn't match. 

## Context

Context here is that my project has two different release cadences for two different repositories right now. We are trying to configure our npm repo deployment with lerna but having two different tagging schemes is throwing it off.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.5.1 |
| `npm --version`  | 6.4.1 |
| `yarn --version` | 1.9.4 |
| `node --version` | 11.2.0 |

| OS | Version |
| --- | --- |
| macOS Mojave | 10.14.4 |

evocateur:
Yeah, that is something we should definitely be doing. PRs welcome!

PR #2601, opened 2020-05-26
Added checking of version prefix, when do "lerna version"

divomen:
A fix for issue https://github.com/lerna/lerna/issues/2056
## Description

Added a match for tagPrefix for collectUpdates in version command (only for fixed-version repo)

## Motivation and Context

The main problem is, that when we use custom tags in a repo, the version command calculates changes from a wrong tag.
Also, there is an open issue: https://github.com/lerna/lerna/issues/2056

## How Has This Been Tested?

I tested it on a real repo.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -187,10 +187,14 @@ class VersionCommand extends Command {
       );
     }
 
+    const opts = Object.assign({}, this.execOpts);
+    if (!this.project.isIndependent() && this.tagPrefix) {
+      opts.match = `${this.tagPrefix}*`;
+    }
     this.updates = collectUpdates(
       this.packageGraph.rawPackageList,
       this.packageGraph,
-      this.execOpts,
+      opts,
       this.options
     ).filter(node => {
       // --no-private completely removes private packages from consideration


divomen:
> This is awesome. Is it possible to add some tests around this?
> 
> https://github.com/lerna/lerna/blob/7abfe43426197fbc7f18c44b0c994324609fc769/utils/collect-updates/__tests__/collect-updates.test.js

Ok, I'll try

evocateur (changes requested):
This is definitely an oversight, thank you for identifying it.

I think we need a more holistic solution, as there are actually four places across the Lerna codebase that call `collectUpdates`, and all of them need this awareness.

We also need tests to validate the change, preventing future regression.

gund:
Any update on this?

This is really a blocker for a cases where Lerna is used multiple times with different tag prefixes and currently will not release anything for subsequent calls after the first one was released...

commands/version/index.js
```diff
@@ -187,10 +187,14 @@ class VersionCommand extends Command {
       );
     }
 
+    const opts = Object.assign({}, this.execOpts);
+    if (!this.project.isIndependent() && this.tagPrefix) {
+      opts.match = `${this.tagPrefix}*`;
```
gund, line 192: It's probably better to use `**` instead to allow for tags that contain forward slashes as a single star will not match those.
```suggestion
      opts.match = `${this.tagPrefix}**`;
```

JamesHenry:
@divomen are you interested in completing this PR?

@gund I see you applied a more comprehensive version of this change to a fork, are you interested in porting that to the main lerna project to supersede this PR?

PR #2603, opened 2020-05-26
remove unused dependency in create which causes compilation issues in some ts projects

dyladan:
## Description
Remove the unused globby dependency in `@lerna/create`

## Motivation and Context

Dependency is unused, so it should be removed.

The old version of `globby` which was in the `package.json` also depended on `glob`, which depended on `@types/glob`, which depended on `@types/node`. If a lerna monorepo contains a typescript package which depends on `zone.js`, it fails to compile because `zone.js` and `@types/node` have conflicting definitions of `global`. A project which depends on lerna in the root of the project will have `@types/node` in its root `node_modules` due to this deep dependency issue. Because of node's module resolution behavior, this `@types/node` will be read during compilation even if a package in the lerna monorepo does not require it.

## How Has This Been Tested?
n/a

## Types of changes

## Checklist:

- [x] ~My code follows the code style of this project.~ n/a
- [x] ~My change requires a change to the documentation.~ n/a
- [x] ~I have updated the documentation accordingly.~ n/a
- [x] ~I have added tests to cover my changes.~ n/a

### DIFF
--- a/commands/create/package.json
+++ b/commands/create/package.json
@@ -41,7 +41,6 @@
     "camelcase": "^5.0.0",
     "dedent": "^0.7.0",
     "fs-extra": "^8.1.0",
-    "globby": "^9.2.0",
     "init-package-json": "^1.10.3",
     "npm-package-arg": "^6.1.0",
     "p-reduce": "^1.0.0",


dyladan:
A test failed in CI which doesn't appear related. It looks like a `core/cli` test which only failed in node 8. Looking at other PRs doesn't show this test to be failing in other places. Is this connected in a way which is not obvious?

dyladan:
It is still unused in this package, so I still think the PR is valid. In current globby, they don't have the problematic `glob` => `@types/glob` => `@types/node` dependency issue that is breaking our installs, so it should be enough to bump the version to the latest in those other places.

JamesHenry:
Thank you @dyladan, I went ahead and removed the globby dep from the create package (and did the necessary lock file update) in #3360

Many thanks for being a part of the lerna community 🙏

PR #2634, opened 2020-06-26
fix(list): force-local option for list command

matheo:
## Description

I'm having troubles trying to run my `lerna exec built` script because the dependency-graph is not correctly built.
I've found this issue in a development branch, while publishing some beta releases of my packages.

I've tracked down the issue to `PackageGraphNode.satisfies`
because `semver('14.0.0-beta.0', '*')` is returning false, and that puts my beta packages as `externalDependencies` instead `localDependencies` while initializing the `PackageGraph`.

So looking at the options, I saw the `forceLocal` parameter that seems supported by bootstrap only, and I added it to the `listable.options` too so we can workaround this bug with `semver` while they have a fix.

## Motivation and Context

The bug was reported 11 days ago without feedback ATM:
https://github.com/npm/node-semver/issues/331

and 26 days ago the first question was posted:
https://github.com/npm/node-semver/issues/329

## How Has This Been Tested?
I've used this patch locally in the node_modules of my beta-branch
and it works as expected :)

## Types of changes

## Checklist:


matheo:
@evocateur a bugfix release this weekend will be highly appreciated!

I confirmed it with `npm test -- --watch commands/list` 
and Travis seems to have something broken with the publish

Thanks again for such a great software!

matheo:
OMG, I just realized that this doesn't fix `lerna exec --since`
semver really screwed this one :(

matheo:
@evocateur there's something we can do to propagate `--force-local` from the `bootstrap`, `exec`, or even the `lerna.json` file to avoid problems building our dependency graph in the branches where we are releasing `beta` versions :-?

it seems that `lerna` needs to provide a built-in solution for this case because semver seems to not compare `latest` `beta` `canary` between each other. Thoughts please?

JamesHenry:

In order to proceed with this a reimplementation of this one, we would want to first see a clear reproduction of any issue being encountered. If you can provide that via a new issue then we will happily consider how best to resolve it.

Thanks again!

PR #2640, opened 2020-07-01
Upgrading @zkochan/cmd-shim to 5.0.0

mesner:
Upgrading @zkochan/cmd-shim to 5.0.0.   

## Description
Change of version within package.json files for the root project and for utils/create-symlink.

## Motivation and Context

We discovered that symbolic links were not being created for a local repository when running `lerna add projectChild --scope=projectParent --loglevel=silly`, with no errors being reported.   Investigation revealed that the promise returned from creating a 'binary' symbolic link (type='exec') in a prior project dependency using the cmd-shim was never being resolved or rejected (windows 10 64 bit running git bash), which halted progress of the pMapSeries in symlink-dependencies, causing early termination.  Upgrading to the latest version of cmd-shim resulted in the successful resolving of the promise and allowed the linking to complete, which itself resulted in creating the link to our projects above.

## How Has This Been Tested?

`npm run test` from a fresh checkout of master results in 829 linting problems due to missing node_module directories in subpackages.  I was able to force a global install of lerna to run lerna bootstrap on this project, but this didn't resolve the issue.

`npm jest` results in all tests passing.

`npm run integration` results in 2 publishing failures.

I'd be happy to try to resolve these test failures, but my suspicion is that they are unrelated to my dependency upgrade.

## Types of changes

## Checklist:

- [] My code follows the code style of this project.
- [] My change requires a change to the documentation.
- [] I have updated the documentation accordingly.
- [] I have added tests to cover my changes.
- [] All new and existing tests passed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -932,6 +932,18 @@
         "@zkochan/cmd-shim": "^3.1.0",
         "fs-extra": "^8.1.0",
         "npmlog": "^4.1.2"
+      },
+      "dependencies": {
+        "@zkochan/cmd-shim": {
+          "version": "3.1.0",
+          "resolved": "https://registry.npmjs.org/@zkochan/cmd-shim/-/cmd-shim-3.1.0.tgz",
+          "integrity": "sha512-o8l0+x7C7sMZU3v9GuJIAU10qQLtwR1dtRQIOmlNMtyaqhmpXOzx1HWiYoWfmmf9HHZoAkXpc9TM9PQYF9d4Jg==",
+          "requires": {
+            "is-windows": "^1.0.0",
+            "mkdirp-promise": "^5.0.1",
+            "mz": "^2.5.0"
+          }
+        }
       }
     },
     "@lerna/describe-ref": {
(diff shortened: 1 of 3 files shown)


yoitsro:
@evocateur Is there anything we can do to push this forward? Just run into the issue myself 😭

revelt:
Is this PR a redundant in light of of https://github.com/lerna/lerna/pull/2473 ? That PR replaces `@zkochan/cmd-shim` with `cmd-shim`.

armano2:
Currently there is issue with upgrading dependencies as lerna still supports node 6 (even if not tested in CI)

we can't upgrade this package without changing required node version, personally i think that lerna should upgrade requirements to node 10 as everything lower than that is no longer maintained. 

this will unlock upgrade of a lot of dependencies not only this one

https://raw.githubusercontent.com/nodejs/Release/master/schedule.svg?sanitize=true
https://nodejs.org/en/about/releases/

evocateur:
[v4.0.0](https://github.com/lerna/lerna/releases/tag/v4.0.0) (now available as `lerna@next`) should resolve this, apologies for the delay.

PR #2642, opened 2020-07-03
fix(docs): add reference to useWorkspaces

andrestone:
This seems to be a very popular use case and there's no mention on the main README.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,8 @@ Run `lerna --help` to see all available commands and options.
       "npmClientArgs": ["--no-package-lock"]
     }
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "useWorkspaces": false
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -158,7 +158,8 @@ Run `lerna --help` to see all available commands and options.
       "npmClientArgs": ["--no-package-lock"]
     }
   },
-  "packages": ["packages/*"]
+  "packages": ["packages/*"],
+  "useWorkspaces": false
```
evocateur, line 162: Why would you explicitly configure a value to `false`? It's confusing the issue. I agree `useWorkspaces` is poorly documented, but it needs more clarity than this.

evocateur (changes requested):

README.md
andrestone: I don't get how 'false' is more or less confusing than 'true', but whatever suits the project best works for me.

evocateur: I mean, `false` is the default. I think people understand how to set boolean properties in JSON files?

What I meant, more specifically, is a separate section that describes the changes between multiple files in the monorepo. It's not just a change to the `lerna.json`, after all, when using Yarn (or npm v7) workspaces.

JamesHenry:
Hi @andrestone thanks a lot for contributing to lerna and its community. I'm going to close this one as it is no longer applicable to the file in question.

We will be building out more guidance around using workspaces in general soon.

Many thanks 🙏

PR #2667, opened 2020-07-23
docs(lerna.json): Update bootstrap command docs

lautarorodriguezmrf:
There is a supported option in the `lerna.json` file which makes `lerna bootstrap` to run `--hoist` option. Adding this parameter to documentation since I found it by diving through open issues and old commits.

## Motivation and Context
`lerna add` currently runs `bootstrap` command when it's done including the new dependency. If the project uses hoisting then the lerna.json is the only place you can make the automatic `bootstrap` command to be run using `--hoist` option.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,8 @@ Run `lerna --help` to see all available commands and options.
     },
     "bootstrap": {
       "ignore": "component-*",
-      "npmClientArgs": ["--no-package-lock"]
+      "npmClientArgs": ["--no-package-lock"],
+      "hoist: true
     }
   },
   "packages": ["packages/*"]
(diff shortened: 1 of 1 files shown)


JamesHenry:
Thanks a lot for contributing to lerna @lautarorodriguezmrf!

I'm sorry this PR sat unaddressed for a long time. I am going to close this as we have finally confirmed our agreement with the previous maintainer in that we will be transitioning away from legacy package management features in lerna, such as `bootstrap`.

Please see the full context on our Lerna 7 Roadmap announcement: https://github.com/lerna/lerna/discussions/3410

Issue #2695, opened 2020-08-19
Non-atomic push fallback condition in version command should be enhanced


fsmaia:
PR #2467 introduced a great feature: when `--atomic` fails due to lack of support, it performs a "non-atomic" push.
However, the fallback condition is not precise and is invoking the fallback even when there is support to atomic.

## Expected Behavior
When running lerna publish, if a git merge happens in the meanwhile, it should reject the push command, and invoke the non-atomic mode only if git doesn't support it.

## Current Behavior
According to the following logs, the atomic push failed due to being behind upstream, which should stop the execution without invoking the fallback, however it didn't:

```
lerna info auto-confirmed 
lerna info execute Skipping releases
lerna info git Pushing tags...
lerna WARN gitPush Warning: Permanently added 'bitbucket.org,18.205.93.0' (RSA) to the list of known hosts.
lerna WARN gitPush error: atomic push failed for ref refs/heads/master. status: 5
lerna WARN gitPush 
lerna WARN gitPush fatal: The remote end hung up unexpectedly
lerna WARN gitPush To bitbucket.org:ifood/portal-packages.git
lerna WARN gitPush  ! [rejected]          master -> master (fetch first)
lerna WARN gitPush error: failed to push some refs to 'git@bitbucket.org:ifood/portal-packages.git'
lerna WARN gitPush hint: Updates were rejected because the remote contains work that you do
lerna WARN gitPush hint: not have locally. This is usually caused by another repository pushing
lerna WARN gitPush hint: to the same ref. You may want to first integrate the remote changes
lerna WARN gitPush hint: (e.g., 'git pull ...') before pushing again.
lerna WARN gitPush hint: See the 'Note about fast-forwards' in 'git push --help' for details.
lerna info gitPush --atomic failed, attempting non-atomic push
...
```

## Possible Solution
Instead of looking for `atomic` in the error log, we could look for: `gitPush fatal: the receiving end does not support --atomic push` or something like `does not support --atomic`.

## Steps to Reproduce (for bugs)
1. Start a lerna publish, with git version higher than 2.4.0
2. Merge a pull-request to the target branch in the meanwhile
3. `isBehingUpstream` should be executed in the version command and it is not (could be a separate issue)
4. The atomic push will fail
5. [Bug] The non-atomic push will be executed when it shouldn't

PR #2696, opened 2020-08-19
Enhance non-atomic push fallback condition in version command

fsmaia:
## Description
Previously, we were checking for `atomic` in the atomic git push error to check if a non-atomic push should be performed.
However, this keyword appears in all failure scenarios, not only when git lacks support to `--atomic` flag.
With this change, this condition is more precise and will fail the lerna publish command properly instead of leading to a unwanted fallback.

## Motivation and Context
✅ Closes: #2695

## How Has This Been Tested?
The current test scenario already covers this well.

## Types of changes

## Checklist:


JamesHenry:

Given the associated issue was closed/not responded to we will assume that no further action is required here.

Thanks again!

PR #2768, opened 2020-11-27
Fixing --no-allow-branch to remove allowBranch option value coming from lerna.json

llecaroz:
## Description
in the of the allowBranch option needs to be removed
without changing lerna.json, which as sourcecode material should never
be changed, --no-allow-branch will now effectively do that from the
lerna command line.

## Motivation and Context
Some projects defined a allowBranch in lerna.json but for ci builds, you want the version to be updated without having the new package.json being pushed in git as this is a ci buld only (don't want any git change to be pushed but a version to be adjusted for example with -SNAPSHOT suffix). 

Even when using --no-git-tag-version --no-push, lerna tries to push in git because the allowBranch option is used in lerna.json. And the lerna.json cannot be touched for some reasons (security & even more because this is a sourcode material). Fixing the use of --no-allow-branch which already exists as option (due to the --allow--branch) will help into removing/disabling the allowBranch lerna.json option without touching the code & finally also getting the lerna version command line working without trying to put changes in git even when --no-git-tag-version --no-push are set.

## How Has This Been Tested?
tests were already existing on --allow-branch in version-allow-branch.test.js
only a "branch should be ignored with --no-allow-branch" was added inside "lerna.json" using the allow-branch-lerna\lerna.json  thru initFixiture, next switching to unmatched. & finally it is able to work as expected with --no-allow-branch set instead of failing with a "Branch 'unmatched' is restricted from versioning" exception.

## Types of changes

## Checklist:


ikeedge (approved):

JamesHenry:
Hi @llecaroz thanks a lot for contributing to lerna!

I am going to close this one as I don't believe we need this option. What you are describing can just as easily be achieved by passing `--allow-branch="*"` because the CLI arguments always take priority over what is in the `lerna.json`. I just verified it on the lerna project itself, which also makes use of `allowBranch` in its lerna.json.

Many thanks 🙏

PR #2773, opened 2020-12-04
fix: broken symlink in windows

### LINKED ISSUE #2539: Lerna bootstrap --hoist broken in certain configurations
#### Description
We had a repository that was using `lerna bootstrap --hoist` to install (dev)dependencies and create symlinks between packages. The dependencies would get hoisted and installed, but no symlinks would be created. Not between dependent packages, but also not in `packages/<name>/node_modules/.bin` directories. What I couldn't understand was why this was consistently failing in our repository, but if I copied the packages and configuration to another directory it worked without issue. After a lot of debugging I found the issue.

#### Symptoms
 1. `.bin` folders are created within `packages/<name>/node_modules`, but they are empty.
 2. No symlinks between local packages are created, regardless of `--force-local`.
 3. Lerna completes successfully, e.g. with exit code 0(!).
 4. This is somewhat harder to spot, but the output of `lerna bootstrap --hoist` is incomplete:

```
lerna notice cli v3.20.2
lerna info Bootstrapping 2 packages
lerna info Installing external dependencies
lerna info hoist Installing hoisted dependencies into root
lerna info hoist Pruning hoisted dependencies
lerna info hoist Finished pruning hoisted dependencies
<MISSING ENTRIES>
```

The missing entries are:
```
lerna info hoist Finished bootstrapping root
lerna info Symlinking packages and binaries
lerna success Bootstrapped 2 packages
```

#### Problem description
After a lot of digging I figured out the problem is in this dependency chain:
[@lerna/create-symlink @3.16.2](https://www.npmjs.com/package/@lerna/create-symlink) > [@zkochan/cmd-shim @3.1.0](https://www.npmjs.com/package/@zkochan/cmd-shim/v/3.1.0) > [mkdirp-promise @5.0.1,  **deprecated**](https://www.npmjs.com/package/mkdirp-promise) > [mkdirp @*](https://www.npmjs.com/package/mkdirp)
At the moment of writing version `3.16.2` is the latest for `@lerna/create-symlink`.

The major problem here is that `mkdirp-promise` states it depends on **any** version of mkdirp, using  `"*"`. The `mkdirp` package itself had a breaking change from version `0.x` to `1.x`: its API returns a `Promise` and **no longer takes a callback parameter**.

So if you get lerna installed with `mkdirp` on some `0.x` version, it works fine. If you're unlucky enough to get a version `1.x` of `mkdirp`, the `package.json` of `mkdirp-promise` says that's fine, but `mkdirp-promise` will then break silently. The problem is that `mkdirp-promise` creates a wrapper `Promise` and then passes a callback to `mkdirp` that will resolve that wrapper `Promise`. Of course a version 1 of `mkdirp` returns a `Promise` of its own and completely ignores any passed callback, so the wrapper `Promise` never resolves. Strangely enough, instead of hanging indefinitely, the process simply exists with an exit code of 0.

I searched if this issue was already reported, and I found #2429 which has a striking overlap. However it doesn't mention the problems I observed, and seems to propose a different solution altogether.

## Expected Behavior
`lerrna bootstrap --hoist` should create symlinks in `.bin` folders, and between local packages (if applicable).

## Current Behavior
Depending on which major version of `mkdirp` you happen to get, lerna will work or break.

## Possible Solution
I think updating `@lerna/create-symlink` to use the latest [@zkochan/cmd-shim](https://www.npmjs.com/package/@zkochan/cmd-shim), e.g. version `4.3.0`, would already fix this issue.

#### Workaround
Ensure a version 0.x of `mkdirp` is installed and used.

 1. Create a new lerna repository containing two packages, one depending on the other.
 2. Ensure a version `1.x` of `mkdirp` gets installed
    For example by putting it in the `devDependencies` of the root `package.json`.
 3. Run `lerna bootstrap --hoist`

There is more information in the **Symptoms** and **Problem description** sections above, please let me know if that's not sufficient.

```json
{
  "packages": [
    "packages/*"
  ],
  "version": "1.1.3"
}

```
</p></details>

```txt
lerna notice cli v3.20.2
lerna info Bootstrapping 2 packages
lerna info Installing external dependencies
lerna info hoist Installing hoisted dependencies into root
lerna info hoist Pruning hoisted dependencies
lerna info hoist Finished pruning hoisted dependencies
<MISSING ENTRIES>
```
</p></details>

## Context
This issue was blocking us from having a working monorepo at all. It took me and some colleagues a few days to track down why lerna was failing, especially since it was the first monorepo I set up.

## Your Environment
We encountered this issue on both Windows 10 and Mac OS with lerna version `3.20.2`. Note that it's very important a version 1.x of the `mkdirp` dependency is resolved/installed for the problem to surface.

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.20.2 |
| `npm --version`  | 6.14.4 |
| `yarn --version` | N/A |
| `node --version` | 12.16.2|

| OS | Version |
| --- | --- |
| MacOS Catalina | 10.15.4 |
| Windows 10 | 1909 Build 18363.720 |

yantze:
## Description
Upgraded `@zkochan/cmd-shim` to ^5.0.0, deprecated the dependency package `mkdirp-promise`, fixed the directory linking problem in windows.

## Motivation and Context
the current `mkdirp-promise@3.0.1` uses `mkdirp@*` and mkdirp uses the new Promise syntax, it does not symlink directories correctly in windows.

fix #2539 

## How Has This Been Tested?

- Create a new lerna repository containing two packages, one depending on the other.
- Ensure a version 1.x of mkdirp gets installed
- For example by putting it in the devDependencies of the root package.json.
- Run `lerna bootstrap --hoist`

## Types of changes

## Checklist:

### DIFF
--- a/utils/create-symlink/package.json
+++ b/utils/create-symlink/package.json
@@ -31,7 +31,7 @@
     "test": "echo \"Run tests from root\" && exit 1"
   },
   "dependencies": {
-    "@zkochan/cmd-shim": "^3.1.0",
+    "@zkochan/cmd-shim": "^4.3.0",
     "fs-extra": "^8.1.0",
     "npmlog": "^4.1.2"
   }


herberttn:
We're having the same problem! Right now we're having to force the `mkdirp` version that `lerna` needs on our root `package.json`, or `lerna` is not able to make the symlinks correctly. It works but it can make the packages that depend on the newer version stop working as well.

The only real solve is to bump `mkdirp` in `lerna`'s tree, which this pr achieves or stop using `--hoist` (which is a whole lot different problem for some projects).

Would be nice to make this pull request happen!

evocateur:
[v4.0.0](https://github.com/lerna/lerna/releases/tag/v4.0.0) is now available as `lerna@next`, which should resolve this issue. Please update if that's not the case. Thanks for making the effort!

Issue #2711, opened 2020-09-12
Do not use red color in console output


otakustay:
Maybe this is a strange feature request but a red console output is strongly frightening, many times I was cheated as our unit tests fail.

![image](https://user-images.githubusercontent.com/639549/92990301-da0ebb80-f50d-11ea-9f4c-b97e6685d3bb.png)

## Expected Behavior

Choose some more friendly colors but not red in `lerna run`

## Current Behavior

Lerna may [use red from color wheel](https://github.com/lerna/lerna/blob/dea9745f7eba81f7c551c1150890fb248b1b9eb6/core/child-process/index.js#L12).

## Possible Solution

Simply remove red from color wheel, keeping cyan, green, yellow, blue and magenta.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.22.1 |
| `npm --version`  | 6.14.8 |
| `yarn --version` | 1.22.4 |
| `node --version` | v14.8.0 |

cybadave:
Bump

PR #2774, opened 2020-12-05
Replace `"red"` color with `"brightBlue"` on package's output prefix

wmaca:
## Description
Replaced color wheel's `"red"` color with `"brightBlue"`.

## Motivation and Context
When running a script/command on multiple packages, the output of each package may be prefixed with the package's name. 

This prefix of each package is colored differently, but the color wheel is limited to six colors in the following order: `"cyan"`, `"magenta"`, `"blue"`, `"yellow"`, `"green"` and `"red"`. The color `"red"` is used every six packages, but this color is generally associated to error messages.

Whenever I see lerna output parts of the text in red, I believe an error happened. But that is not what actually happened.

Only removing the color red would leave the color wheel with even fewer options.
Remove prefix coloring also seems not to be a good alternative, as the colors are helpful when reading the output.

As I solution, I propose to change it to one of the other colors available for chalk. As all of the "solid" colors left would be problematic, I opted for `brightBlue`.

It solves #2711. 

## How Has This Been Tested?
Ran a monorepo locally with more than six packages to the the different prefix color applied.

## Types of changes

## Checklist:

### DIFF
--- a/core/child-process/index.js
+++ b/core/child-process/index.js
@@ -10,7 +10,7 @@ const logTransformer = require("strong-log-transformer");
 const children = new Set();
 
 // when streaming processes are spawned, use this color for prefix
-const colorWheel = ["cyan", "magenta", "blue", "yellow", "green", "red"];
+const colorWheel = ["cyan", "magenta", "blue", "yellow", "green", "white"];
 const NUM_COLORS = colorWheel.length;
 
 // ever-increasing index ensures colors are always sequential
--- a/core/child-process/index.js
+++ b/core/child-process/index.js
@@ -10,7 +10,7 @@ const logTransformer = require("strong-log-transformer");
 const children = new Set();
 
 // when streaming processes are spawned, use this color for prefix
-const colorWheel = ["cyan", "magenta", "blue", "yellow", "green", "white"];
+const colorWheel = ["cyan", "magenta", "blue", "yellow", "green", "blueBright"];
 const NUM_COLORS = colorWheel.length;
 
 // ever-increasing index ensures colors are always sequential


wmaca:
Just out of curiosity. 

I know there are more important PRs than this one and maintainers have limited time available, but am I missing something on this PR?

evocateur:
@wmaca Sorry for the delay.

I must admit I'm not inclined to accept this PR, as "red" is cosmetic in this case, and I don't believe "brightBlue" to be sufficiently distinct from the other blue in the list. Errors are represented by (well-designed) programs by their exit codes, not the color of their logging.

It's a very primitive color wheel, and I don't think Lerna needs to be themable or anything. I dunno.

wmaca:
Hi, @evocateur!

No problems with the delay.

While I do agree exit codes are a good way of signaling errors, not all errors can be signaled that way.

For example, when running webpack in watch mode, the process does not exit, it just logs when errors happen. And generally in red (not necessarily webpack's case).

And regarding the brightBlue, it is rather a lack of options than necessarily the best alternative. I would be glad to suggest a different color.

Let me know if this has clarified anything at all.

If you are totally not inclined to discuss this, I am okay if you simply reject this PR.

Best,
Walter

PVermeer:
> @wmaca Sorry for the delay.
> 
> I must admit I'm not inclined to accept this PR, as "red" is cosmetic in this case, and I don't believe "brightBlue" to be sufficiently distinct from the other blue in the list. Errors are represented by (well-designed) programs by their exit codes, not the color of their logging.
> 
> It's a very primitive color wheel, and I don't think Lerna needs to be themable or anything. I dunno.

Don't understand where you coming from. Red is universally used by designers to warn people (street signs, ui, product labeling, console logging, etc...). Almost all "well-designed" programs adhere to this and display errors in red including browsers, webpack,  jest, etc....

Also Microsoft tends to use different exit codes for programs that succeed in different ways. "Well designed" is very subjective in this matter while using 'red' in ui is widely accepted as a warning sign.

I'm still getting a minor heart attack every time my console turns red when running something in Lerna.

JamesHenry:
Hi Folks 👋 

A lot has changed since this PR was first opened and there is a new, much more powerful task runner implementation available in `lerna run`. It is currently opt-in but will be the default in imminent v6 release.

I would strongly encourage you all to start using the new task runner implementation if you haven't already, but I think this PR is relatively harmless for the legacy runners and so I'm happy to land it now.

Many thanks for contributing to lerna and its community 🙏

PR #2804, opened 2021-01-27
Readme publish option `--force-publish` was missing

MickL:
Within the readme of publish the option `--force-publish` was not described there. Found it randomly in some Github issues.

### DIFF
--- a/commands/publish/README.md
+++ b/commands/publish/README.md
@@ -50,6 +50,7 @@ This is useful when a previous `lerna publish` failed to publish all packages to
 - [`--canary`](#--canary)
 - [`--contents <dir>`](#--contents-dir)
 - [`--dist-tag <tag>`](#--dist-tag-tag)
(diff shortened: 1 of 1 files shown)


christophehurpeau:
@MickL I think it's supposed to be a `version` option not a `publish` version as publish doesn't create versions, and it's documented https://github.com/lerna/lerna/blob/main/commands/version/README.md#--force-publish

evocateur:
Yep, this is already documented, as @christophehurpeau pointed out. Thank you for taking the time.

MickL:
I dont understand how this is documented. It is not part of the publish page and not linked there. @lerna/version does not seem to be related to @lerna/publish to me.

christophehurpeau:
https://github.com/lerna/lerna/tree/main/commands/publish#options

> lerna publish supports all of the options provided by lerna version in addition to the following:

It's in the documentation

Issue #2869, opened 2021-04-09
npmClient does not work when hoist is used


372623460jh:
npmClient does not work when hoist is used

## Expected Behavior
Both npmClient and hoist are valid.

## Current Behavior
When we use command `lerna bootstrap --npm-client tnpm --hoist` to setup npmClient as `tnpm`, hoist will still use `npm` as npmClient. npmClient and hoist cannot work together.

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.22.1 |
| `npm --version`  | 7.6.3 |
| `yarn --version` | - |
| `node --version` | 15.12.0 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Catalina | 10.15.4 |

PR #2870, opened 2021-04-09
fix(): both npmClient and hoist are valid.

372623460jh:
fix: #2869

When we use hoist, `npmGlobalStyle` is equal to `true` and makes `npmClient` always `npm`. The set 
 of `npmClient` won't take effect.

https://github.com/lerna/lerna/blob/6cb8ab2d4af7ce25c812e8fb05cd04650105705f/commands/bootstrap/index.js#L548

https://github.com/lerna/lerna/blob/6cb8ab2d4af7ce25c812e8fb05cd04650105705f/utils/npm-install/npm-install.js#L25

### DIFF
--- a/utils/npm-install/npm-install.js
+++ b/utils/npm-install/npm-install.js
@@ -22,7 +22,7 @@ function npmInstall(
   let cmd = npmClient || "npm";
 
   if (npmGlobalStyle) {
-    cmd = cmd === "yarn" ? "npm" : npmClient;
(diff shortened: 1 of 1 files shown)


wangybg:
good job

JamesHenry:
Hi @372623460jh I am going to close this one as the issue it is associated with was closed as it didn't receive a response.

Many thanks for contributing to lerna! 🙏

PR #3017, opened 2021-10-25
fix: If the includeMergedTags are true, it will filter out --first-parent

Tincole:
If the includeMergedTags are true, it will filter out --first-parent

## Types of changes

## Checklist:


JamesHenry:

In order to proceed with this a reimplementation of this one, we would want to first see a clear reproduction of any issue being encountered. If you can provide that via a new issue then we will happily consider how best to resolve it.

Thanks again!

Issue #1304, opened 2018-03-05
Lerna clean does not remove node_modules folder at the root


bertho-zero:
## Expected Behavior
Lerna should delete the node_modules folder at the root.

## Current Behavior
Lerna clean does not delete the root node_modules folder.

## Steps to Reproduce (for bugs)

1. `lerna bootstrap`
2. `lerna clean`

<details><summary>lerna.json</summary><p>

```json
{
  "name": "oa",
  "version": "1.0.0",
  "description": "OpenAgenda monorepo",
  "main": "index.js",
  "repository": "https://bitbucket.org/openagenda/oa",
  "author": "OpenAgenda Dev Team",
  "license": "See license.txt",
  "private": true,
  "scripts": {
    "bs": "lerna bootstrap"
  },
  "workspaces": {
    "packages": [
      "packages/*"
    ],
    "nohoist": [
      "**/react",
      "**/react-dom",
      "**/redbox-react",
      "**/create-react-class",
      "**/font-awesome"
    ]
  },
  "devDependencies": {
    "lerna": "^2.9.0"
  }
}

```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 2.9.0 |
| `yarn --version` | 1.5.1 |
| `node --version` | 8.9.1 |

| OS | Version |
| --- | --- |
| Ubuntu | 16.04 |

evocateur:
Blowing away `./node_modules` would blow away the `lerna` dev dependency, in your example. That's not ideal.

In general, `lerna clean` is ...kinda cheesy? It's not adding anything that `git clean -fdx packages` can't do, and slower! I don't recommend using `lerna clean`, tbh.

Closing as "wont-fix" because it's not really a command that serves a useful purpose.

bertho-zero:
I would have used it in development, I'm doing a `rm -rf **/node_modules` instead.
I thought I could use `lerna clean` and then make a fresh install only from package.json and lock files.

evocateur:
`git clean -fdx` is way better.

hon2a:
I've just run across this as well through a nasty issue after using `lerna clean` with the abovedescribed expectation for some time.

@evocateur If it's _"not really a command that serves a useful purpose"_, why is it provided at all? Providing a misleading API and then saying it shouldn't be touched, because it's useless anyway, makes no sense whatsoever IMO.

evocateur:
[...]

PR #3027, opened 2021-11-11
mention that "lerna clean" can also clean the root

trusktr:
## Description

`lerna clean` does not remove modules from the root `node_modules` directory, even if you have the `--hoist` option enabled, unless the root is also a package that you have listed in `lerna.json`.

## Motivation and Context

For example [ Lerna clean does not remove node_modules folder at the root, #1304 ](https://github.com/lerna/lerna/issues/1304).

## How Has This Been Tested?

I have been using this approach for a while with no issues. I know that some discussions have mentioned that making the root also be a package managed by Lerna may have unexpected results, but so far I have been fine under all scenarios that I've needed Lerna for.

## Types of changes

- [x] Documentation

## Checklist:

### DIFF
--- a/commands/clean/README.md
+++ b/commands/clean/README.md
@@ -15,4 +15,4 @@ Remove the `node_modules` directory from all packages.
 `lerna clean` accepts all [filter flags](https://www.npmjs.com/package/@lerna/filter-options), as well as `--yes`.
 
 
-> `lerna clean` does not remove modules from the root `node_modules` directory, even if you have the `--hoist` option enabled.
+> `lerna clean` does not remove modules from the root `node_modules` directory, even if you have the `--hoist` option enabled, unless the root is also a package that you have listed in `lerna.json`. Note that making your root also be a package managed by Lerna should work in most cases, but is not fully tested, and there may be edge cases that don't work.


trusktr:
To clean your root node_modules with `lerna clean`, add a `name` to it in `package.json`, and list the path in `lerna.json`. For example:

```js
{
	"packages": [
		"packages/*", // other packages as usual
		"./" // also the root
	],
}
```

Note that this **will** remove any locally-installed version of `lerna` from your root. If you depend on `lerna` being global, this is not a problem. If you'd like to keep `lerna` local, you can easily work around the issue by doing something like the following in your root `package.json`. Suppose you have an `install` script that bootstraps everything after first cleaning everything, then it may look like this:

```js
{
  "script": {
    "install": "lerna clean --yes --ignore name-of-your-root-package && lerna bootstrap"
  }
}
```

If you want to actually clear the root too, then just run `lerna clean` without the `--ignore` and you're good!

commands/clean/README.md
```diff
@@ -15,4 +15,4 @@ Remove the `node_modules` directory from all packages.
 `lerna clean` accepts all [filter flags](https://www.npmjs.com/package/@lerna/filter-options), as well as `--yes`.
 
 
-> `lerna clean` does not remove modules from the root `node_modules` directory, even if you have the `--hoist` option enabled.
+> `lerna clean` does not remove modules from the root `node_modules` directory, even if you have the `--hoist` option enabled, unless the root is also a package that you have listed in `lerna.json`. Note that making your root also be a package managed by Lerna should work in most cases, but is not fully tested, and there may be edge cases that don't work.
```
trusktr, line 18: Not sure this is the best wording. Also, I'm not sure if there's actually any issue with roots being managed by Lerna (I haven't had any). Are there any known issues?

JamesHenry:
Thanks a lot for contributing to the lerna community, @trusktr! I am sorry that this PR went unaddressed for so long.

I am not going to try and dissuade you from a setup you are happy with, but I think we will stop short of documenting setups which the lerna docs suggest not to create. Therefore I am going to close this one for now.

Thanks again 🙏

PR #3067, opened 2022-01-22
dependabot fix downgrade to node-fetch 2.6.7

Pyrolistical:
node-fetch 3.x requires esm which require changes to support


dreamorosi:
Any chance this can be merged?

PR #3097, opened 2022-03-30
feat(version): add `from-package` bump option

luwes:
Thanks for this great project, we've been using Lerna w/ success over at https://github.com/muxinc/elements
Just a little thing came up with versioning and automation, below explains the issue.

## Description
This change adds a new positional parameter `from-package` to keep the versions that are already in the `version`'s of the mono-repo packages. Definitely not set on the naming, it might be confusing with the same positional parameter in the publish command. Just wanted to create this PR and see if there's interest in this functionality? Thanks

## Motivation and Context
It solves an issue where one wants to set the package versions manually locally or in a PR and then let an automated CI run the actual version command generating tags (could use --amend to reuse commit), and the changelogs and release notes. This can be used with --conventional-commits because the positional takes priority when it comes to calculating the versions.

(I guess another way to tackle this issue is to have a new CLI param that disables the recommended-bump from --conventional-commits)

## How Has This Been Tested?
Yes, I ran the test commands on Mac OS 12.3, Node 16.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/__tests__/version-bump.test.js
+++ b/commands/version/__tests__/version-bump.test.js
@@ -83,7 +83,7 @@ describe("version bump", () => {
 
     await expect(command).rejects.toThrow(
       "bump must be an explicit version string _or_ one of: " +
-        "'major', 'minor', 'patch', 'premajor', 'preminor', 'prepatch', or 'prerelease'."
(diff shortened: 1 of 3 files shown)


joebowbeer:
Addresses #3138 ?

joebowbeer:
I don't mind the reuse of `from-package`

In general, I favor improving lerna's ability to publish from CI/CD

PR #3139, opened 2022-05-24
feat: add experimental support to run tasks via Nx

vsavkin:
Nx is the third option in addition to p-map and p-queue

## Description

Currently running `lerna run` will result in Lerna either delegating to `p-map` or `p-queue` to schedule tasks/processes. Lerna has direct dependencies on both `p-map` and `p-queue`.

The industry has advanced since the time of `p-map` and `p-queue`. Tools like Nx and Lage are much more powerful and efficient task orchestrators. They support computation caching and, in case of Nx, config-free task distribution. A lot of teams are moving away from Lerna for this reason.

This PR adds Nx (which is a MIT-licenced extensible build system) as a third option (in addition to `p-map` and `p-queue`). If you have the nx package installed, and you have `nx.json` in your workspace, Lerna can now delegate its `lerna run` commands to Nx. This is optional: you have to opt into it. And it's done in a backwards-compatible way: every `lerna run` command will work but will be significantly faster and can be distributed across multiple machines without any config.

## How Has This Been Tested?

We have tested on several small Lerna repos. For instance, this repo contains a benchmark measuring cache restoration of different monorepo-based tools: https://github.com/vsavkin/large-monorepo

With this change enabled, Lerna is on par with Nx, and is significantly faster than other tools.

Once the beta with this change is out, we are going to add it on several large enterprise Lerna repos to measure the perf and make sure the change handles all edge cases well.


commands/run/README.md
```diff
@@ -57,7 +58,9 @@ $ lerna run watch --stream
 
 ### `--parallel`
 
-Similar to `--stream`, but completely disregards concurrency and topological sorting, running a given command or script immediately in all matching packages with prefixed streaming output. This is the preferred flag for long-running processes such as `npm run watch` run over many packages.
+Similar to `--stream`, but completely disregards concurrency and topological sorting, running a given command or script
```
juristr, line 61: probably not affecting the rendering of the readme, but this newline didn't seem intentional

juristr (commented):
Just couple of observations

PR #3152, opened 2022-06-03
[Test] Fix github-client tests to not use real env

gund:
While trying to setup custom deployment of Lerna I stumbled upon this funny unit test suite that was using real env instead of the mocked one which was causing it to fail in CI because it had github token variable in the env 🤦‍♂️

So here is the fix for it 👌


JamesHenry:
Thanks a lot for this @gund! The Github UI would not allow me to update your branch, so to save your time I have applied this change directly

PR #3154, opened 2022-06-06
chore: update conventional-changelog-core

JamesHenry:
All `npm audit` warnings specific to the workspace itself will be resolved by this update

NOTE: `npm audit` warnings for userland workspaces have already been resolved

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,6 +3,13 @@
   "extends": ["airbnb-base", "plugin:jest/recommended", "plugin:node/recommended", "prettier"],
   "plugins": ["jest", "node"],
   "ignorePatterns": ["website/**/*"],
+  "settings": {
+    "import/resolver": {
+      "node": {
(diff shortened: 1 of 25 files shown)


JamesHenry:
@fahslaj will pick this up to troubleshoot windows specific failures

fahslaj:
I was able to reproduce this on my personal windows machine: 
image

fahslaj:
I've done some digging, and this commit (https://github.com/conventional-changelog/conventional-changelog/pull/820/commits/6de215a8080a4d59c3de7a7d3317699f75654eff), which updated the `get-pkg-repo` package from version ^1.0.0 to ^4.0.0, was included in the conventional-changelog-core release 4.2.3. After the upgrade, the `repoUrl` is coming back from `get-pkg-repo` as `null` on windows, causing the commit url to be omitted entirely from the written changelog. This causes the test failure observed above.

Issue #2776, opened 2020-12-10
Is there any way to just check if my monorepo doesn't contain circular dependencies between packages? 


adybionka:
I'm aware of "--reject-cycles" argument, which I can pass along with the "bootstrap" command. But my goal is just to check if my monorepo doesn't contain circular dependencies between my packages. Simply lerna could analyze the only package.json of all my packages and nothing else.

The problem with bootstrap is that, doesn't work with yarn client. I can force to use npm client from the command line like this `lerna bootstrap --npm-client=npm --use-workspaces=false --reject-cycles`, but still just to check if I have circular dependencies this the approach seems to be overkill. Analyzing only package.json files could take only a few seconds, instead of a few minutes of performing full bootstrap.

Can lerna do what I need? If not, if there is any plan to add this feature? If not, what is the reason? 

Or maybe somebody can recommend other tools which can do what I need?

imflavio:
@adybionka Take a look at this https://github.com/imflavio/standard-monorepo#standard-monorepo-circular-deps

PR #3168, opened 2022-06-11
fix(bootstrap): reject-cycles when using workspaces

tkloht:
## Description
when the --reject-cycles flag is set and the repo is using workspaces,
run `collapseCycles()` to detect cycles before installing root package dependencies.

## Motivation and Context
Currently the --reject-cycles flag does not have any effect when using workspaces.
Without workspaces it is evaluated during `runLifecycleInPackages()`, but as far as I can tell this code-path is never reached with workspaces.
This would also fix the open issue [2776](https://github.com/lerna/lerna/issues/2776)

## How Has This Been Tested?

added a test-case including fixture for yarn workspace with cyclic dependency
for local testing I have published to verdaccio with `e2e:local-publish`.
output before:

```
❯ npx lerna bootstrap --reject-cycles
lerna notice cli v5.1.1
lerna info bootstrap root only

up to date, audited 604 packages in 754ms

52 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```

output after my change:
```
❯ npx lerna bootstrap --reject-cycles
lerna notice cli v999.9.20
lerna ERR! ECYCLE Dependency cycles detected, you should fix these!
lerna ERR! ECYCLE one -> two -> one
```

with this Lerna config:
```
❯ cat lerna.json
{
  "packages": [
    "packages/*"
  ],
  "version": "0.0.0",
  "useWorkspaces": true
}
```

## Types of changes

## Checklist:

- [ ] All new and existing tests passed. (I had issues with some snapshot tests locally that seem unrelated to my changes)

### DIFF
--- a/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/.yarnrc
+++ b/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/.yarnrc
@@ -1,0 +1,1 @@
+workspaces-experimental true
--- a/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/lerna.json
+++ b/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/lerna.json
@@ -1,0 +1,5 @@
+{
+  "version": "1.0.0",
+  "npmClient": "yarn",
+  "useWorkspaces": true
+}
--- a/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/package.json
+++ b/commands/bootstrap/__tests__/__fixtures__/yarn-workspaces-cyclic/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "root",
+  "private": true,
+  "workspaces": [
+    "workspaces/*"
+  ]
+}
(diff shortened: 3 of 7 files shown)


tkloht:
@cds-amal I ended up writing a small CLI tool to do what the reject-cycles flag does, since there seems to be no-one from nrwl willing to review PRs and I don't think this will ever be fixed in Lerna.
You can use it like `npx cyclic-dependencies --reject`, should be the exact same behaviour as `lerna bootstrap --reject-cycles` (without the bootstrapping part). Code is over here https://github.com/tkloht/cyclic-dependencies

ghiscoding:
The thing is that `lerna bootstrap` should be deprecated in the future, so I'm not sure they want to invest time or feature on that command, see comment from Lerna maintainer [here](https://github.com/lerna/lerna/discussions/3140#discussioncomment-3033079)

tkloht:
I think I agree with evocateur that it should be deprecated, and I think checking for cyclic dependencies doesn't need to be part of yarn/npm/pnpm either. In my opinion a small, dedicated program like the one I wrote is better for that.

All that said, currently bootstrap still exists and has a flag to check for cyclic dependencies. I would expect this to work, but it does not. This PR fixes the reject-cycles flag and adds test coverage, it's frustrating that it's not getting reviewed.

JamesHenry (approved):

JamesHenry:
Thanks for your patience on this folks!

JamesHenry:
And thank you to @tkloht for submitting the fix 🙏

PR #3184, opened 2022-06-16
chore(e2e): Add e2e tests for lerna exec

fahslaj:
Add e2e tests to cover lerna exec.

## Description

Add e2e tests to cover the lerna exec command.

## Motivation and Context

This change adds additional confidence in the consistent performance of lerna over time.

## How Has This Been Tested?

I began with failing e2e test cases, added the appropriate flag to the command run in the test, then confirmed that the test succeeded.

## Types of changes

## Checklist:

### DIFF
--- a/e2e/lerna-exec/lerna-exec.spec.ts
+++ b/e2e/lerna-exec/lerna-exec.spec.ts
@@ -16,7 +16,7 @@ expect.addSnapshotSerializer({
   serialize(str) {
     return str
       .replaceAll(/package-\d/g, "package-X")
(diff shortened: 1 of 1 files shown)


JamesHenry (changes requested):
I don't think we want all tests to use `--silent`, because we are not capturing the common/default behavior that users will experience.

Please update to include at least one non-silent example (we have one in lerna run already)

fahslaj:
> 

I added a test without --silent, but needed to use --concurrency 1 in order to make it deterministic.

JamesHenry (approved):

PR #3188, opened 2022-06-16
chore(e2e): Add e2e tests for global filter options

fahslaj:
Add e2e tests to cover global filter options for the list command.

## Motivation and Context

This change adds additional confidence in the consistent performance of lerna over time.

## How Has This Been Tested?

I began with failing e2e test cases, added the appropriate flag to the command run in the test, then confirmed that the test succeeded.

## Types of changes

## Checklist:


fahslaj:
I skipped the option for [--include-merged-tags](https://github.com/lerna/lerna/tree/main/core/filter-options#--include-merged-tags), since it did not seem to do anything with the list command. The only PR I could find that references it is [when it was introduced four years ago](https://github.com/lerna/lerna/pull/1712). In that example, it is used with `lerna changed`, so I'll try to add a test for it when I get to that command.

Issue #3077, opened 2022-02-10
Lerna wrongly calculate topological order of the dependencies


atvorch:
We have a monorepo configuration using lerna + yarn workspaces in our company.
All our packages are written in Typescript and therefore need to be compiled.
There are interdependencies between packages in the monorepo.

We are building packages with `lerna run build --no-bail --stream --concurrency=1 --include-dependencies`

Build fails because Lerna does not follow the dependency graph. 
To be more correct `lerna list --toposort` gives the wrong topological order.

Therefore it tries to build packages before their dependencies are built.
This [issue](https://github.com/lerna/lerna/issues/2731) was already raised by @drochgenius, but for some reason it was closed without any comments or resolution.

## Expected Behavior
Lerna should execute package scripts following the dependency graph.
I would expect the command `lerna ls --toposort` to output packages according to the dependency graph, dependencies before dependants.

## Current Behavior
Lerna does not follow the dependency graph. 
Output of `lerna ls --toposort` gives the wrong topological order of the dependencies.

## Possible Solution
Fix the algorithm for calculation topological order of the dependencies.

## Context
I won't share the whole output of the `lerna ls --graph -a`, cause it's huge.
Here is the small part of our dependency graph generated with `lerna ls --graph -a` for :
```
 {
    "@company/date-inputs": [
    "@angular/common",
    "@company/icon",
    "@company/input",
    "@company/market-config",
    "@company/models",
    "date-fns",
    "ramda"
  ],
  "@company/market-config": [
    "@angular/common",
    "@angular/core",
    "@nguniversal/express-engine",
    "@company/helpers"
  ]
}
```

Now here is the result of `lerna ls --toposort`:

@company/asset
@company/auth-interfaces
@company/cdk
@company/checkbox
@company/content-portal
@company/date
@company/dynamic-loader
@company/icon
@company/locale
@company/logger
@company/match-media
@company/models
@company/newrelic-interfaces
@company/operators
@company/overlay
@company/portal
@company/rcli
@company/scroll
@company/session-management
@company/sticky
@company/styleguide
@company/test-helpers
@company/types-analytics
@company/action-badge
@company/action-sheet
@company/breadcrumb
@company/counter
@company/dropdown
@company/expansion-panel
@company/full-calendar
@company/header
@company/inc-dec
@company/input
@company/intersection-observer
@company/modal
@company/pax-badge
@company/radio-button
@company/range-slider
@company/recaptcha
@company/select
@company/slider
@company/spinner
[...]

PR #3197, opened 2022-06-22
fix(run): run scoped script in toposort order if some packages miss the script

tkloht:
## Description
Runs a topological sort on packages before filtering for packages that include the script to run.

## Motivation and Context
Given you have three packages A, B C where C depends on B and B depends on A.
When you run a scoped command with lerna like this: `lerna run --scope C --include-dependencies my-script` you'd expect the script to be executed in all packages with the script `my-script`, and this should happen in topologically sorted order (A before B before C).
This works when all packages A, B C have the script `my-script`, but it breaks down when B doesn't have the script.
In the current version of lerna, `my-script` would first be executed in C then in A.
This is a significant issue if you consider `my-script` being a build command for example.

I have a hunch that this might address open issues like #3077, even though it contradicts slightly with the description ("All my packages are written in Typescript and therefore need to be compiled"), but in a bigger repository it seems likely that there might be an intermediate package without a build step.

## How Has This Been Tested?
- added a fixture and test to cover the scenario described above - packages A,B,C, only A&C have the script
- renamed test description from --include-filtered-dependencies to --include-dependencies, to align with current API
- adjust prior test for --include-dependencies to check for ordering

## Types of changes

## Checklist:


JamesHenry:
Hi @tkloht thanks a lot for contributing to lerna! I'm sorry that this PR fell through the cracks while we got up to speed on the project after taking over stewardship.

The modern task runner in Nx has been the default for some time now, and is fully aware of topological ordering concerns and optional scripts on packages. Advanced task pipelines can be configured using an `nx.json` file at the root:

https://lerna.js.org/docs/concepts/task-pipeline-configuration

Therefore I am going to close this PR, as we very much recommend folks use the default, modern task runner and move away from the more naive legacy one (powered by `p-map` and `p-queue`, instead of `nx`)

Many thanks again!

tkloht:
All good, thanks for the response 👍

charlesritchea:
@JamesHenry the Nx backend completely fails to determine topological sort in my project and just runs in lexical order for some reason, quite frustrating

PR #3203, opened 2022-06-24
chore: refactor e2e test implementation

JamesHenry:


e2e/utils/fixture.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+  ): Promise<RunCommandResult> {
+    return this.execImpl(command, { ...opts, cwd: this.fixtureWorkspacePath });
+  }
+
+  /**
+   * Resolve the locally published version of lerna and run the `init` command, with an optionally
+   * provided arguments.
+   */
+  async lernaInit(args?: string): Promise<RunCommandResult> {
```
fahslaj, line 91: Minor suggestion, but if you change this method signature to `lernaInit(args: string = "")` then the `||` isn't needed in the interpolation below: `...init ${args}`

PR #3226, opened 2022-07-06
fix: resolve dependency audit issue

jaranin-b:
To solve dependency audit issue:  https://www.npmjs.com/advisories/1080923 

Screen Shot 2565-07-06 at 20 53 20

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,5 @@
 {
   "name": "lerna-monorepo",
-  "version": "4.0.0-monorepo",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
@@ -35,7 +34,7 @@
         "fs-extra": "^9.1.0",
         "get-port": "^5.1.1",
         "get-stream": "^6.0.0",
-        "git-url-parse": "^11.4.4",
+        "git-url-parse": "^12.0.0",
         "glob-parent": "^5.1.1",
         "globby": "^11.0.2",
         "has-unicode": "^2.0.1",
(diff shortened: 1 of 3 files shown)


JamesHenry (changes requested):
Hi @jaranin-b, thanks for this, when updating dependencies you need to make sure you run an `npm install` at the root of the repo so that the lock file gets updated

jaranin-b:
Thanks @JamesHenry. Updated the lock file.

JamesHenry:
@jaranin-b it seems extremely unlikely that this is correct:

image

Please revert and try again. You should simply need to run `npm install` at the root and be using an appropriate version of npm. Please try with npm `8.12.1` to be safe.

ThisIsMissEm (dismissed):
This change looks good and matches my own investigation into this issue (I've reviewed the changelogs of git-up, git-url-parse and parse-url to see if there's any breaking changes that would affect lerna, and as far as I can tell, there is not)

Note: the package-lock still needs to be fixed.

jaranin-b:
Thanks @ThisIsMissEm @JamesHenry I've updated the lock file that generated using npm@8.12.1

JamesHenry:
@jaranin-b I'm sorry but you are still generating 1000s of lines of change in the lockfile which is definitely unexpected:

image

I have recreated the change in #3231 and the change is appropriately small:

image

In order to land this today, I am going to press ahead with #3231 but thank you very much for taking the initiative to submit this PR!

PR #3229, opened 2022-07-06
chore: Add json schema for lerna.json

fahslaj:
Adds a json schema for lerna.json. 

## Motivation and Context

This change improves the developer experience of Lerna. Now, when modifying lerna.json, users can see what properties are valid and a brief description of what they do (taken from the lerna command options documentation). 

Note: All options for commands (as described in the command docs) are valid configuration options in lerna.json. They can be provided at three levels, in order of priority:
1. Command line inline option
2. lerna.json, under a specific command
3. Top level of lerna.json

The json schema for lerna.json supports supplying options at both level #2 and #3 above.

## How Has This Been Tested?

I tested the schema by changing lerna.json in this repo and observing (in my editor) the metadata from the schema. I **did not** manually test every configuration option for every command. I obtained the metadata from the configuration code for each of the Lerna commands (command.js), with some input from the docs. 

## Types of changes

## Checklist:


JamesHenry (changes requested):
@fahslaj The source for this should be within the lerna package so that it is published and can be referenced by other workspaces not just the lerna repo itself. I had already done that on the initial branch: https://github.com/lerna/lerna/compare/main...lerna-json-schema

Please also add the reference to the schema as part of `lerna create` as I had done on that branch.

NOTE: one thing I hadn't yet done on there was update the `files` array in the lerna package's `package.json`, you will need to do that as well please

schemas/lerna-schema.json
```diff
@@ -0,0 +1,1833 @@
...
+      "items": {
+        "type": "string",
+        "uniqueItems": true
+      },
+      "description": "An array of globs where packages can be located.",
+      "default": ["packages/*"]
+    },
+    "useNx": {
+      "type": "boolean",
+      "description": "When true, Lerna will use the nx task runner.",
```
JamesHenry, line 23:
```suggestion
      "description": "When true, Lerna will delegate the implementation details of `lerna run` to the Nx task runner, instead of p-map or p-queue.",
```

schemas/lerna-schema.json
```diff
@@ -0,0 +1,1833 @@
...
+            "registry": {
+              "$ref": "#/$defs/commandOptions/shared/registry"
+            },
+            "noBootstrap": {
+              "$ref": "#/$defs/commandOptions/add/noBootstrap"
+            },
+            "bootstrap": {
+              "$ref": "#/$defs/commandOptions/add/bootstrap"
+            },
+
```
JamesHenry, line 52: Nit: random newline

schemas/lerna-schema.json
```diff
@@ -0,0 +1,1833 @@
...
+            "sort": {
+              "$ref": "#/$defs/globals/sort"
+            },
+            "maxBuffer": {
+              "$ref": "#/$defs/globals/maxBuffer"
+            },
+            "yes": {
+              "$ref": "#/$defs/globals/yes"
+            },
+
```
JamesHenry, line 166: Nit: random newline

JamesHenry:
Ok there are so many newlines, I'm guessing they are actually not random? 😅

JamesHenry:
> > Ok there are so many newlines, I'm guessing they are actually not random? 😅
> 
> Yes, they are not random. I put a new line between the filter options, global options, and command specific options to make it slightly more readable.

Ah ok fair enough, shame we can't force it to stay like that over time, but then again hopefully this shouldn't need too many major updates anyway!

schemas/lerna-schema.json
```diff
@@ -0,0 +1,1833 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "https://lerna.js.org/docs/getting-started#adding-lerna",
+  "title": "JSON schema for Lerna configuration",
+  "type": "object",
+  "properties": {
+    "version": {
+      "type": "string",
+      "description": "The version of the repository, or \"independent\" for an repository with independently versioned packages.",
```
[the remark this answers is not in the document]
fahslaj: Updated 👍

schemas/lerna-schema.json
fahslaj: Updated 👍

fahslaj:
I have moved the lerna-schema.json inside of core/lerna. It will now be published with the lerna package. Running `lerna init` will generate `lerna.json` with the correct schema reference to `node_modules/.../lerna-schema.json`

PR #3232, opened 2022-07-07
docs: add how to section

isaacplmann:
## Description
Adds a features section in the docs.

- Make sure that every command a user is asked to run is in a copy/paste code block, not inline
- Every code block that is terminal output is labeled with "terminal output"
- Configuration doc moved under API Reference
- added DTE article use Nicole's illustrations


fahslaj (dismissed):

JamesHenry:
Thanks @isaacplmann!

Issue #3233, opened 2022-07-07
package scripts not executed since 5.1.3


simllll:
## Current Behavior

we have a mono repo with about 100 packages, 
running `lerna run build` does not execute all build scripts anymore since 5.1.3.

Even though the output is showing that it executes 100 packages, it does only run it in about half of it.
I tested it by putting "exit 1" inside of a build script of one packgae that wasn't executed, it fails (as supposed to) in 5.1.2, but runs wihtour any errors (and the script is therefore not executed) since 5.1.3.

## Expected Behavior

Should execute the defined script in each package, like the output even says.

## Steps to Reproduce

The issue causing this is this change here: https://github.com/lerna/lerna/pull/2874/files

Removing this check, fixes the issue for me.

### Failure Logs / Configuration

<details><summary>lerna.json</summary><p>

```json
{
	"packages": [
		"packages/*",
		"services/*",
		"docs",
		"infrastructure/*"
	],
	"version": "independent",
	"command": {
		"publish": {
			"message": "chore: publish"
		}
	},
	"hoist": true,
	"forceLocal": true,
	"conventionalCommits": true,
	"reject-cycles": true,
	"ignoreChanges": ["**/__tests__/**", "**/*.md", "**/*.tmp"]
}
```

</p></details>

### Environment

 Environment info:

  System:
    OS: Linux 5.13 Ubuntu 21.10 21.10 (Impish Indri)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
  Binaries:
    Node: 16.15.0 - /usr/local/bin/node
    Yarn: 1.22.11 - /usr/local/bin/yarn
    npm: 8.5.5 - /usr/local/bin/npm
  Utilities:
    Git: 2.32.0 - /usr/bin/git
  npmPackages:
    lerna: ^5.1.3 => 5.1.3

simllll:
I ran into the same thing today again, but now I don't have any circular dependencies.. still not sure what's going on (lerna 5.2.0), but it's for sure that even though the package list on lerna run build --no-private says it has exectued the packages, it does not run the commands.

simllll:
Alright, now I found the culrpitn again, it's still a dependency circle, BUT, since the last fix https://github.com/lerna/lerna/pull/2874 it doesn't show any dependecy circles anymore.
running `lerna bootstrap --npm-client=npm --use-workspaces=false --reject-cycles`
with lerna 5.0.0:
```
[...]

PR #3234, opened 2022-07-07
fix(package-graph): ensure to touch all nodes

simllll:
## Description
the fix of not touching a node twice (https://github.com/lerna/lerna/pull/2874) has a side effect so that not all scirpts are executed anymore.

this fixes it by still improving / keeping the performance benefits of the other pr.

fixes https://github.com/lerna/lerna/issues/3233

### DIFF
--- a/core/package-graph/index.js
+++ b/core/package-graph/index.js
@@ -232,10 +232,11 @@ class PackageGraph extends Map {
       }
 
       // Otherwise the same node is checked multiple times which is very wasteful in a large repository
-      if (alreadyVisited.has(topLevelDependent)) {
+      const identifier = `${baseNode.name}:${topLevelDependent.name}`;
+      if (alreadyVisited.has(identifier)) {
         return;
       }
-      alreadyVisited.add(topLevelDependent);
+      alreadyVisited.add(identifier);
 
       if (
         topLevelDependent === baseNode ||


core/package-graph/index.js
```diff
@@ -232,10 +232,12 @@ class PackageGraph extends Map {
       }
 
       // Otherwise the same node is checked multiple times which is very wasteful in a large repository
-      if (alreadyVisited.has(topLevelDependent)) {
-        return;
-      }
-      alreadyVisited.add(topLevelDependent);
+      const identifier = baseNode.name + ':' + topLevelDependent.name;
+			if (alreadyVisited.has(identifier)) {
+				console.log('topLevelDependent', identifier);
```
JamesHenry, line 237: @simllll Please remove the console log

JamesHenry (changes requested):
Thanks @simllll, is there a failing test we can add first and fix with this?

core/package-graph/index.js
simllll: oha, sorry about that

simllll:
Okay, I was trying to break it down further, and I guess I found the real issue here: a dependency circle caused this behavior. Even though the circle is unrelated to this package. So it seems it's not really a bug, it's just a hard to find "issue". Sorry for the troubles

simllll:
I ran into the same thing today again, but now I don't have any circular dependencies.. still not sure what's going on (lerna 5.2.0), but it's for sure that even though the package list on `lerna run build --no-private` says it has exectued the packages, it does not run the commands.

ghiscoding:
@simllll are you by any chance using the new `useNx`? I might have seem something similar when that option is enabled

simllll:
Hi @ghiscoding , nope, not using this option

ghiscoding:
so does this PR fix your latest issue then? if so I'd be willing to push it to Lerna-Lite

simllll:
It does, but I'm not sure if we losse the performance benefits of the earlier patch with this one again... would be great if the one who created the https://github.com/lerna/lerna/pull/2874 PR can look at this.

ghiscoding:
I can confirm that #2874 is causing side effect, we ran into this on a large project which had 43 packages changed, it was only returning 5 of them, so that's a pretty big side effect. Your code change seems to work, so I'll apply it into Lerna-Lite unless something better comes up

JamesHenry:
Based on the feedback above let's get this merged, we'll need to spend some more time trying to understand these changes better though and have some test coverage.

If there are any other issues that get raised then we will likely have to revert the original change

PR #3250, opened 2022-07-14
docs: add type metadata and reorganize

isaacplmann:
Reorganizes a few docs and adds an invisible `type` metadata field for the docs that we might use later.


website/docs/faq.md
```diff
@@ -1,3 +1,9 @@
+---
+id: faq
+title: FAQ
+type: how-to
```
juristr, line 4: do we want to call this "how-to" or "recipe"? I'm fine with both

website/docs/features/cache-tasks.md
```diff
@@ -9,7 +10,67 @@ title: Cache Task Results
...
 It's costly to rebuild and retest the same code over and over again. Lerna uses a computation cache to never rebuild the
-same code twice. This is how it does it.
+same code twice.
+
+## Setup
+
+Lerna via Nx has the most sophisticated and battle-tested computation caching system. It knows when the task you are
+about to run has been executed before, so it can use the cache to restore the results of running that task.
+
+To enable caching for `build` and `test`, edit the `cacheableOperations` property in `nx.json` to include the `build` and `test` tasks:
```
juristr, line 20: The enabling/configuration part for me would already be somewhat more in a recipe. I think it is fine to leave it here for now, but maybe move it more towards the end?

If I look at the "Cache Task Results" as an explainer, then I'd want to have it like

- How does Caching Work?
- Local Computation Caching _(<< more details here)
- Distributed Computation Caching _(<< more details here)
- What is Cached (because it explains the process-level caching etc as a concept and that is is per project etc.. which is important to understand)
   - Maybe integrate some more about the fact that there are inputs including the src files, runtime and args etc
   - outputs are already mentioned

---

Then come the more "recipe" parts like

- Configuring caching (e.g. the nx.json runner etc and how that differs for local vs remote caching with Nx Cloud)
	- Customizing the cache location
- Configuring Cache inputs (all of these should link to the "references" page with all the options you can have in terms of globs and template variables etc)
   - 	Source Code hash Inputs 
   - Runtime hash Inputs
   - Args Hash Inputs
 - Skipping Caching

---

Not sure what others think. @meeroslav? I'm also still in the process of fine-tuning my understanding of this 😃

juristr (changes requested):

website/docs/faq.md
isaacplmann: good point, I'll switch all the `how-to`s to `recipe`

website/docs/features/cache-tasks.md
meeroslav: I agree with this. The leaner tutorial - the better.

I would push everything that doesn't belong to the "happy path" to recipes. We can/should have a link to the recipe.

e.g. "Check in depth caching configuration guide here >>>"

website/docs/features/version-and-publish.md
```diff
@@ -1,6 +1,7 @@
 ---
 id: version-and-publish
 title: Version and Publish
+type: explanation
```
meeroslav, line 4: Isn't this also tutorial?

juristr: I agree that the intro of this docs page starts as a tutorial, this was before we started looking into categorizing things much better. I'd rather convert this more to an explainer and recipe IMHO.

A tutorial is purely learning oriented providing dedicated exercises for the "learner" to go through. I don't think we have any of those and not sure it's worth the effort to create them for Lerna (given the small API surface).

Looking at it from a Explainer/Recipe perspective, I'd have a division like

**Explainer:**
- Versioning (optionally to be added: general topic on versioning, semver etc..)
- Fixed/Locked mode
- Independent mode

**Recipe**
- How to use locked mode versioning
- How to use independent mode versioning (we could also just have one merged)
- How to publish a package

juristr (approved):

PR #3254, opened 2022-07-21
fix: continue if no script in internal package

tomasbjerre:
## Description

In my use case this code:
```
  if (pkg.__isLernaPackage) {
    pkg = pkg.toJSON();
  }
```
Changes `pkg` so that it no longer contains the `scripts`.

My fix is to add a return, like there already is on line 78:
https://github.com/lerna/lerna/blob/c567a29d352bc97dfc4e3dc5dd0799fc35db2026/utils/run-lifecycle/run-lifecycle.js#L78

## Motivation and Context

I was getting this error:
```
info lifecycle @X/Y@1.2.3-featureXXX-213123~prepack: @X/Y@1.2.3-featureXXX-213123
lerna ERR! TypeError: Cannot read properties of undefined (reading 'prepack')
lerna ERR!     at runLifecycle (C:\X\workspace\X\node_modules\@lerna\run-lifecycle\run-lifecycle.js:122:46)
lerna ERR!     at C:\X\workspace\X\node_modules\@lerna\pack-directory\lib\pack-directory.js:50:28
lerna ERR!     at async C:\X\workspace\X\node_modules\p-pipe\index.js:12:19
ERR! lerna Cannot read properties of undefined (reading 'prepack')
```

When doing:
```
npm exec lerna -- publish prepatch --preid=1.2.3-featureXXX-213123 -m 'chore: ny prerelease %s (refs XX-1234)' --registry=https://asdasd/ --dist-tag=dev --no-push --yes --concurrency 1
```

My prepack script is the culprit here, it manipulates the scripts in package.json:
https://gitlab.com/html-validate/semantic-release-config/-/issues/4

One fix is to not change the scripts in prepack. This PR is another alernative fix that avoid a crash when this happens.

## How Has This Been Tested?

I patched the code live in `node_modules`.

## Types of changes

## Checklist:

### DIFF
--- a/utils/run-lifecycle/run-lifecycle.js
+++ b/utils/run-lifecycle/run-lifecycle.js
@@ -102,6 +102,10 @@ function runLifecycle(pkg, stage, options) {
     // we must pass the _actual_ JSON instead of our fancy Package thingy
     // eslint-disable-next-line no-param-reassign
     pkg = pkg.toJSON();
+    if (!pkg.scripts || !pkg.scripts[stage]) {
+      opts.log.info("lifecycle", "No script for internal package %j in %j, continuing", stage, pkg.name);
+      return Promise.resolve();
+    }
   }
 
   // _id is needed by @npmcli/run-script


ghiscoding:
I wonder if this conflicts or does something similar to another PR #3197, I didn't try any of the two PRs but the other PR seem to have the advantage to still follow a topological order even if a dependent script is missing while in your PR I wonder what would happen. Perhaps a merge of both PRs might be useful.. but again I didn't try any of the two so I'm not sure what is best

JamesHenry:
@tomasbjerre I can see that you have had further discussion with @ext (hopefully that's the right github user!) on the Gitlab issue and changed your approach with the tooling a little.

Would one of you mind summarizing where you have got to and if there are any issues with lerna that need to be addressed?

tomasbjerre:
I think you should change Lerna like in this PR. Or change it something like this:

```js
    if (!pkg.scripts || !pkg.scripts[stage]) {
      throw `No script for internal package ${stage} in ${pkg.name}`;
    }
```

The error that currently occurs is not pretty.

ghiscoding:
on the other hand, I just stumble across this piece of code in the publish command which runs the `execScript` and their handling of missing script is different, the context might be different but it's still something to review for comparison sake

https://github.com/lerna/lerna/blob/5cfcd3a3209be547c3f259220453e5fd620fa10b/commands/publish/index.js#L595-L606

ext:
> @tomasbjerre I can see that you have had further discussion with @ext (hopefully that's the right github user!) on the Gitlab issue and changed your approach with the tooling a little.
> 
> Would one of you mind summarizing where you have got to and if there are any issues with lerna that need to be addressed?

From my point of view Lerna is doing something different to how NPM handles this. I am not sure if what the `@html-validate/release-scripts` is doing is up to spec or not but it works fine with vanilla NPM.

* `prepublishOnly` modifies `package.json` before publishing, among other things it removes the `script` block entirely as it is just cruft that an end user of the package does not need (or can use in any meaningful way).
* `postpublish` restores `package.json` to its original version.

With `npm publish` in a non-monorepo setup this works as expected:

1. `prepublishOnly` strips `package.json` before publish, i.e. the published `package.json` (and tarball contents) are stripped.
2. `postpublish` restores `package.json` so the working copy is unmodified.
3. `git status` after `npm publish` shows no modifications.

So for this to work I can infer what NPM is doing is that it reads and caches the contents of `scripts` (and I could very well be playing with fire aka undefined behavior here but it works with all current supported Node/NPM versions). If it wouldn't cache the contents of `scripts` I would not expect `postpublish` to work (i.e. `package.json` would not be restored properly).

I believe Lerna should try to replicate this behavior but if that is not possible this PR makes the situation better as it would no longer crash as it currently is.

For ease of use this is also duplicated for `prepack` and `postpack`, e.g. if the user runs `npm pack` the `package.json` inside the tarball is stripped but restored afterwards. For this to work the scripts looks at two environmental variables:

* `npm_command` (`publish` when `npm publish` is run or `pack` when `npm pack` is run)
* `npm_lifecycle_event` (`prepack`, etc)

I haven't dug deep into Lerna but it seems these are not properly set either (but I could be very wrong on this so please correct me if I'm wrong) and these are well defined in the [documentation](https://docs.npmjs.com/cli/v8/using-npm/scripts#environment).

TL;DR, I think Lerna should update how the scripts are fetched and run. Maybe look into using [@npmcli/run-script](https://github.com/npm/run-script) as well?

ext (approved):

PR #3255, opened 2022-07-22
feat(init): default useNx and useWorkspaces to true for new lerna workspaces

fahslaj:
When initializing lerna for the first time with `lerna init`, workspaces and nx will both be enabled by default. 

## Description

For lerna init:
- default useNx to true
- default useWorkspaces to true
- automatically add nx as a dev dependency if useNx will be true
- automatically add workspaces property to package.json if useWorkspaces will be true
- explicitly set useNx to false if lerna.json already exists and useNx is not set
- explicitly set useWorkspaces to false if lerna.json already exists and useWorkspaces is not set
- fix a bug where useNx was being reset to false whenever lerna init was run on an existing workspace with useNx: true
- refactor ensurePackageJson() and ensureLernaConfig() into one ensureConfig() function to reduce surface area for bugs in configuration that spans both config files (such as workspaces)
- add an informational message upon successful initialization that links the user to the lerna docs

For all commands:
- add a warning to the console when a command is run and lerna does not have either:
  - useWorkspaces: true in lerna.json AND a workspaces property set in package.json
  - packages property set in lerna.json
- add a warning to the console when workspaces are defined in package.json, but lerna.json does not have useWorkspaces: true

For the docs:
- updated the getting started guide to no longer describe needing to install nx manually on a newly created lerna project

For `lerna --help`
- update the link to point to the new docs

## Motivation and Context

This change enables two of lerna's most powerful features by default - the ability to run commands with nx and the ability to make use of Yarn/npm workspaces by default for detecting packages.

## How Has This Been Tested?

Extensive unit, integration, and e2e tests have been added. I have also manually tested the lerna init command for each of these scenarios.

## Types of changes

## Checklist:

### DIFF
--- a/commands/init/index.js
+++ b/commands/init/index.js
@@ -131,8 +131,8 @@ class InitCommand extends Command {
       version = "0.0.0";
     }
 
-    let useNx = config.useNx ?? false;
-    let useWorkspaces = config.useWorkspaces ?? false;
+    let useNx = config.useNx === true;
+    let useWorkspaces = config.useWorkspaces === true;
 
     if (!this.hasExistingLernaConfig) {
       this.logger.info("", "Creating lerna.json");
--- a/commands/init/index.js
+++ b/commands/init/index.js
@@ -110,10 +110,6 @@ class InitCommand extends Command {
 
       setDependency({ name: "lerna", version: this.lernaVersion });
 
-      if (useNx) {
-        setDependency({ name: "nx", version: "14.4.3" });
-      }
-
       return rootPkg.serialize();
     });
 
--- a/commands/init/index.js
+++ b/commands/init/index.js
@@ -48,6 +48,7 @@ class InitCommand extends Command {
 
     return chain.then(() => {
       this.logger.success("", "Initialized Lerna files");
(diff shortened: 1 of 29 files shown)


commands/init/index.js
```diff
@@ -50,46 +51,65 @@ class InitCommand extends Command {
...
-          { indent: 2 }
-        )
-      );
+      chain = chain.then(() => {
+        const pkg = {
+          name: "root",
+          private: true,
+        };
+
+        if (!this.hasExistingLernaConfig) {
```
fahslaj, line 71: The more I look at this logic, the more I realize that the `ensurePackageJSON` and `ensureLernaConfig` functions should be combined. Both config files are dependent on each other, such as for workspaces support.

JamesHenry (changes requested):
Per slack discussion

JamesHenry (approved):

AviVahl:
This installs `nx` and its many dependencies (some of which are requesting exact versions) to any project using lerna, even if they don't use `nx`. :-1: 

Even stuff that I wouldn't expect like `tsconfig-paths` and `@parcel/watcher` are now being installed.

PR #3258, opened 2022-07-22
docs: slim down features docs

isaacplmann:
The Features docs are slimmed down and treated as recipes and a starting point to learn about each feature.

The in depth discussions are moved under Concepts.


juristr (approved):
I love the new structure. Much better!

PR #3263, opened 2022-07-26
feat(publish): include all dependencies in package graph by default, allow no-sort

JamesHenry:
Users can still opt into the old behavior of ignoring `devDependencies` until v6 by passing `--graph-type=dependencies`, or setting it in their `lerna.json`.

---

Note: For the unit test _updates_ it's down to the fact that in the relevant fixture, `package-3` has a devDependency on `package-2`, so it now comes after it in the ordering.


vsavkin:
LGTM

PR #3291, opened 2022-08-13
Update cache-tasks.md

fakership:
change docs

## Description

when i execute command as the docs, i find i can not get the cache result, it always execute as the new program, util i find i should set useNx: true in the example.
so, i think should add the tip to avoid the same thing happen.
## Types of changes

## Checklist:


fakership:
i think this commit is necessary, while the config exist in other page。

PR #3295, opened 2022-08-16
fix(version): Ignore npm lifecycle scripts on package lock update

fahslaj:
Fixes an issue where npm install lifecycle scripts were being run during the version command.

## Description

This change passes the --ignore-scripts option to the `npm install` command used to update the lockfile during the version command. 

## Motivation and Context

This prevents npm lifecycle scripts from running, which can cause CI workflows to fail.

## How Has This Been Tested?

An end to end test has been created to cover the case of having failing lifecycle scripts during a version command. I also manually tested the change against a local environment with install scripts and git commit hooks (with [husky](https://typicode.github.io/husky/#/)).

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/index.js
+++ b/commands/version/index.js
@@ -618,9 +618,11 @@ class VersionCommand extends Command {
       if (fs.existsSync(lockfilePath)) {
         chain = chain.then(() => {
           this.logger.verbose("version", "Updating root package-lock.json");
-          return childProcess.exec("npm", ["install", "--package-lock-only"], this.execOpts).then(() => {
-            changedFiles.add(lockfilePath);
-          });
+          return childProcess
+            .exec("npm", ["install", "--package-lock-only", "--ignore-scripts"], this.execOpts)
+            .then(() => {
+              changedFiles.add(lockfilePath);
+            });
         });
       }
     }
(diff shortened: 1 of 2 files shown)


sluramod:
This actually creates an issue for us. Is there a way to use lerna.json configuration to specify whether or not use --ignore-scripts ?

fahslaj:
@sluramod Running `npm install --package-lock-only` to update the root package-lock.json file is a very recent addition to the implementation details of the version command. It was never meant to run normal install lifecycle scripts, so this PR removed an unintended side-effect. If you need to run additional scripts during or after `lerna version`, the proper way to do so is via [version lifecycle scripts](https://github.com/lerna/lerna/tree/main/commands/version#lifecycle-scripts).

If version lifecycle scripts do not solve your problem, then could you please share more information about your use case? (Specifically, the scripts that need to be run and why they must be done as part of the implementation details of the `lerna version` command)

Another note - this PR only affects the `version` command. Scripts will still be run during a normal `npm install` and during `lerna bootstrap`.

sluramod:
@fahslaj I use package `firebase-tools` from google. when I run `npm install` I get a bunch of dependencies with `"extraneous": true` from that package. When I run `lerna version` from 5.4.3 all of those extraneous dependencies get removed from `package-lock.json`. To fix this I need to run `npm install` after I run `lerna version`, amend lerna's commit with updated `package-lock.json` and move the tag that lerna assigned to the amended commit. 
As you can see it's not very convenient and I can't force the change in third-party package.

PR #3302, opened 2022-08-21
feat(repair): add lerna repair command

JamesHenry:
## Description

Adds a new command: `lerna repair`. This command will run any available migrations which have been designed to bring a lerna workspace up to date with the latest and greatest.

This PR does not ship with a real implementation of such a migration, so instead a `noop` example migration is provided to concretely illustrate how a migration is both written and unit tested within the `lerna` package.

NOTE: the implementation for the `repair` command is intentionally inlined into the `lerna` package. Because of the way migrations are orchestrated, having `repair` be its own package would result in a cyclical dependency between it and the core package. Plus, in general, we are not looking to continue with the pattern of having an ever-increasing number of tiny published packages. 

## Types of changes

## Checklist:


fahslaj (approved):
Looks good. Looking forward to this command!

evocateur:
@JamesHenry Seems like this should have been placed in `/commands/repair`, not `/core/lerna/commands/...`?

/me goes back into cave

PR #3303, opened 2022-08-21
feat(filter-options): add `--ignore-dev-dependencies`

joselucasoliveira:
## Description

This feature adds the filter option `--ignore-dev-dependencies` to exclude all transitive devDependencies when running a command regardless of `--include-dependencies` or `--include-dependents`.

## Motivation and Context

This feature is useful when creating custom and deployment scripts as it makes it possible to filter code dependencies from development environment dependencies.

## How Has This Been Tested?

Changes were manually tested in multiple test workspaces. Comprehensive end-to-end and unit tests have also been added around the new feature. The changes don't affect other areas of the code.

## Types of changes

## Checklist:

### DIFF
--- a/core/filter-options/README.md
+++ b/core/filter-options/README.md
@@ -92,3 +92,7 @@ $ lerna exec --since --include-merged-tags -- ls -la
 ```
 
 Include tags from merged branches when running a command with `--since`. This is only useful if you do a lot of publishing from feature branches, which is not generally recommended.
+
+### `--ignore-dev-dependencies`
+
+Exclude all transitive devDependencies when running a command regardless of `--include-dependencies` or `--include-dependents`.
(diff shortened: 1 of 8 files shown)


DominicFarrington:
@joselucasoliveira the team I work on has exactly the same problem.

@bcabanes we need a feature like this to prevent lerna from installing dev dependencies, is it possible that this could get reviewed by someone?

JamesHenry:
Hi @joselucasoliveira, thanks a lot for contributing to lerna!

We are currently working on an overhaul of the graph which gets constructed behind the scenes which is relevant here. Once we have done that, improvements to the introspection capabilities for the graph is definitely a priority and so we will be open to implementing something like that then.

As it stands, because of the way lerna's filtering was implemented this change would apply this flag to all of the core commands, including where it is inappropriate.

Therefore, I am going to close this PR for now, and as I say we will be happy to revisit this use-case for `lerna list` in a month or two once our graph work is complete.

Thanks again!

Issue #3298, opened 2022-08-18
The lerna was not build all the packages as expected


blood-romantic:
## Current Behavior

- The lerna builds failed, if the `import { Header } from 'db-header';` was added to the top of the App.js that in db-main repo.
- The lerna builds success, if the `import { Header } from 'db-header';` was not added to the top of the App.js that in db-main repo.

[NOTE]: the `db-header` is a dependence of the `db-main`, and the `npx lerna bootstrap` has been executed.

## Expected Behavior

- The lerna builds success, if the `import { Header } from 'db-header';` was added to the top of the App.js that in db-main repo.
- The lerna builds success, if the `import { Header } from 'db-header';` was not added to the top of the App.js that in db-main repo.

## Steps to Reproduce

- add the dependence `import { Header } from 'db-header';` to the top of the App.js that in db-main repo.

- run the command:  `npx lerna run build`

The repo: https://github.com/blood-romantic/monorepo

### Failure Logs / Configuration

```txt

    ✔  db-header:build (3s)

    ✖  db-main:build
       > db-main@0.1.0 build
       > react-scripts build
       
       Creating an optimized production build...
       Failed to compile.
       
       Module not found: Error: Can't resolve 'db-header' in '/Users/gwang2/projects/Dashboard/packages/db-main/src'
```

<details><summary>lerna.json</summary><p>

```json
{
  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
  "useNx": true,
  "useWorkspaces": true,
  "version": "0.0.0"
}

```

</p></details>

<details><summary>lerna-debug.log</summary><p>

</p></details>

### Environment

lerna notice cli v5.4.3

 Environment info:

  System:
    OS: macOS 12.1
    CPU: (10) arm64 Apple M1 Pro
  Binaries:
    Node: 16.14.0 - /usr/local/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.3.1 - /usr/local/bin/npm
  Utilities:
    Git: 2.15.0 - /usr/local/bin/git
  npmPackages:
    lerna: ^5.4.3 => 5.4.3

ghiscoding:
possible duplicate of #3289

JamesHenry:
Hi @blood-romantic this should be resolved in 5.5.1 thanks to @fahslaj

Issue #3289, opened 2022-08-12
inconsistent behavior of lerna run scope


YousefED:
## Steps to Reproduce
- lerna init
- add two packages (a and b), both with the same "dev" command
- make package a depend on b
- run `npx lerna run --scope a dev --stream`

## Current Behavior
if `useNx` is set, it will run the command in a and b

## Expected Behavior

if `useNx` is not set, it will run the command in a only

James-Reilly:
+ 1 also seeing this behavior after updating

MirKml:
I think, without Nx, `lerna run --scope a dev` runs always only `dev` command in `a` package.
If you want to include dependencies, it's necessary to use `--include-dependencies` flag. It is and was pure lerna default, I think.

Maybe Nx presence changed this behaviour.

ifatregev:
**+1**
I also tried to integrate Lerna with NX using the `useNx` flag, and see that `lerna run` behaviour changed.
I want to run only the project which is specified with the `scope` flag, **without dependencies.**

ifatregev:
@YousefED can you please mention in the title that this is happening upon integration with NX?

AlonMiz:
+1 we expect some kind of a flag to exclude dependencies, or even better, preserve the previous behaviour from lerna...

JamesHenry:
Hi folks, this should be resolved in 5.5.1 thanks to @fahslaj

PR #3316, opened 2022-09-02
fix(run): Exclude dependencies with --scope when nx.json is not present

fahslaj:
When nx.json is not present, packages being run will no longer include their dependencies when those dependencies would be filtered out by --scope. Using the flag `--include-dependencies` will cause them to be included again, as expected.

## Description

This preserves the default behavior of `lerna run --scope` when `useNx` is `true` and nx.json does not exist.

## Motivation and Context

https://github.com/lerna/lerna/issues/3289
https://github.com/lerna/lerna/issues/3298

## How Has This Been Tested?

End to end tests have been added that cover this behavior. The functionality was also manually tested.

## Types of changes

## Checklist:

### DIFF
--- a/commands/run/command.js
+++ b/commands/run/command.js
@@ -73,6 +73,11 @@ exports.builder = (yargs) => {
         hidden: true,
         type: "boolean",
       },
+      verbose: {
+        group: "Command Options:",
+        describe: "When useNx is true, show verbose output from dependent tasks.",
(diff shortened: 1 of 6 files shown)


JamesHenry (changes requested):
Per discussion about verbose

JamesHenry (approved):

PR #3327, opened 2022-09-15
fix(publish): ensure zero exit code when EWORKINGTREE warning occurs

pengzai-dev:
fixed  error when run "lerna publish from-package"  in package.json scripts
## Description

when verifyWorkingTreeClean error happend set exitCode to 0
## How Has This Been Tested?
1、add  publish:'lerna publish from-package" to package.json->"scripts"
2、exec "npm run publish" in terminal
3、everything well done

## Types of changes

## Checklist:

### DIFF
--- a/integration/__tests__/lerna-publish-relative-file-spec.spec.ts
+++ b/integration/__tests__/lerna-publish-relative-file-spec.spec.ts
@@ -46,8 +46,8 @@ index SHA..SHA 100644
 --- a/packages/package-1/package.json
 +++ b/packages/package-1/package.json
 @@ -3 +3 @@
--	"version": "1.0.0",
-+	"version": "2.0.0",
+-  "version": "1.0.0",
++  "version": "2.0.0",
 diff --git a/packages/package-2/package.json b/packages/package-2/package.json
 index SHA..SHA 100644
 --- a/packages/package-2/package.json
--- a/integration/__tests__/lerna-publish-transitive.spec.ts
+++ b/integration/__tests__/lerna-publish-transitive.spec.ts
@@ -47,8 +47,8 @@ index SHA..SHA 100644
 --- a/packages/package-1/package.json
 +++ b/packages/package-1/package.json
 @@ -4 +4 @@
--	"version": "1.0.0",
-+	"version": "2.0.0",
+-  "version": "1.0.0",
++  "version": "2.0.0",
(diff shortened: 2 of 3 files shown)


pengzai-dev:
fixed

sirctseb:
@juristr we recently upgraded from version 3.22.1 to version 6. When using `lerna publish from-package --no-git-reset`, we get the issue described here: the publish succeeds, but the process exit code is 128, so our build containers are marked as failures.

Some more context:
1. The issue occurs when there is no .git directory (as it the case in CI, since we don't include the .git directory in the image), and can be reproduced locally by simply removing or renaming the .git directory
2. The issue is introduced in version 4.0.0. Version 3.22.1 works just fine. There are a lot of commits in that release, I haven't found a way to git bisect because I can't `yarn link` lerna directly into my project, and I haven't identified any suspicious changes in the [publish implementation that would explain it](https://github.com/lerna/lerna/compare/f6e7a13e60fefc523d701efddfcf0ed41a77749b..4582c476e07dddddd6b2e3ab6e7f52c1f9eed59a#diff-7707787f29cb401ea4188ed1153fe9cb08c74aa9d82091bd54ac2a7a51999e93), however:
3. the change in this PR resolves our problem.

Do you think this PR should be accepted?

juristr:
@sirctseb I'll bring this issue to the team's attention so we can look into it. There are a couple of potential issues we've already seen that might occur when there's no Git repo set up because Lerna requires Git for many of its operations. Some are legit because the Git information is needed as part of the operation, but for others it should not be required.

sirctseb:
Thanks @juristr. Let me know if we can help with any additional information from our use-case. We do not need (or want) our .git directory in the docker image we us in CI when we call `lerna publish from-package` and my understanding is that that operation does not require any knowledge of git for that operation, with evidence from the fact that version 3 worked with no issue and this message from the code:
> this.logger.notice("FYI", "Unable to verify working tree, proceed at your own risk");

We're exploring workarounds like using `patch-package` to include the change from this PR in our local installation, but would rather get correct working behavior back into your repo.

Thanks!

JamesHenry:
Thank you @pengzai-dev!

JamesHenry:
Thanks for your patience on this, it slipped my mind completely - we will cut a release this week for sure

PR #3332, opened 2022-09-17
fix(run): missing `fs-extra` dependency declaration

fupengl:
## Description

The `fs-extra` dependency is missing when `lerna run` is executed. 
This dependency must be declared and installed

## Types of changes

## Checklist:

### DIFF
--- a/commands/run/package.json
+++ b/commands/run/package.json
@@ -41,6 +41,7 @@
     "@lerna/run-topologically": "file:../../utils/run-topologically",
     "@lerna/timer": "file:../../utils/timer",
     "@lerna/validation-error": "file:../../core/validation-error",
-    "p-map": "^4.0.0"
+    "p-map": "^4.0.0",
+    "fs-extra": "^9.1.0"
   }
 }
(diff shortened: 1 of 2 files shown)


fupengl:
> Thanks @fupengl, but what issue/error have you actually observed please?
> 
> `@lerna/run` depends on `@lerna/profiler` which already depends on `fs-extra` so the dependency is already transitively available.
> 
> I understand the desire for correctness, and that's fine, but it would first be good to understand if this actually caused any issues in practice

If the non-dependency upgrade is used, the module will still not be found. It is recommended to add it. The currently added version is the same as the version number that @lerna/profiler depends on, and there will be no additional impact.

fupengl:
> Thanks @fupengl, but what issue/error have you actually observed please?
> 
> `@lerna/run` depends on `@lerna/profiler` which already depends on `fs-extra` so the dependency is already transitively available.
> 
> I understand the desire for correctness, and that's fine, but it would first be good to understand if this actually caused any issues in practice

fupengl:
My problem is that the dependencies installed through pnpm show that the package cannot be found when executing the run command

JamesHenry:
@fupengl ok cool thanks, I wondering if it might turn out to be a pnpm thing... We can definitely merge this once it's green

JamesHenry (approved):

fupengl:
If it is a case of non-dependent promotion, it needs to be solved by lerna

Issue #3336, opened 2022-09-21
Little regression in run command with useNx: true


ruggi99:
## Current Behavior

I don't have a `nx.json` file.

With lerna `5.5.2` if I run `npx lerna run build` for example it doesn't run in topological order (so dependencies before dependents) but runs in parallel mode. I tried also using `--include-dependencies` and still works in parallel mode.

With lerna `5.4.3` it used to run in topological order even if `nx.json` file was not present. No parallel mode unless specified.

I saw this PR that changed this behavior #3316.

My question is:
Is that by design?

If I have never got a `nx.json` file and I use `"useNx": true` in `lerna.json`, is it right to run in parallel mode instead of topological order by default (as said in previous version it wasn't like this)?

Thanks

## Expected Behavior

I think it's a little regression. 
I expect `run` command to run in topological order if `"useNx": true` in `lerna.json` and no `nx.json` file.
Also because in the PR it is stated that the current behaviour should be preserved, if I haven't misread

## Steps to Reproduce

Create a project with 3 packages. The first with no dependencies, the others with the first ads dependence. Use lerna.json file as above. If I run a script (above there's an example) in all three packages I would expect to run the script in the first package and then in the others.

### Failure Logs / Configuration

<details><summary>lerna.json</summary><p>

```json
{
  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
  "useNx": true,
  "useWorkspaces": true,
  "version": "0.7.0"
}
```

</p></details>

### Environment

System:
    OS: Windows 10 10.0.19044
    CPU: (6) x64 AMD FX(tm)-6300 Six-Core Processor
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.15.1 - C:\Program Files\nodejs\npm.CMD
  Utilities:
    Git: 2.37.2.
  npmPackages:
    lerna: ^5.4.3 => 5.5.2

alasdairhurst:
I saw this error as well. Fixed by adding nx.json

But it seems in the latest version (5.5.4) even with an nx.json file created, it now runs in alphabetical order again, rather than respecting the dependencies and implicitDependencies.

ruggi99:
I temporarily fixed this issue the same day I opened this.
I forgot to update the issue.
I fixed by disabling entirely nx in `lerna.json` file.

[...]

PR #3345, opened 2022-09-27
fix(run): fully defer to Nx for dep detection when nx.json exists

fahslaj:
## Description

- Only sends "targetDependencies" from Lerna to Nx when nx.json does not exist.
- Updates the warning when using `--include-dependencies` to indicate that Lerna will send Nx additional dependencies to run. 
- Updates docs to be more clear on the differing behavior.
- Adds a note to the docs when installing Nx that `lerna run` will behave differently and links to the page describing the differences.

## How Has This Been Tested?

This change has been covered by e2e tests and tested manually.

## Types of changes

## Checklist:

### DIFF
--- a/commands/run/index.js
+++ b/commands/run/index.js
@@ -219,15 +219,10 @@ class RunCommand extends Command {
   }
 
   prepNxOptions() {
-    const { readNxJson } = require("nx/src/config/configuration");
-    const nxJson = readNxJson();
     const nxJsonExists = existsSync(path.join(this.project.rootPath, "nx.json"));
-    const useParallel = this.options.parallel && !nxJsonExists;
 
-    const targetDependenciesAreDefined =
-      Object.keys(nxJson.targetDependencies || nxJson.targetDefaults || {}).length > 0;
     const targetDependencies =
-      this.toposort && !useParallel && !targetDependenciesAreDefined
+      this.toposort && !this.options.parallel && !nxJsonExists
         ? {
             [this.script]: [
               {
@@ -250,7 +245,7 @@ class RunCommand extends Command {
        * To match lerna's own behavior (via pMap's default concurrency), we set parallel to a very large number if
        * the flag has been set (we can't use Infinity because that would cause issues with the task runner).
        */
-      parallel: useParallel ? 999 : this.concurrency,
+      parallel: this.options.parallel && !nxJsonExists ? 999 : this.concurrency,
       nxBail: this.bail,
       nxIgnoreCycles: !this.options.rejectCycles,
       skipNxCache: this.options.skipNxCache,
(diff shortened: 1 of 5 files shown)


alasdairhurst:
Getting a regression in 5.5.4 where i have an nx.json (default file) and suddenly the order of `lerna run` (`lerna run build`) is back to being alphabetical rather than topological order after updating lerna.

JamesHenry:
@alasdairhurst please open an issue with a reproduction

alasdairhurst:
{
	"tasksRunnerOptions": {
		"default": {
			"runner": "nx/tasks-runners/default",
			"options": {
				"cacheableOperations": []
			}
		}
	}
}

alasdairhurst:
> @alasdairhurst please open an issue with a reproduction

We have quite a complex setup (There's 46 modules with a number of interdependencies.) but i'll see if i can find time to set up something and try to build a reproduction to help you out if you're not sure about the root of the issue yet. Since nobody else has reported the issue surprisingly, it seems like it could be slightly more in depth than i was expecting.

Let me know if there's any other info/logs/config that could help you pinpoint the issue that i can provide from my current monorepo.

JamesHenry:
Ah right, no it's ok @alasdairhurst you were right in your initial point. This change here has made it so that the presence of `nx.json` is considered a more significant "buy in" to wanting the benefits of Nx's much smarter task runner.

Nx's task runner doesn't just understand dependencies at a package/project level, it also understands them at a script (a.k.a "target") level.

Therefore when the `nx.json` is present we now expect folks to configure the target relationships (i.e. declare which targets need to be run in order). This is achieved via the `targetDefaults` property in nx.json - which your config doesn't yet have.

`targetDefaults` is covered in this section of the docs: https://lerna.js.org/docs/getting-started#target-dependencies-aka-task-pipelines.

For example, if you have a `build` script in all your packages, here is how you would tell the task runner that `build` scripts need to be run in topological order:

**nx.json**
```json
{
  ...
  "targetDefaults": {
    "build": {
      "dependsOn": [
        "^build"
      ]
    }
  }
}
```

(`^` means "from dependencies")

We are adding a helper command to the lerna CLI which will give you an interactive prompt to help you generate this configuration (and will then update the docs accordingly once that lands), but until then you'll just need to configure your nx.json based on the above.

Please let me know if you face any issues with that and I can help out!

alasdairhurst:
Thanks a lot for the detailed description! I'll look into applying this as you suggested.

The concern i have are that these are probably breaking changes, and i'm not sure a case like this would count as an edge case. I'll explain how i recall that got to this configuration.

I updated to the latest version of lerna when nx got integrated. At this point, i was able to install nx, set "useNx" to true in lerna.json and everything worked as per the migration guide (only followed the required steps), and get the cool new behavior for `lerna run` out of the box. At this point nx.json was not required.

A few weeks ago, the change in behavior i'm seeing currently suddenly appeared (build order changed and everything started running in parallel without taking into account dependencies).  running "nx init" (had to run it in a different project since it was already installed in my monorepo) allowed us to generate the default nx.json as i posted above, and this returned to the previous `lerna run` behaviour.  I believe it was this  (just found it today):  https://github.com/lerna/lerna/issues/3336

Then it seems like adding the nx.json as a solution was a double-edged sword since the behavior reverted again in the last release and impacts when you have the (default) nx.json :)

Note that the nx.json created from nx init, and the nx.json in the lerna migration guide (https://nx.dev/recipe/lerna-and-nx) doesn't include targetDefaults, but at the time that I upgraded, this was the default behavior, now it won't be for users following the same process, and i presume by default if they follow the optional step to create nx.json and leave it as-is then `nx run` probably won't work as expected.

I'll leave the rest up to you guys. Just letting you know my experience and hopefully it is useful.

JamesHenry:
@alasdairhurst thank you for the feedback, we do agree and want to help bridge the gap, so Austin is working on a refinement to it here: https://github.com/lerna/lerna/pull/3349

The new `add-caching` command is also in draft here: https://github.com/lerna/lerna/pull/3350

PR #3348, opened 2022-09-29
feat(version): apply prettier to updated files, if applicable

JamesHenry:
If a user has prettier installed in their workspace, apply their prettier config to the files updated by the `lerna version` command.

The prime example of why this is useful is right here in the lerna repo itself. After this PR, we will no longer have to do a follow up commit after every release.


commands/version/lib/git-add.js
```diff
@@ -1,22 +1,56 @@
...
 module.exports.gitAdd = gitAdd;
 
+function maybeFormatFile(filePath) {
+  let prettier;
+  try {
+    // If the workspace has prettier installed, apply it to the updated files
+    const prettierPath = path.join(workspaceRoot, "node_modules", "prettier");
+    // eslint-disable-next-line import/no-dynamic-require, global-require
+    prettier = require(prettierPath);
+  } catch {
```
fahslaj, line 19: This block could be done only once outside of this function, since it only looks in the workspace root.

PR #3350, opened 2022-09-29
feat(core): add add-caching command

JamesHenry:


core/lerna/commands/add-caching/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+    process.stdout.write("\n");
+
+    this.convertAnswersToNxConfig({ cacheableOperations, targetDefaults, scriptOutputs });
+
+    this.logger.success("add-caching", "Successfully updated task runner configuration in `nx.json`");
+
+    this.logger.info(
+      "add-caching",
+      "Learn more about task runner configuration here: https://lerna.js.org/docs/concepts/task-pipeline-configuration"
```
vsavkin, line 99: I'd also add a note saying --sort, --no-sort and --parallel no longer work and link to the doc we link from the warning.

core/lerna/package.json
```diff
@@ -61,6 +61,7 @@
     "@lerna/version": "file:../../commands/version",
     "@nrwl/devkit": ">=14.8.1 < 16",
     "import-local": "^3.0.2",
+    "inquirer": "^8.2.4",
```
vsavkin, line 64: enquirer is much smaller than inquirer but works the same way

core/lerna/commands/add-caching/index.js
```diff
@@ -0,0 +1,157 @@
...
+    );
+    process.stdout.write("\n");
+
+    log.pause();
+
+    const { targetDefaults } = await inquirer.prompt([
+      {
+        type: "checkbox",
+        name: "targetDefaults",
+        message: "Which of the following scripts need to be run in deterministic/topoglogical order?\n",
```
vsavkin, line 54: Do you think it's worth giving an example: "E.g., builds are but tests are not"

core/lerna/commands/add-caching/index.js
```diff
@@ -0,0 +1,157 @@
...
+        choices: this.uniqueScriptNames,
+      },
+    ]);
+
+    const { cacheableOperations } = await inquirer.prompt([
+      {
+        type: "checkbox",
+        name: "cacheableOperations",
+        message:
+          "Which of the following scripts are cacheable? (Produce the same output given the same input)\n",
```
vsavkin, line 64: You think it's wroth giving an examples (e.g., builds, tests and lints are, serve and start are not)

core/lerna/commands/add-caching/index.js
```diff
@@ -0,0 +1,157 @@
...
+    const scriptOutputs = {};
+
+    for (const scriptName of cacheableOperations) {
+      // eslint-disable-next-line no-await-in-loop
+      scriptOutputs[scriptName] = await inquirer.prompt([
+        {
+          type: "input",
+          name: scriptName,
+          message: `Where within a particular project directory does the "${scriptName}" script write its outputs?\n`,
+          default: "dist",
```
vsavkin, line 78: Do you think it's worth default to dist or ""?

core/lerna/commands/add-caching/index.js
```diff
@@ -0,0 +1,157 @@
...
+
+    const scriptOutputs = {};
+
+    for (const scriptName of cacheableOperations) {
+      // eslint-disable-next-line no-await-in-loop
+      scriptOutputs[scriptName] = await inquirer.prompt([
+        {
+          type: "input",
+          name: scriptName,
+          message: `Where within a particular project directory does the "${scriptName}" script write its outputs?\n`,
```
vsavkin, line 77: Maybe we can say "leave empty" if no files are created?

vsavkin (approved):

core/lerna/commands/add-caching/index.js
JamesHenry: Cool will do

core/lerna/commands/add-caching/index.js
JamesHenry: That makes sense given your comment below yep

PR #3363, opened 2022-10-11
docs: add-caching and version 6 changes

isaacplmann:
Updates docs for Lerna version 6.

- reference `lerna add-caching` instead of `nx init`
- remove references to `--no-sort`
- remove references to `useNx`

This PR does not have a standalone documentation page for `lerna add-caching` or `lerna repair`


vsavkin:
We should also change "If you don't have nx.json run nx init to run lerna add-caching". Basically we should no mention nx init.

PR #3401, opened 2022-10-30
feat(version, publish): Added support for `--dry-run` option

### LINKED ISSUE #51: Feature request: lerna prepublish

meteabduraman:
Added support for `--dry-run` option in `lerna version` and `lerna publish`.

## Description
Declared a new `--dry-run` option of type `boolean`.
Whenever this option is used with one of the above commands (either `version` or `publish`), the confirmations ([`confirmVersions()`](https://github.com/lerna/lerna/blob/e33be4b043f4392311a1b12a1e3de4bdd5e0f42d/commands/version/index.js#L471) and [`confirmPublish()`](https://github.com/lerna/lerna/blob/e33be4b043f4392311a1b12a1e3de4bdd5e0f42d/commands/publish/index.js#L431) functions) automatically get rejected and an info log with `Auto-rejected. This is a dry-run.` appears in the console.

## Motivation and Context
Folkes that would like to use lerna in CIs would enjoy this change (I would).
It will be nice to have a pipeline step that would preview the versions to be changed and add some sort of manual validation to really make sure we're publishing the right versions (that's actually a use case I want to implement in one of the pipelines I work on).

There has been an ongoing discussion since 2016 regarding this feature but not any concrete action from someone.
Maybe this PR turns things around a bit 🤓

closes #51 
and maybe also a step closer to address #1094 

## How Has This Been Tested?
Added 2 unit tests for each command. They passed.
Locally tested with a locally published lerna version.

## Types of changes

## Checklist:

### DIFF
--- a/commands/publish/README.md
+++ b/commands/publish/README.md
@@ -65,6 +65,7 @@ This is useful when a previous `lerna publish` failed to publish all packages to
 - [`--tag-version-prefix`](#--tag-version-prefix)
 - [`--temp-tag`](#--temp-tag)
 - [`--yes`](#--yes)
+- [`--dry-run`](#--dry-run)
 
 ### `--canary`
 
@@ -293,6 +294,16 @@ lerna publish --canary --yes
 
 When run with this flag, `lerna publish` will skip all confirmation prompts.
 Useful in [Continuous integration (CI)](https://en.wikipedia.org/wiki/Continuous_integration) to automatically answer the publish confirmation prompt.
+
+### `--dry-run`
+
+```sh
+lerna publish --dry-run
+# auto-rejects `Are you sure you want to publish the above changes?`
+```
+
+When run with this flag, `lerna publish` will skip all confirmation prompts.
+Useful in [Continuous integration (CI)](https://en.wikipedia.org/wiki/Continuous_integration) to automatically answer the publish confirmation prompt. (In this case, with "no").
 
 ## Deprecated Options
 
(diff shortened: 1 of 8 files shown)


JamesHenry:
Hi @meteabduraman thanks a lot for taking the time to submit this PR!

We already have something a lot more comprehensive in the works for `--dry-run` (including previews of your changes before you make them etc), so this PR is not something we can go forward with.

I spoke about the upcoming changes to `version` and `publish` with `--dry-run` at my talk at Nx Conf a couple of weeks ago including giving a sneak peek on one of my slides, but essentially - watch this space!

Many thanks again

Issue #3311, opened 2022-08-28
Improve error message for an invalid `lerna.json`


ts-web:
## Current Behavior

If a `lerna.json` lacks a `version` field, running `lerna run build` produces this error:
```
lerna ERR! ENOLERNA `lerna.json` does not exist, have you run `lerna init`?
```

## Expected Behavior

A more correct error would say "failed to load lerna.json", at a minimum, and if we're being really helpful: "a `version` property is required but was not found." Maybe a json schema library can generate this.

This issue is a resurrection of https://github.com/lerna/lerna/issues/1546 which was resolved by saying "add the version property" but did not address the misleading error message.

## Steps to Reproduce

1. Delete `version` from `lerna.json`.
2. Run `lerna run build`
3. 

### Environment

 Environment info:

  System:
    OS: Linux 5.15 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 3.2.2 - ~/.nvm/versions/node/v16.15.0/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Utilities:
    Git: 2.37.2 - /usr/bin/git
  npmPackages:
    lerna: ^5.4.3 => 5.4.3

es50678:
i got a similar issue when running any `npx lerna ...` command from within a package directory:

lerna.json:
```
  "packages": [
    "core",
    "infra", // PRIVATE, so doesn't show up in output from `npx lerna list`
    "cli",
    "sdk"
```
root package.json:
```
  "name": "org-dev-ops"
  "workspaces": [
    "core",
    "infra", // PRIVATE, so doesn't show up in output from `npx lerna list`
    "cli",
    "sdk"
  ],
```
from workspace root:
```
❯ npx lerna list
lerna notice cli v8.1.9
lerna info versioning independent
@scopedorg/core
@scopedorg/cli
@scopedorg/sdk
lerna success found 3 packages
```

from core/infra/etc:
```
❯ npx lerna list
lerna notice cli v8.1.9
lerna ERR! ENOLERNA `lerna.json` does not exist, have you run `lerna init`?
```

kerryj89:
[...]

PR #3424, opened 2022-11-11
fix(command): more detailed error message when version cannot be found

amorscher:
When running a lerna command and no version is present in lerna.json configuration following error message is printed:
```sh
lerna ERR! ENOLERNA `lerna.json` does not exist, have you run `lerna init`?
``` 
As described in (https://github.com/lerna/lerna/issues/3311) this is not completely correct.
## Description
Addtional check is added during command validation to prove a version is available in lerna.json file. In addition the check for having a lerna configuration (lerna.json) available is adapted to really check for existance of the config file. This is achieved by setting an additional flag configNotFound in the project object during lerna.json retrival.

## Motivation and Context
Improve error messages for invalid lerna.json to support user finding configuration problems.
The issue is described here (https://github.com/lerna/lerna/issues/3311).

## How Has This Been Tested?
Added a unit test.
Manually built a lerna version.

## Types of changes

## Checklist:


amorscher:
@JamesHenry any chance that the fix will make it  to main?

Thx for having a look!

core/project/index.js
```diff
@@ -66,6 +66,7 @@ class Project {
             // No need to distinguish between missing and empty,
```
JamesHenry, line 66: Obviously I can't speak to the intent behind this comment originally, but your change does make it seem a little nonsensical, given you are adding an `isEmpty` right below a comment saying "no need to distinguish between missing and empty".

We could probably just amend the `this.configNotFound` assignment to check for `Object.keys(loaded.config ?? {}).length === 0`?

JamesHenry (changes requested):
.

core/project/index.js
amorscher: You are right that is a little bit weird. Just changed that and used your code suggestion.

JamesHenry (approved):

JamesHenry:
Thanks @amorscher!

Issue #3418, opened 2022-11-08
Exclude --package-lock-only from npm install


jeeten-parmar:
## Current Behavior
By running the command, `npx lerna version --conventional-commits --no-changelog --no-push --yes`, later on It is using npm install (`npm install --package-lock-only --ignore-scripts`) command internally. Due to certain changes in package lock file, It is failing repeatedly.

I want to exclude it.

## Expected Behavior
I found may be similar issue here: https://github.com/lerna/lerna/issues/3386

<details><summary>lerna.json</summary><p>

```json
{
  "packages": [
    "**"
  ],
  "version": "independent",
  "npmClientArgs": [
    "--legacy-peer-deps"
  ]
}
```

</p></details>

simllll:
Experience the same during lerna publish


Somehow only happening since today, not updated lerna though 🤔

jeeten-parmar:
@simllll We got that error after upgrading Angular from v11 to v14 and related dependencies.

simllll:
Could it be related to npm@9? This was set to latest npm yesterday

simllll:
It is definitely related to npm v9, works with version 8.

paulrobertlloyd:
Just experienced the same error. Downgrading to npm v8 fixed it.

TombolaShepless:
I'm using npm v8.19.2 and this has just started happening today, is there any workaround?

TombolaShepless:
Thanks for getting back to me @ghiscoding but the recommended fix was to downgrade to npm v8, which I'm already on.

ghiscoding:
@TombolaShepless 
as every Node and npm, they start backporting new versions (npm 9.x) into their older version (npm 8.x), so you might need to downgrade to an older version of npm (prior to v8.19.2). 

@simllll 
adding peer legacy might be just a workaround but technically speaking Lerna shouldn't touch at all any of the package manager lock files, it should let the package manager do it. I think Lerna does that now and runs `npm install --package-lock-only` but I could be wrong (that's the code I've put in Lerna-Lite long before Lerna did). Did anyone try running the command directly? Technically speaking it shouldn't fail unless the lock file has to be recreated (if so maybe delete the lock file and rerun the npm install command to get new lock file)

TombolaShepless:
Thanks @ghiscoding I'll give that a try, but what I don't understand is why this has been working for weeks using that exact npm version (8.19.2)?

ghiscoding:
[...]

Issue #3386, opened 2022-10-21
lerna version does not take npmClientArgs into account


alex3683:
## Current Behavior

When i run `lerna version` with `npmClientArgs` set to `["--legacy-peer-deps"]`, it crashes while doing `npm install --package-lock-only --ignore-scripts` because of unmet peer dependencies. I had a look at the sources and can see `npmClientArgs` not being used there.

Error output:
```
lerna ERR! Error: Command failed with exit code 1: npm install --package-lock-only --ignore-scripts
lerna ERR! npm ERR! code ERESOLVE
lerna ERR! npm ERR! ERESOLVE could not resolve
lerna ERR! npm ERR!
lerna ERR! npm ERR! While resolving: eslint-plugin-react-app@6.2.2
lerna ERR! npm ERR! Found: eslint@8.25.0
lerna ERR! npm ERR! node_modules/eslint
lerna ERR! npm ERR!   dev eslint@"^8.11.0" from the root project
lerna ERR! npm ERR!   peer eslint@"^7.5.0 || ^8.0.0" from @babel/eslint-parser@7.19.1
lerna ERR! npm ERR!   node_modules/@babel/eslint-parser
lerna ERR! npm ERR!     @babel/eslint-parser@"^7.16.3" from eslint-config-react-app@7.0.1
lerna ERR! npm ERR!     node_modules/eslint-config-react-app
lerna ERR! npm ERR!       dev eslint-config-react-app@"^7.0.0" from the root project
lerna ERR! npm ERR!   22 more (@typescript-eslint/eslint-plugin, ...)
lerna ERR! npm ERR!
lerna ERR! npm ERR! Could not resolve dependency:
lerna ERR! npm ERR! peer eslint@"6.x" from eslint-plugin-react-app@6.2.2
lerna ERR! npm ERR! node_modules/eslint-plugin-react-app
lerna ERR! npm ERR!   dev eslint-plugin-react-app@"^6.2.2" from the root project
lerna ERR! npm ERR!
lerna ERR! npm ERR! Conflicting peer dependency: eslint@6.8.0
lerna ERR! npm ERR! node_modules/eslint
lerna ERR! npm ERR!   peer eslint@"6.x" from eslint-plugin-react-app@6.2.2
lerna ERR! npm ERR!   node_modules/eslint-plugin-react-app
lerna ERR! npm ERR!     dev eslint-plugin-react-app@"^6.2.2" from the root project
```
## Expected Behavior

[...]

PR #3427, opened 2022-11-14
fix(version): pass npmClientArgs to npm install cmd

amorscher:
- support -- args assign as well

Passes provided npmClient args to used npm install when executing `lerna version -- --legacy-peer-deps`

## Motivation and Context
Problems raised in #3386 and #3418

## How Has This Been Tested?
Added unit test.

## Types of changes

## Checklist:

### DIFF
--- a/commands/version/command.js
+++ b/commands/version/command.js
@@ -263,12 +263,6 @@ exports.builder = (yargs, composed) => {
         delete argv.githubRelease;
         log.warn("deprecated", "--github-release has been replaced by --create-release=github");
       }
-      /* eslint-enable no-param-reassign */
-
-      if (argv["--"]) {
-        log.warn("EDOUBLEDASH", "Arguments after -- are no longer passed to subprocess executions.");
-        log.warn("EDOUBLEDASH", "This will cause an error in a future major version.");
-      }
 
       return argv;
     });
--- a/commands/version/__tests__/version-command.test.js
+++ b/commands/version/__tests__/version-command.test.js
@@ -32,6 +32,8 @@ const initFixture = require("@lerna-test/helpers").initFixtureFactory(
 const { showCommit } = require("@lerna-test/helpers");
 const { getCommitMessage } = require("@lerna-test/helpers");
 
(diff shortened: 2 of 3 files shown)


commands/version/command.js
```diff
@@ -263,12 +263,6 @@ exports.builder = (yargs, composed) => {
         delete argv.githubRelease;
         log.warn("deprecated", "--github-release has been replaced by --create-release=github");
       }
-      /* eslint-enable no-param-reassign */
-
```
amorscher, line 267: Not sure if this causes any other problems?

commands/version/command.js
```diff
@@ -263,12 +263,6 @@ exports.builder = (yargs, composed) => {
...
-      if (argv["--"]) {
-        log.warn("EDOUBLEDASH", "Arguments after -- are no longer passed to subprocess executions.");
-        log.warn("EDOUBLEDASH", "This will cause an error in a future major version.");
-      }
```
amorscher, line 271: Not sure if this causes any other problems?

amorscher:
@JamesHenry any chance that the fix will make it to main?
Thx for having a look.

jeeten-parmar:
@JamesHenry Can you please do the needful? It has blocked our some work so we are also waiting for this change.

commands/version/command.js
fahslaj: This command does not support `--` args passthrough because it is not clear which subprocesses those arguments are being passed to. Just by running `lerna version -- --force`, it is not clear that the `--force` is only relevant for an inner `npm install` command. This is why more specific options (like --npmClientArgs) are preferred for specifying subprocess arguments.

fahslaj:
@amorscher Thank you very much for your work on this PR! As mentioned in a different comment, I do not think it is appropriate to add `--` args forwarding back to the `lerna version` command. Since A different PR has been merged that adds support for `--npmClientArgs` on `lerna version`, I do not believe that this PR is necessary anymore. Please feel free to reach out if you disagree or have any other questions or concerns.

PR #3437, opened 2022-11-23
docs(website): add state of js banner

bcabanes:
It adds the state of JS survey link on the banner for lerna.js.org.


juristr (approved):
👍

PR #3503, opened 2023-01-20
feat(publish): add --include-private option for testing private packages

fahslaj:
## Description

Adds an option `--include-private` to `lerna publish`. This allows packages marked as `"private": true` in package.json to be published. Since npm does not allow publishing packages with `"private": true`, lerna will strip out the property before the publish.

The option accepts a list of package names to be included in the publish. The packages provided are only published if those packages are deemed necessary to publish, as is consistent with lerna's existing publish behavior. Alternatively, a "*" can be provided, telling lerna to allow any private package to be published.

```sh
$ lerna publish from-git --include-private my-private-package my-other-private-package
$ lerna publish from-git --include-private "*"
```

## Motivation and Context

This allows for the ability to publish packages that are currently private, but _will be public in the future_. This is useful for end to end testing. These packages can now be published to a locally hosted npm repository and downloaded for real in the end to end tests, all before the package is ready for publishing on the public npm repository.

Lerna's official stance on publishing private packages (and that you shouldn't do it) wouldn't change with this PR. If a package is meant to be published, then it is by definition "public", which means it should not have `"private": true` in package.json (as described in the [npm docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#private)). However, this strict implementation by lerna leaves no room for the specific use case outlined above - for packages that _will be public, but aren't yet_.

Therefore, this option is meant to be used as a workaround in CI pipelines for testing purposes, not for consistently publishing private packages to actual public repos. 

_Note that this PR is talking about packages with `"private": true` in package.json, which is different than npm's idea of [scoped private packages](https://docs.npmjs.com/about-private-packages)._

## How Has This Been Tested?

This has been tested manually and covered by e2e tests.

## Types of changes

## Checklist:

### DIFF
--- a/e2e/publish/src/from-git-recover-from-error.spec.ts
+++ b/e2e/publish/src/from-git-recover-from-error.spec.ts
@@ -77,7 +77,6 @@ describe("lerna-publish-from-git-recover-from-error", () => {
         lerna WARN ENOLICENSE Packages test-1 and test-2 are missing a license.
         lerna WARN ENOLICENSE One way to fix this is to add a LICENSE.md file to the root of this repository.
         lerna WARN ENOLICENSE See https://choosealicense.com for additional guidance.
-        lerna WARN publish Package is already published: test-1@XX.XX.XX
         lerna success published test-2 XX.XX.XX
         lerna notice 
         lerna notice 📦  test-2@XX.XX.XX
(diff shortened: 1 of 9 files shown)


JamesHenry (changes requested):
Needs manual rebase after the refactor, please ping me if anything is unclear on where it should live now

JamesHenry (approved):

PR #3504, opened 2023-01-21
fix(create): normalize quotes in generated test and lib files

teddybradford:
This change normalizes quotes in generated test and lib files to all be single quotes.

PR #3529, opened 2023-02-06
Update formatting changes

teddybradford:
re: #3504
This PR normalizes quotes and indents in generated test and lib files to all be single quotes and 2-space indents.

### DIFF
--- a/e2e/create/src/create.spec.ts
+++ b/e2e/create/src/create.spec.ts
@@ -95,7 +95,7 @@ describe("lerna-create", () => {
         module.exports = noOptions;
 
         function noOptions() {
-            return "Hello from noOptions";
+          return 'Hello noOptions';
         }
 
       `);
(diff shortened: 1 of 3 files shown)


JamesHenry (changes requested):
Thanks a lot for recreating this @teddybradford!

Looks like the `e2e-publish` is failing: https://cloud.nx.app/runs/GghxN3Qy9i

Excuse the verbose logging we currently have enabled, but if you scroll down within `e2e-publish` you will see the relevant snapshot failures

JamesHenry:
Yes it's the jest snapshots:

image

We have very verbose debugging logging enabled for ourselves at the moment so I'm afraid it is a little buried in there.

To resolve it I ran the relevant e2e tests and passed `-u` to instruct it to update the snapshots.

```sh
npx nx e2e e2e-publish -u
```

To save you time I have gone ahead and committed the result already, let's see if there is anything else outstanding on CI now (my local machine was playing up a bit so it's possible I missed an update or too as well)

JamesHenry (approved):
Thanks again @teddybradford!

teddybradford:
@JamesHenry Awesome! Thanks, James.

PR #3556, opened 2023-02-23
chore(core): remove some ts-ignore comments and any types

amorscher:
Introduce a CommandConfigOptions type as a generic on command. Each concrete command has to provide a specific interface if needed.

## Motivation and Context
Make typescript implementation more typesafe

## Types of changes

## Checklist:


amorscher:
@JamesHenry To explain the appoach a liitle bit: 
So I was trying to make core/command more typesafe by removing ts-ignores and any types. To do this a generic is introduced to have all options explicitly specified for each command. In order to validate this I just updated concrete CommandConfigOptions where tsc reported an error due to strict mode. If the apporach is considered as the way to go, other commands could be updated by setting strict mode in the ts-config and introduce the needed options interface.

libs/core/src/index.ts
```diff
@@ -32,6 +32,7 @@ export { symlinkDependencies } from "./lib/symlink-dependencies";
 export { default as tempWrite } from "./lib/temp-write";
 export { timer } from "./lib/timer";
 export { ValidationError } from "./lib/validation-error";
+export { CommandConfigOptions as CommandConfigOptions } from "./lib/project";
```
JamesHenry, line 35:
```suggestion
export { CommandConfigOptions } from "./lib/project";
```

libs/core/src/lib/command/clean-stack.ts
```diff
@@ -10,5 +10,9 @@ export function cleanStack(err: ExecaError, className: string) {
   }
 
   // nothing matched, just return original error
-  return err;
+  return err.toString();
+}
+
+function isAExecaError(err: ExecaError | string): err is ExecaError {
```
JamesHenry, line 16: I think `isExecaError` reads a bit more naturally

libs/core/src/lib/command/clean-stack.ts
```diff
@@ -1,7 +1,7 @@
 import { ExecaError } from "execa";
 
-export function cleanStack(err: ExecaError, className: string) {
-  const lines = err.stack ? err.stack.split("\n") : String(err).split("\n");
+export function cleanStack(err: string | ExecaError, className: string) {
+  const lines = isAExecaError(err) && err.stack ? err.stack.split("\n") : String(err).split("\n");
```
JamesHenry, line 4: Shouldn't isExecError(err) replace `err.stack` here, rather than be checked in addition to it, based on how you have implemented it?

libs/core/src/lib/command/index.ts
```diff
@@ -208,14 +205,11 @@ export class Command {
   configureProperties() {
     const { concurrency, sort, maxBuffer } = this.options;
 
-    this.concurrency = Math.max(1, +concurrency || DEFAULT_CONCURRENCY);
+    this.concurrency = Math.max(1, +(concurrency ?? 0) || DEFAULT_CONCURRENCY);
```
JamesHenry, line 208: What additional case are you attempting to protect against here?

I believe it would already always do the right thing (given `NaN` and `0` are falsely values, and the result of coercion for values such as `undefined` and `null`), so I think it would be simpler to revert this unless there is something I am missing

JamesHenry (changes requested):

JamesHenry:
Thank you @amorscher, the approach makes sense, please see the feedback above

libs/core/src/index.ts
amorscher: Ok done!

libs/core/src/lib/command/index.ts
amorscher: Typechecker is reporting an error because it is possibly undefined from the defined options. I gave it a default value during assignment

libs/core/src/lib/project/index.ts
```diff
@@ -15,15 +15,38 @@ import { applyExtends } from "./apply-extends";
...
   packages: string[];
   useNx: boolean;
   useWorkspaces: boolean;
   version: string;
   npmClient: string;
+  command?: CommandConfigs;
 }
 
-interface PnpmWorkspaceConfig {
+interface CommandConfig {
```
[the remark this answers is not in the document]
amorscher: OK just forgot to change the name back. I reverted the change.

libs/core/src/lib/command/index.ts
amorscher: Typechecker is reporting an error because concurrency can possibly be null from options. So I assigned a default value to concurrency so it is always set even if undefined from options.

libs/core/src/lib/command/clean-stack.ts
amorscher: Changed the interface to error as there is nothing specific needed from ExecaError. I renamed typeguard to make it clear what is checked.

JamesHenry (approved):
Thank you @amorscher!

PR #3582, opened 2023-03-10
chore: use node 18 and npm 9 as primary versions

JamesHenry:


integration/__tests__/lerna-bootstrap-hoist.spec.ts
```diff
@@ -48,20 +48,20 @@ package-3 cli2 package-2 OK
   expect(lock3).toMatchObject({
     name: "@integration/package-3",
     version: "1.0.0",
-    dependencies: { pify: expect.any(Object) },
+    packages: { "node_modules/pify": expect.any(Object) },
   });
   expect(rootLock).toMatchObject({
     name: "integration",
     version: "0.0.0",
-    dependencies: {
```
JamesHenry, line 56: dependencies in lockfiles was already deprecated in lockfile v2, fully removed in lockfile v3 which is what this PR starts using

Issue #2423, opened 2020-01-21
Use JSON5 or another parser with support for comments for config-files


ELLIOTTCABLE:
## Expected Behavior
Given that larger and larger proportions of the 'code' we write in the last decade have become configuration (see also, [Dhall](https://dhall-lang.org/)), it's increasingly important that we follow good software-engineering practices even *outside* of turing-complete source-code. Towards this, I would like to be able to add comments to my `lerna.json`, as I can with many other configuration-files.

(Not to mention being able to follow other JavaScript practices, such as using trailing commas to improve diff-quality, having multiline strings, etc.)

## Current Behavior
At the moment, including a comment in `lerna.json` causes Lerna to crash with a parsing-error like this:

```sh
$ npx lerna bootstrap
npx: installed 703 in 39.274s
lerna notice cli v3.20.2
lerna ERR! JSONError JSON Error in /Users/ec/Sync/Code/excmd/lerna.json:
lerna ERR! JSONError Unexpected token / in JSON at position 79 while parsing near
    '...tstrap": {         // FIXME: Hoisting w...'
```

## Possible Solution
There's quite a few alternative configuration systems out there; personally, for JavaScript-ecosystem projects, I am partial to [JSON5](https://json5.org/), which is fully backwards-compatible with `JSON.parse` - and in fact, nearly indistinguishable, except for a few tiny nods to human-usability. There's also YAML, although it has security considerations and is probably unnecessary complexity; and then there's simply "switching to a better language" like TOML or Dhall; although I find that to be an obtuse choice if you're operating primarily in the JavaScript ecosystem!

dimitropoulos:
As another alternative: https://github.com/lerna/lerna/pull/2626 which would allow for comments and much more, at the same time as not annoying those people that are annoyed by files that are not json files (json5 is not json) that have a `.json` extension.  I'm not one of those people, but #2626 is a way to keep everyone happy.

ELLIOTTCABLE:
Still relevant, afaik.

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #3603, opened 2023-03-23
fix: handle trailing commas and comments in lerna.json files

JamesHenry:
Currently lerna will hard error if trailing commas and/or comments are present in `lerna.json` files. This PR changes that and allows it to continue in these cases. It will still hard error if there is any other kind of JSON parsing error.

Closes https://github.com/lerna/lerna/issues/2423


donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

Issue #3611, opened 2023-03-24
@lerna/legacy-package-management missing dependency on rimraf


keeganstreet:
## Current Behavior

Error when running lerna after 6.6.0 release

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './bin' is not defined by "exports" in node_modules/rimraf/package.json
```

## Expected Behavior

The new @lerna/legacy-package-management module imports from rimraf, so it probably needs to specify rimraf as a dependency to ensure it gets the correct version

```
RIMRAF_CLI = require.resolve("rimraf/bin");
```

buffcode:
Error and stack trace:
```
$ /builds/node_modules/.bin/lerna bootstrap
node:internal/modules/cjs/loader:488
      throw e;
      ^
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './bin' is not defined by "exports" in /builds/node_modules/rimraf/package.json
    at new NodeError (node:internal/errors:372:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:472:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:753:3)
    at resolveExports (node:internal/modules/cjs/loader:482:36)
    at Function.Module._findPath (node:internal/modules/cjs/loader:522:31)
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:919:27)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at libs/core/src/lib/rimraf-dir.ts (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:3574:26)
    at __init (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:10:56)
    at libs/core/src/index.ts (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:3942:5)
    at __init (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:10:56)
    at libs/commands/add/src/command.ts (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:4613:5)
    at __require (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:13:50)
    at Object.<anonymous> (/builds/node_modules/@lerna/legacy-package-management/dist/index.js:4884:22)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10) {
[...]

PR #3614, opened 2023-03-24
fix: resolving rimraf bin dir

tada5hi:
## Description
Due to the fact that rimraf has changed its binary path, lerna no longer works with the new version 4.4.x. Therefore the cli location is now read from the package.json, as fallback the original module resolving was used.

## Types of changes

## Checklist:

fixes [#3614](https://github.com/lerna/lerna/issues/3611)

### DIFF
--- a/libs/core/src/lib/rimraf-dir.ts
+++ b/libs/core/src/lib/rimraf-dir.ts
@@ -10,7 +10,7 @@ const childProcess = require("@lerna/child-process");
 
 let rimrafBinPath: string | undefined;
 
-export async function useRimrafBinPath(): Promise<string> {
+async function useRimrafBinPath(): Promise<string> {
   if (typeof rimrafBinPath === "string") {
     return rimrafBinPath;
   }


libs/core/src/lib/rimraf-dir.ts
```diff
@@ -1,36 +1,59 @@
...
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const childProcess = require("@lerna/child-process");
 
-// NOTE: if rimraf moves the location of its executable, this will need to be updated
-const RIMRAF_CLI = require.resolve("rimraf/bin");
+let rimrafBinPath: string | undefined;
 
-export function rimrafDir(dirPath: string) {
+export async function useRimrafBinPath(): Promise<string> {
```
JamesHenry, line 13: Please can we not export this to make it clear that it is not used outside this file

```suggestion
async function useRimrafBinPath(): Promise<string> {
```

JamesHenry:
Thanks @tada5hi one small change then this LGTM!

libs/core/src/lib/rimraf-dir.ts
tada5hi: good point ^^, sry :see_no_evil: 
I will fix that asap

JamesHenry: You can commit the suggestion directly in the github UI just FYI

tada5hi: oh, wasn't aware of that ^^. Now i have already made a commit from my fork, but good to know, thanks

JamesHenry (approved):

JamesHenry:
Thanks again!

JamesHenry:
Sadly this didn't actually solve the problem because the bin path is not part of the package's exports and so node blocks it, I am looking into refactoring the whole approach here, I believe we can just use the node API of rimraf instead of all these child process + bin shenanigans

tada5hi:
damm ... i think the problem is that i missed to join the `pck.bin` path with the directory path of require.resolve.

PR #3616, opened 2023-03-24
fix(deps): update to rimraf v4, remove path-exists

JamesHenry:
Refactors rimraf usage to no longer attempt to resolve the bin and invoke it via a child process because the bin is not resolvable in modern Node thanks to package exports. Hopefully using the primary node API should be what we want anyway.

Also removes the unnecessary `path-exists` dependency


JamesHenry:
It failed as hoped prior to that fix PR going in, let's see if it passes now when we sync up!

tada5hi:
@JamesHenry i think that, as you have already noted, the better way is to use the api directly instead of the cli variant :thumbsup: .

PR #3627, opened 2023-03-29
chore(publish): fix failing tests on windows os

amorscher:
Fixes failing publish unit tests on windows operating system

## Description
Fixes failing publish unit tests on windows operating system

## Motivation and Context
Re-activate ignored unit tests to achieve higher test coverage and regression testing.

## Types of changes

## Checklist:


libs/commands/publish/src/lib/publish-command.spec.ts
```diff
@@ -31,12 +31,12 @@ jest.mock("./get-unpublished-packages", () => ({
 jest.mock("./git-checkout");
 
 // lerna version mocks
-jest.mock("@lerna/commands/version/lib/git-push");
-jest.mock("@lerna/commands/version/lib/is-anything-committed", () => ({
+jest.mock("../../../version/src/lib/git-push");
```
fahslaj, line 34: I suspect this will still work as `"@lerna/commands/version/src/lib/git-push"`. Can you please try that for all these imports of version command helpers? It's better if we can keep this as an absolute import since it crosses library boundaries.

This way, the publish tests won't start failing if we ever change the directory structure around the version command.

fahslaj (changes requested):

libs/commands/publish/src/lib/publish-command.spec.ts
amorscher: Thx for the review. This was actually the fix for the failing tests. Using the pathmap here is not working then the mock will not be applied and the tests fail. I also tried to refactor the git-push into a real typescript module but it was still not working.
 I fully understand your point. A quickfix would be the proposed solution for a cleaner one I have to investigate a little more.

amorscher: Uploaded a solution keeping original paths. So what I did is considering the path config from tsconfig. Maybe we could do this on top-level in jest.preset.js? 
I think this should also work but I did not do it, because its only needed by publish tests.

amorscher:
@fahslaj I updated the pull request so you could have a look again.

amorscher:
@fahslaj any updates here!?

fahslaj (approved):

fahslaj:
@amorscher looks good, thank you!!

PR #3629, opened 2023-03-30
chore(import): fix failing tests on windows os

amorscher:
Fixes failing import unit tests on windows operating system

## Description
Fixes failing import unit tests on windows operating system

## Motivation and Context
Re-activate ignored unit tests to achieve higher test coverage and regression testing.

## Types of changes

## Checklist:


amorscher:
@fahslaj @amorscher I found the problem. It is related to different temp dirs on the main and the agents. So on the main we have `C:/Users/runneradmin/AppData/Local/Temp/69d67553712c3557d7449c90945ed8ba/test` on the agent we have `C:\Users\RUNNER~1\AppData\Local\Temp\69d67553712c3557d7449c90945ed8ba\test`. So then the import command tries to extract the relative part out of gitRepoRoot and the project root. But actually the project root is not part of the workspace root. See the code here:

https://github.com/lerna/lerna/blob/f5bd591c592bc6fee7d3a6b2ec996d633236aa6c/libs/commands/import/src/index.ts#L71-L76

We then get something like this `..\..\..\..\..\..\RUNNER~1\AppData\Local\Temp\69d67553712c3557d7449c90945ed8ba\test` as the `targetDirRelativeToGitRoot `. This used to create the import patch which results in an "empty" import as the paths are not considered to be in the git repo. See here:

https://github.com/lerna/lerna/blob/f5bd591c592bc6fee7d3a6b2ec996d633236aa6c/libs/commands/import/src/index.ts#L181-L192

So from the import point of view everthing seems to be ok but actually nothing is imported. This can also happen if you have symlinks involved.

To slove this I just set the temp dir in the ci config to `C:\temp`.
In addition I think I will add a check on this to ensure that the projectRoot is part of the workspaceRoot.
 What do you think?

fahslaj:
> @fahslaj @JamesHenry I found the problem. It is related to different temp dirs on the main and the agents. So on the main we have `C:/Users/runneradmin/AppData/Local/Temp/69d67553712c3557d7449c90945ed8ba/test` on the agent we have `C:\Users\RUNNER~1\AppData\Local\Temp\69d67553712c3557d7449c90945ed8ba\test`. So then the import command tries to extract the relative part out of gitRepoRoot and the project root. But actually the project root is not part of the workspace root. See the code here:
> 
> https://github.com/lerna/lerna/blob/f5bd591c592bc6fee7d3a6b2ec996d633236aa6c/libs/commands/import/src/index.ts#L71-L76
> 
> We then get something like this `..\..\..\..\..\..\RUNNER~1\AppData\Local\Temp\69d67553712c3557d7449c90945ed8ba\test` as the `targetDirRelativeToGitRoot `. This used to create the import patch which results in an "empty" import as the paths are not considered to be in the git repo. See here:
> 
> https://github.com/lerna/lerna/blob/f5bd591c592bc6fee7d3a6b2ec996d633236aa6c/libs/commands/import/src/index.ts#L181-L192
> 
> So from the import point of view everthing seems to be ok but actually nothing is imported. This can also happen if you have symlinks involved.
> 
> To slove this I just set the temp dir in the ci config to `C:\temp`. In addition I think I will add a check on this to ensure that the projectRoot is part of the workspaceRoot. What do you think?

What do you mean by "ensure that the projectRoot is part of the workspaceRoot"? Are you talking about adding this as functionality of Lerna, or just an additional check within the tests themselves?

amorscher:
Well I mean here:

https://github.com/lerna/lerna/blob/f5bd591c592bc6fee7d3a6b2ec996d633236aa6c/libs/commands/import/src/index.ts#L71-79

Something like

` if (this.targetDirRelativeToGitRoot.startsWith("..")) {
      throw new ValidationError("NOTINREPOSITORY", `Project root ${this.project.rootPath} is a subdirectory of git root ${gitRepoRoot}"`);
    }`

amorscher:
@fahslaj
 hey Austin,

Is there something missing from your point of view. Thx.

amorscher:
@fahslaj any updates here?

fahslaj:
@amorscher sorry for the delay. I think the solution you have suggested of throwing a ValidationError in the case above makes sense. If you could please make that change and update this branch from main then we can get this merged as soon as CI passes. Thanks!

amorscher:
> @amorscher sorry for the delay. I think the solution you have suggested of throwing a ValidationError in the case above makes sense. If you could please make that change and update this branch from main then we can get this merged as soon as CI passes. Thanks!

OK should be done!

fahslaj (approved):

fahslaj:
Thank you @amorscher ! Your contributions are greatly appreciated 😄

PR #3633, opened 2023-04-03
Update version-and-publish.md

ruk93:
I think `remixapp` is private, so it doesn't get published to NPM

## Description

Fix typo in documentation

## Types of changes

## Checklist:


fahslaj:
@ruk93 Thank you for opening this PR. However, I do not think this change is appropriate. I believe the existing language is more accurate than the updated language in this PR. The remixapp is an application that can be deployed for public consumption, so it is not necessarily "private" in the purest sense of the word. `"private": true` in the package.json file only indicates that `npm` (and `lerna`) should not try to publish it specifically as an npm package to the npm registry.

PR #3658, opened 2023-04-20
feat: do not include bootstrap, add, link commands by default

JamesHenry:
BREAKING CHANGE: We no longer include the bootstrap, add, and link commands by default. We strongly recommend using your package manager (npm, yarn, pnpm) for package management related concerns such as installing and linking dependencies.

If you want to temporarily continue to use those commands in v7 you can do so by installing the @lerna/legacy-package-management package at the same version as your lerna version.

There will not be any active work done on these commands and you should look to migrate as soon as possible, please check out https://lerna.js.org for further guidance.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -113,8 +113,7 @@ You will need two different terminal windows for this as one of them will contai
 - Run `npm run local-registry start` in Terminal 1 (keep it running)
 - Run `npm adduser --registry http://localhost:4873` in Terminal 2 (real credentials are not required, you just need to
   be logged in. You can use test/test/test@test.io.)
-- Run `npm run local-registry enable` in Terminal 2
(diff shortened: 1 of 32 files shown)


iSuslov:
@JamesHenry where can we read about this breaking changes in `v7`. I used lerna mostly to link between packages with bootstrap which is something I wasn't able to do with `npm`. What should we do now?

JamesHenry:
@iSuslov rest assured there will be plenty of information about v7 once it is ready for use, that is not yet the case.

`npm` has actually supported package linking for many years, that is why we are making this change, to let the package managers do what they are good at and let lerna bring unique value on top and not reinvent the wheel. Part of what we will be doing for v7 of lerna is providing some educational material to highlight this.

In the meantime, I recommend checking out the official npm guide on workspaces: https://docs.npmjs.com/cli/v9/using-npm/workspaces?v=true

PR #3667, opened 2023-04-28
feat(core): convert commands to use nx project graph instead of package graph

fahslaj:


libs/legacy-core/src/lib/run-topologically.spec.ts
fahslaj, line 1: Test cases created by observing Lerna's existing behavior.

packages/lerna/src/commands/publish/lib/get-tagged-packages.ts
fahslaj, line 1: This and `get-unpublished-packages` below are no longer exported from Lerna. If absolutely needed, we can add back and export the old version of the function that uses the package graph.

libs/commands/publish/README.md
```diff
@@ -146,28 +145,6 @@ lerna publish from-package --git-head ${CODEBUILD_RESOLVED_SOURCE_VERSION}
 
 Under all other circumstances, this value is derived from a local `git` command.
 
-### `--graph-type <all|dependencies>`
```
fahslaj, line 149: --graph-type is no longer supported. This was called out as deprecated in the code, giving a warning when used that it would be removed in Lerna v6.

Issue #3579, opened 2023-03-09
Commitlint hook fails with latest version


gvdp:
After upgrading to the latest lerna version in our monorepo, the commitlint hook fails.

Not sure this is something which should be fixed here, but it seems to occur because of some changes in the `package.json` structure which were introduced in v6.5.0. If not it's hopefully a good starting point for others experiencing the same issue.

I think it might have something to do with [this issue](https://github.com/sindresorhus/resolve-pkg/issues/9), because the `resolvePkg` step is where it fails and it seems now there are `exports` in the root package.json file of the lerna dependency.

## Current Behavior

Error when trying to do a commit with `@commitlint/config-lerna-scopes` enabled

## Expected Behavior

Commitlint keeps working as before

## Steps to Reproduce

Created an empty repo where the issue occurs:

https://github.com/gvdp/lerna-commitlint-repro

### Failure Logs / Configuration

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:372:5)
    at validateString (node:internal/validators:120:11)
    at Object.join (node:path:1172:7)
    at getLernaVersion (/Users/gvdp/code/oss/lerna-commitlint/node_modules/@commitlint/config-lerna-scopes/index.js:57:22)
    at /Users/gvdp/code/oss/lerna-commitlint/node_modules/@commitlint/config-lerna-scopes/index.js:33:25
    at async execute (/Users/gvdp/code/oss/lerna-commitlint/node_modules/@commitlint/execute-rule/src/index.ts:20:16)
    at async Promise.all (index 0)
    at async load (/Users/gvdp/code/oss/lerna-commitlint/node_modules/@commitlint/load/src/load.ts:78:3)
    at async main (/Users/gvdp/code/oss/lerna-commitlint/node_modules/@commitlint/cli/src/cli.ts:214:17) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

<details><summary>lerna.json</summary><p>

```json
{
  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
  "useWorkspaces": true,
  "version": "0.0.0",
  "npmClient": "yarn",
  "packages": ["packages/*"]
}
```

</p></details>

### Environment

```
✗ npx lerna info
lerna notice cli v6.5.1

 Environment info:

  System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
  Binaries:
[...]

PR #3670, opened 2023-05-02
fix: export package.json to allow for package resolution via resolve-pkg

backflip:
This change adds lerna's `package.json` file to its exports.

## Description
We currently have an error with `@commitlint/config-lerna-scopes` as reported in https://github.com/lerna/lerna/issues/3579. The error can be tracked down to `@commitlint/config-lerna-scopes` receiving `undefined` when executing `resolvePkg('lerna', {cwd})` in https://github.com/conventional-changelog/commitlint/blob/v17.6.1/%40commitlint/config-lerna-scopes/index.js#L57. This is apparently by design, see link in https://github.com/sindresorhus/resolve-pkg/issues/9#issuecomment-1433250154.

## Motivation and Context
This should fix https://github.com/lerna/lerna/issues/3579.

## How Has This Been Tested?
I used @gvdp's example repo linked in https://github.com/lerna/lerna/issues/3579 and edited lerna's package.json in there.

## Types of changes

## Checklist:

### DIFF
--- a/packages/lerna/package.json
+++ b/packages/lerna/package.json
@@ -124,7 +124,8 @@
     "./commands/publish/lib/*": "./dist/commands/publish/lib/*.js",
     "./commands/version/lib/*": "./dist/commands/version/lib/*.js",
     "./migrations/**/*": "./dist/migrations/**/*",
-    "./commands/diff/lib/*": "./dist/commands/diff/lib/*.js"
+    "./commands/diff/lib/*": "./dist/commands/diff/lib/*.js",
+    "./package.json": "./package.json"
   },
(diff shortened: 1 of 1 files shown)


JamesHenry:
This doesn't feel like the right approach. It would seem more intuitive to me to alter config-lerna-scope's implementation to find one of lerna's actual entrypoints, rather than modifying lerna to fit in with an implementation detail package's (`resolve-pkg` in this case) needs.

The `getLernaVersion` could be updated to `require.resolve` one of the known entry points and then find lerna's package.json relative to that right?

backflip:
@JamesHenry, that is a very valid point, I'll try to fix it there.

PR #3699, opened 2023-05-31
feat(publish): support custom directory per-package

fahslaj:
## Description

Adds support for a "lerna.publish" config option in package.json files. This tells Lerna which directory to publish for the package and if any assets should be copied to the directory before publishing.

## Motivation and Context

This allows for publishing directories outside of the source directory of the package. This is especially helpful for patterns where the output of a package is written to a `dist/packages/...` directory in the root of the workspace.

## How Has This Been Tested?

This has been tested manually and covered by e2e tests

## Types of changes

## Checklist:


matheo:
@fahslaj I wonder if I give a `--directory` to `lerna publish`
it should sync the new version to that dir too when it bumps the source version and commit the new tag.

Currently I have to build my own script to hook `version` globally but now it makes sense to me that it should be the default behavior. WDYT?

matheo:
@fahslaj @JamesHenry I just created this PR to cover the package contents version after I failed to build a custom script, instead it was easier to have all the info inside `lerna` itself: https://github.com/lerna/lerna/pull/3787

PR #3704, opened 2023-06-01
feat(publish): experimental automatic versioning strategy

fahslaj:
## Description

Adds support for an experimental versioning strategy. With this enabled, Lerna will:
- Not commit changes from `lerna version`
- Automatically --force-publish during `lerna version`
- Detect the previous version from git tags instead of package.json files during `lerna publish`
- Use `npm view` to determine the global version instead of the version number in lerna.json

## Types of changes

## Checklist:


JamesHenry:
Will revisit after v7

matheo:
@JamesHenry so how is it working ATM?
I'm using `npm view` in a custom script to sync my `main` but the original implementation didn't consider the conflicts.
For branches I guess the algorithm should check the next bump carefully, not just the latest version but perhaps the next `14.x` bugfix/feat release.

PR #2952, opened 2021-07-23
fix(npm-publish): Allows disabling of strict SSL checks

10hendersonm:
## Description

Alias `strict-ssl` npm configuration setting to `strictSSL` publish `opts`.

## Motivation and Context

Since lerna 4, `npm_config_strict_ssl=false` is not respected by the publish process. My team publishes to an internal private registry for CI builds to avoid polluting our `npm view [package] versions` unnecessarily, but we don't (currently?) have SSL working on that garbage heap.

While `strict-ssl` is resolved out of environment and npm settings in `@lerna/npm-conf`, there's nothing downstream looking for that. [`libnpmpublish`](https://www.npmjs.com/package/libnpmpublish) passes all extra arguments directly down to [`npm-registry-fetch`](https://www.npmjs.com/package/npm-registry-fetch), but `npm-registry-fetch` is reading for `strictSSL` rather than `strict-ssl`.

closes #2942 

## How Has This Been Tested?

I monkey patched\* the changes into a monorepo I was previously unable to publish to our private registry.

Prior to implementing the changes, adding `npm_config_strict_ssl=false` either A) prepended on the command line (`npm_config_strict_ssl=false npx lerna publish ....`) as environment; or B) applied in a root-level `.npmrc` file would have no impact. This can be validated by running the lerna command in VS Code's `JavaScript Debug Terminal` and placing a breakpoint on this line:

https://github.com/lerna/lerna/blob/a47fc294393a3e9507a8207a5a2f07648a524722/utils/npm-publish/npm-publish.js#L91

When in the un-aliased state, the `opts` / `innerOpts` will still show a `strict-ssl` value, but if you `Go to Definition` or `step into` `libnpmpublish`'s `publish` method and follow down through to `npm-registry-fetch`'s default export method, you'll find that it's passing the `opts.strictSSL` property through to [`make-fetch-happen`'s `fetch` method](https://www.youtube.com/watch?v=Pubd-spHN-0):

https://github.com/npm/npm-registry-fetch/blob/8954f61d8d703e5eb7f3d93c9b40488f8b1b62ac/index.js#L126

After implementing the alias change to `npm-publish.js` and re-running, the breakpoint will now show `strict-ssl` **AND** `strictSSL` in the `opts`. You can follow them down the function calls and see the value get passed correctly all the way to `fetch`.

I've included a unit test to validate this functionality performs the aliasing as expected.

\* _You could probably do the same with `npm link`. I just forgot to._

## Types of changes

[...]

PR #3747, opened 2023-06-23
fix: fix strict-ssl mapping for node-fetch-registry during unpublished projects lookup

### LINKED ISSUE #3746: Lerna doest not respect strict-ssl=false in .npmrc for registry lookup
## Current Behavior

This is a followup of #2942 and #2952 .

If you are using self signed certificates the lookup via pacote will always fail with:

> lerna WARN Unable to determine published version, assuming "..." unpublished.

Even if strict-ssl is set to false. This is the same behavior like in #2952 explained. node-fetch-registry is using strictSSL instead of strict-ssl as option name.

```
{
  "code": "SELF_SIGNED_CERT_IN_CHAIN",
  "errno": "SELF_SIGNED_CERT_IN_CHAIN",
  "type": "system"
  "stack": "FetchError: request to ... failed, reason: self signed certificate in certificate chain
    at ClientRequest.<anonymous> (project\node_modules\minipass-fetch\lib\index.js:130:14)
    at ClientRequest.emit (node:events:527:28)
    at TLSSocket.socketErrorListener (node:_http_client:454:9)
    at TLSSocket.emit (node:events:539:35)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)"
}
```

## Expected Behavior

Should be able to lookup in a registry with self-signed certificates.

## Steps to Reproduce

1. Use a registry with self-signed certificates
2. Run `npx lerna publish from-package` to lookup packages there
3. It will always try to publish all packages regardless if they are already there.

### Environment

 Environment info:

  System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 5 5600G with Radeon Graphics
  Binaries:
    Node: 16.15.0 - C:\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.5.5 - C:\nodejs\npm.CMD
  Utilities:
    Git: 2.39.1.
  npmPackages:
    lerna: ^6.4.1 => 6.5.1

CSchulz:
fixes #3746 

## Description
There is a missing mapping of strict-ssl for unpublished projects lookup.
This results into falsy unpublished projects.

## Motivation and Context
We are using lerna to automatically determine which packages are unpublished. We have encountered the issue it tries to publish packages again and again regardless if they are already published.
With debugging I have found out there is an issue using the given strict-ssl attribute. It is not passed to pacote correctly.

In general it is the same issue like here #2952 

## How Has This Been Tested?
I have debugged everything and also tried to write an unit test which should cover this scenario.

## Types of changes

## Checklist:

  - there are some tests failing (perhaps of my Windows environment)

### DIFF
--- a/libs/commands/publish/src/lib/fetch-config.ts
+++ b/libs/commands/publish/src/lib/fetch-config.ts
@@ -1,6 +1,7 @@
 import log from "npmlog";
 
 export interface FetchConfig {
+  [key: string]: unknown;
   fetchRetries?: number;
   log: typeof log;
   registry?: string;
--- a/libs/commands/publish/src/lib/get-projects-with-unpublished-packages.spec.ts
+++ b/libs/commands/publish/src/lib/get-projects-with-unpublished-packages.spec.ts
@@ -82,3 +82,26 @@ test("getProjectsWithUnpublishedPackages with private package", async () => {
     }),
   ]);
 });
+
+test("getProjectsWithUnpublishedPackages with strict-ssl = false", async () => {
+  const cwd = await initFixture("public-private");
+  const packages = await getPackages(cwd);
+  const projectNodes = packages.map(
+    (pkg): ProjectGraphProjectNodeWithPackage => ({
+      name: pkg.name,
+      type: "lib",
+      data: { root: pkg.location },
+      package: pkg,
+    })
+  );
+
+  const opts = {'strict-ssl': false};
+  const pkgs = await getProjectsWithUnpublishedPackages(projectNodes, opts);
+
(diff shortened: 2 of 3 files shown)


JamesHenry:
Thanks @CSchulz, please run `npm run format:write` and commit the results, the formatting check is currently failing

JamesHenry (approved):
Thanks!

Issue #3693, opened 2023-05-20
Get error in gitlab ci


serveyk0-work:
Hi)
I use this command to publish my package,
npx lerna publish prepatch --preid rc --dist-tag rc --yes --force-publish --no-git-reset --loglevel silly --no-progress
however ran into a strange error in gitlab ci. Can you help me with it?
Thank you!

```
$ HUSKY=0 npx lerna publish prepatch --preid rc --dist-tag rc --yes --force-publish --no-git-reset  --loglevel silly --no-progress
lerna sill argv {
lerna sill argv   _: [ 'publish' ],
lerna sill argv   preid: 'rc',
lerna sill argv   'dist-tag': 'rc',
lerna sill argv   distTag: 'rc',
lerna sill argv   yes: true,
lerna sill argv   y: true,
lerna sill argv   'force-publish': true,
lerna sill argv   forcePublish: true,
lerna sill argv   'git-reset': false,
lerna sill argv   gitReset: false,
lerna sill argv   loglevel: 'silly',
lerna sill argv   progress: false,
lerna sill argv   composed: 'publish',
lerna sill argv   lernaVersion: '6.6.2',
lerna sill argv   '$0': 'node_modules/.pnpm/lerna@6.6.2/node_modules/lerna/dist/cli.js',
lerna sill argv   bump: 'prepatch'
lerna sill argv }
lerna notice cli v6.6.2
lerna verb rootPath /builds/serveyk0-work/serveyk0
lerna info versioning independent
lerna info ci enabled
lerna verb packageConfigs Package manager 'pnpm' detected. Resolving packages using 'pnpm-workspace.yaml'.
lerna verb session 7a2d1ba3e4fd13f2
lerna verb user-agent lerna/6.6.2/node@v20.2.0+x64 (linux)
lerna sill argv {
lerna sill argv   _: [ 'publish' ],
lerna sill argv   preid: 'rc',
lerna sill argv   'dist-tag': 'rc',
lerna sill argv   distTag: 'rc',
lerna sill argv   yes: true,
lerna sill argv   y: true,
lerna sill argv   'force-publish': true,
lerna sill argv   forcePublish: true,
lerna sill argv   'git-reset': false,
lerna sill argv   gitReset: false,
lerna sill argv   loglevel: 'silly',
lerna sill argv   progress: false,
lerna sill argv   composed: 'publish',
lerna sill argv   lernaVersion: '6.6.2',
lerna sill argv   bump: 'prepatch'
[...]

PR #3748, opened 2023-06-23
fix(publish): ensure that error code is a number

fahslaj:
## Description

Ensure the error code ends up as a number.

## Motivation and Context

Node will throw a different error if the process error code is not a number, which conceals the original error.

#3693 

## How Has This Been Tested?

This has been tested manually.

## Types of changes

## Checklist:


fahslaj:
Added logging of "errno" if it is not a number:
image

I made it log as an error instead of verbose since most times this case is it, it is a transient error and reproducing it with the "verbose" log setting would be annoying and inconsistent.

JamesHenry (approved):

PR #3762, opened 2023-07-05
Update version-and-publish.md

Kyoloro:
## Description
Document content error: "The remixapp is private"

## How Has This Been Tested?

While I am reading : )

## Types of changes

## Checklist:


JamesHenry:
Thanks @Kyoloro, but this text is intentional when you review the whole block. The app is being given as an example of something which isn’t private by its nature, but that nevertheless does not get published. So the private flag is leveraged despite it not being private.

Kyoloro:
> Thanks @Kyoloro, but this text is intentional when you review the whole block. The app is being given as an example of something which isn’t private by its nature, but that nevertheless does not get published. So the private property is leveraged despite it not being private.

In order not to publish to NPM, set the private = true, although "the remixapp isn't private" . XD

PR #3780, opened 2023-07-14
fix(core): Export cli.js for global installs

### LINKED ISSUE #3779: Lerna **Does Not** Run Locally (When installed globally and not using npx)
# Basic repro:
**The Basics:**
* `npm install -g lerna@latest`
* `mkdir test-proj && cd $_ && lerna init -y`
* `yarn install`

Now your environment will have lerna@latest both globally and within the local project

**Test Case:**
1) Open the file `./test-proj/node_modules/lerna/dist/cli.js`
2) Replace the whole file with:
```
#!/usr/bin/env node
throw new Error('Local does nothing');
```

**Where it will work as expected:**
The package manager variants will work as expected:
* `npx lerna list`
* `yarn lerna list`
* ...

(Which is expected because there is no interaction with the globally installed version)

**Where it fails:**
If you rely on the globally installed version, you get a very different result:
```
$ lerna list
lerna notice cli v7.1.3
lerna success found 0 packages
```

No failure!

## Cause
To understand the cause, we need to look at the globally installed version and we want to edit a file there. Use the following steps:
* `where lerna # -> /Users/me/.nvm/versions/node/v18.16.1/bin/lerna`
* Open the global directory (so in my case it would be `/Users/me/.nvm/versions/node/v18.16.1`)
* Edit the file: `/Users/joseph/.nvm/versions/node/v18.16.1/lib/node_modules/lerna/node_modules/resolve-from/index.js`
* In the resolve-from replace the last catch block (Ln 36-40) with:
```
    try {
      return resolveFileName();
    } catch (error) {
      console.log('Lerna Error: ', error)
      return;
    }
```

Now the error will be clear:
```
$ lerna list
Lerna Error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/cli.js' is not defined by "exports" in /Users/me/develop/test-proj/node_modules/lerna/package.json
    at new NodeError (node:internal/errors:393:5)
    at throwExportsNotFound (node:internal/modules/esm/resolve:358:9)
    at packageExportsResolve (node:internal/modules/esm/resolve:668:3)
    at resolveExports (node:internal/modules/cjs/loader:529:36)
    at Module._findPath (node:internal/modules/cjs/loader:569:31)
    at Module._resolveFilename (node:internal/modules/cjs/loader:981:27)
    at resolveFileName (/Users/me/.nvm/versions/node/v18.16.1/lib/node_modules/lerna/node_modules/resolve-from/index.js:30:39)
    at resolveFrom (/Users/me/.nvm/versions/node/v18.16.1/lib/node_modules/lerna/node_modules/resolve-from/index.js:38:11)
    at module.exports.silent (/Users/me/.nvm/versions/node/v18.16.1/lib/node_modules/lerna/node_modules/resolve-from/index.js:49:54)
    at module.exports.silent (/Users/me/.nvm/versions/node/v18.16.1/lib/node_modules/lerna/node_modules/resolve-cwd/index.js:5:49) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
```
## The fix
Not sure if the exports in the `package.json` are generated but a quick fix is the following:
* Open the file `./test-proj/node_modules/lerna/package.json`
* Add the following line to the `exports` entry of the package:
```
    "./dist/cli.js": "./dist/cli.js",
```

Now run the command:
```
$ lerna list
throw new Error('Local does nothing');
^

Error: Local does nothing
    at Object.<anonymous> (/Users/me/develop/test-proj/node_modules/lerna/dist/cli.js:2:7)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
```

Note: The failure is expected because we still haven't replaced the original contents of the local `dist/cli.js`. To really get it fix put that file back to original.

jd-carroll:
fixes #3779

### DIFF
--- a/packages/lerna/package.json
+++ b/packages/lerna/package.json
@@ -120,6 +120,7 @@
     "./commands/diff/lib/*": "./dist/commands/diff/lib/*.js",
     "./commands/publish/lib/*": "./dist/commands/publish/lib/*.js",
     "./commands/version/lib/*": "./dist/commands/version/lib/*.js",
+    "./dist/cli.js": "./dist/cli.js",
     "./migrations/**/*": "./dist/migrations/**/*",
     "./utils": "./dist/utils/index.js"
   },


JamesHenry (approved):
Thanks a lot for looking into this!

PR #3787, opened 2023-07-19
feat(version): sync the version of the package contents

matheo:
## Description

Usually in our CI we do `lint` -> `test` -> `built` -> `publish`  
but it has been a pain to build package contents (in other directory like `./dist`)
and then sync the generated version by `lerna publish` with a `version` hook in the root of the monorepo.

## Motivation and Context

In case we configure a publish directory different than the source folder  
we will need a custom script to sync the new version, otherwise the old version will be published.

## How Has This Been Tested?

I've tested it in my monorepo with the build output of this branch and it worked:

image

the config interpolation happens earlier than before and sync the `dist` output.

## Types of changes

## Checklist:

### DIFF
--- a/e2e/publish/src/custom-publish-directories.spec.ts
+++ b/e2e/publish/src/custom-publish-directories.spec.ts
@@ -555,4 +555,59 @@ describe("lerna-publish-custom-publish-directories", () => {
       `);
     });
   });
+
+  describe("version", () => {
+    it("should sync the version of the contents directory", async () => {
+      await fixture.updateJson("lerna.json", (lernaJson) => {
+        return {
+          ...lernaJson,
+          command: {
+            publish: {
+              directory: "{workspaceRoot}/dist/{projectRoot}",
+              syncDistVersion: true,
+            },
+          },
+        };
+      });
+
+      const version = randomVersion();
+
+      await fixture.lerna("create package-1 -y");
+      await fixture.updateJson("packages/package-1/package.json", (pkg) => ({
+        ...pkg,
+        main: "main.js",
+        version,
+        scripts: {
+          build: "cp ./lib/package-1.js ../../dist/packages/package-1/lib/main.js",
+          copyManifest: "cp ./package.json ../../dist/packages/package-1/package.json",
+        },
+      }));
+
+      // Make sure dist location exists before running mini build target
+      await ensureDir(fixture.getWorkspacePath("dist/packages/package-1/lib"));
+
+      await fixture.exec("git checkout -b test-main");
+      await writeFile(fixture.getWorkspacePath(".gitignore"), "node_modules\n.DS_Store\ndist");
+      await fixture.exec("git add .gitignore");
+      await fixture.exec("git add .");
+      await fixture.exec('git commit -m "initial-commit"');
+      await fixture.exec("git push origin test-main");
+
+      await fixture.lerna("run build,copyManifest");
+
+      await fixture.lerna(
+        `publish ${version} --registry=http://localhost:4872 --loglevel verbose --concurrency 1 -y`
+      );
+
+      const distVersion = JSON.parse(
+        await fixture.readWorkspaceFile("dist/packages/package-1/package.json")
+      ).version;
+
+      await fixture.exec(`npm unpublish --force package-1@${version} --registry=http://localhost:4872`);
+
+      expect(JSON.parse(await fixture.readWorkspaceFile("packages/package-1/package.json")).version).toEqual(
+        distVersion
+      );
+    });
+  });
 });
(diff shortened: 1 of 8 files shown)


matheo:
@fahslaj @JamesHenry I just found out that `lerna bootstrap` takes in account the package.json `publishConfig.directory` and it's like a central build's `options.outputPath` that should not be configured on command basis but just a single time like in the old bootstrap days.

Lkhmom85 (approved):

libs/commands/publish/src/index.ts
```diff
@@ -263,6 +263,58 @@ class PublishCommand extends Command {
...
+        const assets = this.project.config.command?.publish?.["assets"].map((asset) =>
+          interpolateAsset(asset, interpolateStr)
+        );
+        pkg.lernaConfig = pkg.lernaConfig || {};
+        pkg.lernaConfig.command = pkg.lernaConfig.command || {};
+        pkg.lernaConfig.command.publish = pkg.lernaConfig.command.publish || {};
+        pkg.lernaConfig.command.publish.assets = assets;
+      }
+    });
+
```
matheo, line 317: @fahslaj I just moved this code before `versionCommand` so it has the `contents` resolved to update the package version there with `syncDistVersion`

libs/core/src/lib/package.ts
```diff
@@ -286,6 +287,24 @@ export class Package {
     return writePkg(this.manifestLocation, this[PKG] as any).then(() => this);
   }
 
+  /**
+   * Sync dist manifest version
+   */
+  syncDistVersion() {
```
fahslaj, line 293: As mentioned in my review, this function can be converted to use `async/await` instead of `Promise.resolve` and `.then` chaining.

matheo:
@fahslaj I've tried to run `e2e-publish` locally with no luck, I may need help with the wording :thinking: 

hopefully the e2e works without the conventional commit of my setup,
and I'd like to know what's inside the `test-main` branch because I have no idea of the package `build` setup
and we need to be sure that after building the package, `lerna publish` bumps both `package.json` files

matheo:
@fahslaj I get a weird `init: required argument missing` https://cloud.nx.app/runs/TQp7pFoPgJ

my e2e test should be fine tho, perhaps I don't know which bump will be used without conventional commits :thinking:

fahslaj:
@matheo I've discovered a few things missing from the new e2e test. Can you please ensure that "allow edits by maintainers" is checked on the PR so I can push the fixes?

image

Specifically, what is missing is that the publish command needs to provide a version:
`publish ${version} --registry=http://localhost:4872 --loglevel verbose --concurrency 1 -y`

This version also needs to be set in `packages/package-1/package.json`, then `packages/package-1/package.json` needs to be copied to the dist directory.

matheo:
@fahslaj maintainers allowed now :)

I'm updating the version number only to the "output" folder because sometimes we have executors like `@nx/angular` that generates more output via `ng-packagr` in the `package.json` file so we cannot override it completely.

e2e/publish/src/custom-publish-directories.spec.ts
```diff
@@ -555,4 +555,52 @@ describe("lerna-publish-custom-publish-directories", () => {
...
+      });
+
+      await fixture.lerna("create package-1 -y");
+      await fixture.updateJson("packages/package-1/package.json", (pkg) => ({
+        ...pkg,
+        main: "main.js",
+        scripts: {
+          build: "cp ./lib/package-1.js ../../dist/packages/package-1/lib/main.js",
+        },
+      }));
```
fahslaj, line 580:
```suggestion
      const version = randomVersion();

      await fixture.lerna("create package-1 -y");
      await fixture.updateJson("packages/package-1/package.json", (pkg) => ({
        ...pkg,
        main: "main.js",
        version,
        scripts: {
          build: "cp ./lib/package-1.js ../../dist/packages/package-1/lib/main.js",
          copyManifest: "cp ./package.json ../../dist/packages/package-1/package.json",
        },
      }));
```

e2e/publish/src/custom-publish-directories.spec.ts
```diff
@@ -555,4 +555,52 @@ describe("lerna-publish-custom-publish-directories", () => {
...
+      await fixture.exec('git commit -m "initial-commit"');
+      await fixture.exec("git push origin test-main");
+
+      await fixture.lerna("run build");
+
+      await fixture.lerna(
+        "publish --registry=http://localhost:4872 --loglevel verbose --concurrency 1 -y"
+      );
+      const version = JSON.parse(await fixture.readWorkspaceFile("dist/packages/package-1/package.json")).version;
+
```
fahslaj, line 598:
```suggestion
      const distVersion = JSON.parse(
        await fixture.readWorkspaceFile("dist/packages/package-1/package.json")
      ).version;
```

fahslaj (changes requested):
@matheo for some reason, I was still unable to push to this PR. Instead I've added the changes I've made to the e2e test as comment suggestions. When you get a chance, try those changes and let me know if it still fails for you.

matheo:
@fahslaj Ok, I've synced your changes except `copyManifest` because that's supposed to be done by `lerna publish --sync-dist-version`, Let's see if e2e passes now

fahslaj:
It also appears that CI is failing due to code formatting. Be sure to run `npx nx format:check` and then `npx nx format:write` to make sure the changes match the formatting standards.

fahslaj (approved):
Thank you @matheo !

PR #3805, opened 2023-08-10
feat(init): detect existing package manager and auto-install packages

fahslaj:
## Description

Detects the package manager during `lerna init` by looking for existing lockfiles. Also makes the `init` command perform an installation task after the initialization to ensure Lerna is downloaded from the registry.

## Motivation and Context

This makes it easier to add Lerna to a project that is managed by Yarn or Pnpm.

## How Has This Been Tested?

This has been tested manually and covered by e2e tests.

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/init/src/command.ts
+++ b/libs/commands/init/src/command.ts
@@ -28,7 +28,7 @@ const command: CommandModule = {
       default: false,
     },
     skipInstall: {
-      describe: "Skip installation of npm dependencies after initialization",
+      describe: "Skip installation of dependencies after initialization",
       type: "boolean",
     },
   },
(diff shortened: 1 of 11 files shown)


libs/commands/init/src/command.ts
```diff
@@ -27,6 +27,10 @@ const command: CommandModule = {
       type: "boolean",
       default: false,
     },
+    skipInstall: {
+      describe: "Skip installation of npm dependencies after initialization",
```
JamesHenry, line 31:
```suggestion
      describe: "Skip installation of dependencies after initialization",
```

Seemed too specific in light of the features this PR itself is adding!

JamesHenry (changes requested):
Other than the one nit, please could update the docs to make note of this new behaviour?

JamesHenry (approved):

PR #3823, opened 2023-09-01
fix(version): option to not ignore scripts on lock update

matheo:
## Description

Option to not pass `--ignore-scripts` while updating the lock file after the version bump.

## Motivation and Context

Currently I'm using the old `lerna bootstrap` and I noticed that when versioning/publishing the npm lock file gets incomplete as `lerna version` is ignoring the scripts by default while updating it.

## How Has This Been Tested?

I've used `patch-package` to remove this code locally and it works as expected.

## Types of changes

## Checklist:

### DIFF
--- a/packages/lerna/schemas/lerna-schema.json
+++ b/packages/lerna/schemas/lerna-schema.json
@@ -1003,6 +1003,9 @@
             "gitTagVersion": {
               "$ref": "#/$defs/commandOptions/version/gitTagVersion"
             },
+            "runScriptsOnLockfileUpdate": {
+              "$ref": "#/$defs/commandOptions/version/runScriptsOnLockfileUpdate"
+            },
             "syncDistVersion": {
               "$ref": "#/$defs/commandOptions/version/syncDistVersion"
             },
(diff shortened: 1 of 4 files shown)


matheo:
@fahslaj here a new small one
it is super hard to pick a name for the option when `ignore-scripts` already exists and I wanted the new option to be next to `--npm-client-args` in the docs :sweat_smile:

fahslaj:
Hi @matheo , I think something like `--run-scripts-on-lockfile-update` would be a much more clear name for the argument. The current `--no-scripts-ignore-lock` is a bit ambiguous. You can always add a line to the readme near the `npmClientArgs` option that mentions and links to the new option, so someone looking at `npmClientArgs` can find the new one easily.

packages/lerna/schemas/lerna-schema.json
```diff
@@ -1703,6 +1706,10 @@
           "type": "boolean",
           "description": "During `lerna version`, when true, commit and tag version changes."
         },
+        "runScriptsOnLockfileUpdate": {
+          "type": "boolean",
+          "description": "During `lerna version`, when true, updates the lock file running the lifecycle scripts after the version bump."
```
fahslaj, line 1711: Can you please clarify this description a bit? "updates the lock file running the lifecycle scripts" is unclear. Maybe something like `... when true, runs lifecycle scripts when syncing the lock file after the version bump.`

fahslaj (changes requested):

packages/lerna/schemas/lerna-schema.json
matheo: Done!

fahslaj (approved):
Thank you @matheo !

matheo:
@fahslaj thanks!
any estimation for the next release?

Issue #2554, opened 2020-04-22
lerna version - "error: unable to sign the tag"


ernestostifano:
Lerna fails to sign tags when running `lerna version`.

## Expected Behavior

Tags to be correctly signed as well as commits (commits are successfully signed).

## Current Behavior

```
lerna ERR! Error: Command failed: git tag @scope/my-package@1.0.0-beta.1 -m @scope/my-package@1.0.0-beta.1
lerna ERR! error: gpg failed to sign the data
lerna ERR! error: unable to sign the tag
```
Please note that the same command (`git tag ...`) works fine and succesfully sign tags when manually ran.

## Possible Solution

Some ideas...
* Is `lerna version` running tasks in child processes? Does those processes have access to the GPG agent (GPG Suite in this case)?
* Commit signing seems to be correctly working, that should be a good staring point.
* What user information (name/email) is being used? (GPG key to be used is matched using that info).

## Steps to Reproduce (for bugs)

1. Correctly setup GPG and set commit.gpgSign and tag.gpgSign to true in global git config.
2. Run `lerna version` on project.
3. Follow steps. Process will fail when trying to sign tags.

<details><summary>lerna.json</summary><p>

```json
{
    "packages": ["./packages/*"],
    "version": "independent",
    "useWorkspaces": true,
    "npmClient": "yarn"
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.20.2 |
| `npm --version`  | 6.14.4 |
| `yarn --version` | 1.22.4 |
| `node --version` | 13.12.0 |

| OS | Version |
| --- | --- |
| NAME | VERSION |
| macOS Catalina | 10.15.3 |

carsonfarmer:
I'm seeing this exact same behavior with `npm` (as opposed to `yarn`).

ryexley:
I'm running into the same issue, also using Yarn. Commits are signed with no issues, but tagging fails. We would really prefer to enforce having our tags signed. Has anyone had any success getting tag signing to work without failing?

skyrpex:
Same problem here: the tagging fails when ran using lerna, but works correctly when manually ran.

skyrpex:
Strangely enough, I placed a console log at `@lerna/version/lib/git-tag.js` because I wanted to manually reproduce the problem, so it looks like this:

```
  log.verbose("git", args);
  console.log("git", args, opts);
  return childProcess.exec("git", args, opts);
```

[...]

PR #3832, opened 2023-09-08
Fix crash in `lerna version` when using `gitSignTag`

squidfunk:
When instructing lerna to sign tags, `gpg` often runs out of memory because it isn't designed to be run in parallel ([see similar issue on another repository](https://github.com/sbt/sbt-pgp/issues/168)). Thus, when `signGitTag` is set to true, signing must be carried out serially.

## Description

If `gitTagSign` is set to true, this PR will sign tags one after another, omitting the race condition

## Motivation and Context

Signing git tags

## How Has This Been Tested?

I'm not sure how to test it, as we'd need to somehow mock `gitTag` to check if we call it concurrently or not.

## Types of changes

## Checklist:

## Possibly related

- https://github.com/lerna/lerna/issues/2554

### DIFF
--- a/libs/commands/version/src/index.ts
+++ b/libs/commands/version/src/index.ts
@@ -850,9 +850,14 @@ class VersionCommand extends Command {
     if (await this.hasChanges()) {
       await gitCommit(message, this.gitOpts, this.execOpts);
     }
-    await Promise.all(
-      tags.map((tag) => gitTag(tag, this.gitOpts, this.execOpts, this.options.gitTagCommand))
-    );
+    if (this.gitOpts.signGitTag) {
+      for (const tag in tags)
+        await gitTag(tag, this.gitOpts, this.execOpts, this.options.gitTagCommand)
+    } else {
+      await Promise.all(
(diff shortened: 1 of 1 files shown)


fahslaj:
Hi @squidfunk , do you happen to have a public repo that this error occurs in? I created a small test repo and set up gpg locally and was not able to reproduce the issue you describe.

squidfunk:
Thanks for investigating. I don't currently have a public repo exposing this behavior. The problem only occurs sporadically, and you need to have several CPUs available and run `lerna version` concurrently on several packages so that tags are signed in parallel. I believe it's a race condition. Let me try to create a reproduction.

fahslaj:
@squidfunk Thank you for the extra context. I tested this on a larger repo with many packages and was able to reproduce the issue after all. I've pushed a small formatting change and will merge this as soon as CI passes.

fahslaj (approved):

fahslaj:
Thanks @squidfunk !

squidfunk:
Perfect! Thanks for the quick merge! Sorry that I wasn't more responsive, I had some other things on my plate today, but glad you could reproduce the problem ☺️ Next time I will provide a reproduction right away.

PR #3847, opened 2023-09-23
chore: use import instead of require for bundling diff command

amorscher:
- ensures that typechecks are executed for diff command during build

## Description
using require does not execute type checks during `npx nx run-many -t build --parallel=3`

## Motivation and Context
Ensure typechecking during build for libs as well. Remove warnings/errors from eslint regarding `require` statements

## How Has This Been Tested?
All unit tests and integration tests still pass.

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/diff/src/command.ts
+++ b/libs/commands/diff/src/command.ts
@@ -24,7 +24,7 @@ const command: CommandModule = {
   },
   handler(argv) {
     // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require(".")(argv);
+    return require(".").factory(argv);
   },
 };
 
--- a/libs/commands/diff/src/index.ts
+++ b/libs/commands/diff/src/index.ts
@@ -6,19 +6,19 @@ import { hasCommit } from "./lib/has-commit";
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const childProcess = require("@lerna/child-process");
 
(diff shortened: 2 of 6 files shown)


amorscher:
@fahslaj @JamesHenry What do you think about this approach?
I saw that require is used to include commands into lerna package. I recognized that using this appoach during build has the consequence that no typechecks are applied. During tests they will for sure. Using imports would solve this issue.

amorscher:
@fahslaj @JamesHenry what do you think about this approach? I recognized that during build no typechecks are executed for libs/command libraries. I know for tests typechecks are applied but not during  build. 
Replacing require/module.export statements with import/export  enables typechecking during build.

fahslaj:
@amorscher I like it. When changing how these commands and functions are exported, we should ensure that they can be imported by end users in the same way as before. For example:
```
const diff = require('lerna/commands/diff');
const getLastCommit = require('lerna/commands/diff/lib/get-last-commit');
```
I just tested this branch locally and this appears to be the case with `diff` after the change, so we should be good.

amorscher:
@fahslaj Ok good. Thx for having a look!
So I will apply this to other commands as well. I will do this command after command.

amorscher:
To make it clear that only the diff command is updated for now I will rephrase the commit message.

amorscher:
I will update the commit message to make it clear that only diff command is changed for now

JamesHenry (approved):

Issue #3727, opened 2023-06-13
`lerna version --force-publish` does not update all packages in yarn workspace


ilhan007:
Hello there,
we have an yarn workspace with multiple packages:

- base
- main
- fiori
....
- tools

Link to the project repo: https://github.com/SAP/ui5-webcomponents/

and we rely on lerna for our project release. It works fine and it's very usefule, however we faced unexpected behaviour that I hope you can help us with. We want to synchronise our  packages' versions and we use the `--force-publish` parameter to do so.

## Current Behavior

Currently, we run
```sh
yarn lerna version 1.14.2 --conventional-graduate --force-publish --yes --exact --create-release github 
        
yarn lerna publish from-git --yes
```

but sometimes not all packages' version are updated - some of them are updated to the new patch/minor, some of them remain on the version before running the commands. In the latest case all the packages but one were updated to 1.14.2 - one of them remained on 1.14.1

## Expected Behavior

When running
```sh
yarn lerna version 1.14.2 --conventional-graduate --force-publish --yes --exact --create-release github 
        
yarn lerna publish from-git --yes
```

we expect all the packages to be bumped to 1.14.2 even without changes.

## Steps to Reproduce

Locally it seems to work properly, so it might be related to the environment. It happens on GitHub actions (provided the environment details below), so I don't have a minimal reproducible example, but I hope that you might have seen this in the past and connect it somehow to the environment that we have.

### Failure Logs / Configuration

Link to the log of the  GitHub action
https://github.com/SAP/ui5-webcomponents/actions/runs/5241010719/jobs/9462553912

I post a screenshot of the log just in case. The screenshot shows the commands being executed and the packages that are updated - just one is missing the `tools`, it is not updated to 1.14.2.

The command executed is
`yarn lerna version 1.14.2 --conventional-graduate --force-publish --yes --exact --create-release github`

<img width="1372" alt="Screenshot 2023-06-13 at 13 19 23" src="https://github.com/lerna/lerna/assets/15702139/8a26da68-a086-4a0c-a6e5-788b35d68204">

Link to the `lerna.json` https://github.com/SAP/ui5-webcomponents/blob/release-1.14.2/lerna.json

### Environment

- Operating System Ubuntu 22.0[4](https://github.com/SAP/ui5-webcomponents/actions/runs/5241010719/jobs/9462553912#step:1:4).2 LTS
 - git version 2.40.1
- yarn 1.22.19
- node 16.20.0
- lerna v6.6.1

[...]

PR #3850, opened 2023-09-26
docs(version): udpate force-publish documentation for conventional workflow

amorscher:
## Description
Clarifies behaviour in #3727

## Motivation and Context
#3727

## Checklist:

### DIFF
--- a/libs/commands/version/README.md
+++ b/libs/commands/version/README.md
@@ -293,7 +293,7 @@ lerna version --force-publish=package-2,package-4
 lerna version --force-publish
 ```
 
-When run with this flag, `lerna version` will force publish the specified packages (comma-separated) or all packages using `*`. This option is ignored when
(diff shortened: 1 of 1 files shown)


libs/commands/version/README.md
```diff
@@ -293,9 +293,10 @@ lerna version --force-publish=package-2,package-4
 lerna version --force-publish
 ```
 
-When run with this flag, `lerna version` will force publish the specified packages (comma-separated) or all packages using `*`.
+When run with this flag, `lerna version` will force publish the specified packages (comma-separated) or all packages using `*`. This option is ignored when
```
fahslaj, line 296: >This option is ignored when

Is this meant to be here? It looks like this is stated below in the NOTE

amorscher: Yes that was a mistake. Should be ok now.

fahslaj (approved):

Issue #3867, opened 2023-10-17
Consider calling yarn and pnpm with corepack


trusktr:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

Some users may not have "installed" yarn or pnpm by running `corepack enable`, so when Lerna runs scripts using one of those, they get an error that yarn or pnpm can't be found.

As for yarn, installing it as a local dev dependency doesn't work (installs 1.22, not the latest).

I haven't tried pnpm, but I bet installing as a local dev dependency would be a workaround.

## Motivation

This is beneficial because things would just work without additional install steps.

As an example, I have scripts in package.json for getting started with a monorepo, by running `npm run link`. This links things with corepack yarn. But subsequent Lerna commands fail.

I *could* tell people to run corepack enable. I could also try to automate that, but it could interfere with someone else's flow.

## Suggested Implementation

Replacing `yarn` with `corepack yarn`, same with pnpm, should do the trick.

trusktr:
> I _could_ tell people to run corepack enable. I could also try to automate that, but it could interfere with someone else's flow.

On second thought, that'll be difficult, it requires sudo/admin by default.

trusktr:
Apparently it is more complicated that that! I ran `corepack enable`, so `yarn --version` in PowerShell shows the latest version.

I missed that the output before the above `NX` error shows this:

```
       Error: Cannot find module 'D:\src\lume+lume\node_modules\yarn\bin\yarn.js'
           at Module._resolveFilename (node:internal/modules/cjs/loader:1048:15)
           at Module._load (node:internal/modules/cjs/loader:901:27)
           at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
           at node:internal/main/run_main_module:23:47 {
```

So although `yarn --version` works when I call it in PowerShell, it still doesn't work when Lerna tries.

trusktr:
This is my first time on v7 with Yarn. Before this, I was using Lerna 4 with npm.

But now that I've upgraded, I don't have `npmClient` set but Lerna seems to want to use Yarn as hinted by the NX yarn --version output.

After I set `npmClient: "npm"`, I still see the same yarn --version error, which makes it seem like the option isn't being honored by Lerna based on the docs.

trusktr:
[...]

PR #3877, opened 2023-10-26
feat(version): use corepack for install when enabled

fahslaj:
## Description

Use `corepack` for install commands when it is enabled & use corepack in CI for package manager setup. Update tests to expect pnpm version 8.

When lerna is run via:
```
corepack yarn lerna version
```
then the install command run by the version command will be:
```
corepack yarn install
```

## Motivation and Context

#3867 

## Types of changes

## Checklist:


fahslaj:
I will change this pr to be into the v8 branch after the e2e test fix for yarn v4 is in

e2e/init/src/init-pnpm.spec.ts
```diff
@@ -53,18 +51,7 @@ describe("lerna-init-pnpm", () => {
...
-        CREATE lerna.json
-        UPDATE package.json
-        CREATE .gitignore
-        lerna info Initializing Git repository
-        lerna info Using pnpm to install packages
-        lerna success Initialized Lerna files
-        lerna info New to Lerna? Check out the docs: https://lerna.js.org/docs/getting-started
-
-      `);
+      expect(result.combinedOutput).toContain("lerna success Initialized Lerna files");
```
fahslaj, line 54: Filtering out pnpm install output is much harder on v8. I've opted to decouple this test from the pnpm output by just checking for success instead of comparing to an exact snapshot.

fahslaj: TODO: add comment explaining use of toContain over snapshot

Issue #3392, opened 2022-10-26
Add --skip-unstable option to lerna version command


Stompke:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

Add the option `--skip-unstable` to the lerna version options list.

## Motivation

I want to create a changelog during my command `npx lerna version --conventional-commits --conventional-graduate` that is based off changes between the last major version and current major version, and ignoring the pre-release versions in the middle.

I see that the tool `conventional-changelog` has the option `--skip-unstable` which says "If given, unstable tags will be skipped, e.g., x.x.x-alpha.1, x.x.x-rc.2"

This option `--skip-unstable` added to lerna version in creation of the changelog would be great!

NextThread:
hi, can I work on this

Issue #2248, opened 2019-09-02
CHANGELOGS not generated correctly when using "--conventional-graduate"


kelvincy:
## Expected Behavior

When versioning and publishing a prerelease version we do not want to generate the CHANGELOGS. This is done successfully with the "--no-changelog" for DTA environments. 

But when we do "--conventional-graduate" for the production release to remove the "preid", we want to generate a CHANGELOG for the production version. This is what we expect in the CHANGELOG for production.
```
## [0.0.7](git-tags) (2019-09-02)

### Bug Fixes

- commit 1
- commit 2
```

## Current Behavior

This is what we are currently seeing in the CHANGELOG for production.
```
## [0.0.7](git-tags) (2019-09-02)

**Note:** Version bump only for package @package
```

## Possible Solution

It probably has something to do with the git-tags for the prerelease versions that are already "released" with their changes. That is why the production publish does not see the changes anymore.

## Steps to Reproduce (for bugs)

For prerelease:
1. `lerna version --conventional-prerelease --no-changelog`
2. `lerna publish`

Then for production:
1. `lerna version --conventional-graduate`
2. `lerna publish`

<details><summary>lerna.json</summary><p>

```json
{
  "packages": ["packages/*"],
  "version": "independent",
  "command": {
    "publish": {
      "conventionalCommits": true,
      "ignoreChanges": ["**/*.md"],
      "yes": true
    },
    "version": {
      "conventionalCommits": true,
      "ignoreChanges": ["**/*.md"]
    }
  },
}
```
</p></details>

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.16.4 |
| `npm --version`  | 10 |
| `node --version` | 6 |

thuringia:
I can reproduce this.

This commit message is valid under the conventional commitlint config but does not yield an entry in the changelog, instead creating the "bump only" message:
`build(sweepstake): transpile the styleguide`
It does create when using `conventional-changelog` directly with the conventional preset, this doesn't work in though -> #2138

kelvincy:
Hey @thuringia, I tried with the commit messages `fix` and `feat`. Both of them will not generate CHANGELOGs when performing `lerna version --conventional-graduate` on production.

We have found a fix for this. Not sure if this is a preferred way, but it is currently working for us:

**Prereleases:**
1. `lerna version --conventional-prerelease --preid dev --no-changelog`
2. `git tag -d $(git describe --abbrev=0)`
[...]

PR #3882, opened 2023-10-30
feat(version): add --changelog-skip-unstable option

Juveniel:
## Description

Adds `--changelog-skip-unstable` option that allows to include changelogs from prereleases when publishing a stable release.

## Motivation and Context

fixes: https://github.com/lerna/lerna/issues/3392, https://github.com/lerna/lerna/issues/3119, https://github.com/lerna/lerna/issues/2248

## How Has This Been Tested?

Added unit and integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/integration/__tests__/lerna-publish-conventional-fixed-prerelease.spec.ts
+++ b/integration/__tests__/lerna-publish-conventional-fixed-prerelease.spec.ts
@@ -438,3 +438,285 @@ describe(`lerna publish --conventional-prerelease/graduate fixed w/ changelog`, 
     `);
   });
 });
+
+describe(`lerna publish --conventional-prerelease/graduate fixed w/ stable changelog`, () => {
+  let cwd: string;
+
+  beforeAll(async () => {
+    ({ cwd } = await cloneFixture("normal", "chore: init repo"));
+    await gitTag(cwd, "v1.0.0");
+  });
+
+  test(`release specified stable packages as prerelease, ignoring specified packages`, async () => {
+    const args = [
+      "publish",
+      "--conventional-commits",
+      "--conventional-prerelease",
+      "--yes",
+    ];
+    await commitChangeToPackage(cwd, "package-1", "feat(package-1): Add foo", { foo: true });
+    await commitChangeToPackage(cwd, "package-2", "fix(package-2): Fix bar", { bar: true });
+    await commitChangeToPackage(
+      cwd,
+      "package-3",
+      `feat(package-3): Add baz feature${os.EOL}${os.EOL}BREAKING CHANGE: yup`,
+      { baz: true }
+    );
+
+    const { stdout } = await cliRunner(cwd, env)(...args);
+    expect(stdout).toMatchInlineSnapshot(`
+
+      Changes:
+       - package-1: 1.0.0 => 2.0.0-alpha.0
+       - package-2: 1.0.0 => 2.0.0-alpha.0
+       - package-3: 1.0.0 => 2.0.0-alpha.0
+       - package-5: 1.0.0 => 2.0.0-alpha.0 (private)
+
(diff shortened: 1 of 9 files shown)


integration/__tests__/lerna-publish-conventional-fixed-prerelease.spec.ts
```diff
@@ -438,3 +438,285 @@ describe(`lerna publish --conventional-prerelease/graduate fixed w/ changelog`,
...
+
+
+
+      # [2.0.0-alpha.0](/compare/v1.0.0...v2.0.0-alpha.0) (YYYY-MM-DD)
+
+
+      ### Features
+
+      * **package-1:** Add foo ([SHA](COMMIT_URL))
+
```
fahslaj, line 629: 2.0.0-alpha.0 appears twice in this markdown file. This shouldn't happen, as users looking at version 2.0.0 won't care which specific prerelease version the feature was added in. Instead, we should get the changes from all the prereleases aggregated together with the changes from 2.0.0 itself, so it would look like this:
```
# [2.0.0](/compare/v1.0.0...v2.0.0) (YYYY-MM-DD)


### Features

* **package-1:** Add baz ([SHA](COMMIT_URL))
* **package-1:** Add foo ([SHA](COMMIT_URL))


# [2.0.0-alpha.0](/compare/v1.0.0...v2.0.0-alpha.0) (YYYY-MM-DD)


### Features

* **package-1:** Add foo ([SHA](COMMIT_URL))
```

fahslaj (changes requested):
We should not have duplicate prerelease headers in the generated markdown. Changes in prereleases should be aggregated together with the changes from the major release. See comment in `lerna-publish-conventional-fixed-prerelease.spec.ts`.

integration/__tests__/lerna-publish-conventional-fixed-prerelease.spec.ts
Juveniel: @fahslaj I do agree that this is the expected markup. 

https://github.com/conventional-changelog/conventional-changelog/issues/675 suggests this is an issue with `conventional-changelog`. However, tests in this PR with independent versioning seem to be correct and the prerelease changes are properly aggregated [see here](https://github.com/lerna/lerna/pull/3882/files/245a739118716bcf38849acf6e737c3ff78392fe#diff-dc2f58b3f119430f0a847687e18e51fb3a62e6dd131e0aea813455fcaed7cb18). Any suggestions ?

JamesHenry:
@Juveniel do you have plans to carry this forward please?

Juveniel:
Hey @JamesHenry,

In order to continue with this PR, I need to first fix [the issue](https://github.com/conventional-changelog/conventional-changelog/issues/675) in `conventional-changelog`. Unfortunately I am quite limited on time for the upcoming quarter and will not be able to do it asap. 

In general, I would love for this feature to be pushed forward, as it really is a pain to not be able to aggregate changes from pre-release versions in the actual graduated version. I am even considering switching to `nx release` as a somewhat faster alternative once it goes out of beta, as the issue seems to be resolved with nx.

JamesHenry:
Hi @Juveniel you are right that we are kind of blocked on this and so I am going to close this one for clarity. Please feel free to revisit if there is movement on the conventional changelog side

ecnaidar:
@Juveniel I guess you have seen that while investigating but does this not address the issue you referenced? 
https://github.com/conventional-changelog/conventional-changelog/pull/1066

Is there some specific behaviour that is missing in this case?

Juveniel:
@ecnaidar the above does not fix the issue unfortunately. It is a good step in the right direction, however there is more to be done in the default preset where the passed data is used to produce the content.

PR #3884, opened 2023-10-31
chore: update to Nx 17.0.3

fahslaj:


libs/nx-plugin/src/generators/e2e-project/generator.ts
```diff
@@ -36,16 +35,11 @@ function normalizeOptions(_tree: Tree, options: E2eProjectGeneratorSchema): Norm
...
-    tree,
-    {
-      name: normalizedOptions.name,
-      directory: normalizedOptions.projectDirectory,
-      skipTsConfig: true,
-    },
-    "./",
-    resolve("node_modules/@nx/js/src/generators/library/files")
-  );
+  await libraryGenerator(tree, {
```
fahslaj, line 38: This doesn't appear to have the same behavior since https://github.com/nrwl/nx/pull/18420 . I've fixed the compilation error but I don't think the previous behavior was preserved. 

`nx g @lerna/nx-plugin:e2e-project -d <name>` fails after this change

fahslaj: @JamesHenry Could you please compare this behavior with what's expected out of this generator?

JamesHenry: Pushed the fix

JamesHenry (approved):

PR #3886, opened 2023-11-01
chore(init): use import instead of require

amorscher:
ensures that typechecks are executed for init command during build

## Description
using require does not execute type checks during npx nx run-many -t build --parallel=3

## Motivation and Context
Ensure typechecking during build for libs as well. Remove warnings/errors from eslint regarding require statements

## How Has This Been Tested?
All unit tests and integration tests still pass.

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/init/src/command.ts
+++ b/libs/commands/init/src/command.ts
@@ -32,9 +32,8 @@ const command: CommandModule = {
       type: "boolean",
     },
   },
-  handler(argv) {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require(".")(argv);
+  async handler(argv) {
+    return (await import(".")).factory(argv);
   },
 };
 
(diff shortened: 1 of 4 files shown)


libs/commands/init/src/command.ts
```diff
@@ -32,9 +32,8 @@ const command: CommandModule = {
       type: "boolean",
     },
   },
-  handler(argv) {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require(".")(argv);
+  async handler(argv) {
+    (await import(".")).factory(argv);
```
fahslaj, line 36: @amorscher Does this need to have a `return` before this line or is that not relevant?

amorscher: @fahslaj In this case it is not relevant it would cause a type error as init command is not inheriting from command and thus it is not a promise having functions satisfying handler type ( void | promise<void>).
 Init command is directly executed in its constructor. 

I could add an explicit return after the import. with a comment explaining the reason for it? Or maybe you have a better solution? It is a little bit inconsistent with the general approach. I could also make initcmd a promise like class.

amorscher: I updated it to make the command fulfilling promise signature so it could be returned in the handle method. So I think now it is more consitent with other commands.

fahslaj (approved):
Thanks @amorscher!

PR #3890, opened 2023-11-05
chore(core): use specific logger for type safety

amorscher:
- support ts-compiler option "noPropertyAccessFromIndexSignature": true in  order to use use logger.success("message") instead of
logger['success']

## How Has This Been Tested?
As it is a type change-> compilation is still ok and all available tests still pass.

## Types of changes

## Checklist:


e2e/publish/src/publish-private-packages.spec.ts
```diff
@@ -302,7 +302,7 @@ describe("lerna-publish-private", () => {
         lerna notice cli v999.9.9-e2e.0
         lerna WARN Unable to determine published version, assuming "test-X" unpublished.
         lerna notice from-package No unpublished release found
-        lerna success No changed packages to publish 
+        lerna success publish No changed packages to publish
```
fahslaj, line 305: This prefix in this message doesn't seem to make it more readable or useful. I think it's fine without the prefix:
`lerna success No changed packages to publish` 

The command name in the prefix is mostly useful for the more verbose log levels as it helps debug issues in Lerna. For the success case, I think we want the most readable and concise messages possible.

fahslaj:
@amorscher looks good, just one small comment above about one of the message changes.

amorscher:
@fahslaj I also reverted the other message change in the publish command. Other messages are OK or should I also revert to the previous state.

fahslaj:
@amorscher I personally like how they were before, but I'm less worried about them because they're just for the legacy task runner.

amorscher:
Ok i reverted all of them.

fahslaj (approved):

PR #3901, opened 2023-11-16
chore(exec): enable strict type checking

amorscher:
updated ts-config to enable strict type checking
Use import instead of require to enable type checks during build

## Motivation and Context
Make exec command more typesafe.

## How Has This Been Tested?
All tests still pass.

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/exec/src/command.ts
+++ b/libs/commands/exec/src/command.ts
@@ -1,5 +1,6 @@
 import { filterOptions } from "@lerna/core";
 import type { CommandModule } from "yargs";
+import { factory } from "./index";
 
 /**
  * @see https://github.com/yargs/yargs/blob/master/docs/advanced.md#providing-a-command-module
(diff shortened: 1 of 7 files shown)


libs/commands/exec/src/command.ts
```diff
@@ -69,9 +70,8 @@ const command: CommandModule = {
 
     return filterOptions(yargs);
   },
-  handler(argv) {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require(".")(argv);
+  async handler(argv) {
+    return factory(argv);
```
amorscher, line 74: I did it like that because there is a bug in used yargs version.
If the function is async and we use an import with await. Yargs controll flow continues and skips -- arguments before the command is invoked. With yargs 17.7.2(latest) it is working as expected. I will update yargs with a seperate change.

fahslaj (approved):

fahslaj:
Thanks @amorscher !

PR #3908, opened 2023-11-21
chore(child-process): convert to typescript

amorscher:
## Description
Convert all *.js files to typescript so we have full typechecking support for child-process lib. This enables to use import statements providing type awareness for all commands using lerna's child-process lib.

## Motivation and Context
Make code more typesafe

## How Has This Been Tested?
All tests still pass.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Chore (change that has absolutely no effect on users)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3909, opened 2023-11-21
feat!: stop publishing @lerna/child-process

fahslaj:
## Description

Move @lerna/child-process into a private library.

## Motivation and Context

@lerna/child-process is by no means the best child process tool in the ecosystem. With Lerna v8, we will recommend that users that were previously depending directly on @lerna/child-process to use other tools instead. This also allows us to more easily update the underlying implementation of @lerna/child-process as we will no longer need to worry about breaking changes by updating dependencies like `execa`.

## Types of changes

## Checklist:


amorscher:
@fahslaj ok I think we have to discuss how we proceed with #3908. I was also working a little on the @lerna/child-process package. 

I think we could include types anyway and we just have to adapt paths in tsconfig.base.json.

fahslaj:
@amorscher yep! This PR is the bare minimum to get it to be an internal package instead of external, but the next step will be to convert it fully to typescript.

amorscher:
@fahslaj ok so I could do the conversion when this PR is merged? I have done some of the work already in #3908 .

fahslaj:
@amorscher yes that would be perfect.

Issue #3916, opened 2023-12-04
lerna version not creating correct tags since `lerna@7.3.0` when signing commits and tags


ivml:
## Current Behavior

Since `lerna@7.3.0`, using the `--sign-git-commit` and  `--sign-git-tag` flags causes `lerna version` to always create `0` and `1` as tags in a repository with two projects inside it.

This happens both when running the command locally or in a workflow and might have to do with me setting up my `git config` to sign commits already.

The full command is `yarn lerna version --conventional-commits --message "chore: bump version" --force-git-tag --sign-git-commit --sign-git-tag --yes --loglevel silly`. Running the command without the two flags related to signing fixes the issue. Running the command with `lerna@7.2.0` with the two flags also fixes the issue.

The version bump inside `package.json` seems correct, but the tags are always `0` and `1`:

![image](https://github.com/lerna/lerna/assets/4217267/56f36b98-5914-4001-b781-78f035d22a47)

## Expected Behavior

This is a regression bug. I've found that using `lerna@7.2.0` still creates the tags correctly when signing.

The created tags should reflect the version set in `package.json`. Like so:

![image](https://github.com/lerna/lerna/assets/4217267/447bc289-7500-4686-aadd-bc7891cce55c)

## Steps to Reproduce

- Can you reproduce this on https://github.com/lerna/repro?

Yes. This is visible here - https://github.com/ivml/lerna-repro

The repo has an initial commit setting up 2 projects with a single `md` file inside. For each subsequent commit, I change the `md` files so that it causes a version bump. Then, the next commits do the following:

- a [small change](https://github.com/ivml/lerna-repro/commit/cd91086830cf4595cf7c83dd19d2d40a4865e0d1) to the markdown files while using `lerna@7.2.0`. This correctly causes the [CI/CD to do a minor bump](https://github.com/ivml/lerna-repro/actions/runs/7089382070/job/19293996839#step:7:121).
- another [small change](https://github.com/ivml/lerna-repro/commit/606cf3123940e7f44de01d21ec9fd91af4f1c1f2) which also bumps `lerna` to 8.0.0.  This causes the [CI/CD to create the 0 and 1 tags](https://github.com/ivml/lerna-repro/actions/runs/7089440206/job/19294185655#step:7:126). This commit causes the issue.
[...]

PR #3917, opened 2023-12-08
fix(version): create correct independent tags when using --sign-git-tag

ivml:
Since `lerna@7.3.0`, running `lerna version` with independent versioning and the `--sign-git-tag` flag would always create `0`, `1`, ... `N` as tags instead of the actual version.

## Description

A recent fix done in #3832 introduced a new issue. In TypeScript, `for (const tag in tags)` would iterate through the array of tags. However, `for...in` returns a list of keys on the `tags` object, meaning that it iterates through the indices of the array, rather than the values. This means that regardless of what version number we are creating a tag for, that value will never be used - its index in the `tags` array will be used instead.

The issue is a bit niche, as it only happens when the repo is configured to use `independent` versioning and the `--sign-git-tag` flag is used.

## Motivation and Context

This is an issue because your project might have a specific version (e.g. `1.0.0`), but `lerna version` will create the tag `0` for it. This is especially crucial when using `--conventional-commits`, as that flag uses the latest tag in order to evaluate the next version bump.

The issue is described in more detail in #3916.

## How Has This Been Tested?
For this fix, I wrote some E2E tests in `e2e/version/src/sign-git-tag.spec.ts` that bump the version and use the `--sign-git-tag` flag. You can see the tests [failing](https://github.com/ivml/lerna/actions/runs/7131053286/job/19418821077#step:6:796) when using `for...in`, but [pass](https://github.com/ivml/lerna/actions/runs/7131253669/job/19419414299) when using `for...of`.

Things get a bit interesting - since the E2E tests run actual `git` commands, the tests need a GPG key, otherwise they will fail. For this reason, I have changed the `ci.yml` workflow to generate a GPG key, store its ID and revocation file, configure git to use the key, and finally, revoke and delete the key after the tests pass.

I have a feeling this might be overkill, and it probably makes testing locally harder, as it implies the user has setup a key themselves.

Therefore, **I would love some feedback on the tests** - whether it was a good idea, whether the implementation was overkill (especially for a one-word change), and whether there are alternatives (e.g. mocking git signing somehow).

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
 
       - name: Generate and configure GPG for signing commits and tags in E2E tests
         run: |
-          # Generate a PGP key for test@example.com and store the output from stderr
+          # Generate a GPG key for test@example.com and store the output from stderr
           GPG_OUTPUT=$(echo "Key-Type: default
           Key-Length: 2048
           Subkey-Type: default
@@ -81,7 +81,7 @@ jobs:
           # Find and extract the revocation file path from sdterr
           REVOCATION_FILE=$(echo "$GPG_OUTPUT" | grep '.rev' | tr '\n' ' ' | awk -F "'" '{print $4}')
 
-          # Get the PGP key ID and the full fingerprint
+          # Get the GPG key ID and the full fingerprint
           export GPG_KEY_ID=$(gpg --list-secret-keys --keyid-format LONG | grep sec | awk '{print $2}' | cut -d'/' -f2)
           export GPG_FULL_KEY_ID=$(gpg --list-secret-keys --keyid-format LONG | grep "$GPG_KEY_ID" | grep -v "sec" | awk '{print $1}' | cut -d'/' -f2)
 
@@ -104,8 +104,10 @@ jobs:
           NX_AGENT_NAME: ${{ matrix.agent }}
 
       - name: Revoke and delete GPG key
+        # It's important that we always run this step, otherwise the key will remain active if any of the steps above fail
+        if: ${{ always() }}
         run: |
-          # As instructed in the revocation file, the colon needs to first be removed
+          # As instructed in the text of revocation file, there is a colon that needs to be removed manually
           sed -i "s/:-----BEGIN PGP PUBLIC KEY BLOCK-----/-----BEGIN PGP PUBLIC KEY BLOCK-----/" $REVOCATION_FILE
 
           # Revoke the key and delete it
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,35 @@ jobs:
           git config --global user.email test@example.com
           git config --global user.name "Tester McPerson"
 
+      - name: Generate and configure GPG for signing commits and tags in E2E tests
+        run: |
+          # Generate a PGP key for test@example.com and store the output from stderr
+          GPG_OUTPUT=$(echo "Key-Type: default
+          Key-Length: 2048
+          Subkey-Type: default
+          Subkey-Length: 2048
+          Name-Real: Tester McPerson
+          Name-Email: test@example.com
+          Expire-Date: 0
+          %no-protection" | gpg --pinentry-mode loopback --batch --generate-key 2>&1)
+
+          # Find and extract the revocation file path from sdterr
+          REVOCATION_FILE=$(echo "$GPG_OUTPUT" | grep '.rev' | tr '\n' ' ' | awk -F "'" '{print $4}')
+
+          # Get the PGP key ID and the full fingerprint
+          export GPG_KEY_ID=$(gpg --list-secret-keys --keyid-format LONG | grep sec | awk '{print $2}' | cut -d'/' -f2)
+          export GPG_FULL_KEY_ID=$(gpg --list-secret-keys --keyid-format LONG | grep "$GPG_KEY_ID" | grep -v "sec" | awk '{print $1}' | cut -d'/' -f2)
+
+          # Export fingerprint and the path to the revocation file to GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -74,6 +103,15 @@ jobs:
         env:
           NX_AGENT_NAME: ${{ matrix.agent }}
 
+      - name: Revoke and delete GPG key
+        run: |
```
JamesHenry, line 107: We would want this to always run, even in error cases, right? So we need to add the if always syntax to this step

.github/workflows/ci.yml
```diff
@@ -66,6 +66,35 @@ jobs:
           git config --global user.email test@example.com
           git config --global user.name "Tester McPerson"
 
+      - name: Generate and configure GPG for signing commits and tags in E2E tests
+        run: |
+          # Generate a PGP key for test@example.com and store the output from stderr
```
JamesHenry, line 71:
```suggestion
          # Generate a GPG key for test@example.com and store the output from stderr
```

.github/workflows/ci.yml
```diff
@@ -66,6 +66,35 @@ jobs:
...
+          Subkey-Length: 2048
+          Name-Real: Tester McPerson
+          Name-Email: test@example.com
+          Expire-Date: 0
+          %no-protection" | gpg --pinentry-mode loopback --batch --generate-key 2>&1)
+
+          # Find and extract the revocation file path from sdterr
+          REVOCATION_FILE=$(echo "$GPG_OUTPUT" | grep '.rev' | tr '\n' ' ' | awk -F "'" '{print $4}')
+
+          # Get the PGP key ID and the full fingerprint
```
JamesHenry, line 84:
```suggestion
          # Get the GPG key ID and the full fingerprint
```

.github/workflows/ci.yml
```diff
@@ -74,6 +103,15 @@ jobs:
         env:
           NX_AGENT_NAME: ${{ matrix.agent }}
 
+      - name: Revoke and delete GPG key
+        run: |
+          # As instructed in the revocation file, the colon needs to first be removed
+          sed -i "s/:-----BEGIN PGP PUBLIC KEY BLOCK-----/-----BEGIN PGP PUBLIC KEY BLOCK-----/" $REVOCATION_FILE
```
JamesHenry, line 109:
```suggestion
          sed -i "s/:-----BEGIN GPG PUBLIC KEY BLOCK-----/-----BEGIN GPG PUBLIC KEY BLOCK-----/" $REVOCATION_FILE
```

JamesHenry (changes requested):
Thanks a lot for this @ivml!

Out of curiosity was this GPG key gen approach copied from/inspired by another open source repo?

JamesHenry:
Sorry just read your note about the potential difficulty for local test runs, I agree.

Maybe we could exclude this spec if `process.env.CI !== 'true'`?

.github/workflows/ci.yml
ivml: This one is not a typo. If you were to open a revocation file in a text editor, it would look something like this:

```
This is a revocation certificate for the OpenPGP key:

pub   rsa2048 2023-12-11 [SC]
      81CFAE4BD0034D6D1AD443408F11F369BDA6B541
uid          Tester McPerson <test@example.com>

A revocation certificate is a kind of "kill switch" to publicly
declare that a key shall not anymore be used.  It is not possible
to retract such a revocation certificate once it has been published.

Use it to revoke this key in case of a compromise or loss of
the secret key.  However, if the secret key is still accessible,
it is better to generate a new revocation certificate and give
a reason for the revocation.  For details see the description of
of the gpg command "--generate-revocation" in the GnuPG manual.

To avoid an accidental use of this file, a colon has been inserted
before the 5 dashes below.  Remove this colon with a text editor
before importing and publishing this revocation certificate.

:-----BEGIN PGP PUBLIC KEY BLOCK-----
Comment: This is a revocation certificate

iQE2BCABCAAgFiEEgd+uS9ADTW0a1ENAjxHzab2msUEFAmV2yBoCHQAACgkQjxHz
...

-----END PGP PUBLIC KEY BLOCK-----
```

ivml:
> Maybe we could exclude this spec if process.env.CI !== 'true'?

Thanks, that is a good suggestion. I attempted to do that in https://github.com/lerna/lerna/pull/3917/commits/a8378c3278cc3c67996a677d92c4852e0784c3da but I am unsure if there are not better alternatives for the syntax.

> Out of curiosity was this GPG key gen approach copied from/inspired by another open source repo?

There is not a particular example that I am aware of.

Signing a commit in a workflow is something that can be done relatively easy if you have generated a key, uploaded it to your account and made some secrets. I have used this gist as an example https://gist.github.com/vansergen/88eb7e71fea2e3bdaf6aa3e752371eb7

Generating the key inside the workflow, storing it, and revoking it was way less intuitive. I had a bunch of failed attempts in my fork that took small incremental changes to get to here (that I have cleaned up). I mostly had an idea, and took it to Copilot for validation and error deciphering. For example, it was that back-and-forth that told me that the revocation file path is written to `stderr` instead of `stdout` after successfully generating a key.

.github/workflows/ci.yml
JamesHenry: Ah sorry that was me just not being familiar enough with these specs/standards

JamesHenry:
Thanks a lot @ivml!

.github/workflows/ci.yml
ivml: To be honest it's an easy thing to miss - your other suggestions were indeed typos, so it made sense for this to also be a typo :) It's good to double-check in these situations, so your feedback was spot on!

PR #3929, opened 2024-01-03
fix(add-caching): explicitly set targetDefaults for all scripts

fahslaj:
## Description

- Explicitly sets `targetDefaults` for all scripts, not just those selected in the prompts.
- Updates how cacheable targets are configured to align with Nx 17 expectations.

## Motivation and Context

This allows `lerna run` to correctly identify when the user has configured Nx via `lerna add-caching`, even if they didn't select any scripts to run in order. This also fixes `add-caching` for Nx 17.

## How Has This Been Tested?

This has been tested manually on a fresh workspace.

## Types of changes

## Checklist:


packages/lerna/src/commands/add-caching/index.ts
```diff
@@ -55,16 +64,23 @@ class AddCachingCommand extends Command {
...
+      (scriptName) => nxJson.targetDefaults?.[scriptName]?.dependsOn?.length
+    );
     const { targetDefaults } = await inquirer.prompt<{ targetDefaults: UserAnswers["targetDefaults"] }>([
       {
         type: "checkbox",
         name: "targetDefaults",
         message:
           "Which scripts need to be run in order? (e.g. before building a project, dependent projects must be built.)\n",
         choices: this.uniqueScriptNames,
+        default: existingTargetDefaults,
```
fahslaj, line 77: This allows inquirer to pre-select all targets that already have dependsOn configured

image

packages/lerna/src/commands/add-caching/index.ts
```diff
@@ -74,13 +90,13 @@ class AddCachingCommand extends Command {
         message:
           "Which scripts are cacheable? (Produce the same output given the same input, e.g. build, test and lint usually are, serve and start are not.)\n",
         choices: this.uniqueScriptNames,
+        default: existingCacheableOperations,
```
fahslaj, line 93: Same as above; this allows inquirer to pre-select all targets that have `"cache": true` already configured

packages/lerna/src/commands/add-caching/index.ts
```diff
@@ -108,29 +126,7 @@ class AddCachingCommand extends Command {
...
-    nxJson.tasksRunnerOptions["default"].runner =
-      nxJson.tasksRunnerOptions["default"].runner || "nx/tasks-runners/default";
-    nxJson.tasksRunnerOptions["default"].options = nxJson.tasksRunnerOptions["default"].options || {};
-
-    if (nxJson.tasksRunnerOptions["default"].options.cacheableOperations) {
-      this.logger.warn(
-        "add-caching",
-        "The `tasksRunnerOptions.default.cacheableOperations` property already exists in `nx.json` and will be overwritten by your answers"
-      );
-    }
```
fahslaj, line 130: Adding the default task runner seems like excess config. Nx should default this for us when this is missing.

The meaningful removal here is of `cacheableOperations`, which moved to `targetDefaults`.

website/docs/api-reference/configuration.md
```diff
@@ -69,10 +69,7 @@ Find the available options in [the API docs](/docs/api-reference/commands).
 {
   "tasksRunnerOptions": {
     "default": {
-      "runner": "nx/tasks-runners/default",
-      "options": {
-        "cacheableOperations": ["build", "test"]
-      }
+      "runner": "nx/tasks-runners/default"
```
JamesHenry, line 72: Shouldn’t we remove this here too for simplicity?

website/docs/api-reference/configuration.md
```diff
@@ -99,16 +98,16 @@ Find the available options in [the API docs](/docs/api-reference/commands).
 Everything in Nx is customizable, including running npm scripts. Most of the time you will either use the default runner
 or the `@nrwl/nx-cloud` runner.
 
-### cacheableOperations
```
JamesHenry, line 102: We can remove the paragraph on runner above it’s outdated

website/docs/concepts/how-caching-works.md
```diff
@@ -199,7 +199,6 @@ option for the task runner in `nx.json`:
   "tasksRunnerOptions": {
     "default": {
       "options": {
-        "cacheableOperations": ["build", "test"],
```
JamesHenry, line 202: This whole config is outdated too, cache directory is set at the top level, no needs for the object

website/docs/concepts/task-pipeline-configuration.md
```diff
@@ -32,7 +32,6 @@ Note, you can also change the default in `nx.json`, like this:
     "default": {
       "runner": "nx/tasks-runners/default",
       "options": {
-        "cacheableOperations": [],
```
JamesHenry, line 35: Same again, outdated

packages/lerna/src/commands/add-caching/index.ts
```diff
@@ -156,6 +159,22 @@ class AddCachingCommand extends Command {
...
     writeJsonFile(nxJsonPath, nxJson);
   }
+
+  private async configureGitIgnore(): Promise<void> {
+    try {
+      await execa("git", ["check-ignore", ".nx/cache"]);
+      // .nx/cache is already ignored - no need to update .gitignore
+    } catch (e) {
+      try {
+        await writeFile(joinPathFragments(workspaceRoot, ".gitignore"), "\n.nx/cache\n", { flag: "a+" });
```
JamesHenry, line 169: I’m not familiar with passing this a+ flag, I would have gone for appendFile, I’m on mobile and very late but assuming this makes this equivalent?

JamesHenry (changes requested):
See comments

packages/lerna/src/commands/add-caching/index.ts
fahslaj: Yes, this is equivalent to using `appendFile`. I went ahead and changed it to use `appendFile` since that's more readable.

website/docs/concepts/task-pipeline-configuration.md
fahslaj: Removed

website/docs/concepts/how-caching-works.md
fahslaj: I've removed this whole section. We're pretty clear that it's Nx doing the caching in the docs and users can find the Nx docs if they need to do advanced things like changing the cache location.

I think we should minimize the amount Nx-specific docs on the Lerna site as they will get out of date quickly.

fahslaj:
@JamesHenry I've updated the docs you've pointed out. I removed a couple of the outdated sections instead of just updating their config. To reduce the maintenance burden, I think we should keep the more advanced Nx configuration options to the Nx docs and link to the Nx site instead of duplicating them in the Lerna docs.

JamesHenry (approved):

PR #3933, opened 2024-01-10
chore(docs): add explicit example of using environment variables on windows

olton:
The specifics of using environment variables in Windows. User must frame the variable in `%`. For example:
```sh
$ lerna watch -- lerna run build --scope=%LERNA_PACKAGE_NAME% --include-dependents
```


JamesHenry (approved):
Thanks @olton

Issue #3837, opened 2023-09-12
Customize "@" separator character when creating git tag


Drew-Kimberly:
- [x] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

I'd like to be able to specify a character other than `@` to be used as the separator when creating a git tag.

Ex: ~~`my-pkg@0.1.0`~~ -> `my-pkg_0.1.0`

## Motivation

The `@` character can cause problems in other systems when referencing a git tag directly is necessary.

## Suggested Implementation

Looks like it's hardcoded to `@` [here](https://github.com/lerna/lerna/blob/main/libs/commands/version/src/index.ts#L845). Add a new optional `version-separator` option to the `version` command that defaults to `@` and gets threaded through the program.

PR #3951, opened 2024-02-05
feat(version): custom tag-version-separator for independent projects

### LINKED ISSUE #3902: Custom tag template when independent versioning
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

We have a mono-repo with individual versioning.
We use `lerna version --conventional-commits --git-tag-version false` to bump package versions and generate changelogs.
After that we use `lerna publish from-package` to release the packages.

By default, lerna generates tags in the format `@scope/package-name@X.Y.Z`.
We disable the auto-generated lerna tags and as part of our release process, we create git tags in the format `package-name/vX.Y.Z` (without the `@scope`).

With custom tag names, `lerna changed`, `lerna diff` and `lerna version` successfully detects which packages are changed since the last release. Unfortunately, when generating the changelogs, it ignores the custom tags and searches for tags with the default name.

I would like to have an option to change the tag format that lerna uses.

## Motivation

There are two reasons we prefer the different format:
1. The `@scope` is the same for all our packages, so no need to add it. It only adds noise.
2. Some GUI tools (e.g. GitLens) visually group the list of tags into folders. They use `/` as a separator. With the default format it will create one folder `@scope` with all tags in it. With our custom format it will create individual folders for each package, with the tags just for that package.

## Suggested Implementation

While investigating why `lerna version` is not able to generate the correct change logs, I saw that it uses `conventional-recommended-bump` which uses `git-semver-tags`. The latter uses the `@scope/package-name@X.Y.Z` format. Lerna itself does not care about the names of the tags, as long there is tag for every release.

For independent versioning `--tag-version-prefix` is ignored. We can make use of it and add a placeholders for package name and scope, for example:
- `%s` for the `"@scope"` and `%p` for `"package-name"`.
- `--tag-version-prefix '%s/%p@'` for the default tag names.
- `--tag-version-prefix '%p/v'` for our custom naming.

Lerna will use the prefix when generating tags.
The prefix can be provided as `tagPrefix` to `conventional-recommended-bump` and `git-semver-tags`.

## Alternate Implementations

For now, we'll stick with the default lerna tag names, but really hope to consider adding such a feature.
Thanks!

JamesHenry:
## Description

Allows users to customize the tag version separator for independent projects in cases where they do not want to use the default of `@`.

Example usage can be seen in the e2e test for this project.

## Motivation and Context

Closes #3902 #3837

## How Has This Been Tested?

e2e and unit tests added.

## Types of changes

## Checklist:

### DIFF
--- a/e2e/version/src/tag-version-separator.spec.ts
+++ b/e2e/version/src/tag-version-separator.spec.ts
@@ -1,0 +1,92 @@
+import { Fixture, normalizeCommitSHAs, normalizeEnvironment } from "@lerna/e2e-utils";
+
+expect.addSnapshotSerializer({
+  serialize(str: string) {
+    return normalizeCommitSHAs(normalizeEnvironment(str));
+  },
+  test(val: string) {
+    return val != null && typeof val === "string";
+  },
+});
+
+describe("lerna-version-tag-version-separator qqqq", () => {
+  let fixture: Fixture;
+
+  beforeEach(async () => {
+    fixture = await Fixture.create({
+      e2eRoot: process.env.E2E_ROOT,
+      name: "lerna-version-tag-version-separator",
+      packageManager: "npm",
+      initializeGit: true,
+      lernaInit: { args: [`--packages="packages/*" --independent`] },
+      installDependencies: true,
+    });
+    await fixture.lerna("create package-a -y");
+    await fixture.lerna("create package-b -y");
+    await fixture.updateJson("lerna.json", (json) => ({
+      ...json,
+      command: {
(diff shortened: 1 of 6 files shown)


briananstett:
I created a new issue for the GitHub Release failure created from this change. I believe the issue is with this [line of code](https://github.com/lerna/lerna/blob/main/libs/commands/version/src/lib/create-release.ts#L26)

https://github.com/lerna/lerna/issues/3974

PR #3971, opened 2024-03-09
fix(list): flush output before exiting

### LINKED ISSUE #3970: `lerna list` output is truncated at 8192 bytes when `exec`ed
## Current Behavior

When executing `lerna list` from within a Node process via `exec` or `spawn`, the stdout is truncated at 8192 bytes.

## Expected Behavior

All stdout output should be captured.

This appears to have been introduced with bafe090, because the `process.exit` terminates the process handle before the output buffers are finished flushing.

Reference: https://stackoverflow.com/a/59322701/27578

## Steps to Reproduce

Run this code in a monorepo which has a large enough number of packages to push the output of `list` past 8192 bytes.

```
const { exec } = require("child_process");

exec("npx lerna@8.1.2 list --all --json", (error, stdout, stderr) => {
  console.log(`stdout: ${stdout}`);
});
```

Other variantions tried:

```
const { exec } = require("child_process");

const process = exec("npx lerna@8.1.2 list --all --json");

let stdout = "";
process.stdout.on("data", (chunk) => (stdout += chunk));

process.on("close", () => {
  console.log(stdout);
});
```

```
const { spawn } = require("child_process");

const process = spawn("npx", ["lerna@8.1.2", "list", "--all", "--json"]);

let stdout = "";
process.stdout.on("data", (chunk) => (stdout += chunk));

process.on("close", () => {
  console.log(stdout);
});
```

These snippets work as expected with `lerna@8.1.0`.

smerrill:
## Description
Fixes #3970.

## Motivation and Context
Without this code, any code trying to consume the output of `lerna list` would only get the first 8k.

## How Has This Been Tested?
It has not yet been tested other than the simple repro I put into #3970 .

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/list/src/index.ts
+++ b/libs/commands/list/src/index.ts
@@ -37,6 +37,6 @@ export class ListCommand extends Command {
       this.result?.count === 1 ? "package" : "packages"
     );
 
-    process.exit(0);
+    process.stdout.end(() => process.exit(0));
   }
 }


JamesHenry (approved):

JamesHenry:
Thanks a lot @smerrill!

Issue #3974, opened 2024-03-12
No GitHub release is created when using a 'tag-version-separator' other than '@'


briananstett:
## Current Behavior
After the [8.1.0](https://github.com/lerna/lerna/releases/tag/v8.1.0), the `version` command does not create a GitHub release when using a custom version seperator.

## Expected Behavior
I would expect there to be a new GitHub release associated with the new git tag.

## Steps to Reproduce
1. Create a new commit for one of your packages.
```bash
git commit -m'fix(package1): lerna version seperator test'
```
2. Create a new version of your package
```
$ npx lerna version --tag-version-separator '-' --conventional-commits --create-release github --yes
lerna notice cli v8.1.0
lerna info versioning independent
lerna info Looking for changed packages since package1g@1.0.7
lerna info getChangelogConfig Successfully resolved preset "conventional-changelog-angular"

Changes:
 - package1: 1.0.7 => 1.0.8

lerna info auto-confirmed 
lerna info execute Creating releases...
lerna info git Pushing tags...
lerna success version finished
```

If I repeat this step but use the '@' seperator, everything works as expected and I can see there was a `fetch` call presumably to GitHub to create a GitHub release.
```
$ git commit -m'fix(package1): lerna version seperator
$ npx lerna version --tag-version-separator '@' --conventional-commits --create-release github --yes
lerna notice cli v8.1.0
lerna info versioning independent
lerna info Looking for changed packages since package1-1.0.8
lerna info getChangelogConfig Successfully resolved preset "conventional-changelog-angular"

Changes:
 - package1: 1.0.8 => 1.0.9

lerna info auto-confirmed 
lerna info execute Creating releases...
lerna info git Pushing tags...
(node:54189) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
lerna success version finished
```
I believe the issue with [this line of code](https://github.com/lerna/lerna/blob/main/libs/commands/version/src/lib/create-release.ts#L26) in the `create-release.ts` file. It has a conditional that's checking if the incoming tag matches a string which contains a  hard coded `@` character. From some basic testing, something like this seems to resolve the issue.

```javascript
const tag = name === "fixed" ? tags[0] : tags.find((t) => t.startsWith(`${name}${separator}`));
```

I'm happy to make PR if it is indeed that simple of a fix.

### Failure Logs / Configuration

<details><summary>lerna.json</summary><p>

```json
{
[...]

PR #3979, opened 2024-03-21
fix(version): create release when using custom tag-version-separator

briananstett:
## Description
After the [8.1.0](https://github.com/lerna/lerna/releases/tag/v8.1.0), the version command does not create a GitHub release when using a custom version seperator. This was because the `createRelease` function had a hard coded `@` in tag name filter.

## Motivation and Context
Fixes: https://github.com/lerna/lerna/issues/3974

## How Has This Been Tested?
To test, I create a test repo made changes to a packages and commit them. I then ran:
```
node /tmp/lerna/dist/packages/lerna/dist/cli.js version --tag-version-separator '-' --conventional-commits --create-release github --yes
``` 
And was able to see the new GitHub release created with the appropriate tag

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/version/src/lib/create-release.ts
+++ b/libs/commands/version/src/lib/create-release.ts
@@ -16,21 +16,26 @@ export function createReleaseClient(type: "github" | "gitlab") {
 
 export function createRelease(
   client: ReturnType<typeof createReleaseClient>,
-  { tags, releaseNotes }: { tags: string[]; releaseNotes: { name: string; notes: string }[] },
+  {
+    tags,
+    releaseNotes,
+    tagVersionSeparator,
+  }: { tags: string[]; tagVersionSeparator: string; releaseNotes: { name: string; notes: string }[] },
   { gitRemote, execOpts }: { gitRemote: string; execOpts: ExecOptions }
 ) {
   const repo = parseGitRepo(gitRemote, execOpts);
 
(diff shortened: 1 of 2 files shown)


libs/commands/version/src/lib/create-release.ts
```diff
@@ -17,20 +17,21 @@ export function createReleaseClient(type: "github" | "gitlab") {
 export function createRelease(
   client: ReturnType<typeof createReleaseClient>,
   { tags, releaseNotes }: { tags: string[]; releaseNotes: { name: string; notes: string }[] },
-  { gitRemote, execOpts }: { gitRemote: string; execOpts: ExecOptions }
+  { gitRemote, execOpts }: { gitRemote: string; execOpts: ExecOptions },
+  separator: string
```
ghiscoding, line 21: I don't think it's a good idea to add yet another argument, I think it might be better to include it in either 2nd or 3rd argument, this makes it easier to extend later if other arguments are to be added in the future

ghiscoding: looks good

JamesHenry (approved):

JamesHenry:
Thank you very much!

briananstett:
@JamesHenry do you know when the next release might be? Looking forward to getting this fix out so I can use it.

PR #4013, opened 2024-05-27
feat(publish): enable throttling when publishing modules

### LINKED ISSUE #3962: Support publish throttling
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

I'd like Lerna to enable large monorepo maintainers that don't use `independant` as their versioning scheme to be able to publish dozens of packages to NPM within their release engineering process, while respecting the NPM throttling limits.

## Motivation

For large monorepos, consisting of dozens of NPM packages that are updated as a group, it is important to avoid being throttled by NPM when publishing. NPM enforces non-documented throttling limits that can ban an account for hours, leaving the publishing process half completed, and resulting in a poor developer and customer experience.

## Suggested Implementation

Enable a `--throttling` option on the `lerna publish` command to define either the throttling mechanism in requests per second and/or per minute to the registry.

Ayowel:
The aim is to provide a way to fix #3962 .

This is opened to validate the implementation design and get feedback on the specific values that should be used as defaults.

Currently known limits when uploading to NPM:

* Burst limit of 25/500 modules per 24h (depending on account/already-published modules)

## Description

Add three publishing options:

* `--throttle` to enable/disable throttling
* `--throttle-size`: the 'bucket size' to use when throttling
* `--throttle-delay`: Once the bucket is full, the minimum time between the end of a module upload and the start of an other module's upload

Throttling is enabled by default when publishing to NPM. The throttling implementation is designed to allow changes in the implementation/maintaining different implementations without too much trouble

## Motivation and Context

See #3962: NPM throttles module publishing, which results in partial publish which are a pain to deal with

## How Has This Been Tested?

Tested on ubuntu with node 20.9.0:

* With the new testfile `libs/commands/publish/src/lib/throttle-queue.spec.ts`
* By publishing to a local verdaccio instance

The removal from draft of this PR will wait for the upload to NPM of a 10000+ modules monorepo

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/publish/src/index.ts
+++ b/libs/commands/publish/src/index.ts
@@ -88,6 +88,9 @@ interface PublishCommandConfigOptions extends CommandConfigOptions {
   rejectCycles?: boolean;
   distTag?: string;
   preDistTag?: string;
+  throttle?: boolean;
+  throttleSize?: number;
+  throttleDelay?: number;
 }
 
 class PublishCommand extends Command {
@@ -974,10 +977,18 @@ class PublishCommand extends Command {
     };
     process.on("log", logListener);
 
-    const queue =
-      this.conf.get("registry") !== "https://registry.npmjs.org/"
-        ? new ImmediateQueue()
-        : new TailHeadQueue();
+    let queue;
+    if (
+      this.options.throttle ||
+      (this.options.throttle === undefined && this.conf.get("registry") !== "https://registry.npmjs.org/")
+    ) {
+      queue = new TailHeadQueue(
+        this.options.throttleSize !== undefined ? this.options.throttleSize : 20,
+        (this.options.throttleDelay !== undefined ? this.options.throttleDelay : 30) * 1000
+      );
+    } else {
+      queue = new ImmediateQueue();
+    }
     const mapper = pPipe(
       ...[
         (pkg: Package) => {
--- a/libs/commands/publish/src/index.ts
+++ b/libs/commands/publish/src/index.ts
@@ -978,12 +978,9 @@ class PublishCommand extends Command {
     process.on("log", logListener);
 
     let queue;
-    if (
-      this.options.throttle ||
-      (this.options.throttle === undefined && this.conf.get("registry") !== "https://registry.npmjs.org/")
-    ) {
+    if (this.options.throttle) {
       queue = new TailHeadQueue(
-        this.options.throttleSize !== undefined ? this.options.throttleSize : 20,
+        this.options.throttleSize !== undefined ? this.options.throttleSize : 25,
         (this.options.throttleDelay !== undefined ? this.options.throttleDelay : 30) * 1000
       );
     } else {
--- a/libs/commands/publish/src/index.ts
+++ b/libs/commands/publish/src/index.ts
@@ -51,6 +51,7 @@ import { getTwoFactorAuthRequired } from "./lib/get-two-factor-auth-required";
 import { gitCheckout } from "./lib/git-checkout";
 import { interpolate } from "./lib/interpolate";
 import { removeTempLicenses } from "./lib/remove-temp-licenses";
+import { ImmediateQueue, TailHeadQueue } from "./lib/throttle-queue";
 import { verifyNpmPackageAccess } from "./lib/verify-npm-package-access";
 
 module.exports = function factory(argv: Arguments<PublishCommandConfigOptions>) {
@@ -973,6 +974,10 @@ class PublishCommand extends Command {
     };
     process.on("log", logListener);
 
+    const queue =
+      this.conf.get("registry") !== "https://registry.npmjs.org/"
+        ? new ImmediateQueue()
+        : new TailHeadQueue();
     const mapper = pPipe(
       ...[
         (pkg: Package) => {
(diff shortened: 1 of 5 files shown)


JamesHenry:
@Ayowel this comment wasn't entirely clear to me:

> The removal from draft of this PR will wait for the upload to NPM of a 10000+ modules monorepo

Are you going to be reporting back to us once you have done this? Is that repo in question a real thing? Or something built just to test this functionality?

I have pulled in the latest main so that you can get initial CI feedback

JamesHenry:
It failed code formatting, you can run `npm run format:write` and commit the result

Ayowel:
> Are you going to be reporting back to us once you have done this? Is that repo in question a real thing? Or something built just to test this functionality?

The repo is a real thing, I'm having trouble with the specifics of lerna's configuration as some configuration works when testing locally but not the rest of the time.

It appears that the NPM limit is not fixed (I was at 25 publish per day before getting timed-out initially, I'm allowed ~500/day at the moment and I'm not clear why the limit changed), so I'll move throttling to an off-by-default state to honor the principle of least astonishment (and proper throttling configuration will be the user's responsibility).

I'll rebase and format once I figure this out

Ayowel:
Looks good to me

Ayowel:
Not sure if I should keep rebasing each time master is updated, let me know if i shouldn't

libs/commands/publish/src/lib/throttle-queue.ts
```diff
@@ -0,0 +1,90 @@
...
+   * @param f A function that will be executed once the queue is useable
+   * @return A promise that wraps the value returned by f
+   */
+  queue<T>(f: () => Promise<T>): Promise<T>;
+}
+
+/**
+ * A queue that immediately executes any promise it receives
+ */
+export class ImmediateQueue implements Queue {
```
JamesHenry, line 16: I don't think we need this concept, let's just not leverage the Queue if not throttling

libs/commands/publish/src/index.ts
```diff
@@ -973,14 +977,25 @@ class PublishCommand extends Command {
...
+    }
     const mapper = pPipe(
       ...[
         (pkg: Package) => {
           const preDistTag = this.getPreDistTag(pkg);
           const tag = !this.options.tempTag && preDistTag ? preDistTag : opts.tag;
           const pkgOpts = Object.assign({}, opts, { tag });
 
-          return pulseTillDone(npmPublish(pkg, pkg.packed.tarFilePath, pkgOpts, this.otpCache))
+          return pulseTillDone(
```
JamesHenry, line 996: See previous comment, let's just use the old logic if throttle is not enabled

libs/commands/publish/src/lib/throttle-queue.ts
```diff
@@ -0,0 +1,90 @@
...
+  private queue_size: number;
+  private queue_period: number;
+  private allowance: number;
+  private last_end: number[];
+
+  /**
+   * @param size The number of items that may run concurrently
+   * @param period The time between the end of the execution of an item and the start of the execution of the next one (ms)
+   */
+  constructor(size = 20, period = 30 * 1000) {
```
JamesHenry, line 38: I'd prefer this class to require this data to be passed in and only have a set of default values defined once for maximum clarity

libs/commands/publish/src/index.ts
```diff
@@ -973,14 +977,25 @@ class PublishCommand extends Command {
     };
     process.on("log", logListener);
 
+    let queue;
+    if (this.options.throttle) {
+      queue = new TailHeadQueue(
+        this.options.throttleSize !== undefined ? this.options.throttleSize : 25,
```
JamesHenry, line 983: These default values are easy to miss, please extract them to name `const`s

JamesHenry (changes requested):

libs/commands/publish/src/lib/throttle-queue.ts
Ayowel: The idea is to not have to think about the queue when using it. With this "abstraction", the queue can be used transparently and its instantiation can be delegated to a function no matter what the user-configured value is (it also allows to implement new queue types transparently as this specific implementation seems to work well-enough for my need with NPM but other environments/users might have different needs).

libs/commands/publish/src/index.ts
Ayowel: How do you envision this change ? A ternary that would look like `has_queue ? queue.queue(() => npmPublish(pkg, pkg.packed.tarFilePath, pkgOpts, this.otpCache)) : npmPublish(pkg, pkg.packed.tarFilePath, pkgOpts, this.otpCache)` ?

JamesHenry: Yeah, exactly, thank you

libs/commands/publish/src/lib/throttle-queue.ts
JamesHenry: For sure, I get it, it's just one more abstraction and one more moving piece to be tested, and in this case it is modifying the existing common case of not needing throttling. I'd like to keep that path unchanged but still support your case

libs/commands/publish/src/index.ts
Ayowel: Wasn't sure where to declare them as file-global consts do not appear to be a thing in the project so I just added the const declaration right before they are used.

JamesHenry (approved):
Thank you, @Ayowel!

Issue #4040, opened 2024-07-07
Regression: cannot run lerna since 8.1.5 with MODULE packages


92dev:
#4033 seems to break dependancies when running lerna using `npm run`
This happened when migrating the `gauge` pacakge 
See https://github.com/lerna/lerna/blob/9e822ca9b27f6e798290058386f8c5e58b120c5b/libs/core/src/lib/npmlog/gauge/wide-truncate.ts

This probably used to work since it was a `.js` file and not a `.ts` like now with the migration
For now I workaround the issue by just specifying that `lerna` has to be under `8.1.5` - but I really prefer to work with latest

## Current Behavior

```
> dev
> lerna run --stream --parallel dev,storybook

..../node_modules/lerna/dist/index.js:319
    var stripAnsi = require("strip-ansi");
                    ^

Error [ERR_REQUIRE_ESM]: require() of ES Module ..../node_modules/strip-ansi/index.js from ..../node_modules/lerna/dist/index.js not supported.
Instead change the require of ..../node_modules/strip-ansi/index.js in ..../node_modules/lerna/dist/index.js to a dynamic import() which is available in all CommonJS modules.
    at libs/core/src/lib/npmlog/gauge/wide-truncate.ts (/..../node_modules/lerna/dist/index.js:319:21)
    at __require (..../node_modules/lerna/dist/index.js:12:50)
    at libs/core/src/lib/npmlog/gauge/render-template.ts (..../node_modules/lerna/dist/index.js:453:24)
    at __require (..../node_modules/lerna/dist/index.js:12:50)
    at libs/core/src/lib/npmlog/gauge/plumbing.ts (..../node_modules/lerna/dist/index.js:660:26) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.14.0
```

## Expected Behavior

For it to work like the past

## Steps to Reproduce

Use a package scheme such as:
```
{
  "name": "lerna-regression",
  "private": true,
  "workspaces": [
    "packages/*"
  ],
  "scripts": {
    "dev": "lerna run --stream --parallel dev,storybook",
    "build": "lerna run --stream build",
    "type-check": "lerna run type-check"
  },
  "devDependencies": {
    "lerna": "^8.1.3",
    "prettier": "^3.2.5",
    "prettier-plugin-tailwindcss": "^0.6.1"
  }
}
```

with some module package:
```
{
  "name": "@lerna-regression/core",
  "private": true,
  "version": "1.0.0",
  "type": "module",
[...]

PR #4045, opened 2024-07-11
fix: set explicit strip-ansi dependency

Brooooooklyn:
- Fix https://github.com/lerna/lerna/issues/4040


Brooooooklyn:
You guys should deprecate  `8.1.5` and `8.1.6` and set the npm latest tag to `8.1.4` since these 2 versions are totally unusable.

JamesHenry (approved):

JamesHenry:
Thank you @Brooooooklyn and sorry for missing this in the original PR, it was very unfortunate that npm deprecated that group of packages in one go. I am on vacation without my laptop until very late on Thursday, but can cut a new release on Friday and mark those versions as deprecated as you say

JamesHenry:
8.1.7 is the latest and contains this fix, thanks again @Brooooooklyn

PR #4070, opened 2024-08-25
chore: update repo to nx 19.6.2

JamesHenry:
## Types of changes

## Checklist:


xmahle:
Could we get this in? Would be nice to get rid of the`Server-Side Request Forgery in axios` security alert.. 🙇

JamesHenry:
@xmahle I will need to take another look at this later in the week, there are some strange jest issues occurring.

Regarding an axios vulnerability, naturally the only user input you will be giving this developer tooling is your own. I trust you will not be looking to exploit the vulnerability on your own machine :) These kind of vulnerabilities only apply when applying the tooling in a different context.

Nevertheless, you also should not be blocked by lerna in updating the nx copy that appears in your node_modules, lerna depends on a very open range, so you should be able to bump `nx` and `@nx/devkit` (make sure you bump they both and that they always match version number) no problem.

PR #4095, opened 2024-10-16
fix: drop strip-ansi in favor of native stripVTControlCharacters

mcmxcdev:
## Description
- uninstall `strip-ansi` dependency from multiple `package.json`
- replace `strip-ansi` import with `util.stripVTControlCharacters` in `wide-truncate.ts`
- replace `strip-ansi` import with `util.stripVTControlCharacters` in `index.spec.ts`

## Motivation and Context
References https://github.com/es-tooling/ecosystem-cleanup/issues/122

## Types of changes

## Checklist:


JamesHenry:
Thank you @mcmxcdev! Looks like the snapshot change you made causes the tests to fail. Was that unintentional?

JamesHenry (changes requested):
See comment

JamesHenry:
I have marked this as a fix because it affects the runtime, and even though it should be like for like I think it's good to call it out in the release notes (which will happen automatically for a fix)

mcmxcdev:
> Thank you @mcmxcdev! Looks like the snapshot change you made causes the tests to fail. Was that unintentional?

This was definitely unintenional and was caused by this setting in VSCode:

image

Not gonna solve it for every dev, but it might be a good idea to introduce a `settings.json` in the repo that disables this setting.

JamesHenry:
Thank you @mcmxcdev!

Issue #2060, opened 2019-04-29
lerna publish --canary publishes broken packages


bimbiltu:
`lerna publish --canary` only looks at HEAD to determine what packages changed. This means that a package can be published without correct dependencies if those dependencies were changed in an earlier commit. I believe that `--canary` only operating on the HEAD commit is intentional, but I'm not sure how to achieve the "nightly release" behavior that this flag is supposed to provide.

I read through a few issues about the `--canary` and this may be a duplicate but I decided to create a new issue to share my use case and describe the problems im running into.

## Expected Behavior

Every package published by `lerna publish --canary` should have dependencies that align with the state of the repo when the publish was run.

## Current Behavior

Since `lerna publish --canary` only looks at the HEAD for changes it can publish broken packages. For an example, see the "Steps to Reproduce" section

## Possible Solution

One solution of this could be to add `--changed` to `lerna publish` like https://github.com/lerna/lerna/issues/962 suggested

As a workaround I could use `--force-publish "*"` or try to replicate the `--since` flag using `--force-publish` and `lerna list --since`. I could also stop using the `--canary` flag but then I would lose its ability to automatically increment the `-alpha.X` suffix.

## Steps to Reproduce (for bugs)

Do the following in a monorepo with package-a and package-b where package-a depends on package-b
1. Commit a new feature to package-b and run `lerna publish --canary`. lerna releases `package-a@1.0.0-alpha.0` and `package-b@1.0.0-alpha.0`. `package-a` will depend on `package-b@1.0.0-alpha.0` as expected.
2. Commit a new feature to package-a and run `lerna publish --canary`. lerna releases `package-a@1.0.0-alpha.1` that depends on `package-b@1.0.0`. `package-a` will be broken if it uses a feature introduced in a previous commit and therefore is only available in package-b's canary

<details><summary>lerna.json</summary><p>

I was using a pretty basic config:
```json
{
  "packages": [
    "packages/*"
  ],
  "npmClient": "npm",
  "version": "1.0.1"
}
```
</p></details>

## Context

I have a monorepo containing a few apps and libraries that are used by the apps. I want to run `lerna publish --canary` in CI for each commit to the master branch and a lot of the apps end up getting broken canaries as they are not using the correct versions of the libraries.

## Your Environment

| Executable | Version |
| ---: | :--- |
| `lerna --version` | 3.13.4 |
| `npm --version`  | 6.4.1 |
| `yarn --version` | 1.13.0 |
[...]

PR #4124, opened 2024-11-15
feat(publish): replace +{SHA} with .{SHA} in lerna publish --canary

MatusKnizka:
Replace +{SHA} with .{SHA} in lerna publish --canary

## Description
Instead of using plus (+) sign in package name provided by --canary flag, use dot.

## Motivation and Context
It's not possible to publish --canary version into jfrog artifactory, as jfrog is cutting part of filename, due to + (plus) character. So, instead of using +, I would suggest to use dot.

https://github.com/lerna/lerna/issues/2060

## How Has This Been Tested?
Unit tests
Integration tests
Run publish command

## Types of changes

## Checklist:


JamesHenry:
I'm really sorry for the delay on a formal resolution here @MatusKnizka. This is not a PR we can proceed with directly because it's mixing a couple of different concerns IMO and is a breaking change which would required wider clean up.

The `+` is significant in the `semver` spec, it's called "build metadata": https://semver.org/#spec-item-10

Importantly, it is just metadata and _not_ used to differentiate versions during installs and dependency resolution:

`1.0.0-alpha.0+abc123` and `1.0.0-alpha.0+def456` will _not_ be treated any differently by your package manager - they are effectively the same version - `1.0.0-alpha.0`.

Therefore, it does make sense that jfrog is "cutting it" as you describe.

Please can you share more about your use-case? Are you trying to publish over the same version with different build metadata? Are you saying that lerna is not incrementing the `.0` after the alpha upon subsequent canary releases?

JamesHenry (changes requested):
See comment

Issue #4167, opened 2025-03-21
Lerna v8.2.0+ removes package-lock.json dependencies on lerna version.


kbrzozowski-clgx:
## Current Behavior

After update of lerna to 8.20.1, packages which has `peer: true` in `package-lock.json` are removed from the file on `lerna version ${version} --no-git-tag-version --force-publish --yes`. It causes npm install to crash as `package-lock.json` is changed.

## Expected Behavior

`lerna version ${version} --no-git-tag-version --force-publish --yes` doesn't remove these packages.

## Steps to Reproduce
1. Install `deck.gl` package with `npm install deck.gl`
2. Use `lerna version ${version} --no-git-tag-version --force-publish --yes` command to bump version.
3. `arcgis` dependencies which have `peer:true` are removed from package-lock

### Failure Logs / Configuration

<details><summary>lerna.json</summary><p>

```{
  "$schema": "node_modules/lerna/schemas/lerna-schema.json",
  "version": "0.160.0-SNAPSHOT",
  "hoist": true
}
```

lumirlumir:
I'm having the same issue.  

Running `npx lerna version` removes unnecessary dependencies from `package-lock.json`.

mm-georges:
I'm having the same issue on the lastest version. I had to change back to 8.1.9

JamesHenry:
Hi folks, I’m afraid these are not valid steps to reproduce the issue. Please can someone provide steps to reproduce and I’ll take a look?

JamesHenry:
Actually sorry I think I had a local env issue, I think I did reproduce now, let me take a look

lumirlumir:
@JamesHenry If you still need a valid one, please ping me.

xofromthemoon:
Same here. :(

jaysoo:
PR to fix this: https://github.com/lerna/lerna/pull/4175

The local workaround for now is to set `npm_config_legacy_peer_deps=false` environment variable.

PR #4168, opened 2025-03-23
fix: incorrectly removes `package-lock.json` dependencies on `lerna version`

lumirlumir:
## Description

@JamesHenry  Hello,  

I've fixed the issue mentioned in [#4167](https://github.com/lerna/lerna/issues/4167).

resolve: #4167

## Motivation and Context

Currently, running `npx lerna version (version) --no-git-tag-version` results in the removal of incorrect dependencies from the `package-lock.json`.

I was able to reproduce the issue from [#4167](https://github.com/lerna/lerna/issues/4167) in a similar manner. **This issue occurs in Lerna `v8.2.0` and `v8.2.1`**, but it works as expected in `v8.1.9`.

This is essentially a partial revert of the changes made in [#4151](https://github.com/lerna/lerna/pull/4151). (https://github.com/lerna/lerna/compare/v8.1.9...v8.2.0)

The bug was caused by the plugin isolation code: `process.env["NX_ISOLATE_PLUGINS"] = "false";`.

## How Has This Been Tested?

After fixing the code, running `npx lerna version (version) --no-git-tag-version` now works correctly, just as it did in Lerna `v8.1.9`.

## Types of changes

## Checklist:

### DIFF
--- a/libs/core/src/lib/cli.ts
+++ b/libs/core/src/lib/cli.ts
@@ -1,6 +1,3 @@
-// Plugin isolation is not relevant to lerna
-process.env["NX_ISOLATE_PLUGINS"] = "false";
-
 import dedent from "dedent";
 import os from "node:os";
 import yargs from "yargs";
--- a/libs/core/src/lib/test-helpers/create-project-graph.ts
+++ b/libs/core/src/lib/test-helpers/create-project-graph.ts
@@ -1,7 +1,6 @@
 // We always need fresh copies of the graph in the unit test fixtures
 process.env.NX_DAEMON = "false";
 process.env.NX_CACHE_PROJECT_GRAPH = "false";
-process.env.NX_ISOLATE_PLUGINS = "false";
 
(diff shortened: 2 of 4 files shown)


mm-georges:
FYI: I'm not using the command `--no-git-tag-version` but I'm also facing the issue.

lumirlumir:
@mm-georges 

Umm, I believe the issue can also be resolved with the solutions I provided.  

I looked into the differences between `v8.1.9` and `v8.2.0`, and the only problem was the use of `process.env["NX_ISOLATE_PLUGINS"] = "false";`.

JamesHenry:
@lumirlumir Undoing a fix for another issue isn't the solution I'm afraid, all we would do is return to the other issue. A member of our team is looking into this one this week, we can reproduce, we just need to diagnose it now

JamesHenry:
Thanks a lot for your patience here folks, this is the fix: https://github.com/lerna/lerna/pull/4175

PR #4169, opened 2025-03-24
chore(deps): solve axios 1.8.2 vulnurability

oyershov:
## Description
🔒 Fix High Severity Vulnerability in axios via nx Dependency Bump

## Motivation and Context
This PR addresses a [high severity vulnerability](https://github.com/advisories/GHSA-jr5f-v2jv-69x6) in the axios package (<1.8.2), which is vulnerable to possible SSRF and credential leakage via absolute URLs.

Although axios is not a direct dependency of this project, it is brought in transitively via @nrwl/nx. To resolve this, I’ve bumped the nx dependencies from 20.4.5 to ^20.5.0, which includes an updated axios version that is no longer affected by this vulnerability.

## How Has This Been Tested?

1. Build passes: `npm run build`.
2. Unit tests pass: `npm test`.
3. Integration tests pass: `npm run integration`.
4. `npm audit` results no high severity axios vulnerability. 

### Before
Screenshot 2025-03-24 at 14 50 29

### After
Screenshot 2025-03-24 at 15 33 54

## Types of changes

## Checklist:


JamesHenry:
Thanks @oyershov but this does not do anything. Lerna depends on a range of nx, so the issue is already resolved by simply updating any consuming lockfiles.

Additionally, it's important to note that Lerna (nor Nx) is not subject to this violation because you do no run it as a server exposing it to outside influence. They are developer tools, not long running processes with connection to outside requests.

JamesHenry:
Re-reading this, I believe I may have misunderstood your intention with this PR - were you only trying to resolve it for this repo specifically? If so, thank you, but as I say it's not a pressing issue for the reasons I gave above, and the way to fully "resolve" it would be to run `nx migrate latest` rather than bumping the packages in this way

PR #4179, opened 2025-04-11
chore: kill legacy linting setup and migrate from globby to tinyglobby

JamesHenry:


benmccann:
Thanks for this change @JamesHenry! Any chance we could get it published as a new version?

JamesHenry:
`8.2.3` is out now

Issue #2668, opened 2020-07-23
Conventional commits do not respect '!' on types


alexforsyth:
Lerna conventional commits do not respect '!' as denoting a breaking change when part of the 'types' on a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#examples). Additionally, It seems like including the '!' may make lerna not recognize that commit as a valid conventional commit in some cases. [This repo](https://github.com/alexforsyth/lerna-conventional-commits) also highlights this problem. 

Repro repo: 
https://github.com/alexforsyth/lerna-conventional-commits

## Expected Behavior

The generated changelog should list breaking changes, when a '!' is included on a type. 

ex. 2 from [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#examples)
```
refactor!: drop support for Node 6
```
Should be equal to using the `BREAKING CHANGE: ` phrase in the footer.

## Current Behavior

The generated changelog does not include the change when '!' is included on a type. 

ex.
```
refactor!: drop support for Node 6
```

## Steps to Reproduce (for bugs)
[This repo is a live example](https://github.com/alexforsyth/lerna-conventional-commits) that highlights this problem.

alexforsyth:
Bump

kamsar:
The default settings uses [conventional-changelog-angular](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular) which is not the conventional commits spec, and doesn't support !.

It will work if you add `--changelog-preset conventionalcommits`, see https://github.com/lerna/lerna/issues/2138

Also in your repro, you're explicitly specifying the bump as `minor`, which overrides using conventional commits to determine the desired bump - maybe that's what you want, though.

https://github.com/alexforsyth/lerna-conventional-commits/blob/b575a9278be8ea053a8e1617014fe9be679d1272/package.json#L12

kamsar:
This can also be specified in `lerna.json`: https://github.com/lerna/lerna/tree/master/commands/version#--create-release-type

mrchief:
I'm seeing the same issue with lerna 3.2 and v4. Adding 

```
{
  "changelogPreset": {
    "name": "conventionalcommits",
    "issueUrlFormat": "{{host}}/{{owner}}/{{repository}}/issues/{{id}}"
  }
}
```

has no effect.

mrchief:
Digging further I found this: 
[...]

PR #4189, opened 2025-05-16
docs(version): update conventional-commits info in version docs

maribethb:
## Description
Adds information about using conventional-commits with `lerna version`. In particular, adds details about how the use of the `--conventional-commits` flag interacts with the `--changelog-preset` flag.

View the markdown preview of this change [here](https://github.com/maribethb/lerna/blob/d3104b1d3c42e235b4db2ddc4c55bf9ecee03eba/libs/commands/version/README.md)

## Motivation and Context
The previous docs stated that passing `--conventional-commits` flag would have lerna bump versions according to the conventionalcommits.org spec. However, this was not actually true, because the conventionalcommits.org spec allows using `!` in the header to denote a breaking change. If you only set the `--conventional-commits` flag, you would then be surprised to learn that your changes marked with `!` did not cause a new major version bump. You can see many examples of this confusion in https://github.com/lerna/lerna/issues/2668.

This change updates the documentation to explain the use of the `angular` spec as default, and explains how to change the spec used. It also updates the `--changelog-preset` option to explain that you need to install the spec used as a devDependency if you don't use the default.

## How Has This Been Tested?
This is a documentation change only.

## Types of changes

## Checklist:

### DIFF
--- a/libs/commands/version/README.md
+++ b/libs/commands/version/README.md
@@ -184,6 +184,10 @@ If the preset exports a builder function (e.g. `conventional-changelog-conventio
 }
 ```
 
+If you use any preset other than the default, you should also install the matching `conventional-changelog` package as a devDependency.
+
+The value of this argument will also be used to determine the version bump for each package if the [`--conventional-commits`](#--conventional-commits) flag is passed.
+
 ### `--changelog-entry-additional-markdown`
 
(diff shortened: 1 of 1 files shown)


JamesHenry:
Thanks very much for this!

PR #4216, opened 2025-08-08
refactor: use native object spreading

MichaelDeBoey:
~~[`structuredClone`](https://nodejs.org/api/globals.html#structuredclonevalue-options) is natively available as of Node v17, so no need to have an extra dependency for this anymore~~

We could use Object spreading since we're only accessing root properties

## Types of changes

## Checklist:

---

I'm sure people from the wider @e18e ecosystem cleanup (like @43081j, @benmccann, @Fuzzyma, @outslept & @talentlessguy) will be very happy to see these kind of changes as well

### DIFF
--- a/libs/core/src/lib/command/index.ts
+++ b/libs/core/src/lib/command/index.ts
@@ -1,6 +1,5 @@
 import { ProjectFileMap } from "@nx/devkit";
 import { ExecOptions as NodeExecOptions } from "child_process";
-import cloneDeep from "clone-deep";
 import dedent from "dedent";
 import { daemonClient } from "nx/src/daemon/client/client";
 import os from "os";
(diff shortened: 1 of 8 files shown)


JamesHenry (approved):
Thank you!

libs/core/src/lib/command/index.ts
```diff
@@ -93,7 +92,7 @@ export class Command<T extends CommandConfigOptions = CommandConfigOptions> {
     log.pause();
     log.heading = "lerna";
 
-    const argv = cloneDeep(_argv);
+    const argv = structuredClone(_argv);
```
outslept, line 95: Function objects cannot be duplicated by the structured clone algorithm; attempting to throws a DataCloneError exception. I believe this is why tests are failing

43081j:
do we even need a deep clone? 👀 

looking at the code, we `delete` some properties and access some others, but it looks like they're always root properties?

so we could just `{...argv}` maybe... or even better, just type it as `readonly` and refactor so we never mutate `argv`

JamesHenry:
@43081j I'm sure you're right, this code will be very old

JamesHenry (approved):

JamesHenry:
Thanks again @MichaelDeBoey

PR #4242, opened 2025-11-03
fix: expand version range to include nx v22.x

JamesHenry:


tay1orjones:
I'm glad to see this and would love to see it get in soon as we're running into the issues described and fixed over in 

* https://github.com/nrwl/nx/pull/32784

We want to use OIDC but we can't update to lerna v9 because of this blocker.

JamesHenry:
@tay1orjones sure, one of the reasons this stalled this week was because it uncovered some pretty wonky behaviour around git versions that is covered by very old tests (from many years before we took over).

Essentially Nx's dependency graph got _more_ accurate with regards to local dependencies and lerna's old tests here are expecting something extremely loose/inaccurate.

We've discussed it internally and we're classing this as a bug fix and not a breaking change, as much as anything because there are now _many_ better options for specifying a local dependency than a git/github reference back to the same repo. Indeed our hypothesis is that these tests and expected behaviour might well be before things like `file:` and `workspace:` were even a thing.

I am now looking into the other failures

JamesHenry:
Looks like the remaining failures were a Windows specific regression: https://github.com/nrwl/nx/pull/33408

I will get this branch updated once an Nx patch is available with that in

PR #4260, opened 2026-01-20
chore: remove dead link in README.md

msakrejda:
## Description
This link appears to no longer be working. I'm proposing that lerna drops it to avoid confusion. Of course, if another reference source is available, we can use that instead.


JamesHenry:
Thank you!

PR #4263, opened 2026-01-22
fix(deps): add missing ci-info dependency

mhassan1:
## Description

In https://github.com/lerna/lerna/commit/d51e344e016ebc2308a2185ca7e660a0cb47d143, we added code that uses the `ci-info` package, but we did not add a dependency on `ci-info` to `package.json`; as a result, strict environments (e.g. Yarn Plug'n'Play) show an error like this:
```txt
Error: lerna tried to access ci-info, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

## Motivation and Context

See above.

## How Has This Been Tested?

N/A

## Types of changes

## Checklist:

### DIFF
--- a/e2e/repair/src/repair.spec.ts
+++ b/e2e/repair/src/repair.spec.ts
@@ -56,6 +56,7 @@ describe("lerna-repair", () => {
       - nx: 22-0-0-release-version-config-changes (Updates release version config based on the breaking changes in Nx v22)
       - nx: 22-0-0-consolidate-release-tag-config (Consolidates releaseTag* options into nested releaseTag object structure)
       - nx: 22-1-0-update-nx-wrapper (Updates the nx wrapper.)
+      - nx: 22-7-0-add-polygraph-to-git-ignore (Adds .nx/polygraph to .gitignore)
       ---------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


oscarmarina:
Hi,

I've been testing the new [npm install-strategy = "linked"](https://github.com/npm/cli/pull/8996) and Lerna fails consistently because it attempts to require('ci-info') in libs/core/src/lib/oidc.ts. Since it's not explicitly declared in Lerna's dependencies, the "linked" strategy (which is very strict with phantom dependencies) blocks access to it.

Example/Reproduction Branch:

- https://github.com/oscarmarina/lerna-netlify-test/blob/feature/install-strategy-linked/package.json#L113

**Note:** In this branch, I've had to manually add ci-info to my root package.json as a workaround to keep the CI running. If I remove that line, Lerna immediately fails with the following trace:

Error: Cannot find module 'ci-info'
Require stack:

```sh
 .../node_modules/lerna/dist/index.js
 .../node_modules/lerna/dist/cli.js
    at libs/core/src/lib/oidc.ts (.../node_modules/lerna/dist/index.js:5703:30)
```

Thanks for the fix!

JamesHenry:
I will fix the snapshot and sync up

PR #4264, opened 2026-01-22
feat(core): add bun as supported package manager

Ch-Valentine:
## Description

Adds Bun as a supported package manager to Lerna, enabling users to leverage Bun's native performance for monorepo management.

## Motivation and Context

Bun has gained significant adoption for its speed. This PR enables Lerna to work seamlessly with Bun workspaces alongside npm/yarn/pnpm.

**Key Changes:**
- Automatic `bun.lockb` detection
- `bunx` and `bun run` support
- Lockfile updates during versioning
- Full E2E test coverage
- Comprehensive documentation

## How Has This Been Tested?

- [x] Unit tests added and passing
- [x] 4 E2E test files created (init, version, run, publish)
- [x] Format, lint, TypeScript checks passing
- [x] Integration tests

## Types of changes

## Checklist

- [x] My code follows the code style of this project
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [x] I have read the **CONTRIBUTING** document
- [x] I have added tests to cover my changes
- [x] All new and existing tests passed

---

> [!NOTE]
> **Medium Risk**
> Touches version commits (lockfile delete/regenerate) and package-manager detection; behavior is well covered by unit and E2E tests but lockfile handling is inherently sensitive.
> 
> **Overview**
> Adds **Bun** as a fourth `npmClient` option (`npm`, `yarn`, `pnpm`, `bun`) across init, install/run scripting, versioning, schema, docs, CI, and E2E.
> 
> **Init** detects `bun.lock` / `bun.lockb` (ahead of other lockfiles), recognizes `bunx` via `process.versions.bun` and path segments (including `.bun` cache), and writes `"npmClient": "bun"` when appropriate.
> 
> **Core execution** invokes `bun` directly when Corepack is on (Corepack has no Bun), sets `BUN_CONFIG_REGISTRY` for registry overrides, and routes `bun install` / `bun run` without yarn-only flags.
> 
> **Version** refreshes Bun lockfiles by deleting the stale file, running `bun install --lockfile-only` (with optional `--ignore-scripts`), restoring the old lockfile on failure, and staging `bun.lock` or `bun.lockb`.
> 
> **E2E fixtures** gain Bun registry setup (`bunfig.toml`, `BUN_CONFIG_REGISTRY`), `bunx` for init/lerna, and new specs for init, version lockfile, legacy `run`, and publish.
> 
> **Docs/schema** extend `lerna-schema.json` and add a **Using Bun with Lerna** recipe.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8adacd97a0cd6406130330dee05b4323141d9d37. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Ch-Valentine:
@JamesHenry Hey! Can you take a look on the PR and review it, please ?

vcherevatyi-smartling:
@JamesHenry Hey, any updates regarding this PR ? We'd like to use the bun with the lerna in our company ?

JamesHenry (changes requested):
Sorry for the delay on this one folks, there's a few things missing before this can be eligible, the most pressing of which is that this doesn't actually install `bun` anywhere

(I'll let CI run so that you can see the status of other checks in advance)

AI-JamesHenry:
> [!NOTE]
> 🤖 This review was performed by [@AI-JamesHenry](https://github.com/AI-JamesHenry), an AI assistant account guided and overseen by [@JamesHenry](https://github.com/JamesHenry).

Great work on this — the approach is solid and follows the established patterns well. Here's a detailed review with some things that need addressing before this can merge.

---

## Blocking Issues

### 1. `path.includes("bun")` false-positives on Ubuntu

`detectInvokedPackageManager` checks `invoker.path.includes("bun")` — this will match on **every Ubuntu CI runner** (and any path containing "ubuntu", "abundance", etc.). This needs a more specific check, e.g. matching on path segments rather than substring.

### 2. Missing `case "bun":` in `fixture.ts: lernaInit()`

The `install()` and `lerna()` methods in the Fixture class get bun cases, but `lernaInit()` has no `case "bun":` and falls through to `default: throw`. The init E2E test calls `fixture.lernaInit()` directly, which would throw.

### 3. No registry configuration for bun in E2E tests

The fixture's `install()` method for bun just runs `bun install` without any registry configuration, so E2E tests would hit the real npm registry instead of the local Verdaccio instance.

**How Nx solved this (in `nrwl/nx`):** Bun uses dedicated environment variables — no `bunfig.toml` needed:

```javascript
process.env.BUN_CONFIG_REGISTRY = registry;  // e.g. http://localhost:4873
process.env.BUN_CONFIG_TOKEN = authToken;
```

Both `bun install` and `bunx` respect these env vars automatically. This pattern is proven in Nx's e2e infrastructure and can be adopted directly here. The env vars just need to be set alongside the existing `npm_config_registry` / `YARN_REGISTRY` vars in Lerna's e2e setup scripts.

### 4. `fixture.ts: inferPackageManagerFromExistingFixture()`

Only checks for `pnpm-workspace.yaml` and `yarn.lock`, defaults to npm. Missing `bun.lockb`/`bun.lock` detection — so any fixture restoration would misidentify bun projects as npm.

---

## Should Fix

### 5. Corepack compatibility

All PM invocations go through `exec-package-manager.ts` which wraps calls with `corepack` when `COREPACK_ROOT` is set. Corepack doesn't support bun, so if a user has corepack enabled, `corepack bun install` would fail. This needs either a guard (skip corepack wrapping for bun) or documentation.

### 6. `npm-install.ts` flag handling

Has yarn-specific flag logic (`--non-interactive`, `--mutex`, skipping `--global-style`). Bun may need its own branch here, or at minimum verification that the default npm-style flags don't break bun.

### 7. `get-npm-exec-opts.ts` / registry env vars

Sets `npm_config_registry` which bun may not respect. Bun needs `BUN_CONFIG_REGISTRY` instead. Without this, registry configuration silently breaks for bun users outside of E2E tests too.

### 8. `update-lockfile-version.ts`

Directly reads and modifies `package-lock.json` as JSON to update nested version fields. Bun's binary lockfile (`bun.lockb`) can't be JSON-parsed. The text-based `bun.lock` is JSONC but has a different structure. This file may need a bun-aware path or should just skip direct manipulation and rely on `bun install --lockfile-only`.

### 9. `bun publish` needs explicit `--registry` flag

Per Nx's implementation, `bun publish` requires an explicit `--registry` CLI flag — the `BUN_CONFIG_REGISTRY` env var alone isn't sufficient for publish operations.

### 10. Unit test gaps

- No unit tests for the `detectPackageManager()` / `detectInvokedPackageManager()` changes
- No unit test for `npmRunScriptStreaming` with bun (only `npmRunScript` is covered)

---

## Minor / Nice to Have

### 11. Lockfile detection priority

Bun lockfile detection is ordered before yarn/pnpm/npm in `detectPackageManager()`. This means projects with both `bun.lockb` and another lockfile would prefer bun. This is a design choice worth documenting.

### 12. No programmatic config reading for bun

Bun doesn't support `bun config get registry` ([oven-sh/bun#7140](https://github.com/oven-sh/bun/issues/7140)), so any code that tries to read the configured registry programmatically would need to skip bun.

### 13. `fixture.ts: setNpmRegistry()`

Has a pnpm special case writing store-dir to `.npmrc`. Bun doesn't need file-based config (env vars are sufficient per Nx's approach), but this method should be aware of bun to avoid writing unnecessary npm config.

---

Overall this is a great foundation — the approach and structure are right. The main gaps are around the E2E infrastructure (registry config, fixture methods) and the `"bun"` substring matching. Happy to help work through any of these!

Ch-Valentine:
Hey @JamesHenry, thanks a lot for such a detailed review! Really appreciate the time spent on this.

I went through all 13 points and here is the current status:

Blocking issues (1-4) — all addressed:
  - Reworked path detection to use exact segment matching instead of includes(). Added explicit test
  for "ubuntu" false-positive.
  - Added case "bun": to lernaInit() with bunx and BUN_CONFIG_REGISTRY env var.
  - All fixture methods (install(), lerna(), lernaInit()) now pass BUN_CONFIG_REGISTRY via env, same
  approach as Nx uses.
  - inferPackageManagerFromExistingFixture() now checks for both bun.lockb and bun.lock.

Should fix (5-10) — mostly addressed:
  - Corepack — added npmClient !== "bun" guard in exec-package-manager.ts.
  - Bun install is called with clean args, no yarn-specific flags. Covered by unit test.
  - get-npm-exec-opts.ts now sets BUN_CONFIG_REGISTRY alongside npm_config_registry.
  - Lockfile update uses bun install --lockfile-only instead of trying to parse binary lockfile.
  - Regarding bun publish — actually Lerna doesn't shell out to bun publish at all. It uses
  libnpmpublish library which makes direct HTTP requests to registry, so the registry URL is passed
  programmatically. This means npmClient setting doesn't affect publish behavior, and no --registry flag
  is needed.
  - Added tests for detectPackageManager, detectInvokedPackageManager, both npmRunScript and
  npmRunScriptStreaming with bun.

Minor (11-13) — addressed:
  - Documented lockfile detection priority in the docs.
  - No code tries to read bun config programmatically, so nothing to do here.
  - setNpmRegistry() has a comment explaining bun uses env vars, not file-based config.

Also fixed the format:check failure (trailing whitespace in docs).

@JamesHenry regarding bun installation in CI — added oven-sh/setup-bun@v2 to the GitHub Actions setup. Let me know if anything else is missing!

libs/commands/version/src/index.ts
```diff
@@ -797,7 +797,9 @@ class VersionCommand extends Command {
           client,
           [
             "install",
-            "--lockfile-only",
+            // pnpm supports --lockfile-only to update the lockfile without a full install;
+            // bun does not have this flag — bun install updates the lockfile by default.
+            ...(client === "pnpm" ? ["--lockfile-only"] : []),
```
JamesHenry: This does not seem to be true:

https://bun.sh/docs/pm/lockfile#generate-a-lockfile-without-installing

JamesHenry:
Please sync with latest main whenever you have chance, it resolved an e2e failure

libs/commands/version/src/index.ts
Ch-Valentine: Yeah, Thanks for the catch! I made self review, and of course with AI, and did this mistake in the last commit =/

Ch-Valentine: How can I run all these CI on my workspace to not distract you every push to the branch ? set up nx for my forked repo and run all tests there? I just never used the nx before

JamesHenry: We have a `CONTRIBUTING.md` which outlines everything. Let me know if anything is incorrect

JamesHenry: You can always just copy and paste the commands from the ci.yml too

Ch-Valentine: Oh, that’s it? That’s easy. I thought the logic would be much more complicated because of tests for different platforms (Windows, Unix). Thanks! I pushed new commit and tests are passed. Please review it when you have a chance

Ch-Valentine:
Hey @JamesHenry!

Can you review the PR, please? All tests are passed locally, and I hope they will also pass on the CI

libs/commands/init/src/index.ts
```diff
@@ -245,13 +245,16 @@ export class InitCommand {
...
   private detectPackageManager(): PackageManager | null {
-    const packageManager = existsSync("yarn.lock")
-      ? "yarn"
-      : existsSync("pnpm-lock.yaml")
-      ? "pnpm"
-      : existsSync("package-lock.json")
-      ? "npm"
-      : null;
+    const packageManager =
+      existsSync("bun.lockb") || existsSync("bun.lock") || existsSync("bunfig.toml")
```
JamesHenry: Please remove the `bunfig.toml` for now, it just doesn't feel right since all the other package managers are based on lock files, not e.g. `.npmrc` or `.yarnrc.yml`

libs/commands/init/src/index.ts
JamesHenry, line 269: This comment is now in the wrong place

JamesHenry: And incomplete

libs/core/src/lib/corepack/exec-package-manager.ts
```diff
@@ -8,7 +8,7 @@ function createCommandAndArgs(npmClient: string, args: string[]) {
   let command = npmClient;
   const commandArgs = [...args];
 
-  if (isCorepackEnabled()) {
+  if (isCorepackEnabled() && npmClient !== "bun") {
```
JamesHenry, line 13: Would be worth a specific comment explaining this one

JamesHenry (commented):
There is a subtle bug regarding lockfiles in `libs/commands/version/src/index.ts` verified by Codex and Claude:

  1. A repo has bun.lockb (the old binary format from Bun <1.2)
  2. Line 805: lockfilePath resolves to /root/bun.lockb (it exists)
  3. Line 814: bun.lockb is deleted from disk via fs.unlinkSync
  4. Line 817-826: bun install --lockfile-only runs — but if the user has upgraded to Bun 1.2+, Bun now generates bun.lock (text
  format) instead of bun.lockb
  5. Line 833: updatedPath finds bun.lock (the new file) and adds it to changedFiles
  6. The deletion of bun.lockb is never added to changedFiles

  Then in git-add.ts:66-68, with granularPathspec: true (the default), only the explicitly listed files are staged. So git add --
  bun.lock runs, but the deletion of bun.lockb is never staged. The version commit includes the new bun.lock but the old bun.lockb
  remains tracked in git.

  The fix would be to also add the original lockfilePath to changedFiles when it differs from updatedPath:

```ts
  const updatedPath = candidates.find((p) => fs.existsSync(p));
  if (updatedPath) {
    changedFiles.add(updatedPath);
  }
  // If the lockfile format changed (e.g. bun.lockb → bun.lock),
  // stage the deletion of the old file too.
  if (updatedPath !== lockfilePath) {
    changedFiles.add(lockfilePath);
  }
```

  This works because git add -- <deleted-file> stages the deletion. The non-granular path (git add -- .) would also work but is a
  heavier change that affects the broader staging behavior.

libs/commands/init/src/lib/init-command.spec.ts
```diff
@@ -214,4 +221,128 @@ describe("InitCommand", () => {
...
+
+  describe("detectPackageManager", () => {
+    afterEach(() => {
+      mockedExistsSync.mockReset();
+    });
+
+    it.each([
+      ["bun.lockb", "bun"],
+      ["bun.lock", "bun"],
+      ["bunfig.toml", "bun"],
```
JamesHenry: See other comment

Ch-Valentine:
@JamesHenry Thank you for the detailed review and a good catch! Let me some time to investigate these issues, make double check and then fix them!

JamesHenry (changes requested):
We're very close now, thanks for all your work on this!

Ch-Valentine:
@JamesHenry Hey! Sorry for the delay, I'll take a look it today to finally make the PR ready

JamesHenry:
@Ch-Valentine thanks, e2e now failing, I imagine you were relying on the now removed logic in the test

Ch-Valentine:
@JamesHenry yeah, but maybe my solution wasn't the best one, feel free to propose changes to unblock the PR, ok ?

JamesHenry:
@Ch-Valentine I'll get this one over the line for you, sorry it's taken so long overall

Ch-Valentine:
@JamesHenry thank you so much for the help! And for accepting this feature! I had a pleasure!

PR #4397, opened 2026-07-21
chore(deps): bump js-yaml to 4.3.0 and tar to 7.5.20

i62navpm:
## Description

Bumps two runtime dependencies of the `lerna` package to their latest patched releases:

| Dependency | From | To |
| --- | --- | --- |
| `js-yaml` | 4.2.0 | 4.3.0 |
| `tar` | 7.5.16 | 7.5.20 |

Both bumps stay within the current major; the diff is the two `package.json` declarations (plus the root devDependency on `js-yaml`) and the corresponding `package-lock.json` entries, regenerated with npm 12 using `--package-lock-only`.

## Motivation and Context

Both dependencies are flagged by `npm audit` / `yarn audit` at their current pins:

- **`js-yaml` 4.2.0** is vulnerable to CVE-2026-59869 (merge-key chains can force quadratic CPU consumption), patched in 4.3.0.
- **`tar` 7.5.16** is vulnerable to CVE-2026-59873 (**critical** — decompression/parse DoS via unlimited input, patched in 7.5.19), CVE-2026-59874 (high — infinite loop on a negative tar entry size, patched in 7.5.18), CVE-2026-59871 and CVE-2026-59875 (moderate — PAX parsing crashes, patched in 7.5.18/7.5.17).

Because `lerna` pins exact versions, every consumer inherits these findings: the released `lerna` 9.0.7 pins `js-yaml` 4.1.1 and `tar` 7.5.11, so downstream projects currently see one critical and two high audit findings via `lerna` alone (that is how we ran into this). Bumping here clears them at the source with the next release.

## How Has This Been Tested?

- `npm audit` no longer reports the advisories above for `js-yaml`/`tar`.
- Both releases are semver-compatible patch/minor updates within the majors already in use; no `js-yaml` or `tar` API surface changes are involved.

## Types of changes

## Checklist:

---

> [!NOTE]
> **Low Risk**
> Dependency-only semver-compatible security patches; no logic or API usage changes in this diff.
> 
> **Overview**
> Pins **`js-yaml`** from 4.2.0 to **4.3.0** (root devDependency and `packages/lerna`) and **`tar`** from 7.5.16 to **7.5.20** in `packages/lerna`, with matching **`package-lock.json`** updates. No application source changes—only version declarations and lockfile entries.
> 
> These are same-major patch/minor bumps intended to clear **`npm audit`** advisories (YAML merge-key CPU abuse for `js-yaml`; decompression/parse and PAX-related issues for `tar`) for consumers of the published `lerna` package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7cea9d339a56b2e476eeb27aa76f13319f91a76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


JamesHenry (approved):

JamesHenry:
Thanks @i62navpm, these are the frustrating kind of alerts that have zero relevance to lerna's use-case, but happy to help reduce the noise

slorber:
Hey

@i62navpm & @JamesHenry, was just wondering: is there a good reason to use a pinned dependency version instead of a `^` semver range? 

Now that 4.3.0 is also vulnerable, it's not convenient to lock us on that exact version.

I understand it's noise, we all face the same problems unfortunately, but semver ranges can help and I don't think js-yaml release breanking changes on minor/patch versions, so it should be safe

JamesHenry:
> is there a good reason to use a pinned dependency version instead of a ^ semver range

Yes, the Shai-Hulud saga really shone a spotlight on how unsafe loose ranges are. You can always override in your own package manager so I really don't understand why people get so antsy about these irrelevant reports. You are in control?

It we use a loose range and your project pulls in the latest and that version turns out to be malware, that is a much worse scenario, right? It's much more likely a version that is fresh is genuinely malicious than one that is outdated.

JamesHenry:
I say irrelevant because 99% of the "vulnerabilities" that get reported in the context of Lerna or related build tooling require you to pwn your own machine, or have an attacker already have access, in which case exploiting build tooling is the last thing they'll be doing

JamesHenry:
I see you opened an issue about it so responded with this there as well so as not to have the discussion take place on a resolved PR

PR #4415, opened 2026-08-25
fix(core): keep modern writer template when preset ships a legacy writer guard

### LINKED ISSUE #4414: normalizeLegacyWriterOptions mis-triggers conventionalcommits@10.4's legacy-writer guard even with a modern writer
`lerna publish`/`lerna version` breaks with `conventional-changelog-conventionalcommits@10.4.0`, even though lerna bundles a modern `conventional-changelog-writer` (9.2.0, via `conventional-changelog@8.1.0`) that the guard added in conventional-changelog/conventional-changelog#1539 is meant to leave alone.

**Cause:** lerna's own preset normalization (`normalizeLegacyWriterOptions`) unconditionally Handlebars-compiles any string `writer.mainTemplate` it finds and uses that as the template, overwriting the correct function-based `writer.template` the preset already provides. Since 10.4.0's `mainTemplate` is intentionally a "poison pill" string meant to fail loudly *only* when read by an old (v8) writer, lerna's own pre-processing trips it before its modern writer even runs:

```
Missing helper: "conventional-changelog-conventionalcommits requires conventional-changelog-writer@9 or newer (conventional-changelog@8 or newer). Your changelog tooling loaded an older writer which cannot render this preset. Update the tooling or use an older major version of the preset."
```

**Repro:** `lerna.json` with `"changelogPreset": "conventional-changelog-conventionalcommits"`, preset pinned to `10.4.0`, run `lerna version --conventional-commits`.

**Expected:** changelog generation succeeds, since lerna's bundled writer is v9.2.0 (modern).
**Actual:** throws the error above during the `writer.write()` step.

**Workaround:** pin the preset to `9.3.1` (last version before the guard) or fix normalization to only compile `mainTemplate` when the writer doesn't already have `writer.template`.

- lerna: 10.0.1
- conventional-changelog-conventionalcommits: 10.4.0 (works: <=10.3.0, 9.3.1)
- Node: 24

lazerg:
## Description

`normalizeLegacyWriterOptions` compiles any string `writer.mainTemplate` it finds and assigns the result to `writer.template`. Presets built on `@conventional-changelog/template` ship such a string on purpose: it is a guard that renders to a "you need conventional-changelog-writer@9 or newer" error, and it sits next to the real function-based `template`. Since the guard is only meant to be read by an old writer, normalizing it throws away the working template and makes the bundled v9 writer render the error instead.

The normalization now leaves the writer options alone when `writer.template` is already a function.

## Motivation and Context

`lerna version --conventional-commits` with `conventional-changelog-conventionalcommits@10.4.0` fails with:

```
Missing helper: "conventional-changelog-conventionalcommits requires conventional-changelog-writer@9 or newer (conventional-changelog@8 or newer). Your changelog tooling loaded an older writer which cannot render this preset. Update the tooling or use an older major version of the preset."
```

Fixes #4414

## How Has This Been Tested?

Added a unit test with a fixture preset that carries both a guard `mainTemplate` and a modern `template`. It reproduces the error on `main` and passes here. Also ran `nx test core`, `nx test commands-version`, `nx test commands-publish`, lint and `format:check`.

## Types of changes

## Checklist:

---

> [!NOTE]
> **Low Risk**
> Narrow bug fix in preset normalization for conventional commits; behavior for legacy-only string templates is unchanged.
> 
> **Overview**
> Fixes changelog generation for modern conventional-changelog presets (e.g. `conventional-changelog-conventionalcommits@10.x`) that expose both a string `writer.mainTemplate` (legacy writer guard) and a function `writer.template`.
> 
> **`normalizeLegacyWriterOptions`** no longer compiles `mainTemplate` into `template` when `template` is already a function, so Lerna’s v9 writer keeps the real template instead of rendering the guard error message.
> 
> Adds a **`legacy-guard-preset`** fixture and an **`updateChangelog`** test that assert the changelog entry is produced from the function template, not the guard string.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 392bcfc3610039d14b4a7c70dc04cc5cde6c800c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/src/lib/conventional-commits/__fixtures__/fixed/scripts/legacy-guard-preset.js
+++ b/libs/core/src/lib/conventional-commits/__fixtures__/fixed/scripts/legacy-guard-preset.js
@@ -1,0 +1,25 @@
+"use strict";
+
+const parserOpts = require("./parser-opts");
+const whatBump = require("./what-bump");
+
+// Simulates a new v8+ preset that also ships the legacy writer guard, i.e. a
+// string mainTemplate which throws when rendered by a pre-v9 writer
+module.exports = function createPreset() {
+  return {
+    parser: parserOpts,
+    writer: {
+      mainTemplate: "{{[this preset requires conventional-changelog-writer@9 or newer] true}}",
+      template: (context) =>
+        [
+          `<a name="${context.version}"></a>`,
+          `## <small>${context.version} (${context.date})</small>`,
+          ...context.commitGroups.flatMap((group) => group.commits.map((commit) => `* ${commit.header}`)),
+          "",
+        ].join("\n"),
+      groupBy: `type`,
+    },
+    commits: { merges: false },
+    whatBump,
+  };
+};
--- a/libs/core/src/lib/conventional-commits/__fixtures__/fixed/scripts/legacy-guard-preset.js
+++ b/libs/core/src/lib/conventional-commits/__fixtures__/fixed/scripts/legacy-guard-preset.js
@@ -3,20 +3,25 @@
 const parserOpts = require("./parser-opts");
 const whatBump = require("./what-bump");
 
-// Simulates a new v8+ preset that also ships the legacy writer guard, i.e. a
-// string mainTemplate which throws when rendered by a pre-v9 writer
+// Simulates a new v8+ preset that also ships the legacy writer guard:
+// a string mainTemplate which only renders on a pre-v9 writer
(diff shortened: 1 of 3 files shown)


JamesHenry (approved):

JamesHenry:
Thanks @lazerg!

Issue #3273, opened 2022-08-01
Support npmjs.com security keys for 2FA


Frederick888:
- [ ] I'd be willing to implement this feature ([contributing guide](https://github.com/lerna/lerna/blob/main/CONTRIBUTING.md))

## Description

When publishing a package to npmjs.com, if the account uses security keys for 2FA, offer a https://www.npmjs.com/login/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx link where user can use security key to obtain an OTP.

## Motivation

At the moment the only reason for me to have a software authenticator in my npmjs.com account is to publish packages with lerna. It'd be great if I can keep only the hardware tokens which are more secure.

## Suggested Implementation

I'm not sure how lerna interacts with npmjs.com, but it'd be nice if we can have something similar to `npm publish`/`yarn publish`.

`npm publish`:

```
npm notice Publishing to https://registry.npmjs.org/
npm notice Open https://www.npmjs.com/login/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx to use your security key for authentication or enter OTP from your authenticator app
This operation requires a one-time password.
Enter OTP:
```

`yarn publish`:

```
info Two factor authentication enabled.
info Open https://www.npmjs.com/login/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx to use your security key for authentication or enter OTP from your authenticator app
question npm one-time password:
```
![Screenshot_20220801_223821](https://user-images.githubusercontent.com/4507647/182150242-4dd3c153-408e-485a-ab27-200b2d72e49d.png)
![Screenshot_20220801_223856](https://user-images.githubusercontent.com/4507647/182150267-daf30c15-b19c-4ce8-92b9-903f16b36944.png)

JamesHenry:
Thanks a lot @Frederick888! I had missed this before but it was also just raised on twitter. We can definitely add this 👍

sim31:
Now that new TOTP enrolments in npmjs are disabled this becomes even more relevant. Had to use access token just to publish packages using lerna.

TimothyJones:
@JamesHenry I've raised a PR that adds support for security keys when publishing - #4417

PR #4417, opened 2026-08-31
feat(core): Add support for authenticating with a security key when publishing

TimothyJones:
Hi! I ran into #3273 today, and this is a fix for it.

For full disclosure, Claude wrote most of the code in this PR, but I (a human) have personally read it, am happy with it, and have manually tested it. I'm happy to stand behind this change.

The change is based on what npm's CLI does, but adapted (I think) for Lerna's style. Please let me know if there's anything that needs changing, or if I've missed anything.

I have manually tested it against the repository that I couldn't publish on the current version of lerna, and it correctly spawns the browser, and after authenticating with a security key press, it then publishes successfully.

One note - this PR changes the auth that lerna requests from `legacy` to `web`. I was not able to test whether the old style of auth still works correctly, although I believe it should.

I also wasn't able to run the e2e tests on main before I made changes (and, uh, they still didn't run after, I did check). Hopefully these run in CI.

PR template below. With the exception of the first two paragraphs of the description I (a human) wrote it. For the two paragraphs in the description that Claude suggested, I copyedited so that they read better.

----

## Description

When an npm account uses a security key / passkey for 2FA rather than an authenticator app, the registry responds to publish (and dist-tag) requests with an EOTP error carrying `authUrl` and `doneUrl` instead of expecting a typed one-time password. Lerna's `otplease` helper treated every EOTP error as the classic flow and prompted "This operation requires a one-time password:", which users with security keys have no way to satisfy.

To fix this, I borrowed the behaviour from npm CLI: when the EOTP body carries a web-auth challenge, print and open `authUrl` in the browser (if configured), then poll `doneUrl` until the registry returns a token, and retry the request with that token as the `otp`. The existing OTP cache and semaphore ensure the challenge is completed once and the token reused for the remaining packages.

This also required changing the `authType` to `'web'`, which is the current default in npm. I've followed the behaviour of npm's cli as much as possible.

## Motivation and Context

Users with security keys and no authenticator set in their npm account are currently unable to publish with Lerna. It's possible to use an access token as a workaround. This was reported a couple of years ago in #3273, and now that npm is tightening up their requirements, this is going to become more of a problem.

## How Has This Been Tested?

Automated tests are provided, and I've manually tested it to successfully publish a repository I was previously unable to publish. I was unable to test the OTP flow manually (as you can't set an authenticator on your npm account once you've set up a security key), but I believe it should work.

I couldn't get the e2e tests to run on main with no changes, so I wasn't able to run them after the changes.

## Types of changes

## Checklist:

- [x] All new and existing tests passed

^ With the exception of e2e, which I couldn't get to pass even on main

---

> [!NOTE]
> **Medium Risk**
> Changes default registry auth negotiation from legacy to web and extend shared OTP handling used on publish and dist-tag; explicit `--otp` and legacy auth-type still opt out, but untested edge cases around classic 2FA are possible.
> 
> **Overview**
> Adds **browser-based two-factor authentication** for npm accounts that use a security key or passkey, fixing publishes that previously stalled on a typed OTP prompt.
> 
> When the registry returns an `EOTP` error with `authUrl` and `doneUrl`, **`otplease`** now completes that web challenge (print URL, optionally open a browser, poll `doneUrl` for a token) instead of always prompting for a numeric OTP. Tokens are **cached and deduplicated** with the existing OTP semaphore, so multi-package publishes still authenticate once. The same path applies to **publish and dist-tag** registry calls that use `otplease`.
> 
> **`lerna publish`** advertises **`npm-auth-type: web`** and **`npm-command: publish`** by default (matching current npm), and falls back to **legacy** when `--otp` is set or `auth-type` is not `web`. npm-conf defaults and allowed **`auth-type`** values now include **`web`**.
> 
> Publish docs under **`--otp`** explain the security-key flow and **`browser`** / legacy opt-out.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3dc73c7a55496f21cd5538947623ac17e2a9af2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/libs/core/src/lib/npm-conf/defaults.ts
+++ b/libs/core/src/lib/npm-conf/defaults.ts
@@ -58,7 +58,7 @@ export const defaults = {
   also: null,
   audit: true,
   "audit-level": "low",
-  "auth-type": "legacy",
+  "auth-type": "web",
   "bin-links": true,
   browser: null,
   ca: null,
--- a/libs/core/src/lib/web-auth.ts
+++ b/libs/core/src/lib/web-auth.ts
@@ -11,7 +11,7 @@
  * - https://github.com/npm/npm-profile/blob/main/lib/index.js (webAuthOpener / webAuthCheckLogin)
  */
 
-import * as childProcess from "@lerna/child-process";
+import childProcess from "node:child_process";
 import fetch, { FetchOptions } from "npm-registry-fetch";
 import log from "./npmlog";
 
@@ -77,9 +77,7 @@ export async function getWebAuthOneTimePassword(
 
   // Opening the browser is best-effort: the URL has already been printed, so a missing or
   // failing opener (e.g. an SSH session) must not prevent the user from completing the challenge.
-  openInBrowser(authUrl, opts["browser"]).catch((err: Error) => {
-    log.verbose("web-auth", `Unable to open browser automatically: ${err?.message || err}`);
-  });
+  openInBrowser(authUrl, opts["browser"]);
 
   return pollForToken(doneUrl, opts);
 }
@@ -88,27 +86,53 @@ export async function getWebAuthOneTimePassword(
  * Open a URL with the user's default browser, honouring the npm `browser` config:
  * `false` disables opening entirely, a string is used as the opener command,
  * anything else uses the platform default (`open`, `start`, or `xdg-open`).
+ *
+ * This is deliberately fire-and-forget:
+ * - the child is unref'd so that an opener which stays in the foreground (as `xdg-open` may when it has to
+ *   launch the browser itself) cannot keep lerna alive after publishing has finished
+ * - failures are only logged, and never affect `process.exitCode`. `@lerna/child-process` is intentionally not
+ *   used here because it propagates a child's non-zero exit status onto `process.exitCode`, which would turn an
+ *   otherwise successful publish into a failed lerna run whenever no opener is available (SSH, containers, CI).
  */
-function openInBrowser(url: string, browser: unknown): Promise<unknown> {
+function openInBrowser(url: string, browser: unknown): void {
   if (browser === false) {
-    return Promise.resolve();
+    return;
   }
+
+  const onError = (err: unknown) => {
+    log.verbose("web-auth", `Unable to open browser automatically: ${(err as Error)?.message || err}`);
+  };
 
   // URLs are only ever passed through as a single argument, so escape anything that could be
   // interpreted by a shell (`"` etc) - the registry URL is expected to already be well-formed.
   const target = encodeURI(url);
 
-  if (typeof browser === "string") {
-    return childProcess.exec(browser, [target]);
-  }
+  try {
+    const child =
+      typeof browser === "string"
+        ? childProcess.spawn(browser, [target], { stdio: "ignore", windowsHide: true })
+        : process.platform === "win32"
+          ? // `start` is a cmd.exe builtin, so a shell is required. The empty quoted first argument is the
+            // window title, without which `start` would treat the URL as the title.
+            childProcess.spawn("start", ['""', `"${target}"`], {
+              shell: true,
+              stdio: "ignore",
+              windowsHide: true,
+            })
+          : childProcess.spawn(process.platform === "darwin" ? "open" : "xdg-open", [target], {
+              stdio: "ignore",
+            });
 
-  if (process.platform === "win32") {
-    // `start` is a cmd.exe builtin, so a shell is required. The empty quoted first argument is the
-    // window title, without which `start` would treat the URL as the title.
-    return childProcess.exec("start", ['""', `"${target}"`], { shell: true });
+    child.on("error", onError);
+    child.on("exit", (code) => {
+      if (code) {
+        onError(new Error(`opener exited with code ${code}`));
+      }
+    });
+    child.unref();
+  } catch (err) {
+    onError(err);
   }
-
-  return childProcess.exec(process.platform === "darwin" ? "open" : "xdg-open", [target]);
 }
 
 async function pollForToken(doneUrl: string, opts: Record<string, unknown>): Promise<string> {
--- a/libs/core/src/lib/web-auth.ts
+++ b/libs/core/src/lib/web-auth.ts
@@ -1,0 +1,141 @@
+/**
+ * Web-based two-factor authentication ("web OTP") for registry requests.
+ *
+ * When an npm account uses a security key / passkey (WebAuthn) instead of an authenticator app,
+ * the registry cannot be satisfied with a typed one-time password. Instead, the EOTP error it
+ * returns carries an `authUrl` (to be opened in a browser, where the user completes the challenge)
+ * and a `doneUrl` (to be polled until it yields a token that can be sent as the `otp` on a retry).
+ *
+ * Adapted from:
+ * - https://github.com/npm/cli/blob/latest/lib/utils/auth.js (otplease)
+ * - https://github.com/npm/npm-profile/blob/main/lib/index.js (webAuthOpener / webAuthCheckLogin)
+ */
+
+import * as childProcess from "@lerna/child-process";
+import fetch, { FetchOptions } from "npm-registry-fetch";
+import log from "./npmlog";
+
+export interface WebAuthChallenge {
+  authUrl: string;
+  doneUrl: string;
+}
+
+/**
+ * Delay between polls of the `doneUrl` when the registry does not send a `retry-after` header.
+ */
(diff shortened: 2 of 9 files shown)


libs/core/src/lib/web-auth.ts
```diff
@@ -0,0 +1,141 @@
...
+    return childProcess.exec(browser, [target]);
+  }
+
+  if (process.platform === "win32") {
+    // `start` is a cmd.exe builtin, so a shell is required. The empty quoted first argument is the
+    // window title, without which `start` would treat the URL as the title.
+    return childProcess.exec("start", ['""', `"${target}"`], { shell: true });
+  }
+
+  return childProcess.exec(process.platform === "darwin" ? "open" : "xdg-open", [target]);
```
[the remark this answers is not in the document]
TimothyJones: I agree, this looks like an actual problem to me. It looks like lerna's child-process is maybe not intended for processes that we don't mind failing. Can a maintainer advise what the best solution is here?

libs/core/src/lib/npm-conf/defaults.ts
```diff
@@ -58,7 +58,7 @@ export const defaults = {
   also: null,
   audit: true,
   "audit-level": "low",
-  "auth-type": "legacy",
+  "auth-type": "web",
```
ghiscoding, line 61: changing this default seems like a breaking change, why not keeping the "legacy" default?

TimothyJones: Can't keep the `legacy` default, as npm won't offer the security key auth flow on `legacy`.

I had understood that `web` would work with both (this [page here](https://docs.npmjs.com/accessing-npm-using-2fa) says "If you have configured to use TOTP, you will see an TOTP prompt instead"). Is there a flow I'm missing that would make this a breaking change?

TimothyJones: If we need to keep it defaulting to legacy, we could leave the default and let users enable `web` if they want this feature.

In that case, ideally, the OTP prompt would let them know that they can set the auth type to `web` for a security key, but as far as I can tell there's no way we can detect that they're not configured for OTP.

libs/core/src/lib/web-auth.ts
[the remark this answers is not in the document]
TimothyJones: I "fixed" it by running our own childprocess call here, but that's almost certainly not correct. We could use the same spawner that npm's cli uses - https://www.npmjs.com/package/@npmcli/promise-spawn , but that's an extra dependency, and probably not designed for use outside npmcli. Thoughts?

libs/core/src/lib/npm-conf/defaults.ts
TimothyJones: Actually, a nicer option might be to leave it as `legacy` and have the prompt say "Enter OTP (or leave blank to open in the browser)" - and if it's left blank, then it'll redo the auth dance with `web`.

This would still be a reasonably straightforward experience for users, but wouldn't require a change to the default. What do you think?

ghiscoding: you can check what Codex AI wrote in the replicated Lerna-Lite [PR](https://github.com/lerna-lite/lerna-lite/pull/1406), its initial analysis was that the main part of the PR was mainly correct except for the change of default from `legacy` to `web` and I was in agreement with it, so it implemented the PR but kept `legacy` as the default... Typically any change made to a default config becomes a breaking change, so... avoiding such change is better I think, though in my case I now always use npm publishing provider anyway. Note that these are just suggestions, I have no say in the final review

TimothyJones:
Is a maintainer coming to look at this?

JamesHenry (changes requested):
Thanks a lot for this PR @TimothyJones.

I tested it locally against a fake registry that implements the `authUrl` / `doneUrl` handshake. The core flow works. The headers are correct (`npm-auth-type: web`, `npm-command: publish`, `user-agent: lerna/...`). The OTP cache and the semaphore work with the web token. `--otp` correctly forces `legacy`.

I agree with the change of the default `auth-type` to `web`. This is the npm default since npm 9 and the npm challenge page accepts both a security key and a TOTP code. Users can restore the old prompt with `auth-type=legacy` in `.npmrc` so there's always that escape hatch.

These changes are required before merge please.

**1. Do not send the failed request options to the `doneUrl` poll**

`pollForToken` spreads `opts` into the GET request. For `npm-dist-tag.add`, `opts` contains `method: "PUT"`, `body`, and `headers`. `minipass-fetch` rejects a GET request with a body. The error is:

```
TypeError: Request with GET/HEAD method cannot have body
```

This error stops the run during `--temp-tag` retagging. Packages then stay on the `lerna-temp` tag. My agent reproduced this with a second web challenge during the dist-tag step.

Remove the request-specific keys before the poll. Also remove the stale `otp`.

```ts
async function pollForToken(doneUrl: string, opts: Record<string, unknown>): Promise<string> {
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
  const { body, headers, method, otp, ...fetchOpts } = opts;

  while (true) {
    const res = await fetch(doneUrl, { ...fetchOpts, method: "GET", cache: false } as unknown as FetchOptions);
    // ...
  }
}
```

Add a unit test in `web-auth.spec.ts`. In the test, pass `opts` that contain `method: "PUT"`, `body`, `headers`, and `otp`. Assert that the fetch call does not contain `body`, `headers`, or `otp`.

**2. Do not prompt for a typed OTP before publish when `authType` is `web`**

In `publishPacked()`, this code runs when `--verify-access` is set and the account has 2FA enabled:

```ts
if (this.twoFactorAuthRequired) {
  chain = chain.then(() => this.requestOneTimePassword());
}
```

`requestOneTimePassword()` calls `getOneTimePassword("Enter OTP:")`. The validator accepts only digits. A user with a security key cannot answer this prompt. The run stops before the first request.

Skip this step when the auth type is `web`. Let the registry challenge drive the flow.

```ts
if (this.twoFactorAuthRequired && this.conf!["get"]("authType") !== "web") {
  chain = chain.then(() => this.requestOneTimePassword());
}
```

This change will break the existing test "prompts for OTP when option missing, account-level 2FA enabled, and verify access is true" in `publish-command.spec.ts`. Update that test. Set `auth-type` to `legacy` in it, for example with `npm_config_auth_type=legacy`, so it still covers the legacy prompt. Then add a second test for the default (`web`). Mock `getTwoFactorAuthRequired` to return `true`. Run with `--verify-access`. Assert that `getOneTimePassword` is not called.

**3. Consume the body of the `202` response**

`pollForToken` does not read the body of a `202` response. `npm-profile` always reads the body. An unread body can hold the keep-alive socket. Read and discard the body before the retry.

```ts
if (res.status === 202) {
  await res.buffer().catch(() => undefined);
  // ...
}
```

**4. Match the npm prompt behaviour**

npm prints the URL, then shows `Press ENTER to open in the browser...`, and polls `doneUrl` at the same time. This PR opens the browser immediately. Please match npm.

- Print the URL.
- Start the `doneUrl` poll.
- Show the prompt `Press ENTER to open in the browser...`.
- Open the browser only when the user presses ENTER.
- Cancel the prompt when the poll returns a token.
- Do not show the prompt when `browser` is `false`, or when `stdin` is not a TTY. In those cases, only print the URL and poll.

npm implements this as a `Promise.all` of the opener and the poll, with a shared `AbortController`, in `npm-profile`'s `webAuthOpener` and npm's `lib/utils/open-url.js` (`openUrlPrompt`). Use the same structure.

Add unit tests in `web-auth.spec.ts` for: ENTER opens the browser; a token from the poll cancels the prompt without opening the browser; `browser=false` shows no prompt.

**5. Test coverage**

Please add an e2e test with a small fake HTTP registry. The test must:

- return a `401` with `www-authenticate: OTP` and an `authUrl` / `doneUrl` body for the first publish PUT
- return `202` and then `200 { token }` from `doneUrl`
- accept the second publish PUT with `npm-otp`
- also return a challenge for a dist-tag PUT when `--temp-tag` is set

This test would have caught issue 1.

Verdaccio cannot return an OTP challenge, so this test cannot use the shared local registry. Start a small `node:http` server inside the test on a random port. Write the fixture `.npmrc` with `registry=http://127.0.0.1:<port>/`, an `_authToken` for that host, and `browser=false`. Use `Fixture.create({ installDependencies: false })`. Run the CLI through a small wrapper script that sets `process.stdin.isTTY` and `process.stdout.isTTY` to `true` before it imports `cli.js`, because `otplease` rethrows without a TTY. Close the server in `afterEach`.

PR #4432, opened 2026-09-17
fix: bump conventional-changelog so --conventional-commits can resolve conventional-commits-filter

### LINKED ISSUE #4427: `lerna version --conventional-commits` fails with a module resolution error
## Current Behavior

Running `lerna version --conventional-commits` fails with a module resolution error:
```
lerna ERR! ERR_MODULE_NOT_FOUND: Cannot find package 'conventional-commits-filter'
imported from ../node_modules/@conventional-changelog/git-client/dist/index.js
```

## Expected Behavior

`lerna version --conventional-commits` should generate the changelog and create the GitHub release without crashing. (It worked in`8.x` and `9.x`)

## Steps to Reproduce

- Install `lerna@10.0.1`
- Run `npx lerna version --conventional-commits --create-release github --yes`
- The error occurs

## Temporary workaround

Adding the following code to `package.json` fixes the issue

```
"overrides": {
  "lerna": { "conventional-changelog": ">=8.1.3" }
}
```

Or install `conventional-commits-filter` as a devDependency

## Suggested Fix

Bump the Lerna's `conventional-changelog` dependency from `8.1.0` to `8.1.3`. See https://github.com/conventional-changelog/conventional-changelog/releases/tag/conventional-changelog-v8.1.3

00200200:
Fixes #4427

`lerna version --conventional-commits` dies on nested installs with `ERR_MODULE_NOT_FOUND` for `conventional-commits-filter`. `conventional-changelog@8.1.0` left that as an optional peer of git-client.

I bumped it to `8.1.3`, which depends on the package for real.

Tested with:
`npx nx test lerna -- src/conventional-changelog-deps.spec.ts`


JamesHenry (approved):
Thank you!

